Issue #1, opened 2019-08-22
cargo run --example ggez fails


debris:
error:

`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ResourceNotFound("ui.png", [("/Users/marek/projects/iced/target/release/examples/resources", FilesystemError("Path \"ui.png\" is not valid: must be an absolute path with no references to parent directories")), ("/Users/marek/Library/Application Support/ggez.iced", FilesystemError("Path \"ui.png\" is not valid: must be an absolute path with no references to parent directories")), ("/Users/marek/Library/Preferences/ggez.iced", FilesystemError("Path \"ui.png\" is not valid: must be an absolute path with no references to parent directories"))])', src/libcore/result.rs:999:5
`

debris:
I'm not sure what is the best way to fix it is and how ggez attaches resources to the project

hecrj:
`ggez` loads resources based on the executable location.

In order to run the example, you need to create a `resources` symbolic link in `target/debug/examples` pointing to `examples/resources`. In Fish:

```fish
ln -s (pwd)/examples/resources target/debug/examples/.
```

PR #2, opened 2019-08-22
migrated Image and ProgressBar widgets

debris:
WIP, I need to add renderers to this PR, but I had a problem with running the example. It fails with #1

### DIFF
--- a/src/widget/image.rs
+++ b/src/widget/image.rs
@@ -1,0 +1,139 @@
+//! Displays image to your users.
+
+use crate::{
+    Style, Node, Element, MouseCursor, Layout, Hasher, Widget,
+    Rectangle, Point,
+};
+
+use std::hash::Hash;
+
+/// A widget that displays an image.
+/// 
+/// It implements [`Widget`] when the associated [`core::Renderer`] implements
+/// the [`image::Renderer`] trait.
+///
+/// [`Widget`]: ../../core/trait.Widget.html
+/// [`core::Renderer`]: ../../core/trait.Renderer.html
+/// [`image::Renderer`]: trait.Renderer.html
+/// # Example
+///
+/// ```
+/// use iced::Image;
+///
+/// let image = Image::new("image");
+/// ```
+pub struct Image<I> {
+    image: I,
+    source: Option<Rectangle<u16>>,
+    style: Style,
+}
+
+impl<I> std::fmt::Debug for Image<I> {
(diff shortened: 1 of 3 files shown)


src/widget/image.rs
```diff
@@ -0,0 +1,139 @@
...
+///
+/// [`Widget`]: ../../core/trait.Widget.html
+/// [`core::Renderer`]: ../../core/trait.Renderer.html
+/// [`image::Renderer`]: trait.Renderer.html
+/// # Example
+///
+/// ```
+/// use iced::Image;
+///
+/// let image = Image::new("image");
```
hecrj: I would probably write something like:

```
# let my_image_handle = String::from("some_handle");
let image = Image::new(my_image_handle);
```

That way, we remark the fact that the handle is generic.

hecrj: I would probably write something like:

```rust
# let my_image_handle = String::from("some_handle"); // This line will be hidden in docs
let image = Image::new(my_image_handle);
```

That way, we remark the fact that the handle is generic.

src/widget/progress_bar.rs
```diff
@@ -0,0 +1,117 @@
...
+/// It implements [`Widget`] when the associated [`core::Renderer`] implements
+/// the [`button::Renderer`] trait.
+///
+/// [`Widget`]: ../../core/trait.Widget.html
+/// [`core::Renderer`]: ../../core/trait.Renderer.html
+/// [`progress_bar::Renderer`]: trait.Renderer.html
+/// # Example
+///
+/// ```
+/// use coffee::ui::ProgressBar;
```
hecrj: Should be `iced::ProgressBar`.

src/widget/progress_bar.rs
```diff
@@ -0,0 +1,117 @@
...
+    Style, Node, Element, MouseCursor, Layout, Hasher, Widget,
+    Point, Rectangle,
+};
+
+use std::hash::Hash;
+
+/// A widget that displays a progress of an action.
+/// 
+/// It implements [`Widget`] when the associated [`core::Renderer`] implements
+/// the [`button::Renderer`] trait.
```
hecrj: `progress_bar::Renderer` here.

hecrj:
I fixed #1 in #3. Let me know if you have any further issues.

hecrj:
I fixed the issues and merged the PR manually.

Thank you again! :tada:

Issue #7, opened 2019-09-05
Perform benchmarks


hecrj:
As mentioned in #4, in order to avoid performance regressions, it would be great to run some benchmarks on the different parts of the API.

The most expensive operation is clearly `UserInterface::build`. This will be the main focus for many of the optimizations we will implement in the future. We should benchmark this first.

`UserInterface::update` and `UserInterface::draw` will also get slower the more widgets and events a user interface has to handle. It would be great to have some numbers!

TGotwig:
I can try to have a look 🕵🔎

hecrj:
@TGotwig Awesome! I really appreciate the help. Let me know if you need anything.

TGotwig:
CPU is constantly on `3.x` on MacOS, although nothing is happening.
<img width="815" alt="Aktivitätsanzeige (Meine Prozesse) 2019-09-14 14-47-26" src="https://user-images.githubusercontent.com/30773779/64908309-a16fb280-d6fe-11e9-860a-0c5cce31e80e.png">

lachlansneff:
Looks to me like the event loop it's using is continuously polling.

hecrj:
Yes, the example is drawing the UI every frame. This is expected.

Benchmarking the `ggez` example is not what this issue is about! `iced` is renderer-agnostic. We should benchmark `iced`.

PR #15, opened 2019-09-12
Add `env_logger` to tour example

TGotwig:
Relates #7


src/logger.rs
```diff
@@ -0,0 +1,19 @@
...
+// use flexi_logger::{Logger, opt_format};
+impl Logger {
+    /// Creates a new [`Logger`].
+    ///
+    /// [`Logger`]: struct.Logger.html
+    pub fn log(msg: &str) {
+        let now: DateTime<Utc> = Utc::now();
+        info!("{}: {}\n", now, msg);
+    }
+}
```
hecrj: This is normally implemented by a particular logging implementation. For instance, [`env_logger`](https://docs.rs/env_logger/0.6.2/env_logger/) already prints the time.

I think we do not need a `Logger`. We can use `log` macros directly, once we need them.

examples/tour/main.rs
```diff
@@ -33,6 +37,19 @@ pub fn main() -> ggez::GameResult {
...
+            TermLogger::new(LevelFilter::Warn, Config::default(), TerminalMode::Mixed).unwrap(),
+            WriteLogger::new(LevelFilter::Info, Config::default(),
+                OpenOptions::new()
+                    .append(true)
+                    .create(true)
+                    .open(".log")
+                    .unwrap()
+            ),
+        ]
+    ).unwrap();
```
hecrj: We should probably initialize the logger before building the context.

.gitignore
```diff
@@ -1,3 +1,6 @@
+.*
+!/.gitignore
+!/.travis.yml
```
hecrj: Ignoring all hidden files by default seems like a bit too much 😅 

You can ignore files locally by editing: `.git/info/exclude`.

Cargo.toml
```diff
@@ -20,6 +20,9 @@ features = ["winit"]
 [dependencies]
 stretch = "0.2"
 twox-hash = "1.5"
+chrono = "0.4.9"
+log = "0.4.8"
+simplelog = "0.7.2"
```
hecrj: I think [`env_logger`](https://docs.rs/env_logger/0.6.2/env_logger/) is a simpler alternative for the example. We should also add it as a development dependency instead.

src/logger.rs
TGotwig: Ah yeah right, forgot to edit this 🙈

.gitignore
TGotwig: All hidden files except `.gitignore` and `.travis.yml` 😉
But yeah maybe it's better to use `.git/info/exclude` for that 🙂

Cargo.toml
TGotwig: Hmm but `env_logger` can't write a log-file 😬

hecrj: You are right! However, I think it is not a necessary feature for the example and we should probably try to keep it as simple and straightforward as possible.

TGotwig:
Did all the changes, squashed, and force-pushed 😇

Cargo.toml
```diff
@@ -20,6 +20,7 @@ features = ["winit"]
 [dependencies]
 stretch = "0.2"
 twox-hash = "1.5"
+log = "0.4.8"
```
hecrj: We are not using the dependency for anything yet... We should probably add it once we need it.

examples/tour/main.rs
```diff
@@ -12,6 +12,8 @@ use ggez::filesystem;
 use ggez::graphics;
 use ggez::input::mouse;
 
+extern crate env_logger;
```
hecrj: `extern crate` is not necessary when using the 2018 edition of Rust.

examples/tour/main.rs
```diff
@@ -33,6 +35,7 @@ pub fn main() -> ggez::GameResult {
         true,
     );
 
+    env_logger::init();
```
hecrj: This should probably be the first thing we do in `main` :)

PR #16, opened 2019-09-15
Add performance-test for UserInterface::build with Criterion

TGotwig:
Relates #7
Runnable by `cargo bench`


TGotwig:
No problem, probably can look at the weekend or so 😊

erlend-sh:
Could this be merged as-is for a start?

hecrj:
@erlend-sh I will definitely merge this once we start working on benchmarks.

For now, the API is currently changing and we don't want to maintain an empty benchmark.

TGotwig:
Oh the API has changed a lot already 🙀
I'll close this and maybe give it another try after this month or so 🐱

Issue #8, opened 2019-09-05
Implement a low-level renderer


hecrj:
Both the renderer in [`coffee`] and the [`ggez`] one in the examples are using a high-level abstraction. This makes it really easy to build a somewhat simple renderer, but makes interesting features (like scrollables) complicated.

I would like to implement a renderer using a lower-level API. This renderer could accept different kinds of configuration, and also support widgets with many styling options.

I am familiar with [`wgpu`], and supporting all the platforms natively and also the web sounds great! In any case, I think there are many interesting similar efforts that may be worth investigating and contributing to!

[`coffee`]: https://github.com/hecrj/coffee
[`ggez`]: https://github.com/ggez/ggez
[`wgpu`]: https://github.com/gfx-rs/wgpu-rs

lachlansneff:
I'd like to help with this. I'm writing an application that is essentially a form of cad and I'd like to use iced as the gui for it.

olanod:
What would be the approach here, draw widgets in terms of the low level API directly or create higher level 2d drawing abstractions first that are then used to draw the widgets? I don't know the wgpu API but might seem a bit to low level?
Some time ago I read discussions([like here](https://github.com/draw2d/rfcs/issues)) about this kind of stuff and the need for 2d drawing APIs, I have little idea about low level graphics programming but what I got from @nical creator of [`lyon`](https://github.com/nical/lyon)(btw dependency in `ggez`) is that games and UIs might have different needs on how they draw 2d.   
`iced` would be mainly used for UIs but it would still be good to consider both cases by allowing widgets that have different renderer to be composed but hopefully share a common ground? For example I'd like to create a map widget and render tiles in a 2d/3d surface, the kind of APIs needed seem to be closer that of a game but mixes dialogs and is embedded on a normal UI, creators of such a widget usually bundle a lot of code because they create things in terms of the low level API when it seems there could be several things to be shared with the UI rendering code(e.g. text!).

Anyway maybe a good start could be something like Lyon? apparently it can [render using the wgpu backend ](https://github.com/nical/lyon/tree/master/examples/wgpu_basic) still more layers might need to go in the middle. Nical approach is more bottoms up, creating solid foundations, iced is starting from the top, having an opinionated but elegant design that UI developers are already used to and comfortable with. Now we are missing badass APIs that can fit in the middle that solve different use cases and can be used by widget developers in a composable way to create great experiences :smiley:

hecrj:
[...]

PR #17, opened 2019-09-21
Basic web support (core, native, and web crates)

### LINKED ISSUE #6: Decouple built-in widgets from their style
Some of the built-in widgets have part of their style hard-coded, like `height`.

We should be able to move this logic into the `Renderer` trait of each widget. Now, there are different ways of doing this:

  - constants, like `Renderer::HEIGHT`
  - simple methods, like `Renderer::height`
  - a single complex method to decorate a provided style, like `Renderer::style`

We should think about the pros and cons of each before proceeding.

### LINKED ISSUE #10: (V)DOM renderer
Iced working on the web would be awesome, the plan of creating a low level renderer with wgpu-rs is great but could take some time on your side and theirs(last time I checked web support was not there?) so DOM rendering  in the mean time?(side question, would it be easy to mix renderers? like a map component rendered with a wgpu renderer and the rest of the UI with the DOM one?).

The flexbox model of iced should make it easy to create a DOM renderer. For example with [Dodrio](https://github.com/fitzgen/dodrio), the experimental virtual DOM library showcased by Mozilla, should facilitate things and help with performance at the same time instead of using web-sys APIs directly. Structs implement the render trait of dodrio and build the DOM elements, so would it be just implementing the dodrio trait for each widget and put some DOM there?

### LINKED ISSUE #12: Constants applicable to all renderers
Hi !

As you may have noticed, I'm currently writing a `ncurses`-based renderer for iced on my spare time, and am using my own fork of `iced` for that purpose.

The only modification I did was to change some defaults. As it happens, the current constants are assuming pixel unit when pancurses of course works with lines and columns. This makes the current default on some Widgets (e.g. Checkbox, Slider, and Buttons) completely wrong. 

Even if not considering my use case, which I agree is fairly uncommon, I am not sure what these constants were there for in the first place, because they assume some sort of mandatory smaller screen resolution.

What do you think would be the correct way of not enforcing those defaults upon every renderer implementor ?

hecrj:
Fixes #10, resolves #6, and closes #12.

This PR adds basic web support and splits the current crate into multiple subcrates in order to set up the foundations of a cross-platform GUI toolkit.

The new subcrates are:

 - __`iced_core`__ contains the reusable Iced widgets and defines their API. It is meant to be used to implement a compatible runtime, like `iced_native` and `iced_web`.
- __`iced_native`__ builds flexbox layouting and widget logic on top of `iced_core` while staying renderer-agnostic. This is basically what the `iced` crate was until now.
- __`iced_web`__ builds a web runtime on top of `iced_core`. This mostly consists of a simple mapping of widgets to VDOM nodes and a simple runtime to wire messages.

Once we have a low-level renderer (#8), the top-level `iced` crate will become an end-user-oriented, cross-platform GUI toolkit that can be used directly without any need to choose for a particular renderer or worrying about any runtime wiring. It will use `iced_native` and `iced_web` to automatically choose which runtime to use based on the compilation target.

# Changelog

There have been many additional changes and improvements. Please, let me know if I missed anything!

## Added
- `iced_core`, `iced_native`, and `iced_web` subcrates.
- The `Length` type. It unifies methods like `width`, `fill_width`, etc.

## Changed
- `iced` becomes `iced_native`. The current `iced` crate will be empty until we have a first prototype of a low-level renderer.
- Widgets expose their fields publicly, simplifying the `Renderer` traits in `iced_native`.
- Widgets no longer hold a default `Style`. Related to #6 and #12,
- All `Renderer` traits now have total control to produce a `Node`. Related to #6 and #12.
- The tour example lives in its own crate now and it can be run both on native and web.

## Removed
- `Color` generic parameter in widgets. The new `Color` type is used now. This keeps widgets as reusable as possible. It may complicate implementing some runtimes, but I think a color mapper should be easily implementable in situations where colors are limited.

# Pending work before merge
- [x] Think about the `stretch` feature for enabling conversions in `iced_core`. We should probably move this logic to `iced_native`.
- [x] Review documentation.


web/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+    fn new(ui: impl UserInterface<Message = Message> + 'static) -> Self {
+        Self {
+            ui: RefCell::new(Box::new(ui)),
+        }
+    }
+
+    fn update(&mut self, message: Message) {
+        let mut ui = self.ui.borrow_mut();
+
+        // TODO: Resolve futures and publish resulting messages
```
hecrj: Could we use something like [`futures::executor::LocalPool`](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.18/futures/executor/struct.LocalPool.html) for this? We would need to poll it periodically it seems.

web/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+{
+    fn render<'a, 'bump>(
+        &'a self,
+        bump: &'bump bumpalo::Bump,
+    ) -> dodrio::Node<'bump>
+    where
+        'a: 'bump,
+    {
+        let mut ui = self.ui.borrow_mut();
+        let element = ui.view();
```
hecrj: We need a `RefCell` here because some widgets need mutable references to `State` and `dodrio`'s render does not provide a mutable reference. It's understandable, and I have some ideas to remove the `State` structs in the near future.

In the meantime, we may need to clarify the `README` implementation details.

web/src/bus.rs
```diff
@@ -0,0 +1,40 @@
+use crate::Application;
+
+use std::rc::Rc;
+
+#[derive(Clone)]
+pub struct Bus<Message> {
+    publish: Rc<Box<dyn Fn(Message, &mut dyn dodrio::RootRender)>>,
+}
```
hecrj: This is a composable "publisher" of messages. I am not entirely happy with the name `Bus`... Any ideas?

web/src/element.rs
```diff
@@ -0,0 +1,62 @@
...
+where
+    A: 'static,
+    B: 'static,
+{
+    fn node<'b>(
+        &self,
+        bump: &'b bumpalo::Bump,
+        bus: &Bus<B>,
+    ) -> dodrio::Node<'b> {
+        self.widget.node(bump, &bus.map(self.mapper.clone()))
```
hecrj: I really like how the `Widget` trait composes here. I have some ideas to achieve something similar in `iced_native`.

web/src/bus.rs
AlisCode: I think the word for "something to pass Messages" is a *channel*.

hecrj: Yeah. I believe this is more of a message _handler_ than a channel, as it updates the application on `publish`. We will also probably need a way to defer updates to make them asynchronous.

I think we should wait untill we have more information/use cases before we find a final name.

PR #23, opened 2019-10-23
Add `ROADMAP`

hecrj:
This PR adds a `ROADMAP.md` file describing the current state of Iced and some of the most important next steps we should take before it can become a production-ready GUI library.

I will open individual issues for each of these steps so we can further discuss them and/or propose new ideas!

Additionally, we should also update the `README` implementation details and include the crate graph there. I think it can help contributors grasp the different parts of Iced and see the bigger picture.


ROADMAP.md
```diff
@@ -0,0 +1,287 @@
...
+    }
+}
+```
+
+### Event subscriptions
+Besides performing async actions on demand, most applications also need to listen to events passively. An example of this could be a WebSocket connection, where messages can come in at any time.
+
+The idea here is to also follow [Elm]'s footsteps. We can add a method `subscriptions(&self) -> Subscription<Message>` to `Application` and keep the subscriptions alive in the runtime.
+
+The challenge here is designing the public API of subscriptions, so users can build their own, and _subscription diffing_, or basically detecting when a subscription is added/changed/removed. For this, we can take a look at the source code of [Elm] for inspiration.
```
memoryruins: The flow of

> ... so users can build their own, and _subscription diffing_, or basically ...

made the message slightly unclear to me. Is the following the intended meaning?
```suggestion
The challenge here is designing the public API of subscriptions. Users should be able to create and subscribe to events, and we should be able to support _subscription diffing_, i.e. detecting when a subscription to an event is added, changed, or removed. For this, we can take a look at [Elm] for inspiration.
```

ROADMAP.md
```diff
@@ -0,0 +1,287 @@
...
+
+### Layers
+Currently, Iced assumes widgets cannot be laid out on top of each other. We should implement support for multiple layers of widgets.
+
+This is a necessary feature to implement many kind of interactables, like dropdown menus, select fields, etc.
```
memoryruins:
```suggestion
This is a necessary feature to implement many kinds of interactables, like dropdown menus, select fields, etc.
```

ROADMAP.md
```diff
@@ -0,0 +1,287 @@
...
+`iced_native` will need to group widgets to perform layouting and process some events first for widgets positioned on top.
+
+`iced_wgpu` will also need to process the scene graph and sort draw calls based on the different layers.
+
+### Animations
+Allow widgets to request a redraw at a specific time.
+
+This is a necessary feature to render loading spinners, a blinking text cursor, GIF images, etc.
+
+[`winit`] allows controlling the event loop in a flexible way. We may be able to use [`ControlFlow::WaitUntil`](https://docs.rs/winit/0.20.0-alpha3/winit/event_loop/enum.ControlFlow.html#variant.WaitUntil) for this purpose.
```
memoryruins:
```suggestion
[`winit`] allows flexible control of its event loop. We may be able to use [`ControlFlow::WaitUntil`](https://docs.rs/winit/0.20.0-alpha3/winit/event_loop/enum.ControlFlow.html#variant.WaitUntil) for this purpose.
```

ROADMAP.md
```diff
@@ -0,0 +1,287 @@
...
+
+### Canvas widget
+A widget to draw freely in 2D or 3D. It could be used to draw charts, implement a Paint clone, a CAD application, etc.
+
+As a first approach, we could expose the underlying renderer directly here, and couple this widget with it ([`wgpu`] for now). Once [`wgpu`] gets WebGL or WebGPU support, this widget will be able to run on web too. The renderer primitive could be a simple texture that the widget draws to.
```
memoryruins:
```suggestion
As a first approach, we could expose the underlying renderer directly here, and couple this widget with it ([`wgpu`] for now). Once [`wgpu`] gets WebGL or WebGPU support, this widget will be able to run on the web too. The renderer primitive could be a simple texture that the widget draws to.
```

ROADMAP.md
hecrj: Hmm... Close, but not exactly what I meant! A subscription could produce different kinds of events. What about something like this?

```suggestion
The challenge here is designing the public API of subscriptions. Ideally, users should be able to create their own and the GUI runtime should keep them alive while performing _subscription diffing_ (i.e. detecting when a subscription is added, changed, or removed). For this, we can take a look at [Elm] for inspiration.
```

In any case, I am aware the term "subscription" is a bit broad here. I am basically referring to Elm's subscriptions, which are similar to a background listener that can produce events concurrently.

hecrj:
Thank you for the suggestions @memoryruins! I have integrated almost all of them. :heart:

I really appreciate the effort. I am not a native speaker and I still have a lot to learn!

ROADMAP.md
memoryruins: > What about something like this?

Sounds good! The clarification helps ^^

PR #37, opened 2019-11-02
Text input widget

### LINKED ISSUE #25: Text input widget
A widget where the user can type raw text and passwords.

This widget will have a quite complex event logic, potentially coupled with text rendering. Some examples are:
- Text cursor positioning based on a mouse click
- Text selection
- Clipboard interaction (copy & paste)
- Text editing shortcuts (go to start, go to end, etc.)

It also needs scrollable / clippable text support (see #24), as the introduced text may not fit the input and scrolling based on the text cursor will need to be implemented.

Additionally, the text cursor should blink at a regular interval, which can be considered an _animation_.

I think we should start simple here, and build a text widget with the most basic functionality: click to focus and type to fill. We can improve the implementation with time as the library matures.

The end-user API could look like this:

```rust
pub enum Message {
    TextChanged(String),
}

let text_input = text_input::State::new();
let text = "Hello";

TextInput::new(&mut text_input, &text, Message::TextChanged);
```

hecrj:
Depends on #35. Fixes #25.

[Text input section in tour example](https://gfycat.com/alertcalmcrow-rust-gui)

This PR adds a `TextInput` widget that can be used to allow users to interact with the application by inputting text. It features:

- Basic cursor movement. The arrow keys can be used to move the cursor between each `char`. It does not account for graphemes yet!
- An `on_submit` method to produce a message when the `Enter` key is pressed while the input is focused.
- Placeholder support.
- Configurable `padding`, `width`, and `size`.

# Pending work
- [x] Implement cursor-based horizontal scrolling when the content overflows.


victorstewart:
Any thoughts on...

- placeholder text
- emoji support
- ability to calculate the size of a string contains emojis and line breaks
- a flag to just resize font to accommodate the element width once boundary is reached
- attributes for the inputed text to be displayed with (color, spacing, alignment etc)

and I assume there's some easy way to launch the keyboard on iOS when you begin editing?

PR #39, opened 2019-11-04
Fix Tour for HiDPI (for discussion)

Matthias-Fauconneau:
The tour example sizes needs to be adapted to screen resolution.
It seems to use a lot of hardcoded pixel sizes, e.g wgpu default text size.
This PR is only to start the discussion : 
 It fixes some sizes hardcoded to my scale (288ppi) until it is clear how and where resolution independence should be implemented.

I think best would be to use physical units :
- font size: in pt 
- layout size: in em (relative to the text)

### DIFF
--- a/examples/tour.rs
+++ b/examples/tour.rs
@@ -76,7 +76,7 @@ impl Application for Tour {
         }
 
         let element: Element<_> = Column::new()
-            .max_width(Length::Units(540))
+            .max_width(Length::Units((540.0*3.0) as u16))
             .spacing(20)
             .padding(20)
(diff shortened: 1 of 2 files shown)


hecrj:
Thanks!

I have some plans to implement this properly (obtaining the HiDPI factor from `winit`) very soon. The idea is to define all the GUI using logical pixels/units and transform to physical pixels when rendering primitives.

The changes are a bit involved (we should probably branch from #38) and we may need to rethink a bit how text caching/measuring works.

hecrj:
@Matthias-Fauconneau I have implemented generic HiDPI support in #40. As always, feedback is greatly appreciated!

Issue #30, opened 2019-10-23
Layers


hecrj:
Currently, Iced assumes widgets cannot be laid out on top of each other. We should implement support for multiple layers of widgets.

This is a necessary feature to implement many kind of interactables, like dropdown menus, select fields, etc.

`iced_native` will need to group widgets to perform layouting and process some events first for widgets positioned on top.

`iced_wgpu` will also need to process the scene graph and sort draw calls based on the different layers.

manokara:
How would the depth of the widgets be determined? Maybe with something like this?

```rust
enum Depth {
    Above,
    Below,
    Topmost,
}

fn view(&mut self) -> Element<Message> {
    // ...
    Layer(SomeWidget::new(), Depth::Above)
    // ...
}
```

Then the current depth would have to be kept in the layouting state. Given a combobox for example, would it be best for the choice list to be part of the drawing code and the widget set to be on top (and `ComboBox::new` actually returns `Layer(ComboBox {...}, Depth::Above`) or the choice list be a separate widget, attached to the ComboBox's layout? Or maybe there's a better approach?

hecrj:
I do not have a particular approach in mind yet. There is definitely an interesting amount of exploration ahead of us here.

I do know using something like absolute indices (like `z-index` in CSS) can be quite painful and does not compose _at all_.

A relative `Layer` widget sounds like a good idea to me, as I think it should compose no matter where you decide to nest it. [`elm-ui` follows a similar approach](https://package.elm-lang.org/packages/mdgriffith/elm-ui/latest/Element#nearby-elements) and the author [gave a great talk in elm-conf](https://www.youtube.com/watch?v=Ie-gqwSHQr0) with many interesting ideas. I recommend watching it for inspiration!

> Given a combobox for example, would it be best for the choice list to be part of the drawing code and the widget set to be on top (and ComboBox::new actually returns Layer(ComboBox {...}, Depth::Above) or the choice list be a separate widget, attached to the ComboBox's layout? Or maybe there's a better approach?

I think layers (or at least __event__ layers) will need to exist as a first-class concept in `iced_native`. As of now, they cannot simply be another widget because the runtime needs to be aware of them. In other words, `update` needs to be aware that layers on top should capture mouse events.

I am not yet sure about how we will end up implementing this. I do have the feeling that the different methods in the `Widget` trait are inherently coupled (`node` defines layout, which in turn affects `update`, `draw`, and `hash_layout`). Layers will make this coupling even more apparent.

[...]

PR #43, opened 2019-11-06
Fix integer overflow with nested clip primitives

hecrj:
Fixes an integer overflow when nested clip primitives went out of screen boundaries.

Thanks to @memoryruins for catching this! :heart:

### DIFF
--- a/wgpu/src/renderer.rs
+++ b/wgpu/src/renderer.rs
@@ -265,8 +265,8 @@ impl Renderer {
                 if width > 0.0 && height > 0.0 {
                     let clip_layer = Layer::new(
                         Rectangle {
-                            x: x.max(0.0).ceil() as u32,
-                            y: y.max(0.0).ceil() as u32,
+                            x: x.max(0.0).floor() as u32,
+                            y: y.max(0.0).floor() as u32,
                             width: width.ceil() as u32,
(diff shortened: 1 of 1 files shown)


wgpu/src/renderer.rs
```diff
@@ -254,24 +254,31 @@ impl Renderer {
...
-                self.draw_primitive(content, layers);
-
-                layers.push(new_layer);
+                let x = bounds.x - layer.offset.x as f32;
+                let y = bounds.y - layer.offset.y as f32;
+                let width = (bounds.width + x).min(bounds.width);
+                let height = (bounds.height + y).min(bounds.height);
+
+                // Only draw visible content on-screen
+                // TODO: Also, check for parent layer bounds to avoid further
```
memoryruins: Is the`TODO` referring to a parent layer that is already defined/accessible?
or is it referring to part of #30 to be implemented and the `TODO` is a reminder for later?

hecrj: Yes, the parent layer should be the `layer` variable.

Implementing this `TODO` should be a simple rectangle intersection check. In fact, I think this is implementable for all primitives.

I have not decided to go for it yet because performance is not currently my main concern at this point. I prefer to keep code manageable and simple for now, while we are figuring things out.

PR #45, opened 2019-11-07
Minor clean-up

memoryruins:
This PR utilizes std's impls for text_input and makes use of `mem::take` in the todo example.

### DIFF
--- a/core/src/widget/text_input.rs
+++ b/core/src/widget/text_input.rs
@@ -125,17 +125,12 @@ impl Value {
     }
 
     pub fn until(&self, index: usize) -> Self {
-        Self(self.0[..index.min(self.len())].iter().cloned().collect())
+        Self(self.0[..index.min(self.len())].to_vec())
     }
 
     pub fn to_string(&self) -> String {
-        let mut string = String::new();
-
-        for c in self.0.iter() {
-            string.push(*c);
-        }
-
-        string
+        use std::iter::FromIterator;
+        String::from_iter(self.0.iter())
     }
 
     pub fn insert(&mut self, index: usize, c: char) {
--- a/examples/todos.rs
+++ b/examples/todos.rs
@@ -33,7 +33,7 @@ impl Application for Todos {
             Message::CreateTask => {
                 if !self.input_value.is_empty() {
                     self.tasks.push(Task::new(self.input_value.clone()));
-                    self.input_value = String::new();
+                    self.input_value.clear();
                 }
             }
             Message::TaskChanged(i, completed) => {


memoryruins:
Ah the life of only having nightly installed on this desktop has gotten the best of me; `mem::take` stabilizes in `1.40` 🤦. In that case,  we can use `std::mem::replace(&mut self.text_input, String::default())` or I can revert that commit altogether.

examples/todos.rs
```diff
@@ -32,8 +32,9 @@ impl Application for Todos {
             }
             Message::CreateTask => {
                 if !self.input_value.is_empty() {
-                    self.tasks.push(Task::new(self.input_value.clone()));
-                    self.input_value = String::new();
+                    // Move the user's input into a task, replacing input_value with an empty string.
+                    let task = Task::new(std::mem::take(&mut self.input_value));
+                    self.tasks.push(task);
```
hecrj: I think the original version is clearer. I'd bet the compiler will optimize the clone away.

However, It'd be nice if Rust could figure this out by itself:

```rust
self.tasks.push(Task::new(self.input_value));
self.input_value = String::new();
```

I don't see why this shouldn't work. Do you know if there is a related RFC for this? Maybe it's just very hard to implement correctly?

memoryruins: > I'd bet the compiler will optimize the clone away.

I'm unsure what optimizations like those are permitted around non-copy heap allocated types, but it currently does not optimize away https://rust.godbolt.org/z/AsnV2q

> Maybe it's just very hard to implement correctly?

It would likely complicate move and drop analysis, for both the compiler and users. Currently, when `input_value` moves out, `self` is invalidated, and during assignment, the previous non-copy value is dropped.

That said,

> I think the original version is clearer.

I agree, and now think its more correct. By cloning, `Task`'s string will then have the exact capacity it needs, and if we also use `clear`, `Application` will can re-use `input_value`'s current buffer.

```rust
self.tasks.push(Task::new(self.input_value.clone()));
self.input_value.clear()
```

Thus the use of take/replace was not a good choice for multiple reasons, and clarity in example code is key :)

hecrj: > I'm unsure what optimizations like those are permitted around non-copy heap allocated types, but it currently does not optimize away https://rust.godbolt.org/z/AsnV2q

Oh, wow! I didn't know about Compiler Explorer. This is amazing!

> I agree, and now think its more correct. [...]

Yeah, I like the `clear` version better than the original. It's great to see the simplest solution is also the best one.

core/src/widget/text_input.rs
```diff
@@ -125,17 +125,12 @@ impl Value {
...
     pub fn to_string(&self) -> String {
-        let mut string = String::new();
-
-        for c in self.0.iter() {
-            string.push(*c);
-        }
-
-        string
+        use std::iter::FromIterator;
+        String::from_iter(self.0.iter())
```
memoryruins, line 133: Note to anyone looking in from the changelog: this could have been `self.0.iter().collect()` :)
<sub> _micro-refactoring intensifies_ </sub>

PR #53, opened 2019-11-12
Initial iOS Xcode example

simlay:
This currently breaks at runtime due the same error as this: https://github.com/gfx-rs/gfx/issues/3086 is also going to break when https://github.com/rust-windowing/winit/issues/1082 is more complete.

I also added more integration endpoints.


simlay:
Closing in favor of using: https://github.com/iced-rs/ios-examples

PR #57, opened 2019-11-15
Touch support

simlay:
I've got mixed feelings about some of this - like that each touch enum has a position in it. I could have done it more similar to how the mouse events work.

I've tested the todo and tour examples on my iPhone 8 with iOS 13.2.2, the things that don't work are:
* Text input (though, the cursor ends up in the text box) - I think we need to signal to winit to pop up the keyboard.
* Touch scrolling on a scrollable widget. Given that the logic around cursor and touch movement not having the previous location, this didn't seem very easy to do. Fortunately, the side scrollbar works.

Here's what the tour example looks like (up to just before the rust image):

output

### DIFF
--- a/native/src/widget/text_input.rs
+++ b/native/src/widget/text_input.rs
@@ -16,7 +16,7 @@ use crate::keyboard;
 use crate::layout;
 use crate::mouse::{self, click};
 use crate::text;
-use crate::touch::{self, Touch};
+use crate::touch;
 use crate::{
     Clipboard, Element, Hasher, Layout, Length, Point, Rectangle, Size, Widget,
 };
@@ -249,10 +249,7 @@ where
     ) -> event::Status {
         match event {
             Event::Mouse(mouse::Event::ButtonPressed(mouse::Button::Left))
-            | Event::Touch(Touch {
-                phase: touch::Phase::Started,
-                ..
-            }) => {
+            | Event::Touch(touch::Event::FingerPressed { .. }) => {
                 let is_clicked = layout.bounds().contains(cursor_position);
 
                 self.state.is_focused = is_clicked;
--- a/native/src/widget/text_input.rs
+++ b/native/src/widget/text_input.rs
@@ -16,6 +16,7 @@ use crate::keyboard;
 use crate::layout;
 use crate::mouse::{self, click};
 use crate::text;
+use crate::touch::{self, Touch};
 use crate::{
     Clipboard, Element, Hasher, Layout, Length, Point, Rectangle, Size, Widget,
 };
@@ -247,7 +248,11 @@ where
         clipboard: Option<&dyn Clipboard>,
     ) -> event::Status {
         match event {
-            Event::Mouse(mouse::Event::ButtonPressed(mouse::Button::Left)) => {
+            Event::Mouse(mouse::Event::ButtonPressed(mouse::Button::Left))
+            | Event::Touch(Touch {
+                phase: touch::Phase::Started,
+                ..
+            }) => {
                 let is_clicked = layout.bounds().contains(cursor_position);
 
                 self.state.is_focused = is_clicked;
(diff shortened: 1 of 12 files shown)


wgpu/Cargo.toml
```diff
@@ -15,5 +15,5 @@ wgpu_glyph = { version = "0.5", git = "https://github.com/hecrj/wgpu_glyph", bra
 raw-window-handle = "0.3"
 image = "0.22"
 glam = "0.8"
-font-kit = "0.4"
+font-kit = { git = "https://github.com/simlay/font-kit", branch = "add-ios-support" }
```
hecrj: This is waiting for https://github.com/servo/font-kit/pull/110 to get merged.

winit/src/application.rs
```diff
@@ -143,39 +143,6 @@ pub trait Application {
...
-                    window.set_cursor_icon(conversion::mouse_cursor(
-                        new_mouse_cursor,
-                    ));
-
-                    mouse_cursor = new_mouse_cursor;
-                }
-
-                // TODO: Handle animations!
-                // Maybe we can use `ControlFlow::WaitUntil` for this.
-            }
```
hecrj: We should be able to revert this once https://github.com/rust-windowing/winit/issues/1082 is resolved.

cryptoquick:
It looks like a lot of progress has been made since last year, how's this looking?

hecrj:
Now that #181 landed, I think we can get this ready for merging.

@simlay We should rebase and remove all the merge commits. Let me know if you need any help. I could also take it to the finish line from here.

simlay:
@hecrj Well, I squashed the commits (it was a pretty brutal rebase) but it seems that https://github.com/servo/font-kit/issues/115 font-kit has moved to using pathfinder_geometry to draw stuff which breaks the build for aarch64-apple-ios (all modern iPhones).

Looks like there's a bunch of issues similar to:
```no method named `min` found for type `pathfinder_simd::scalar::I32x2` in the current scope```

I might submit a PR to the pathfinding repo if it's not too hard.

I haven't been keeping up with all the widgets this project has added since I opened the PR so I'm not sure which widgets don't have touch support.

File: native/src/widget/text_input.rs [hunk 3]
```diff
@@ -202,7 +202,10 @@ where
             Event::Mouse(mouse::Event::Input {
                 button: mouse::Button::Left,
                 state: ButtonState::Pressed,
-            }) => {
+            })
+            | Event::Touch(touch::Touch::Started { .. }) => {
+                self.state.is_focused =
+                    layout.bounds().contains(cursor_position);
```
hecrj: This statement shouldn't be necessary.

winit/Cargo.toml
```diff
@@ -22,8 +22,9 @@ version = "0.1.0-alpha"
 path = "../native"
 
 [dependencies.window_clipboard]
-git = "https://github.com/hecrj/window_clipboard"
-rev = "22c6dd6c04cd05d528029b50a30c56417cd4bebf"
+git = "https://github.com/simlay/window_clipboard"
+branch = "ios-unimplemented-stub"
+#rev = "22c6dd6c04cd05d528029b50a30c56417cd4bebf"
```
hecrj: I just merged https://github.com/hecrj/window_clipboard/pull/1, so we should be able to depend on my fork again.

simlay:
@hecrj This good to merge?

matthieugayon:
Does that PR also add IOS support for iced ? 
Might be totally unrelated, but is that project relevant for that PR ? And for IOS support in general ?  
https://github.com/BrainiumLLC/cargo-mobile

hecrj:
I think I've brought this up to speed.

I have ended up turning the `Touch` struct back into an enum in 3bdf931925067acbaabf040f6c437a54640ed1a0, given that this is how we handle mouse events now.

Proper support for the latest widgets is missing, but I imagine we can add that in further PRs. For instance, #650 seems to be adding touch support to the `PaneGrid` and the `PickList`. It may be interesting to rebase that PR on top of this one, once we get this merged.

As always, let me know what you think!

hecrj (approved):
I think we can let this ship sail! We can iron out any potential issues later.

Thank you everyone! :tada:

mtsr:
Great news!

PR #62, opened 2019-11-17
Async actions

### LINKED ISSUE #28: Async actions

hecrj:
Closes #28.

This PR adds first-class support for asynchronous actions (i.e. non-blocking background work) leveraging the recently stabilized `futures`! :tada:

A `Command` can be handed to the runtime for execution. As the futures comprising it finish, the produced messages will be handed back to `Application::update`.

The `todos` example has been improved to showcase this feature. Now, it automatically saves and loads its state using `serde_json`.

### Added
- `Command<T>`, which represents a batch of `futures` producing a result of type `T`.
- `Application::new`, which must return a `(Self, Command<Message>)`. This allows applications to perform an async action at startup.

### Changed
- `Application::update` now must return a `Command<Message>`. `Command::none` can be used if no async action is needed.
- `Application::run` is a static method now, thanks to the new `Application::new` method.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,9 @@ iced_web = { version = "0.1.0-alpha", path = "web" }
 
 [dev-dependencies]
 env_logger = "0.7"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+directories = "2.0"
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 wasm-bindgen = "0.2.51"
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -6,3 +6,10 @@ edition = "2018"
 description = "The essential concepts of Iced"
 license = "MIT"
 repository = "https://github.com/hecrj/iced"
+
+[features]
+# Exposes a future-based `Command` type
+command = ["futures"]
+
+[dependencies]
+futures = { version = "0.3", optional = true }
--- a/core/src/command.rs
+++ b/core/src/command.rs
@@ -1,0 +1,49 @@
+use futures::future::{BoxFuture, Future, FutureExt};
+
+pub struct Command<T> {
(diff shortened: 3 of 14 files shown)


daxpedda:
I see that you used `ThreadPool` as the executor. Would you consider making this only the default but allowing to pass a different executor with the help of [`futures::task::Spawn`](https://docs.rs/futures/0.3.1/futures/task/trait.Spawn.html)?

That way a user can implement his own executor or use libraries that support `Spawn`, hopefully tokio or async-std at some point.

hecrj:
Hey @daxpedda, thanks for taking a look!

> That way a user can implement his own executor or use libraries that support Spawn, hopefully tokio or async-std at some point.

What would be the use cases for this?

hecrj:
@daxpedda I have created a new issue referencing your comment in #63. We can keep discussing things there!

PR #68, opened 2019-11-25
Allow `Application` configuration with `Settings`

hecrj:
This PR allows basic configuration (just the window size, for now) of an `Application` using the new `Settings` type.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -1,0 +1,43 @@
+//! Configure your application.
+
+/// The settings of an application.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct Settings {
+    /// The [`Window`] settings.
+    ///
+    /// They will be ignored on the Web.
+    ///
+    /// [`Window`]: struct.Window.html
+    pub window: Window,
+}
+
+/// The window settings of an application.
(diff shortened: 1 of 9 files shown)


src/settings.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/// The window settings of an application.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Window {
+    /// The size of the window.
+    pub size: (u32, u32),
+
+    /// Whether the window should be resizable or not.
+    pub resizable: bool,
+}
```
ab0v3g4me: Adding the `title` field and `fullscreen` field would be nice as well. But the user should be able to dynamically change the size.

hecrj: The title is meant to be dynamic, `Application::title` already addresses that.

I think we should research a bit before adding fullscreen support. I rarely use GUI apps that start in fullscreen mode by default.

I am also not sure if making the size dynamic, like the title, is a good idea. Applications resizing themselves can cause bad UX. I'd like to gather some use cases first.

ab0v3g4me: Wanted to say that the user should be able to switch between normal windows size, to maximized, or to fullscreen. I also remember reading some solutions that recommended making the window size a bit bigger than the screen, hence the fullscreen.

PR #83, opened 2019-11-28
render gradient background

debris:
hey, this is my attempt to implement gradient background for `Quad`s

Screenshot 2019-11-28 at 18 00 00

At first, I need to admit that I know absolutely **nothing** about graphics programming and today it was a first time ever I played with shaders.

That said, I had a lot of fun playing with `iced` and learning new things. This branch is working, but the code is quite messy. Yet, before improving it, I would like to get some feedback on my approach to the problem. 

1. How is gradient rendering done?

    So this implementation is heavily inspired by [this stackoverflow](https://stackoverflow.com/a/47379071/2699089) answer.

    Rendering consists of the following steps:
    - creating a dummy texture with all colors and all sequences of colors (aka. gradients) - `ColorCache`
    - binding the texture to all the quads that are being rendered
    - fragment shader reading the color from the texture

2. What can be improved (apart from the quality)?

    I know that `iced` is heavily inspired by elm. I've never used `Elm`, but according to the [docs](https://package.elm-lang.org/packages/mdgriffith/elm-ui/latest/Element-Background) gradient background has a second param - `angle`. If this pr is ok, I'll add an `angle` param to it


hecrj:
Hey @debris,

> How is gradient rendering done?

To be completely honest, I am not sure. There are probably many valid strategies for this, each one with its own tradeoffs.

Initially, my idea was to support bicolor gradients (which I think are the most common ones) in `Background` by adding another `Color` in `Quad` and changing the fragment shader appropriately.

Now, using a texture as a color lookup table is an interesting strategy. It seems it could reduce instance size considerably when dealing with many quads.

> What can be improved (apart from the quality)?

I think we should try to keep things simple for now. As you noticed, the changes are quite involved. Generating and maintaining a color cache seems a bit overkill at this point. I'd keep the code simple for now and only implement bicolor gradient support without lookup tables.

At the same time, we can start researching and planning potential optimizations here. I have a feeling this will be related with other parts of the renderer in the future (think about image backgrounds). I'd love it if anyone could take the lead here.

Research and design are the hard parts. We can start a discussion on [Zulip](https://iced.zulipchat.com/) or open an issue. Of course, if anyone else has experience or ideas for this, feel free to chime in!

debris:
> I think we should try to keep things simple for now. As you noticed, the changes are quite involved. Generating and maintaining a color cache seems a bit overkill at this point. I'd keep the code simple and implement bicolor gradient support without lookup tables for the time being.

Agreed. I'll simplify the code and open a new pr. 

> Research and design are the hard parts. We can start a discussion on Zulip or open an issue. Of course, if anyone else has experience or ideas for this, feel free to chime in!

I'll hit you up over there next week. :)

PR #90, opened 2019-11-30
Load images from memory and image viewer example

### LINKED ISSUE #76: Image::from_bytes factory
Hello!
I think a good idea, to have **Image::from_bytes** factory. This will allow to include the image bytes in an executable binary or load it manually.

Example:
```rust
use iced::{Image, image::ImageFormat};
const MY_IMAGE: &'static [u8] = include_bytes!("resources/img.png");

fn get_img() -> Image {
  Image::from_bytes(MY_IMAGE, ImageFormat::PNG)
}
```

hecrj:
Closes #76.

[Image viewer - Iced](https://gfycat.com/qualifiedacidichamadryad-rust-gui)

This PR implements an `image::Handle` type which allows displaying images from data living in-memory and from the filesystem as well. Additionally, the PR adds basic image cache trimming to `iced_wgpu`.

An image viewer example has been built to showcase a simple way to load remote images asynchronously.

### Added
- `image::Handle` type with `from_path` and `from_memory` methods.
- `image::Data` enum representing different kinds of image data.
- `Color::from_rgb8` to easily build a `Color` from its hexadecimal representation.

### Changed
- `Image::new` takes an `Into<image::Handle>` now instead of an `Into<String>`.

### Fixed
- `Image` widget not keeping aspect ratio consistently.


ejmahler:
1. I tried to compile this branch, via the following cargo.toml entry:

`iced = { git = "https://github.com/hecrj/iced", branch = "feature/image-from-bytes" }`

 and got this compile error:

```
error: failed to select a version for `libc`.
    ... required by package `winit v0.20.0-alpha4 (https://github.com/rust-windowing/winit?rev=709808eb4e69044705fcb214bcc30556db761405#709808eb)`
    ... which is depended on by `iced_winit v0.1.0-alpha (https://github.com/hecrj/iced?branch=feature/image-from-bytes#1747eb27)`
    ... which is depended on by `iced v0.1.0-beta (https://github.com/hecrj/iced?branch=feature/image-from-bytes#1747eb27)`
    ... which is depended on by `sandpile_fractal v0.1.0 (D:\Documents\GitHub\SandpileFractal)`
versions that meet the requirements `^0.2.64` are: 0.2.66, 0.2.65, 0.2.64

all possible versions conflict with previously selected packages.

  previously selected package `libc v0.2.33`
    ... which is depended on by `cocoa v0.19.1`
    ... which is depended on by `gfx-backend-metal v0.4.1`
    ... which is depended on by `wgpu-native v0.4.1`
    ... which is depended on by `wgpu v0.4.0`
    ... which is depended on by `iced_wgpu v0.1.0 (https://github.com/hecrj/iced?branch=feature/image-from-bytes#1747eb27)`
    ... which is depended on by `iced v0.1.0-beta (https://github.com/hecrj/iced?branch=feature/image-from-bytes#1747eb27)`
    ... which is depended on by `sandpile_fractal v0.1.0 (D:\Documents\GitHub\SandpileFractal)`

failed to select a version for `libc` which could resolve this conflict
```

Is this something I need to fix in my end? I'd like to test this out, but I currently can't.

2. From looking at the source, it looks like you're using the image-rs crate under the hood. Are you opposed to just doing a public export of the image crate? I think that will be vastly more flexible than forwarding to `load_from_memory`, and I suspect that if you don't, you're gong to end up recreating image-rs's public API to cover all the different ways they let users create images. It will also let you sidestep the problem of error handling for image formats, instead letting image-rs handle parsing errors.

If you help me out to get this compiling and give me a week or so to get a demo working, I'd be glad to write a version of this that directly passes in an image.

hecrj:
> From looking at the source, it looks like you're using the image-rs crate under the hood. Are you opposed to just doing a public export of the image crate?

One of the main focus of Iced is simplicity and I think the `image` crate can be quite daunting, especially if you are a beginner.

Additionally, we have to consider that this is a dependency of a specific renderer (which turns out to be the default) and Iced also supports the `web`. We need to control what do we expose and where.

For now, I'd like to gather use cases and keep things simple.

ejmahler:
That's a reasonable approach.

There is one suggestion I have if you're planning to keep this API: right now `from_bytes` doesn't convey what format it expects the data to be in, but it also doesn't return a Result<> to let you know that you passed in an invalid format -- it just fails silently if you don't pass in the correct format. One way to fix this is to rename `from_bytes` to `decode_from_bytes` or `parse_from_bytes`, which would convey the fact that we expect the image data to be in some sort of encoded image format.

hecrj:
@ejmahler I think we could call it `image::Handle::from_memory`. I think we could also have an `image::Handle::from_rgba_pixmap` or similar.

However, we cannot return a `Result` here because `iced_native` does not really perform any decoding/parsing. We would have to move the `image` dependency to `iced_native` and expose an API useful enough for renderers to avoid duplicating work (i.e. parsing the `Image` again).

For now, an `Image` renderer implementation needs to be fault-tolerant. Just like you can provide a path to a file that does not exist, you can also provide invalid image data.

It's true that, in this case, `iced_wgpu` currently simply fails silently and does not show anything, but in the near future we should show an indicator that an image failed to load.

ejmahler:
I was able to get iced working, and I made a proof of concept of this feature. check it out if you're curious:

https://github.com/ejmahler/SandpileFractal

You can see the actual rendering code where it converts an image buffer to an image handle here: https://github.com/ejmahler/SandpileFractal/blob/master/src/render.rs#L71

In this particular case, it would be more convenient to pass the uncompressed RGB data directly through to iced, but it doesn't look like it has any kind of performance impact. Or if there is one, it hidden by the fact that the work is being done in a background thread, so the UI keeps feeling responsive.

Screenshot:
image

clarkmoody:
@ejmahler Amazing!

hecrj:
@ejmahler That's awesome! I wonder if writing the image using BMP format would speed things up.

I guess we can avoid implementing pixmap methods for the time being, as advanced users like you can leverage the `image` crate like this. I think this satisfies the same use cases as your suggestion of exposing the `image` crate directly, but with a performance cost.

I don't think we should focus too much on performance yet, so I am okay with this for now.

PR #92, opened 2019-11-30
Add `From` impls for `Length`, `Point`, and `Size`

nvzqz:
With these, it might make sense to start accepting `Into<Length>`, `Into<Point>`, and `Into<Size>` as function arguments for simplicity.


hecrj:
Nice! Thank you. :clinking_glasses:

PR #93, opened 2019-11-30
render borders using shader

debris:
This pr enables quad shader to render borders. This enables `iced` to easily add border param to all elements (e.g. `Button`) and makes the border more smooth than 2 quads rendered on top of each other.

e.g. 

before:

Screenshot 2019-11-30 at 11 07 39

now:

Screenshot 2019-11-30 at 11 07 13


hecrj:
This is great, @debris. Beside improving quality, it should reduce the amount of primitives too. Thank you!

I have noticed the border color was affecting the background color when the border width was `0`. I have been able to fix it, but I am not sure it's the best way to do it. I know GPUs do not like branching.

Please, take a look and let me know!

debris:
> I have noticed the border color was affecting the background color when the border width was `0`. I have been able to fix it, but I am not sure it's the best way to do it. I know GPUs do not like branching.

good catch! but it's easily solvable even without branching :)

debris:
I forced push to this branch, I hope you don't mind. 

changes overwritten by the push

```frag
    vec4 c = bg_color;
    float alpha = d > 0 ? clamp(radius + s - d, 0, 1) : 1;

    if(frame_width > 0) {
        float inner_radius = radius - frame_width;
        float mix_factor = smoothstep(inner_radius - s, inner_radius + s, d);
        c = mix(c, frame_color, mix_factor);
    }

    return vec4(c.xyz, c.w * alpha);
```

```frag
    float inner_radius = radius - frame_width;
    float alpha = 1.0 - smoothstep(radius - s, radius + s, d);
    float mix_factor = smoothstep(inner_radius - s, inner_radius + s, d);
    vec4 c = mix(bg_color, frame_color, min(frame_width, mix_factor));
    return vec4(c.xyz, c.w * alpha);
```

debris:
Ah, I just realised that `border_radius` affects the rendering in a similar way. Fix on the way :)

hecrj:
Oh, I see. I thought `min` and `max` also performed branching, but I guess they are way easier to optimize by the compiler.

> Ah, I just realised that border_radius affects the rendering in a similar way. Fix on the way :)

Yeah, I think this is currently a bug present in `master` :sweat_smile:

debris:
I think it should be fine now

PR #94, opened 2019-11-30
Improve window setting

hatoo:
Hi, I'm interested in running iced in [vst-rs](https://github.com/RustAudio/vst-rs) environment.

For this, there are 2 things to do.
1. (Windows only) Initialize a Window with the given parent window and with other settings.
2. Run event loop discontinuously. The current `run` method will not return until close. This behavior is not undesirable.

This PR handles 1.
To create a window in VST, we must initialize `winit` like below.
```
let wb = winit::window::WindowBuilder::new()
            .with_title("A fantastic window!")
            .with_decorations(false)
            .with_resizable(false)
            .with_parent_window(parent)
            .with_inner_size(LogicalSize {
                width: 400.0,
                height: 300.0,
            });
```

To do the same thing in `iced`, we need support setting options `with_decorations` and `with_parent_window`.
This is what is the PR does.

For `with_decorations`, it's just straightforward to implement because it is supported in all platforms in `winit`.
For `with_parent_window` it's not straightforward since `with_parent_window` is Windows-specific API. 
So I introduced a simple platform-specific module to implement.


src/settings/mod.rs
```diff
@@ -1,5 +1,15 @@
...
 
+#[cfg(target_os = "windows")]
+#[path = "windows.rs"]
+pub mod platform;
+
+#[cfg(not(target_os = "windows"))]
+#[path = "not_windows.rs"]
+pub mod platform;
+
+pub use platform::PlatformSpecific;
```
hecrj: I would only expose `PlatformSpecific` settings in `iced_winit` for the time being.

hatoo: Sure.
I made iced's platform-specific settings use iced_winit's `PlatformSettings`.

Cargo.toml
```diff
@@ -34,6 +34,9 @@ iced_wgpu = { version = "0.1.0", path = "wgpu" }
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 iced_web = { version = "0.1.0", path = "web" }
 
+[target.'cfg(target_os = "windows")'.dependencies.winapi]
+version = "0.3.6"
```
hecrj: I think this should not be necessary now.

src/settings.rs
```diff
@@ -19,13 +19,21 @@ pub struct Window {
 
     /// Whether the window should be resizable or not.
     pub resizable: bool,
+
+    /// Whether the window should have a border, a title bar, etc.
+    pub decorations: bool,
+
+    /// Platform specific Setting.
+    pub platform_specific: iced_winit::settings::PlatformSpecific,
```
hecrj: The idea is to simply not expose this for now and make advanced users use `iced_winit` directly instead.

src/settings.rs
```diff
@@ -37,6 +45,8 @@ impl From<Settings> for iced_winit::Settings {
             window: iced_winit::settings::Window {
                 size: settings.window.size,
                 resizable: settings.window.resizable,
+                decorations: settings.window.decorations,
+                platform_specific: settings.window.platform_specific,
```
hecrj: Then we'll simply use `iced_winit::settings::PlatformSpecific::default()` here.

Cargo.toml
hatoo: Oops, I forgot to remove it.
I've removed now.

src/settings.rs
hatoo: Make sense!
I've removed extended settings from iced's setting.

hatoo:
Thanks for your work!
It's OK!
Please merge this!

PR #95, opened 2019-11-30
Make scrolling behave like you'd expect it to

Friz64:
This PR adds proper scrolling logic to the `scrollable` widget. 
Here i define a "floater" to be the element that you can drag around with your mouse to scroll.
The new behaviour is the following:
- if the user clicks off the floater, the floater jumps and attaches centered around the cursor
- if the user clicks on the floater, the floater stays where it is and attaches to the cursor position

### DIFF
--- a/native/src/widget/scrollable.rs
+++ b/native/src/widget/scrollable.rs
@@ -102,6 +102,17 @@ impl<'a, Message, Renderer> Scrollable<'a, Message, Renderer> {
     }
 }
 
+fn scroll_percentage(
+    bounds: Rectangle,
+    scroller_bounds: Rectangle,
+    scroller_grabbed_at: f32,
+    cursor_position: Point,
+) -> f32 {
+    (cursor_position.y + bounds.y
+        - scroller_bounds.height * scroller_grabbed_at)
+        / (bounds.height - scroller_bounds.height)
+}
+
 impl<'a, Message, Renderer> Widget<Message, Renderer>
     for Scrollable<'a, Message, Renderer>
 where
@@ -195,12 +206,13 @@ where
                             }
                         };
 
-                        let scroll_percentage = (cursor_position.y + bounds.y
-                            - scroller_bounds.height * scroller_grabbed_at)
-                            / (bounds.height - scroller_bounds.height);
-
                         self.state.scroll_to(
-                            scroll_percentage,
+                            scroll_percentage(
+                                bounds,
+                                scroller_bounds,
+                                scroller_grabbed_at,
+                                cursor_position,
+                            ),
                             bounds,
                             content_bounds,
                         );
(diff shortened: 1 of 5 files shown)


Friz64:
Ohhh, i completely forgot about the null render! Lemme fix this really quickly

wgpu/src/renderer/widget/scrollable.rs
```diff
@@ -16,15 +17,57 @@ fn scrollbar_bounds(bounds: Rectangle) -> Rectangle {
...
+        bounds: Rectangle,
+        content_bounds: Rectangle,
+        offset: u32,
+    ) -> (Rectangle, Rectangle) {
+        let background_bounds = background_bounds(bounds);
+        let scroller_bounds =
+            scroller_bounds(bounds, content_bounds, background_bounds, offset);
+
+        (background_bounds, scroller_bounds)
+    }
```
hecrj: It may be a good idea to split this method into two: `scrollbar_bounds` and `scroller_bounds`.

Maybe we could calculate the scroller bounds in `iced_native` from the `scrollbar_bounds`? It shoud simplify the trait a bunch.

wgpu/src/renderer/widget/scrollable.rs
```diff
@@ -16,15 +17,57 @@ fn scrollbar_bounds(bounds: Rectangle) -> Rectangle {
...
+            && background_bounds.contains(cursor_position)
+        {
+            Some(if scroller_bounds.contains(cursor_position) {
+                ScrollbarGrab::Scroller
+            } else {
+                ScrollbarGrab::Background
+            })
+        } else {
+            None
+        }
```
hecrj: I think this method could also live in `iced_native` entirely. There is nothing renderer-specific here, I think.

This should also allow us to not expose the `ScrollbarGrab` type in `iced_native`.

native/src/widget/scrollable.rs
```diff
@@ -357,11 +393,20 @@ impl State {
     }
 
     /// Returns whether the scrollbar is currently grabbed or not.
-    pub fn is_scrollbar_grabbed(&self) -> bool {
-        self.scrollbar_grabbed_at.is_some()
+    pub fn currently_grabbed(&self) -> bool {
```
hecrj: Maybe `is_scroller_grabbed` is a better name?

native/src/widget/scrollable.rs
```diff
@@ -357,11 +393,20 @@ impl State {
...
 }
 
+#[derive(Debug, Clone, Copy)]
+/// What the mouse is grabbing on the scrollbar
+pub enum ScrollbarGrab {
+    /// The mouse is grabbing the background
+    Background,
+    /// The mouse is grabbing the scroller
+    Scroller,
+}
```
hecrj: I think this type describes a part of the scrollbar. It seems you can have one even when the user isn't grabbing anything.

I think naming it someting like `ScrollbarItem` and then implementing a `ScrollbarItem::from_cursor_position` (the current `Renderer::scrollbar_grab`) would keep it opaque, more intuitive, and safer to use.

hecrj:
Oh, I forgot to mention that it'd be great if you could squash the `WIP` commits too into one with a more descriptive message.

wgpu/src/renderer/widget/scrollable.rs
Friz64: > Although, maybe we could calculate the scroller bounds in `iced_native` from the `scrollbar_bounds`? It shoud simplify the trait a bunch.

In order to calculate the scroller bounds in `iced_native` we would need to know the `SCROLLBAR_MARGIN` which is defined in the renderer.

Friz64:
Thank you for those changes, the idea of a `Scrollbar` type greatly improves the code quality! I just corrected a little oversight in the documentation, it now looks good to me, are we ready to merge this?

hecrj:
Awesome! Yes, let's merge this! :tada:

PR #101, opened 2019-12-01
Added path to image hash so it updates when path changes

### LINKED ISSUE #100: Image widget not updating size when new image is loaded
When providing a new path for an image, the image widget will load the picture, but the size will stay the same as the previous image until changing the window size as seen below:

Before
2019-12-01_15:31:52_1020x791

After
2019-12-01_15:32:03_1025x792

After digging around a bit, it seems to be an issue with the way the image widget is cached. Simply adding `self.path.hash(state)` to the `hash_layout` method like below fixes it.
```
fn hash_layout(&self, state: &mut Hasher) {
    self.path.hash(state);
    self.width.hash(state);
    self.height.hash(state);
}
```

Seems like a trivial fix to me, so if it looks good to you, I can throw up a PR.

Shootertrex:
Fixes #100. Image widget should now update its size when given a new image path.

### DIFF
--- a/native/src/widget/image.rs
+++ b/native/src/widget/image.rs
@@ -100,6 +100,7 @@ where
     }
 
     fn hash_layout(&self, state: &mut Hasher) {
+        self.path.hash(state);
         self.width.hash(state);
         self.height.hash(state);
     }


hecrj:
Awesome. Thank you! :tada:

PR #102, opened 2019-12-02
Implemented preliminary styling for native widgets.

Maldela:
I was playing around with styling the different iced examples and this is what I came up with. 
To make this work I had to extend some of the widgets and change the view function of the Application and Sandbox traits to supply an ApplicationStyle object which can be used to construct widgets. This work is by far not finished yet. My next steps would be adding the possibility to load platform specific plugins to have a native look across different platforms and splitting the ApplicationStyle object into some sort of color palette to theme custom widgets and a set of styles for the built in widgets.
Consider this more a RFC then a PR. I just wanted to get some feedback about my approach so far.


hecrj:
Hey @Maldela, thank you for exploring this! I really appreciate the effort here!

That said, I do not think we are yet ready to write code to implement this feature. There is a lot to consider, discuss, and design. As the [contributing guide] states, please get in touch! I'd love it if you joined [our Zulip server] and we started a conversation there. I really think code will be the easy part here.

Your current approach is a very direct way of implementing support for styling, but it ties style properties to each particular widget in `iced_native`, which is meant to be renderer-agnostic. I am aware that the current `iced_native` implementation has the same issue, but a well-designed styling solution should try to address this issue.

[contributing guide]: https://github.com/hecrj/iced/blob/master/CONTRIBUTING.md
[our Zulip server]: https://iced.zulipchat.com/

PR #106, opened 2019-12-04
Use cfg and path attribute instead of cfg_attr for IntelliJ Rust support

Gohla:
PR for IntelliJ Rust plugin support by replacing `cfg_attr` with `cfg` and `path`, as per https://github.com/hecrj/iced/issues/104


hecrj:
Awesome! Thank you! :tada:

PR #107, opened 2019-12-04
Implemented `From<Color` for `Background`

Maldela:
A small change that makes creating a `Background` from a `Color` easier.


hecrj:
Cool! Thank you! :clinking_glasses:

PR #108, opened 2019-12-05
Improve text input behavior

### LINKED ISSUE #105: Keyboard Shortcuts in Text Boxes
Would be great to have Ctrl+Arrows jump words, and Shift+Arrows select text.

hecrj:
Fixes #105.

[Text input behavior](https://gfycat.com/fakeuntriedgalah)

This PR implements click-based cursor positioning in `TextInput`, using a binary search with the text measure function. Additionally, it adds supports for the `Home` and `End` keys.

### DIFF
--- a/native/CHANGELOG.md
+++ b/native/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `image::Handle` type with `from_path` and `from_memory` methods. [#90]
 - `image::Data` enum representing different kinds of image data. [#90]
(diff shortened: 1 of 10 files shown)


artursapek:
nice

Friz64:
You should also be able to use CTRL + {Right,Left} to jump entire words.

hecrj:
@Friz64 I was meaning to implement that in another PR, but I just went ahead and added it here.

I have also made the jump and started using [`unicode-segmentation`](https://github.com/unicode-rs/unicode-segmentation) to properly account for grapheme clusters and word boundaries. The implementation is still a bit naive, but good enough for now.

PR #111, opened 2019-12-06
Svg and icon support

Maldela:
This is the first step for proper icon support.
As you requested, I changed the `resvg` backend to `raqoste`. Unforunately this adds some code as the `raqoste` crate does not have a function to write a png encoded image to memory. (See the last commit)
 I'll open a feature request about that with them.


Maldela:
I opened an [issue](https://github.com/jrmuizel/raqote/issues/95) about the missing functionality.

Maldela:
Yes there are definitely limitations in the current implementation. I'll look into creating a new primitive and maybe use gpu rendering for that. Could take me a while, though.

Maldela:
I implemented a new `Svg` primitive that handles svg files exclusively. It shares a lot of code with the `Image` primitive. It uses `resvg` and `raqote` to create a texture with the size of the viewport and draws that texture like the `Image` primitive does.
I also implemented a theme icon loader and made it a global static with `lazy_static`. Let me know what you think about that.

Maldela:
Icon is implemented in a way that it does create a `Svg` primitive, if its file path ends in .svg or .svgz. Otherwise it tries to create an `Image` primitive. I wanted the `Icon` widget to be able to handle svg and png files.
The `Svg` primitive is now fully viewport aware. The viewport size is handled in its draw and upload functions.
The icon loader being a lazy_static was quick and dirty way of getting it working. I can scrap that for now.

wgpu/src/svg.rs
```diff
@@ -0,0 +1,604 @@
...
+            if !handle.path.is_file() {
+                let mem = Memory::NotFound;
+
+                let _ = self.cache.borrow_mut().insert(&handle, mem);
+            }
+
+            let mut opt = resvg::Options::default();
+            opt.usvg.path = Some(handle.path.clone());
+            opt.usvg.dpi = handle.dpi as f64;
+            opt.usvg.font_size = handle.font_size as f64;
```
hecrj: Could we use `DrawTarget::set_transform` to set the DPI scaling in `upload`? Everything should scale nicely and we will avoid the weird `96.0` and `20.0` magic constants.

native/src/widget/icon.rs
```diff
@@ -0,0 +1,100 @@
...
+    hash::Hash,
+    path::{Path, PathBuf},
+};
+
+/// A simple icon_loader widget.
+#[derive(Debug, Clone)]
+pub struct Icon {
+    path: PathBuf,
+    size: Length,
+}
```
hecrj: I would name this widget `Svg` and let it be configurable both in width and height.

wgpu/src/svg.rs
Maldela: Those constants are for usvg which takes them to create the node tree. I'm not entirely sure how they affect the image creation but I think they influence how `<text>` tags in the svg file are processed.

native/src/widget/icon.rs
hecrj: After thinking a bit about it, it may make sense to merge this with `Image` as we had it before. We will probably want to keep the SVG aspect ratio after all.

We can check the extension as you were doing and have an additional `Memory` variant in `image` to deal with caching aware of DPI and viewport dimensions.

wgpu/src/svg.rs
hecrj: The idea is to leave everything in default settings and simply apply a uniform scaling to everything.

native/src/widget/icon.rs
Maldela: As I said, that widget can handle svg as well as png or other files since some icons still come in png format. So I think naming it `Svg` would be rather misleading about its capabilities.
I also thought that since icons are usually squares, giving it width and height would be unnecessary and a single size attribute simpler.

Maldela: I was thinking about the image widget too. I would like to give the user more control about aspect ratio of the widget. Maybe an enum with `KeepAspectRatio`, `Stretch` and `KeepAspectRatioCrop` at least. But icons don't need that. 
Also having a separate widget makes integrating themed icons easier in the future.

hecrj: What an icon is depends on each particular application.

To keep things simple, I was proposing that we should instead expose a widget that is able to render only SVG with configurable dimensions. One that can be used to render SVG directly, not just icons.

This way, you could easily create a function `icon` in your application that chooses `Svg` or an `Image` widget based on the path, sets square dimensions, and returns an `Element`.

wgpu/src/svg.rs
Maldela: Usvg does not use the dpi setting to create bigger images. It's just for font rendering inside the icon.

hecrj: Right, I don't think we should alter that setting. Applying a transformation to the `DrawTarget` should be enough.

Maldela: Sure, I can change it to keep the standard settings. I think this is a corner case if one at all anyway.

native/src/widget/icon.rs
Maldela: Now the `Icon` widget does all that for you. You don't have to create an `icon` function yourself.
It's very easy to use.

hecrj: Convenient, but not simple.

If our renderer supports SVG, we need a widget that can leverage it to its full extent, not just squares.

Maldela: I could easily merge the svg capability into `Image`. If simplicity is the main goal here, I could try to merge the two pipelines too. That shouldn't be too hard.

hecrj: I am not completely sure about it.

I like the idea of having a dedicated `Svg` widget. It makes everything more explicit and forces users to keep in mind `Svg` and `Image` have different performance implications. It may also help us in the long run to feature-gate some particular widgets and reduce binary size. I'd say let's keep it as different widgets for now.

However, I believe we can reuse the `image` pipeline by adding a new variant to the `Memory` enum to handle the `Svg` primitive. This should reduce the scope of the changes too.

For instance:

```rust
enum Memory {
    Host(HostMemory),
    Device {
        bind_group: Rc<wgpu::BindGroup>,
        width: u32,
        height: u32,
    },
    NotFound,
    Invalid,
}

enum HostMemory {
    Image(image::ImageBuffer<image::Bgra<u8>, Vec<u8>>),
    Svg { /* ... */ },
}
```

Maldela:
Good catch with the missing rerasterization. I would have liked the svg feature to be included in the default features but it displaying a placeholder text works too.
I think this can be merged.

Maldela:
Regarding the rasterization being slow on resize: Maybe we could rasterize the svg to a slightly too large viewport and downscale. Then only rerasterize, if the resized svg becomes larger than the viewport.

hecrj:
> I would have liked the svg feature to be included in the default features but it displaying a placeholder text works too.

Yeah, I know. However, the `resvg` dependency was increasing around 1MB the binary size in release mode. I think it's fine behind a feature gate. It can be easily enabled if you need it.

It does not display a placeholder text when disabled, that happens because I added some conditional compilation to the `svg` example. If you try to use an `Svg` widget without the `svg` feature enabled, you will get a compiler error telling you that the `wgpu` renderer does not implement `svg::Renderer`.

> Regarding the rasterization being slow on resize: Maybe we could rasterize the svg to a slightly too large viewport and downscale. Then only rerasterize, if the resized svg becomes larger than the viewport.

It may be worth a try. Downscaling can make things look wrong though, so we should keep that in mind. The cool thing about the current design is that any change in the caching strategy should only affect the `image::vector` module. In any case, I think we should explore/tackle this in another PR.

Maldela:
> In any case, I think we should explore/tackle this in another PR.
Agreed. This is fine for now.
We could also try to rasterize in another thread and display a blank texture or a texture that finished for another viewport size as long as the thread is running.

hecrj:
Thanks again for the patience and the work :clinking_glasses: 

Let's merge this!

PR #122, opened 2019-12-14
Event subscriptions

### LINKED ISSUE #29: Event subscriptions

### LINKED ISSUE #109: Allow defining global shortcuts
I would like to add keyboard shortcuts and it does not look like it's possible out of the box, without writing my own wrapping widget. 

IMO it does not seems out of scope of the "batteries included" aspect of iced.

### LINKED ISSUE #123: None of the async examples are actually async
There are multiple functions in the example applications which are marked as `async` but don't use `.await` at all, instead doing blocking I/O. This means that if there are more tasks running than there are threads in the thread pool, the executor will only run the ones that got there first, as they are blocking the worker threads. For example in [examples/pokedex.rs](https://github.com/hecrj/iced/blob/master/examples/pokedex.rs#L150) the `search` function is marked as async but makes HTTP requests using reqwest's sync API, which defeats the purpose of using async functions. In order to make use of the new async functions, they need to only use async I/O.

hecrj:
Closes #29, closes #109, and fixes #123.

<p align="center">
<a href="https://gfycat.com/granularenviousgoitered-rust-gui">Stopwatch - Iced</a>
</p>

This PR adds support for event subscriptions. Event subscriptions are inspired by [Elm's subscriptions](https://guide.elm-lang.org/effects/time.html) and can be used to listen to external events that do not necessarily occur as a response to some interaction or command. For instance, you can use event subscriptions to create a websocket connection, listen to incoming messages, and even handle reconnects graciously.

Event subscriptions are built on top of [`futures::stream::Stream`](https://docs.rs/futures/0.3.1/futures/stream/trait.Stream.html). The messages produced by a subscription are provided to `Application::update` and, of course, can be used to change the state of the `Application`.

The `iced_winit` shell keeps a subscription alive as long as the user returns it in the new `Application::subscription` method. Therefore, unsubscribing simply consists in not returning the subscription anymore.

Moreover, subscriptions can be created by libraries and/or end-users using the new `subscription::Recipe` trait (name subject to change), which requires just a couple of methods.

For now, there is only one built-in `Subscription` in `iced_native`: `subscription::events`. This subscription will notify an application of all the runtime events that occur, producing [`Event`](https://docs.rs/iced_native/0.1.0/iced_native/enum.Event.html) values. It can be useful if you want to listen to keyboard or mouse events (and most likely window events soon):

<p align="center">
<a href="https://gfycat.com/littlesilkyalabamamapturtle">Events - Iced</a>
</p>

Two examples have been added that showcase event subscriptions:

- `events`, which uses the new built-in `subscription::events`.
- `stopwatch`, which showcases a custom event subscription that produces messages periodically at a configurable time interval.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,16 +35,16 @@ iced_wgpu = { version = "0.1.0", path = "wgpu" }
 iced_web = { version = "0.1.0", path = "web" }
 
 [dev-dependencies]
+iced_native = { version = "0.1", path = "./native" }
+iced_wgpu = { version = "0.1", path = "./wgpu" }
 env_logger = "0.7"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 directories = "2.0"
-reqwest = "0.9"
-rand = "0.7"
-chrono = "0.4"
 futures = "0.3"
-iced_native = { version = "0.1", path = "./native" }
-iced_wgpu = { version = "0.1", path = "./wgpu" }
+reqwest = "0.9"
+async-std = { version = "1.3", features = ["unstable"] }
+rand = "0.7"
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 wasm-bindgen = "0.2.51"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,8 +42,8 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 directories = "2.0"
 futures = "0.3"
-reqwest = "0.9"
 async-std = { version = "1.3", features = ["unstable"] }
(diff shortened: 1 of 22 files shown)


clarkmoody:
Demo app works on my system, thanks!

Testing my reflexes now.
Screen Shot 2019-12-14 at 12 21 52 AM

PR #140, opened 2020-01-02
Mesh2D primitive for rendering arbitrary geometry in `iced_wgpu`

artursapek:
image


artursapek:
@hecrj those changes seem great to me. Thanks for taking it to the finish line.

PR #141, opened 2020-01-02
Progress bar widget

Songtronix:
This will add a very primitive progressbar:
image

What additionally can be done:
- [X] make it customizable by accepting colors.
- [ ] allow the use of a border
- [ ] optional label in the middle showing the percentage
-  [ ] make progressbar render on the web


native/src/widget/progressbar.rs
```diff
@@ -0,0 +1,166 @@
...
+/// ```
+/// # use iced_native::Progressbar;
+/// 
+/// let value = 50.0;
+/// Progressbar::new(0.0..=100.0, value);
+/// ```
+///
+/// ![Default Progressbar](https://user-images.githubusercontent.com/18618951/71662391-a316c200-2d51-11ea-9cef-52758cab85e3.png)
+#[allow(missing_debug_implementations)]
+pub struct Progressbar {
```
hecrj: I think we should name it `ProgressBar` and also change the example name to `progress_bar`.

native/src/widget/progressbar.rs
```diff
@@ -0,0 +1,166 @@
...
+/// Your [renderer] will need to implement this trait before being
+/// able to use a [`Progressbar`] in your user interface.
+///
+/// [`Progressbar`]: struct.Progressbar.html
+/// [renderer]: ../../renderer/index.html
+pub trait Renderer: crate::Renderer {
+    /// Returns the height of the [`Progressbar`].
+    ///
+    /// [`Progressbar`]: struct.Progressbar.html
+    fn height(&self) -> u32;
```
hecrj: Maybe we could make the `height` configurable at the widget level and make this a `default_height` instead.

native/src/widget/progressbar.rs
Songtronix: Sure, will change.

native/src/widget/progressbar.rs
Songtronix: :ok_hand:

native/src/widget/progressbar.rs
Songtronix: Should I rename the module ``progress_bar`` too?

hecrj: Yes!

Songtronix:
Hmm I can't get the tour to compile to wasm. Either my setup is messed up for some dependency isn't wasm compatible anymore.
```Rust
error[E0599]: no method named `as_sock` found for type `&std::net::TcpStream` in the current scope
   --> /home/songtronix/.cargo/registry/src/github.com-1ecc6299db9ec823/net2-0.2.33/src/ext.rs:855:25
    |
855 |         do_connect(self.as_sock(), addr)
    |                         ^^^^^^^ method not found in `&std::net::TcpStream`
    |
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following trait defines an item `as_sock`, perhaps you need to implement it:
            candidate #1: `ext::AsSock`
```

hecrj:
This is probably a dev dependency, maybe `surf`? Rust will try to compile all the dev dependencies even if they are not used by the example you are trying to build.

Although it would be great to use a cross-platform HTTP client, maybe we could make the dev dependency conditional based on the target for the time being.

Songtronix:
Okay by disabling almost all dev dependencies helped. I'll try to get the progressbar working on the web.

Songtronix:
Well implementing the progressbar for the web is harder than I thought, instead I've fixed the text_input for the web to support being a password field.

Songtronix:
Good idea to move the default height over to the render. I'm fine with merging it :ok_hand:

PR #144, opened 2020-01-06
Add MS paint example

### LINKED ISSUE #18: A MS paint clone as example
Would be a fun way to show how to do things with iced. :)

hatoo:
Since #140  has landed,
I realized that we can make MS paint like example by using lyon now.

This PR resolves #18


artursapek:
hah, awesome

examples/paint.rs
```diff
@@ -0,0 +1,275 @@
...
+                        &mut buffer,
+                        |pos: lyon::math::Point, _: StrokeAttributes| {
+                            Vertex2D {
+                                position: pos.to_array(),
+                                color: [0.0, 0.0, 0.0, 1.0],
+                            }
+                        },
+                    ),
+                )
+                .unwrap();
```
hecrj: The widget generates strokes every time the cursor is moved while dragging. It emulates a pencil. This means that strokes will add up very quickly and will be very close to each other.

I am not sure implementing a pencil like this is the best way to showcase how to combine the new `Mesh2D` primitive with `lyon`. My intution tells me that something like a buffer of pixels would work better for this particular case.

Maybe we could implement line/rectangle/circle tools instead?

hatoo: I agree with that "pencil" is not a very good example which demonstrates the combination of `iced` and `lyon`.
I choose a bezier tool as an example to demonstarate it.

hatoo:
Thanks! Codes became more decent.
Please go ahead!

PR #147, opened 2020-01-09
Update `winit` to `0.20`

hecrj:
This PR updates `winit` to `0.20` in `iced_winit`.

There is currently an issue on macOS where the `CursorMoved` event is reporting a logical coordinate instead of a physical one (https://github.com/rust-windowing/winit/issues/1371).

# Pending work
- [x] https://github.com/rust-windowing/winit/issues/1371
- [x] https://github.com/rust-windowing/winit/pull/1378

### DIFF
--- a/winit/src/application.rs
+++ b/winit/src/application.rs
@@ -116,10 +116,7 @@ pub trait Application: Sized {
 
             window_builder = window_builder
                 .with_title(&title)
-                .with_inner_size(winit::dpi::LogicalSize {
-                    width: f64::from(width),
-                    height: f64::from(height),
-                })
+                .with_inner_size(winit::dpi::LogicalSize { width, height })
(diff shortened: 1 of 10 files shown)


winit/src/application.rs
```diff
@@ -301,7 +300,8 @@ pub trait Application: Sized {
                 ..
             } => match window_event {
                 WindowEvent::CursorMoved { position, .. } => {
-                    let position = position.to_logical::<f64>(scale_factor);
+                    let position =
+                        position.to_logical::<f64>(size.scale_factor());
 
                     events.push(Event::Mouse(mouse::Event::CursorMoved {
                         x: position.x as f32,
```
Imberflur: Would it be desirable to mirror `winit` and use `f64` here?

hecrj:
Included in #181.

PR #154, opened 2020-01-12
Texture atlas

Maldela:
I finally came around to implementing the texture atlas for images. I'm using guillotiere to manage the atlas space. In the current implementation the atlas only grows if necessary. It never shrinks.
I also wrote a small build script that compiles shaders at compile time using shaderc. Manually compiling shaders during development was super annoying.
Let me know what you think.


wgpu/Cargo.toml
```diff
@@ -22,3 +23,8 @@ glam = "0.8"
 font-kit = "0.4"
 log = "0.4"
 resvg = { version = "0.8", features = ["raqote-backend"], optional = true }
+guillotiere = "0.4"
+debug_stub_derive = "0.3"
+
+[build-dependencies]
+shaderc = "0.6"
```
hecrj: I'd like to keep dependencies at bay. `shaderc` is quite a heavy dependency just to build some shaders. Forcing every contributor/user to compile them from source feels wrong.

I feel our current approach is better: recompile the shaders manually and version control the SPIR-V files. Of course, if you are working on shaders heavily, you are free to create a local `Makefile` or whatever adapts well to your workflow in order to increase productivity.

We should also implement `Debug` manually instead of relying in a dependency with a custom derive macro.

wgpu/src/image.rs
```diff
@@ -20,6 +23,9 @@ pub struct Pipeline {
     instances: wgpu::Buffer,
     constants: wgpu::BindGroup,
     texture_layout: wgpu::BindGroupLayout,
+    #[debug_stub="ReplacementValue"]
+    allocator: AtlasAllocator,
+    atlas: wgpu::Texture,
```
hecrj: I think both of these fields are quite coupled. It may be a good idea to couple them in a new opaque type: `image::Atlas`. This way we should be able to narrow the scope and control mutation nicely.

wgpu/src/image.rs
```diff
@@ -244,44 +282,76 @@ impl Pipeline {
...
+                    }
                 }
-            };
+            }
+        }
+
+        let atlas_width = self.allocator.size().width as f32;
+        let atlas_height = self.allocator.size().height as f32;
+
+        let texture = device.create_bind_group(&wgpu::BindGroupDescriptor {
```
hecrj: Do we need to create a texture view and a bind group every frame?

wgpu/src/image.rs
```diff
@@ -292,57 +362,55 @@ impl Pipeline {
...
-                                    resolve_target: None,
-                                    load_op: wgpu::LoadOp::Load,
-                                    store_op: wgpu::StoreOp::Store,
-                                    clear_color: wgpu::Color {
-                                        r: 0.0,
-                                        g: 0.0,
-                                        b: 0.0,
-                                        a: 0.0,
-                                    },
+                let mut render_pass = encoder.begin_render_pass(
```
hecrj: One of the main advantages of using an atlas is that it should allow us to render many images in a single draw call.

We should upload all the instance data and create only one render pass. Then, bind the corresponding atlas and draw the instances.

wgpu/src/image/raster.rs
```diff
@@ -153,9 +70,139 @@ impl Cache {
...
+                    allocation = a;
+                    break;
+                }
+
+                allocator.grow(allocator.size() * 2);
+            }
+
+            let new_atlas_size = allocator.size();
+
+            if new_atlas_size != old_atlas_size {
```
hecrj: The resize logic seems to be duplicated. The new `Atlas` type will probably help here.

wgpu/src/image.rs
Maldela: No

Maldela:
Sorry it took me so long to update the PR. I think the new one addresses all your requests.

hecrj:
You actually went ahead and implemented a texture array approach! And it seems to even split big images in multiple allocations! Awesome! :tada:

There is a lot to unpack here, so it may take me a while to go over all the changes. I'd appreciate it if you could share some details about the general implementation and any challenges, tradeoffs, future improvements or anything in particular we should be aware of.

How have you tested this? It would be great to create an example application to try these changes and, at the same time, showcase how the library can be used to create an actual application. Maybe some kind of image viewer?

I just took a quick glance and I think I can take it to the finish line from here, if you don't mind of course. It should be a matter of mostly move and rename some stuff.

Thanks for the amazing work! :clinking_glasses:

Maldela:
Just rebased on current master. Go ahead and rename or move anything you like.

wgpu/src/image.rs
```diff
@@ -231,139 +268,167 @@ impl Pipeline {
...
+                    bindings: &[wgpu::Binding {
+                        binding: 0,
+                        resource: wgpu::BindingResource::TextureView(
+                            &self.texture_atlas.view(),
+                        ),
+                    }],
+                });
+
+            self.texture_version = texture_version;
+        }
```
hecrj: Not sure how expensive creating a bind group is, but now we are only re-creating it when necessary instead of every frame.

wgpu/src/image.rs
```diff
@@ -231,139 +268,167 @@ impl Pipeline {
...
+            let end = (i + Instance::MAX).min(total);
+            let amount = end - i;
+
+            encoder.copy_buffer_to_buffer(
+                &instances_buffer,
+                (i * std::mem::size_of::<Instance>()) as u64,
+                &self.instances,
+                0,
+                (amount * std::mem::size_of::<Instance>()) as u64,
+            );
```
hecrj: Recreating a vertex buffer every frame is expensive. Instead, we create one with a fixed size and reuse it every frame, uploading data from a staging buffer using offsets when needed.

wgpu/src/image/raster.rs
```diff
@@ -153,10 +61,45 @@ impl Cache {
...
 
-        self.map.retain(|k, _| hits.contains(k));
+        self.map.retain(|k, memory| {
+            let retain = hits.contains(k);
+
+            if !retain {
+                if let Memory::Device(entry) = memory {
+                    atlas.remove(entry);
+                }
+            }
```
hecrj: I reverted the smart deallocation on drop, as it added a bunch of `Rc` and `RefCell` complexity. I think we should keep deallocation centralized and easy to reason about.

wgpu/src/texture/atlas.rs
```diff
@@ -0,0 +1,361 @@
...
+                | wgpu::TextureUsage::COPY_SRC
+                | wgpu::TextureUsage::SAMPLED,
+        });
+
+        let texture_view = texture.create_default_view();
+
+        Atlas {
+            texture,
+            texture_view,
+            layers: vec![Layer::Empty, Layer::Empty],
```
hecrj: Apparently, there is no way to create a texture array with an `array_layer_count` of `1`. When we do this, we get a bunch of Vulkan validation errors because the created texture is a simple 2D one and the sampler is not compatible.

I am not sure if this is a limitation of the modern graphics backends or a `wgpu` design decision. We should probably bring it up in [the `wgpu` matrix channel](https://riot.im/app/#/room/#wgpu:matrix.org).

In any case, we need to start with at least 2 layers for now.

wgpu/src/texture/atlas.rs
```diff
@@ -0,0 +1,361 @@
...
+                    height,
+                    0,
+                    &allocation,
+                    encoder,
+                );
+            }
+            Entry::Fragmented { fragments, .. } => {
+                for fragment in fragments {
+                    let (x, y) = fragment.position;
+                    let offset = (y * width + x) as usize * 4;
```
hecrj: I have simplified this a bunch. It turns out you can use `copy_buffer_to_texture` to upload a sub-region of an image quite easily.

wgpu/src/texture/atlas.rs
```diff
@@ -0,0 +1,361 @@
...
+                    height: SIZE,
+                    depth: 1,
+                },
+            );
+        }
+
+        self.texture = new_texture;
+        self.texture_view = self.texture.create_default_view();
+    }
+}
```
hecrj: Overall, I really like the simplicity here.

wgpu/src/image/raster.rs
Maldela: Wasn't sure about that on either. Automatic deallocation just seemed kind of "rusty".

wgpu/src/texture/atlas.rs
Maldela: Maybe we should decrease the atlas size then or make it configurable somehow. 4096 squared pixels times two is already 128 MB of data if I calculated that right.

hecrj: It's 2048 currently.

hecrj: In the long run, `wgpu` should offer a way to check available VRAM so we can choose a proper value at runtime.

Maldela: 2048 is fine, I think.

Maldela:
If you're happy with it, then I'm happy with it as well.

hecrj:
I have changed the renderer so it doesn't create a useless `image::Pipeline` (and therefore, an empty atlas) when both `image` and `svg` flags are disabled.

I think this ship is ready to sail. Thank you for all the work here! :tada:

PR #155, opened 2020-01-13
Remove Clone bound on Application::Message

ejmahler:
I'm currently writing an application where I want to create a message that should not be cloned.

It's a `Box<RenderState>` where `RenderState` contains a bunch of wgpu handles to GPU resources. I *could* wrap it in `Arc<Mutex<>>` and pass it through a message that way, but being able to pass it as a box achieves the same thing more clearly.

I can't do this in master, because the Message associated type has a bound on Clone. But it turns out that nothing actually takes advantage of this bound, and IMO it's logical that the system should only ever be moving messages, not cloning them.

This change will put a greater restriction on how Message is used, but it's a restriction that the system appears to be currently handling fine.

### DIFF
--- a/web/src/bus.rs
+++ b/web/src/bus.rs
@@ -8,9 +8,16 @@ use std::rc::Rc;
 ///
 /// [`Application`]: trait.Application.html
 #[allow(missing_debug_implementations)]
-#[derive(Clone)]
 pub struct Bus<Message> {
     publish: Rc<Box<dyn Fn(Message, &mut dyn dodrio::RootRender)>>,
+}
+
+impl<Message> Clone for Bus<Message> {
+    fn clone(&self) -> Self {
+        Self {
+            publish: Rc::clone(&self.publish),
+        }
+    }
 }
 
 impl<Message> Bus<Message>
--- a/web/src/bus.rs
+++ b/web/src/bus.rs
@@ -15,7 +15,7 @@ pub struct Bus<Message> {
 
 impl<Message> Bus<Message>
 where
-    Message: 'static + Clone,
+    Message: 'static,
 {
     pub(crate) fn new() -> Self {
         Self {
--- a/native/src/element.rs
+++ b/native/src/element.rs
@@ -171,7 +171,7 @@ where
     /// ```
     pub fn map<F, B>(self, f: F) -> Element<'a, B, Renderer>
     where
-        Message: 'static + Clone,
+        Message: 'static,
         Renderer: 'a,
         B: 'static,
         F: 'static + Fn(Message) -> B,
@@ -269,7 +269,6 @@ impl<'a, A, B, Renderer> Map<'a, A, B, Renderer> {
 
 impl<'a, A, B, Renderer> Widget<B, Renderer> for Map<'a, A, B, Renderer>
 where
-    A: Clone,
     Renderer: crate::Renderer,
 {
     fn width(&self) -> Length {
@@ -309,8 +308,7 @@ where
         );
 
         original_messages
-            .iter()
-            .cloned()
+            .drain(..)
             .for_each(|message| messages.push((self.mapper)(message)));
     }
 
--- a/src/application.rs
+++ b/src/application.rs
@@ -83,7 +83,7 @@ pub trait Application: Sized {
     /// The type of __messages__ your [`Application`] will produce.
     ///
     /// [`Application`]: trait.Application.html
-    type Message: std::fmt::Debug + Send + Clone;
+    type Message: std::fmt::Debug + Send;
 
     /// Initializes the [`Application`].
     ///
--- a/src/sandbox.rs
+++ b/src/sandbox.rs
@@ -81,7 +81,7 @@ pub trait Sandbox {
     /// The type of __messages__ your [`Sandbox`] will produce.
     ///
     /// [`Sandbox`]: trait.Sandbox.html
-    type Message: std::fmt::Debug + Send + Clone;
+    type Message: std::fmt::Debug + Send;
 
     /// Initializes the [`Sandbox`].
     ///
--- a/web/src/element.rs
+++ b/web/src/element.rs
@@ -38,8 +38,8 @@ impl<'a, Message> Element<'a, Message> {
     /// [`Element`]: struct.Element.html
     pub fn map<F, B>(self, f: F) -> Element<'a, B>
     where
-        Message: 'static + Clone,
-        B: 'static + Clone,
+        Message: 'static,
+        B: 'static,
         F: 'static + Fn(Message) -> B,
     {
         Element {
@@ -82,8 +82,8 @@ impl<'a, A, B> Map<'a, A, B> {
 
 impl<'a, A, B> Widget<B> for Map<'a, A, B>
 where
-    A: 'static + Clone,
-    B: 'static + Clone,
+    A: 'static,
+    B: 'static,
 {
     fn node<'b>(
         &self,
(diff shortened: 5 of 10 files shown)


hecrj:
Native widgets like `Button` require this bound. There are also some features that will need this bound in the future, like a time travelling debugger. It's a safe default.

Why do you want to send GPU resources through a `Message`? They are supposed to contain simple data. If it's the result of some computation, using `Arc` is the proper way to do it.

However, I'd advise you against using a `Mutex`. Chances are what you are trying to do can be achieved using a better data model.

ejmahler:
I have a Command that triggers a compute shader, and a message that returns the results.

Commands have a bound of static on their futures, which means the futures need to take ownership of their parameters -- so the future needs ownership over the compute resources. That means either box or arc.

Arc is problematic though, because my command needs *mutable* access to its GPU resources. So a mutex would be required if I used arc. So the simplest thing to do is box. When starting the computation, I move the box into the future, and when the computation, it moves the box back via the message. The nice thing about this design is that it also prevents multiple commands from accessing the compute resources at once, at compile time. Because you *have* to move the box containing the resources into the command, there's no possibility of reusing it to issue another command later which will wait on a mutex.

hecrj:
How does a `Box` let you move data from your `Application` into a future? What is left behind while the future is running? Isn't what you need an `Option` and `Option::take`?

Why do you need to share a `RenderState` between invocations of the `Command`? Could the `Command` create the state every time from scratch based on some other input? What does your application do?

I feel like you are trying to own state that shouldn't be yours. In other words, I think you should design your renderer as a separate runtime and then have a communication layer between it and your GUI.

One way to achieve this is by making your renderer live in a `Subscription` and communicate with your app using some kind of channel. The renderer will own its state and will be able to mutate it and the application won't see it (no need to move it or borrow it). It will also be able to notify the application of progress by returning messages (a `Subscription` is really just a `Stream`).

hecrj:
In any case, this PR doesn't really remove the bounds found in the widgets themselves, just the `Application`. You will hit the same issue at some point.

If the `Subscription` idea sounds too complicated, you may be able to leverage `Arc::try_unwrap` or `Arc::get_mut` to obtain ownership or a mutable reference, respectively. No need for a `Mutex`.

hecrj:
I gave it some more thought, and I think your changes are worth merging.

Although I would tackle your particular use case differently, I think non-cloneable messages can be useful for other scenarios.

For instance, you can achieve what you want __and still be able to use widgets that need a `Clone` requirement__ by modelling your `Message` like this:

```rust
#[derive(Debug)]
enum Message {
    Interaction(Interaction),
    RenderingFinished(RenderState),
}


#[derive(Debug, Clone, Copy)]
enum Interaction {
    ButtonPressed,
    // ... Other interactions you may be interested in!
}
```

Notice that `Message` does not implement `Clone`, so you should be able to use a `RenderingFinished` without an `Arc`.

Additionaly, I have removed the `Clone` requirement for `Element::map`, so you should be able to turn any `Element<Interaction>` into an `Element<Message>` easily:

```rust
impl Application for Example {
    // ...

    fn view(&mut self) -> Element<Message> {
        let content: Element<Interaction> =
            Button::new(&mut self.increment_button, Text::new("Increment"))
                .on_press(Interaction::IncrementPressed)
                .into();

        content.map(Message::Interaction)
    }
}
```

If we end up implementing a time travelling debugger, we will most likely do it by using a trait extension. We can ask for `Message: Clone` or similar there!

Finally, I have removed `Clone` requirements from `Sandbox` and some web widgets. Let me know what you think, and we can merge this!

ejmahler:
> How does a Box let you move data from your Application into a future? What is left behind while the future is running? Isn't what you need an Option and Option::take?
> 
> Why do you need to share a RenderState between invocations of the Command? Could the Command create the state every time from scratch based on some other input? What does your application do? A bit more details would help me understand your use case and give you further guidance!

The ultimate goal is to show a fractal render in-progress. I have a fractal that takes 100k iterations to complete. I'd like to make a system where you can request that it run N iterations and then return incremental progress back to the UI.

So the design is, there's a Command to run N iterations, and when it's done it sends a message back that it's done. The message handler will spawn 2 commands when this happens. The first command converts the fractal data into an Image handle for rendering, and the second command is to continue computations. Finally, when the fractal is completely done, it either just drops the GPU state right there, or stores it on the GUI object. I want to try the former first.

So I don't want to recreate the render state every time because I imagine the steps will be quite short (N might be 1 to 10).

> From what I gathered, I feel like you are trying to own state that shouldn't be yours. In other words, I think you should design your renderer as a separate runtime and then have a communication layer between it and your GUI.
> 
> One way to achieve this is by making your renderer live in a Subscription and communicate with your app using a channel (mpsc should work). The renderer will own its state and will be able to mutate it and the application won't see it (no need to move it or borrow it). It will also be able to notify the application of progress by returning messages (a Subscription is really just a Stream).

I had never heard of subscriptions before. I'm willing to try this out!

hecrj:
Thanks for the details! Sounds really cool. If you end up open-sourcing it, I'd love to try it out!

I think event subscriptions are the best way to do what you want. They are basically the `Stream` version of `Command`, and can be used nicely to model asynchronous systems that need to persist and report progress like WebSocket connections, HTTP downloads, etc.

Event subscriptions were merged recently in #122.

web/src/bus.rs
```diff
@@ -8,11 +8,18 @@ use std::rc::Rc;
...
 #[allow(missing_debug_implementations)]
-#[derive(Clone)]
 pub struct Bus<Message> {
     publish: Rc<Box<dyn Fn(Message, &mut dyn dodrio::RootRender)>>,
 }
 
+impl<Message> Clone for Bus<Message> {
+    fn clone(&self) -> Self {
+        Self {
+            publish: Rc::clone(&self.publish),
```
ejmahler: It looks like you discovered the same thing I did, where #[derive(Clone)] seemsto generate a wonky implementation of clone here, which only kicks in if Message is Clone -- even though all the fields are trivially cloneable unconditionally. And the solution isto just manually implement an unconditional clone.

ejmahler:
I'll be honest, your last example there is over my head. I'm not super clear on the difference between interactions, elements, messages, and `Element<Interaction>` and `Element<message>`

One concern is, how discoverable do you think it is to have to put element interactions inside an inner message? I can imagine the following sequence:

1. The developer makes a message, which works for something trivial
2. The developer tries to add a button press, which gives them a compile error about needing clone, so on autopilot, the developer adds a Clone bound on their whole message struct
3. The developer never learns that they can remove the Clone bound.

I suppose it's fine if they do - something like this would probably be a pretty advanced usage. It it's worth thinking about.

ejmahler:
Ah, I see my confusion. I thought `interaction` was a built-in thing I hadn't seen before, and content.map basically tells the content widget to map its messages via the function you supply?

How many people will discover that you can do something like `content.map`, and how many people will understand what they're reading when they find it in someone else's code?

hecrj:
@ejmahler The Elm Architecture composes quite nicely. The whole idea is that your `Message` type does not need to be flat. In fact, nesting messages is an interesting technique that can be used to compose different smaller applications into a bigger one. The [`Element::map` documentation](https://docs.rs/iced_native/0.1.0/iced_native/struct.Element.html#method.map) shows an example of this.

> The developer never learns that they can remove the Clone bound.

The `Clone` bound is still needed, but only by the message type you use directly in your widgets. If your usage is advanced enough that you need an un-cloneable `Application::Message`, you can simply nest the bound. I don't think this strategy is that uncommon.

ejmahler:
Yeah, I agree that the idea of nesting messages should feel pretty natural. I developed something for this a while ago and I reached for nesting messages almost immediately.

This looks good then.

hecrj:
> How many people will discover that you can do something like content.map, and how many people will understand what they're reading when they find it in someone else's code?

`Element::map` is no different than `Option::map` or `Iterator::map`.

The `tour` and `todos` examples use it!

PR #163, opened 2020-01-18
Cull images and free VRAM

### LINKED ISSUE #160: Element rendering state support
I am writing an app that loads more than 400 PNGs into a scroll view.

But, it seems that VRAM is out of memory because too many images are being loaded.

I want to unload an image outside the rendering range of the scroll view.
Is there a way to know the display status of children items in scroll view?

hecrj:
Fixes #160.

This PR changes the `iced_wgpu::image::raster::Cache` to unload images from the GPU when they are no longer visible and new ones have been uploaded.

The change may cause applications to block more oftenly when hiding/displaying different images (like a scrolling image gallery). However, they are much less likely to crash due to running out of video memory. We should be able to improve this strategy in #154 by trying to deallocate before creating a new layer instead.

In the long run, we should load and upload images asynchronously to avoid dropping frames.

### DIFF
--- a/core/src/rectangle.rs
+++ b/core/src/rectangle.rs
@@ -16,6 +16,18 @@ pub struct Rectangle<T = f32> {
     pub height: T,
 }
 
+impl<T: std::ops::Add<Output = T> + PartialOrd + Copy> Rectangle<T> {
+    /// Returns true if the current [`Rectangle`] intersects with the given one.
+    ///
+    /// [`Rectangle`]: struct.Rectangle.html
+    pub fn intersects(&self, b: &Rectangle<T>) -> bool {
+        self.x < b.x + b.width
+            && self.x + self.width > b.x
+            && self.y < b.y + b.height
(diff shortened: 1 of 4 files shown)


cryptoquick:
This feature will be really useful in one of the apps I'm building. At least, it's a great start, as you noted. Does any more work need to be done here before it's merged in?

hecrj:
@cryptoquick This is a bit related to the texture atlas work implementation in #154. The new storage strategy may allow us to do a better job at invalidating cache regions.

Therefore, I think we should unify these changes in that PR first. I will work on it once I start batching work again in `iced_wgpu`.

cryptoquick:
Glad to hear, it all sounds close!

hecrj:
Closing, as I will most likely tackle this once I move the texture atlas logic to `iced_graphics` when implementing `Image` and `Svg` support for `iced_glow` (see #354).

PR #168, opened 2020-01-25
Button hover event

sumibi-yakitori:
Add a hover event to the button.

Hover events are called continuously when the mouse cursor moves,
but that is the same for other GUI frameworks.

### DIFF
--- a/native/src/widget/button.rs
+++ b/native/src/widget/button.rs
@@ -32,7 +32,7 @@ pub struct Button<'a, Message, Renderer: self::Renderer> {
     state: &'a mut State,
     content: Element<'a, Message, Renderer>,
     on_press: Option<Message>,
-    on_hovered: Option<Message>,
+    on_hover: Option<Message>,
     width: Length,
     height: Length,
     min_width: u32,
@@ -58,7 +58,7 @@ where
             state,
             content: content.into(),
             on_press: None,
-            on_hovered: None,
+            on_hover: None,
             width: Length::Shrink,
             height: Length::Shrink,
             min_width: 0,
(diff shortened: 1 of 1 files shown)


hecrj:
Thanks for the suggestion!

What is the use case for this? I think hover events are normally used to change styling, and the `button::StyleSheet` should already take care of that.

I am wary of adding this because we do not want users to rely on `on_hover` to change layout either, as I feel it would be quite unintuitive. A button should not produce any effects until it is pressed.

sumibi-yakitori:
I want to display the detailed information of the selected list item in another view.
Often used when there are many items in the list. Displaying items only on hover is faster than clicking on details in other views.

We understand that this usage is not intended for consumers who use touch devices.
But
If there are no basic features supported by other GUI frameworks,
I don't think there is much freedom to create various professional tools, so please add them if possible.

Also,
This may be a somewhat outdated use case, but tooltips are still useful.

I apologize for the strange English.

sumibi-yakitori:
For example, something like this.

Screen Shot 2020-01-25 at 16 42 33

hecrj:
Yes, I think your use case is completely valid.

However, I do not believe a `Button::on_hover` is the best strategy for it. We want widgets to be as simple as possible and with a clear purpose. We want the library to tackle use cases directly.

In my opinion, a better alternative would be to create a `SelectList` widget with an `on_item_hover` method or similar. This approach could tackle this particular use case and many others (lazily rendering items based on scrolling offset, for instance). I understand this requires a lot more effort and exploration but it should be worth it in the long run.

For the GitHub example you posted, a `Tooltip` widget would be the way to go, but that would need layer support (see #30).

sumibi-yakitori:
I think I could understand your design philosophy.

As you say, I still don't have enough time to understand the source code of iced.

I will try to create a simple custom widget that supports hover event.

myfreeweb:
I solved this by making a custom container widget that handles hover events (and might handle others in the future):

https://github.com/myfreeweb/waysmoke/commit/aa74cc51cea3dad000211cde4eae99affd126e29

Nicely composable :)

PR #169, opened 2020-01-26
Enable custom fonts for text input

sumibi-yakitori:
I made custom fonts available for `iced_native::widget::text_input::TextInput`

This fix may have something wrong with it because of the lack of understanding of iced.


sumibi-yakitori:
My email address was anonymous, so I will withdraw it once.

PR #175, opened 2020-02-02
Make iced_winit Clipboard public

Imberflur:
While it would be simple to re-implement, making this type public would be convenient for having clipboard access without using the `Application` trait.


hecrj:
I "accidentally" force-pushed the wrong branch to your `master` branch when trying to fix the compiler errors :sweat_smile:

In any case, this seems acceptable. I reopened the PR in #178, fixed the issues and merged it. Thank you :clinking_glasses:

PR #176, opened 2020-02-03
Support for changing the window position at application startup

sumibi-yakitori:


sumibi-yakitori:
I will rebase this branch and take this PR down once because this PR has gotten dirty.

PR #179, opened 2020-02-05
chore: ignore config files

orzyyyy:
Related to https://stackoverflow.com/questions/37586216/step-by-step-interactive-debugger-for-rust

Follow the guide above, run tour only needs pressing F5.


Songtronix:
I think the best way would be to add `.vscode` and similar to `.gitignore` and let the developer deal with their setup instead of having to maintain various workflows.
However it might be nice to collect these configs somewhere for people to get up and running fast.

orzyyyy:
What about ignoring `.vscode` and add this usage into documentation?
For me, I prefer to use vscode and press F5 to start the tour, but it seems to be a hard work to unity the editor and its workflow. If the repository doesn't want to maintain these workflows, just ignore these config files that should be maintained userself locally.

hecrj:
> I think the best way would be to add .vscode and similar to .gitignore

This can be added locally in `.git/info/exclude`.

> add this usage into documentation

This stills has the burden of maintainance and poses other questions: why do we offer docs for VSCode and not other editors? Why not also IntelliJ, vim, and emacs? Where do we draw the line?

orzyyyy:
What about ignoring all config from editors? For I don't like `.vscode` would get into the workflow of git. When I want to change branch, I must execute `git stash` and pop after.

hecrj:
@orzyyyy I am not against that, but I think it's a solution that doesn't scale well.

Instead, if a developer uses VSCode, they can add `.vscode` to `.git/info/exclude` to ignore all the editor configuration. This avoids specifically listing editor/workflow details in `.gitignore`.

PR #181, opened 2020-02-08
Update `winit` to `0.21`

Imberflur:
Builds on #147 by bumping to `0.21` and rebasing on `master`

### DIFF
--- a/winit/src/application.rs
+++ b/winit/src/application.rs
@@ -138,10 +138,7 @@ pub trait Application: Sized {
 
             window_builder = window_builder
                 .with_title(&title)
-                .with_inner_size(winit::dpi::LogicalSize {
-                    width: f64::from(width),
-                    height: f64::from(height),
-                })
+                .with_inner_size(winit::dpi::LogicalSize { width, height })
                 .with_resizable(settings.window.resizable)
                 .with_decorations(settings.window.decorations)
                 .with_fullscreen(conversion::fullscreen(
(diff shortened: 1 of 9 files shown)


winit/src/application.rs
```diff
@@ -459,17 +474,7 @@ fn build_user_interface<'a, A: Application>(
...
-
-    (
-        physical_size.width.round() as u16,
-        physical_size.height.round() as u16,
-    )
-}
-
 // As defined in: http://www.unicode.org/faq/private_use.html
-// TODO: Remove once https://github.com/rust-windowing/winit/pull/1254 lands
 fn is_private_use_character(c: char) -> bool {
```
Imberflur: Should this be removed?

winit/src/application.rs
```diff
@@ -337,18 +337,22 @@ pub trait Application: Sized {
...
                 WindowEvent::CloseRequested => {
                     *control_flow = ControlFlow::Exit;
                 }
                 WindowEvent::CursorMoved { position, .. } => {
+                    let position =
+                        position.to_logical::<f64>(size.scale_factor());
+
                     events.push(Event::Mouse(mouse::Event::CursorMoved {
                         x: position.x as f32,
                         y: position.y as f32,
```
Imberflur: Would is be desirable to mirror `winit` and use `f64` here?

winit/src/application.rs
hecrj: I prefer to keep this for now. `winit` is not filtering exactly the same characters.

winit/src/application.rs
hecrj: Probably, but we are currently using single precision almost everywhere.

I believe we need to think a bit more about it and eventually tackle it in a different PR.

Imberflur:
Yay! It's great to see this merged :)

PR #183, opened 2020-02-09
Integration with existing `wgpu` projects

hecrj:
[Hello triangle](https://gfycat.com/nicemediocrekodiakbear)

This PR enhances the flexibility of `iced_wgpu` by allowing a `Renderer` to draw to any given `TextureView` while reusing an existing  `Device` and `CommandEncoder`. Therefore, integrating `iced` with existing `wgpu` projects is now possible! :tada:

I have created an example (`integration`) which showcases how this can be achieved with `iced_winit` and `iced_wgpu`. It is important to note that the process should be the same with a different shell and/or renderer:

1. Initialize data structures
2. Map shell events to iced events
3. Process iced events
4. Update the user interface
5. Draw the user interface

This process, although very flexible, is currently not as straightforward as I would like it to be. However, I prefer to wait, see how folks implement it and gather more use cases before designing further abstractions.


parasyte:
This might actually _mostly_ fit my requirements! I'll give it a try. :)

PR #187, opened 2020-02-14
Add `Node::move_to` and make `Node::align` public

daxpedda:
To help create custom widgets.

### DIFF
--- a/native/src/layout/node.rs
+++ b/native/src/layout/node.rs
@@ -54,10 +54,9 @@ impl Node {
         &self.children
     }
 
-    /// Aligns item according to [`Align`] and [`Size`].
-    /// 
-    /// [`Align`]: ../enum.Align.html
-    /// [`Size`]: ../struct.Size.html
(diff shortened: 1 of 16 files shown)


native/src/layout/node.rs
```diff
@@ -54,7 +54,11 @@ impl Node {
         &self.children
     }
```
hecrj: Instead of breaking the existing methods, I would add a `move_to` method. Also, I think we should use `Point` instead of `Size` because we are not setting the `bound` of the `Node` but its `position`.

Therefore:

```rust
pub fn move_to(&mut self, position: Point) {
    self.bounds.x = position.x;
    self.bounds.y = position.y;
}
```

I think this is more explicit because having a `Node` not positioned at the origin is unusual.

daxpedda: Sound much better, thanks.

daxpedda:
Thanks, the documentation is much better!

PR #189, opened 2020-02-14
Grid widget

daxpedda:
This relies on #187, but this reliance can easily be circumvented.

Currently table has no settings and shows no other visuals then it's content.
Any feedback would be appreciated.


wgpu/src/renderer/widget/table.rs
```diff
@@ -0,0 +1,30 @@
...
+
+impl table::Renderer for Renderer {
+    fn draw<Message>(
+        &mut self,
+        defaults: &Self::Defaults,
+        layout: Layout<'_>,
+        cursor_position: Point,
+        rows: &[Vec<Element<'_, Message, Self>>],
+    ) -> Self::Output {
+        let mut mouse_cursor = MouseCursor::OutOfBounds;
```
daxpedda: I don't actually know what this does. I just copied it from `Column` and `Row`.

wgpu/src/renderer/widget/table.rs
```diff
@@ -0,0 +1,30 @@
...
+        let mut primitives = Vec::new();
+
+        for (row, layout) in rows.iter().zip(layout.children()) {
+            for (cell, layout) in row.iter().zip(layout.children()) {
+                let (primitive, new_mouse_cursor) =
+                    cell.draw(self, defaults, layout, cursor_position);
+
+                if new_mouse_cursor > mouse_cursor {
+                    mouse_cursor = new_mouse_cursor;
+                }
```
daxpedda: Same as above.

native/src/widget/table.rs
```diff
@@ -0,0 +1,188 @@
...
+        &self,
+        renderer: &mut Renderer,
+        defaults: &Renderer::Defaults,
+        layout: Layout<'_>,
+        cursor_position: Point,
+    ) -> Renderer::Output {
+        renderer.draw(defaults, layout, cursor_position, &self.rows)
+    }
+
+    fn hash_layout(&self, state: &mut Hasher) {
```
daxpedda: I saw in other implementations a line like this:
```rust
0.hash(state);
```
Not sure what it is supposed to do.

daxpedda:
Rebased.
But the API definitely needs a discussion.

hecrj:
As [I said in the Zulip], a `Table` widget is tricky to get right. We should answer some of the questions I raised there first.

I think we are missing a __clear use case__ here. The widget implementation seems to be laying out content smartly, but I think it's hard to predict exactly how. For instance, what happens if a cell is very wide?

Additionally, I think the current widget is more of a `Grid` widget than a `Table` widget, as it does not support headers nor display cell separators. This is fine! I think a `Grid` widget is a better way to start exploring this kind of layout.

Overall, I believe we should reduce a bit the scope of this, compromise functionality, and tackle a specific use case directly. For instance, maybe we could try to implement a `Grid` with a fixed column width (or number of columns) provided on creation. Users can `push` elements directly (like in a `Column`) and the `Grid` will automatically wrap elements smartly and keep track of each row height. I think this has a clearer scope with a more predictable behavior. We should also showcase this in a cool example.

Finally, do not worry about the `iced_web` version too much. We can tackle that in another PR in the future.

[I said in the Zulip]: https://iced.zulipchat.com/#narrow/stream/213316-general/topic/table.20widget

daxpedda:
> For instance, what happens if a cell is very wide?

Providing a max width for the whole grid would be a good start, what do you think?
> For instance, maybe we could try to implement a `Grid` with a fixed column width (or number of columns) provided on creation. Users can `push` elements directly (like in a `Column`) and the `Grid` will automatically wrap elements smartly and keep track of each row height.

That sounds great! We can do it like HTML and provide the ability to put settings on the items themselves by simply adding `push_with_settings` or something similar later.
> We should also showcase this in a cool example.

Gonna work on that when we finish the API.
Should I put it in tour or into a specific example? Or should I make a dedicated one?
> Finally, do not worry about the `iced_web` version too much. We can tackle that in another PR in the future.

I don't mind, I'm using the web target more right now then the desktop one.

wgpu/src/renderer/widget/table.rs
hecrj: Mouse cursors are ordered so a `Pointer` cursor has higher priority than an `Idle` cursor. This allows us to select the most important cursor of the children.

native/src/widget/table.rs
hecrj: That is a leftover that we should fix eventually.

The idea is that we have to include a unique identifier for this widget in the resulting hash. We can achieve this with:

```rust
std::any::TypeId::of::<Grid<'static, ()>>().hash(state);
```

hecrj:
> We can do it like HTML and provide the ability to put settings on the items themselves by simply adding push_with_settings or something similar later.

Let's keep it simple for now!

> Should I put it in tour or into a specific example? Or should I make a dedicated one?

I'd make a dedicated one. The `tour` is already complex enough. I will update it eventually to showcase all the new stuff.

native/src/widget/grid.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// [`Grid`]: struct.Grid.html
+    pub fn new(columns: usize) -> Self {
+        Self::from_vec(columns, Vec::new())
+    }
+
+    /// Create a new [`Grid`] with the given [`Element`]s.
+    ///
+    /// [`Grid`]: struct.Grid.html
+    /// [`Element`]: ../struct.Element.html
+    pub fn from_vec(
```
hecrj: I'd call this `with_children`.

native/src/widget/grid.rs
```diff
@@ -0,0 +1,198 @@
...
+    {
+        self.elements.push(element.into());
+        self
+    }
+
+    /// Adds [`Element`]s to the [`Grid`].
+    ///
+    /// [`Element`]: ../struct.Element.html
+    /// [`Grid`]: struct.Grid.html
+    pub fn push_vec(
```
hecrj: For consistency, I wouldn't include this for now as no other widget has it. I am not sure this is something we want to encourage.

native/src/widget/table.rs
daxpedda: Done.

daxpedda:
> I'd make a dedicated one. 

On it.

native/src/widget/grid.rs
```diff
@@ -0,0 +1,198 @@
...
+        for (column, element) in (0..self.columns).cycle().zip(&self.elements) {
+            let layout = element.layout(renderer, limits).size();
+            layouts.push(layout);
+
+            if let Some(column_width) = column_widths.get_mut(column) {
+                *column_width = column_width.max(layout.width);
+            } else {
+                column_widths.insert(column, layout.width);
+            }
+        }
```
hecrj: Looks cool! Gotta love iterators.

I think it'd be interesting to also be able to provide a fixed column width on creation and then compute the number of columns based on the `limits` instead of the children. In this case, the `Grid` would have `Fill` width behavior instead.

src/widget.rs
```diff
@@ -45,6 +45,12 @@ mod platform {
     pub type Column<'a, Message> =
         iced_winit::Column<'a, Message, iced_wgpu::Renderer>;
 
+    /// A container that distributes its contents in a grid.
+    ///
+    /// This is an alias of an `iced_native` grid with a default `Renderer`.
+    pub type Grid<'a, Message> =
+        iced_winit::Grid<'a, Message, iced_wgpu::Renderer>;
```
hecrj: I would move this after `Row`.

native/src/widget/grid.rs
daxpedda: Done.

native/src/widget/grid.rs
daxpedda: Done.

src/widget.rs
daxpedda: Done.

native/src/widget/grid.rs
daxpedda: Should I remove fixed column then and just take column width?
If we want both to be possible, how would we approach that:
-  pass an enum?
-  make two different `new_*` functions?
-  what should be the feault?

I don't particularly understand the use case here, is it like an image gallery? Can you give me some examples? It would also help me make the actual grid example, because I was thinking of something table-like, which is, obviously, more table then grid.

daxpedda: Looking at the other widgets I figured we could just do `fn columns(self, columns: usize) -> Self`, basically following the builder pattern here.

ajax-crypto:
I have been following this project for quite some time now, just want to understand the rationale behind a grid widget vs. a grid layout. Shouldn't grid be a layout strategy rather than a widget? (I might be completely wrong here) I would expect table/tree (with multiple columns, lazy loading, pinning, etc.) to be widgets. 
In future, I am looking forward to a tree widget, with the above mentioned features and a form of grid layout (desktop target). (Using Qt/C++ right now, planning to move to Rust).

hecrj:
@ajax-crypto A widget is just a node in the view tree. They usually represent interactive and visual elements, but they can also be used to implement containers with different layouts (like `Column` and `Row`).

Flutter, for instance, has widgets like `Padding`, `GridView`, `Column`, ` Row`...

native/src/widget/grid.rs
daxpedda: I hope this is better now.

native/src/widget/grid.rs
```diff
@@ -0,0 +1,276 @@
...
+            columns: None,
+            column_width: None,
+            elements,
+        }
+    }
+
+    /// Sets a fixed amount of columns for the [`Grid`].
+    ///
+    /// [`Grid`]: struct.Grid.html
+    pub fn columns(mut self, columns: usize) -> Self {
```
daxpedda: 1.  Should I make a note in the documentation about how 0 is handled?
2.  I could also return an error when a 0 is passed.
3.  We could also make the type here NonZeroUsize.
4.  Is `usize` fine?

native/src/widget/grid.rs
```diff
@@ -0,0 +1,276 @@
...
+    for Grid<'a, Message, Renderer>
+where
+    Renderer: self::Renderer,
+{
+    fn width(&self) -> Length {
+        if self.columns.is_some() {
+            Length::Shrink
+        } else {
+            Length::Fill
+        }
```
daxpedda: I hope I got that right?

native/src/widget/grid.rs
daxpedda: So I made it like this:
-  **Defined only column:** determine column width by widest element in that column
-  **Defined only column width:** determine number of columns by dividing max width by column width
-  **Defined both:** number of columns and column width is respected as expected
-  **Defined none:** elements will be displayed exactly as in row

Anything we want to tweak here?

native/src/widget/grid.rs
```diff
@@ -0,0 +1,276 @@
...
+///         Element::from(Text::new("Second row, first column")),
+///         Element::from(Text::new("Second row, second column")),
+///     ],
+/// )
+/// .columns(2);
+/// ```
+#[allow(missing_debug_implementations)]
+pub struct Grid<'a, Message, Renderer> {
+    columns: Option<usize>,
+    column_width: Option<u16>,
```
hecrj: I would group this in an enum:

```rust
enum Strategy {
    Columns(usize),
    ColumnWidth(u16),
}
```

Removing the possibilities of "no strategy" or "two strategies" altogether.

Then, we can make `with_columns` and `with_column_width` the only constructor methods and remove `with_children` for now.

native/src/widget/grid.rs
hecrj: If `0` is passed, I would skip layout and draw nothing.

native/src/widget/grid.rs
hecrj: I think it should be the other way around. When we have a fixed column width, chances are we will not be able to fill all the available space.

native/src/widget/grid.rs
```diff
@@ -0,0 +1,276 @@
...
+        }
+    }
+
+    fn height(&self) -> Length {
+        Length::Shrink
+    }
+
+    fn layout(&self, renderer: &Renderer, limits: &Limits) -> Node {
+        if self.elements.is_empty() {
+            Node::new(Size::ZERO)
```
hecrj: We can return early here and avoid further nesting.

native/src/widget/grid.rs
```diff
@@ -0,0 +1,276 @@
...
+                        grid_height,
+                    ));
+                    nodes.push(node);
+                    row_height = row_height.max(size.height);
+                }
+
+                grid_height += row_height;
+                let grid_width = (columns as f32) * column_width;
+
+                Node::with_children(Size::new(grid_width, grid_height), nodes)
```
hecrj: Maybe we should split this in a couple of private functions.

native/src/widget/grid.rs
daxpedda: Alright.

native/src/widget/grid.rs
daxpedda: I'm not sure what you mean, this should be returning right there, writing `return` here would not change the behaviour as far as I can see.

native/src/widget/grid.rs
daxpedda: My assumption here is if we use `with_columns` and `with_column_width` we won't need this function anymore anyway.

native/src/widget/grid.rs
daxpedda: > avoid further nesting

I thought you are talking about the implementation and not about the actual code, I see now.
Done.

native/src/widget/grid.rs
daxpedda: I handled this case in `fn layout` now.

native/src/widget/grid.rs
daxpedda: Done.

native/src/widget/grid.rs
daxpedda: Done.

daxpedda:
Sorry I was busy for some time now too.

About the example, would love to take a shot at it, but I would like you to tell me what kind of use case you had in mind for the grid widget.

hecrj:
> Sorry I was busy for some time now too.

Absolutely no worries. I am already very grateful you took the time to contribute!

> About the example, would love to take a shot at it, but I would like you to tell me what kind of use case you had in mind for the grid widget.

The use case I had in mind is visualization of any dynamic list of items with small amounts of data. Image galleries, dashboards, and menus are the first real examples that come to mind.

I think examples should be interesting and showcase real use cases without being overly complicated. I think something like an image gallery or a dashboard may be a bit too much...

Maybe we could build something like a `github_explorer`. An example that has a `TextInput` accepting a GitHub username which, once submitted, displays in a grid the user's top 10 repositories. Or maybe something similar but using crates.io? I am just brain-storming. Feel free to propose any other ideas!

Hope this clarifies a bit the use cases I had in mind.

victorstewart:
just to pipe up from the peanut gallery, what I've love to see and use here is something akin to the beautiful collection views in iOS just without all the unnecessary junk. 

So... a callback to generate a view element for a given index, recycle those views, callbacks when elements become visible, have the views be bounds aware so they can conduct animations at boundaries, feed an explicit layout map to the collection, allow items to be placed anywhere and of any size, the ability to tap on one and remove it from the tree and focus it forefront consuming the whole screen then put it back in its place...

I use such collections on iOS for all the core visual experiences in my app so I've hacked them extensively.

daxpedda:
I don't belive I'm going to be able to contribute to Iced for a while :disappointed:.
Everybody is free to reuse my code that I produced here.

hecrj:
@daxpedda No need to close this! Also, no worries whatsoever. I will work on merging your work soon.

Luni-4:
@daxpedda 

Is this PR abandoned or it can be reopened? I'm using your widget in a project and it is really useful, thanks a lot for your work! It would be great if it could be merged upstream also :smiley:

daxpedda:
I can't reopen it because I already deleted my repo and everything, sorry about that!
Personally I'm not gonna continue this anymore, anybody can feel free to copy the code from here any make a PR.

This is kinda annoying, but maybe this helps:
https://patch-diff.githubusercontent.com/raw/hecrj/iced/pull/189.patch
https://patch-diff.githubusercontent.com/raw/hecrj/iced/pull/189.diff

Luni-4:
Oh, I'm sorry to hear that! I could try to open a new PR in the next days using your patch, rebased on the current master, since your changes have already been approved previously.

Issue #77, opened 2019-11-27
Could on_press accept an Fn as well as a message instead of a message only


rebo:
Currently TextInput and slider etc allow an Fn to be run that results in a message. This is useful to filter messages that go back into the update.

Could the same be implemented for button.on_press as well?  It might be useful to moderate the message that button sends depending on context.

hecrj:
Thanks for the suggestion!

Have you tried something like this?

```rust
button.on_press(if some_context { Message::A } else { Message::B })
```

Is there any reason this may not work for your use case? Could you share a bit more about it? Why is a closure necessary?

rebo:
Hey hecrj, the other reason for having a closure is that non-iced functions can be triggered. For instance I could implement a counter like the below and keep all my state in the counter without having to store anything in the App model. Useful for encapsulation of local self contained state (i.e.  React Hooks)

```rust
    fn view(&mut self) -> Element<Message> {

        let (count, count_access) = use_istate(||0);
      
        Column::new()
            .padding(20)
            .push(
                Button::new(Text::new("Increment"))
                    .on_press_callback(||count_accesss.set(count + 1)),
            )
            .push(Text::new(count.to_string()).size(50))
            .push(
                Button::new(Text::new("Decrement"))
                .on_press_callback(||count_accesss.set(count - 1)),
            )
            .into()
    }
```

hecrj:
Iced is inspired by Elm and The Elm Architecture.

Basically, state mutations are centralized in `update` and `view` is a pure function producing a description of the GUI. The whole runtime is built based on the assumption that your view code is pure. This guarantee this is what will allow us to build a time travelling debugger in the future!

Hooks (or any kind of mutable callback) are based on side-effects in view code, which is directly against the core foundations of the library.

That said, widgets themselves _can_ have local state, and using something like hooks may be an interesting strategy to track internal state. This should be completely hidden from end-users, though.

rebo:
Hey sure I understand and I actually like the Elm Architecture very much, just find it a little overkill if you have to run everything through it. It's good that you can have internal states on widgets though.

dmarcuse:
I'd also be interested in this to support the case where messages can't be cloned: I use an enum for my `Message` type, and while the values yielded on button presses are unit variants, some of the other variants contain more complicated data that can't be copied. I could start wrapping things in `Arc`, but that adds significantly more work.

hecrj:
[...]

PR #191, opened 2020-02-16
Avoid Clone bound on Message

Eroc33:
As dmarcuse mentioned in #77 if you have some message variants with uncloneable contents you currently end up having to wrap things in Arc which isn't very ergonomic. Having widgets accept Fn instead of Message avoids this issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ impl Counter {
                 // The increment button. We tell it to produce an
                 // `IncrementPressed` message when pressed
                 Button::new(&mut self.increment_button, Text::new("+"))
-                    .on_press(Message::IncrementPressed),
+                    .on_press(|| Message::IncrementPressed),
             )
             .push(
                 // We show the value of the counter here
@@ -128,7 +128,7 @@ impl Counter {
                 // The decrement button. We tell it to produce a
                 // `DecrementPressed` message when pressed
                 Button::new(&mut self.decrement_button, Text::new("-"))
-                    .on_press(Message::DecrementPressed),
+                    .on_press(|| Message::DecrementPressed),
             )
     }
 }
--- a/examples/bezier_tool/src/main.rs
+++ b/examples/bezier_tool/src/main.rs
@@ -353,7 +353,7 @@ impl Sandbox for Example {
             .push(
                 Button::new(&mut self.button_state, Text::new("Clear"))
                     .padding(8)
-                    .on_press(Message::Clear),
+                    .on_press(|| Message::Clear),
             );
 
         Container::new(content)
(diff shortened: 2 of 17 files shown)


hecrj:
I am not convinced we should make the overall API more complicated because of the `Arc` use case.

You can currently have non-cloneable messages just fine by separating your UI interactions from your non-cloneable messages. I detailed how to do this here: https://github.com/hecrj/iced/pull/155#issuecomment-573523405

dmarcuse:
To me, the API doesn't really seem any more complicated - it's just a closure as opposed to a value. I think making it more consistent with other widgets (slider, text input, etc) which accept functions makes the API more ergonomic as well. As you noted [here](https://github.com/hecrj/iced/pull/155#issuecomment-573523405) it is possible to define a second message type that satisfies the `Clone` bound, but this isn't really obvious (at least not to me) and takes a lot more effort than it would to use closures.

I understand that having a way to introduce non-determinism is a goal of the project, but seeing as other widgets already use closures, I don't think it's worth sacrificing ergonomics for.

Eroc33:
Thanks for pointing out that alternative way to structure Messages. I think that *will* work for my case. However I do agree with dmarcuse that using closures is a much more obvious/discoverable way to allow this. If you're determined not to use closures for this would you be willing to accept a PR adding an example of the pattern discused in https://github.com/hecrj/iced/pull/155#issuecomment-573523405 (assuming I can think of a non-contrived example)?

hecrj:
> it's just a closure as opposed to a value

I think this is a big distinction. There is no need for a widget to accept a closure with no arguments just so we can construct many messages. A `Clone` bound is exactly meant to convey this!

> I think making it more consistent with other widgets (slider, text input, etc) which accept functions makes the API more ergonomic

I think it would be inconsistent! Widgets use closures when they provide some data that is relevant to construct a message. Most of the time, you are not even meant to use the closure syntax:

```rust
let text_input = TextInput::new(
    state,
    "Type something to continue...",
    value,
    StepMessage::InputChanged,
)
```

Forcing our users to type `on_press(|| Message::ButtonPressed)` just for buttons does not seem right! A different user could ask: why does `on_press` accept a closure instead of requiring `Clone`? In the end, this seems very subjective to me.

> takes a lot more effort than it would to use closures

Simplicity isn't always convenient. Structuring your `Message` type properly is key in an application that uses the Elm Architecture.

Moreover, there are also some long term features that will force messages to satisfy `Clone`, like a time travel debugger. Although we may add that as a requirement in a different trait, I believe it's something we should keep in mind: there are many reasons why a `Clone` bound may be necessary and giving it up at the root level will have consequences in the long run.

> If you're determined not to use closures for this would you be willing to accept a PR adding an example of the pattern discused in #155 (comment) (assuming I can think of a non-contrived example)?

Of course! Feel free to join [the Zulip server](https://iced.zulipchat.com/) and share your ideas there!

PR #192, opened 2020-02-17
Customize Checkbox

clarkmoody:
Add `size`, `spacing`, and `text_size` properties to Checkbox.


native/src/widget/checkbox.rs
```diff
@@ -53,11 +56,22 @@ impl<Message, Renderer: self::Renderer> Checkbox<Message, Renderer> {
             is_checked,
             on_toggle: Box::new(f),
             label: String::from(label),
+            size: 20,
             width: Length::Shrink,
+            spacing: 15,
+            text_size: 20,
```
clarkmoody: Would be really nice to pull these from an underlying renderer.

hecrj: Yes, I think we can add this to the `Renderer` trait using associated constants.

clarkmoody:
Those changes look good to me! Let's merge.

Issue #114, opened 2019-12-07
Menu bars support plan


crlf0710:
For windowed games and casual applications, sometimes there's need to add a menu bar. What's the best practice here? Should an addon or something be created?

fenhl:
Should this be reopened given #1047?

4nte:
+1 reopen this?

wyatt-herkamp:
[Tao](https://github.com/tauri-apps/tao) hard forked winit adding menus and #124 Should Iced move the winit impl to Tao. 


Now I know this leaves open how would we implement that into glutin.

tseli0s:
May I ask what's the progress on this? Should I wait or create my own menu implementation from buttons?

breynard0:
[iced_aw](https://github.com/iced-rs/iced_aw) does have a menu implementation. Perhaps it's worth having one in iced, but that one will do until then.

karolisr:
If you can get it to work on Linux, I will be very very happy! P.S. as I mentioned in your PR, Windows support is basically painless assuming the you get the hotkey functionality to work which may be a bit difficult. It is not clear to me if you can do it with `winit` + `muda` on Linux without changing `winit`. I hope I am very wrong.

PR #195, opened 2020-02-19
Quit application when `Cmd+Q` is pressed on macOS

hecrj:
This is a nice default until we tackle #114.

### DIFF
--- a/winit/src/application.rs
+++ b/winit/src/application.rs
@@ -347,6 +347,19 @@ pub trait Application: Sized {
                     WindowEvent::CloseRequested => {
                         *control_flow = ControlFlow::Exit;
                     }
+                    #[cfg(target_os = "macos")]
+                    WindowEvent::KeyboardInput {
+                        input:
+                            winit::event::KeyboardInput {
+                                virtual_keycode:
+                                    Some(winit::event::VirtualKeyCode::Q),
+                                state: winit::event::ElementState::Pressed,
+                                ..
+                            },
+                        ..
+                    } if modifiers.logo() => {
+                        *control_flow = ControlFlow::Exit;
+                    }
                     #[cfg(feature = "debug")]
                     WindowEvent::KeyboardInput {
                         input:


cryptoquick:
Sweet! Just tested it also, works great, and I'm very glad to see this minor UX improvement.

GyulyVGC:
It's nice to have this shortcut by default on macOS, but what if I'd like to override its behaviour?
I tried setting up a subscription to capture the COMMAND+Q shortcut... it works on Windows/Linux but not on macOS.
In my opinion, `cmd+Q` should trigger a CloseRequested Event, so that a user can intercept it and perform additional operations before invoking `window::close()`.

hecrj:
@GyulyVGC  See https://docs.rs/iced/latest/iced/settings/struct.Settings.html#structfield.exit_on_close_request.

GyulyVGC:
Yeah, I noticed it and set it to false, the 'x' button of the window works as expected triggering the actions I defined, but the Cmd+Q shortcut closes the app immediately even in this case... will check better

tarkah:
Can confirm that Cmd+Q doesn't send a CloseRequested event and still force closes the app even with the exit on close requested set to false.

hecrj:
That's strange. We are only breaking here:

https://github.com/iced-rs/iced/blob/a057f8811bfc47afc4271f05b92263a19122d888/winit/src/application.rs#L411-L413

And here:

https://github.com/iced-rs/iced/blob/a057f8811bfc47afc4271f05b92263a19122d888/winit/src/application.rs#L565-L569

Maybe it's `winit` stopping the event loop?

GyulyVGC:
I take advantage of this moment to also point out that Ctrl+C on terminal doesn't trigger the event, and this could lead to unexpected behaviors if developers don't know it

GyulyVGC:
It's even weirder considering that `Cmd+Q` keeps closing the app even if those lines are commented out:
https://github.com/iced-rs/iced/blob/master/winit/src/application.rs#L599-L608

PR #200, opened 2020-02-21
Color Enhancements

clarkmoody:
A few enhancements to the color functionality. Adds the `palette` feature flag that adds dependency and tie-in to the [palette](https://github.com/Ogeon/palette) color management library.

Might wait for a [significant change](https://github.com/Ogeon/palette/pull/176) in palette to land before proceeding.

### Color

- [x] Validation of RGBA values. Limit to `[0,1]`
- [x] Inversion
- [x] Convert to/from palette's `Srgba` type

### Example App

- [x] Simple example app that dynamically generates a color palette based on your input color

Screen Shot 2020-04-27 at 4 28 12 PM


core/src/color.rs
```diff
@@ -33,11 +36,21 @@ impl Color {
...
     };
 
+    /// New Color with range checks
+    pub fn new(r: f32, g: f32, b: f32, a: f32) -> Color {
+        Color {
+            r: clamp(r),
+            g: clamp(g),
+            b: clamp(b),
+            a: clamp(a),
+        }
```
hecrj: Maybe debug assertions are a better idea? I think clamping here normally indicates a logic error from the user of the library.

core/src/color.rs
```diff
@@ -80,16 +93,123 @@ impl Color {
...
+    pub fn invert(&mut self) {
+        self.r = clamp(1.0f32 - self.r);
+        self.b = clamp(1.0f32 - self.g);
+        self.g = clamp(1.0f32 - self.b);
+    }
+
+    /// Return an inverted Color
+    pub fn inverse(self) -> Color {
+        Color::new(1.0f32 - self.r, 1.0f32 - self.g, 1.0f32 - self.b, self.a)
+    }
```
hecrj: These are useful!

core/src/color.rs
```diff
@@ -80,16 +93,123 @@ impl Color {
...
+pub struct HSLColor {
+    /// Hue, 0.0 - 360.0
+    pub h: f32,
+    /// Saturation, 0.0 - 1.0
+    pub s: f32,
+    /// Lightness, 0.0 - 1.0
+    pub l: f32,
+    /// Transparency, 0.0 - 1.0
+    pub a: f32,
+}
```
hecrj: Here we need to decide how much of color conversion logic we want to include in the library.

My idea was to offer a simple sRGB `Color` type as the library boundary and let users choose freely how to deal with color conversions ([`palette`](https://docs.rs/palette/0.5.0/palette/index.html) is pretty nice). Maybe we could add a feature gate to automatically implement conversions for `Color` from `palette`'s types.

clarkmoody: Ooh, `palette` looks nice! I didn't do too much searching for existing color libraries, but I assumed there was something out there.

Feature gate to bring in the much more advanced features is a way better approach than re-implementation ;-)

Two-way conversion would be very cool. Perhaps I'll change the direction of this branch to look into that idea.

clarkmoody: Tie-in to palette's `Srgba` in 59f6fa4718d8e416bf137967994843f2426e8ad4

manokara:
Wouldn't `palette` make more sense for the feature flag since it just enables color conversions through the library? I think naming it `colors` gives off the idea that color management in general is optional.

clarkmoody:
@manokara Renamed in 2e463fe

core/src/color.rs
```diff
@@ -80,16 +115,109 @@ impl Color {
...
+    }
+
+    #[cfg(feature = "palette")]
+    /// Convert from [`Color`] to palette's [`Srgba`] type
+    ///
+    /// [`Color`]: struct.Color.html
+    /// [`Srgba`]: ../palette/rgb/type.Srgba.html
+    pub fn into_srgba(self) -> Srgba {
+        Srgba::new(self.r, self.g, self.b, self.a)
+    }
```
hecrj, line 135: I think we can remove these two. The `From` traits should be enough.

core/src/color.rs
```diff
@@ -33,11 +39,40 @@ impl Color {
...
+        debug_assert!(
+            (0.0f32..=1.0f32).contains(&g),
+            "Green component must be on [0, 1]"
+        );
+        debug_assert!(
+            (0.0f32..=1.0f32).contains(&b),
+            "Blue component must be on [0, 1]"
+        );
+        debug_assert!(
+            (0.0f32..=1.0f32).contains(&a),
```
hecrj, line 57: I believe the compiler should be able to infer the floating-point precision here, so we may be able to remove the `f32` suffixes.

core/src/color.rs
clarkmoody: Fixed in d464d1899c9032e4e1b502ff6b9c3a86f41976c7

core/src/color.rs
clarkmoody: Fixed in daf1c639afaba989e8c4551a01fe0058ef9d1824

Ogeon:
Hey! I noticed the mentioning of my PR and just want to drop by and say that it's already possible to have conversion automatically derived in version 0.5. That includes two way conversion and `From` and `Into`. Just have a look at [`FromColor`](https://docs.rs/palette/0.5.0/palette/trait.FromColor.html) and [`IntoColor`](https://docs.rs/palette/0.5.0/palette/trait.IntoColor.html). The third example, with `CssRgb`, should be close to your use case.

The changes in 0.6 will make that feature more flexible, remove derived `From` and `Into` because of incompatibility, and also add automatic clamping after converting. Not to mention unlocking future improvements. But this only matters if you would implement `FromColor` and `IntoColor`. Otherwise you don't have to worry about it or wait for 0.6. :slightly_smiling_face:

It's cool to see Palette being integrated like this and Iced is a really interesting project! Let me know if you have any questions or requests for how to make things smoother. :grin:

clarkmoody:
Ready for review here

examples/color_palette/src/main.rs
```diff
@@ -0,0 +1,643 @@
...
+                        TextInput::new(
+                            &mut self.lch_text_state,
+                            "",
+                            &mut self.lch_text_value,
+                            |_s| Message::TextInput,
+                        )
+                        .width(Length::Units(150))
+                        .size(14)
+                        .padding(2),
+                    ),
```
hecrj, line 422: I think we could get rid of duplication if we use a `ColorPicker` struct to contain the state of the 3 sliders and we use a custom trait to abstract the specifics of each color space:

```rust
struct ColorPicker<C: ColorSpace> {
    sliders: [slider::State; 3],
    color_space: PhantomData<C>,
}

trait ColorSpace: Sized {   
    const COMPONENT_RANGES: [RangeInclusive<f32>; 3];

    fn new(a: f32, b: f32, c: f32) -> Self;

    fn components(&self) -> (f32, f32, f32);

    fn to_string(&self) -> String;
}
```

I believe this should allow us to implement `ColorPicker::view`, like so:

```rust
impl<C: ColorSpace> ColorPicker<C> {
    fn view(&mut self, color: &C) -> Element<C> {
        let (a, b, c) = color.components();
        // ...
    }
}
```

Then, in `Application::view` we could simply:

```rust
Column::new()
    // ...
    .push(self.rgb.view(&srgb).map(Message::RgbColorChanged))
    .push(self.hsl.view(&hsl).map(Message::HslColorChanged))
    // ...
```

I know this may be a bit involved! No pressure. I can try to give it a shot myself.

examples/color_palette/src/main.rs
```diff
@@ -0,0 +1,643 @@
...
+                });
+                path.line_to(Point {
+                    x: anchor.x + box_width,
+                    y: anchor.y + box_height,
+                });
+                path.line_to(Point {
+                    x: anchor.x,
+                    y: anchor.y + box_height,
+                });
+            });
```
hecrj, line 485: I added [`Path::rectangle`](https://docs.rs/iced_wgpu/0.2.1/iced_wgpu/widget/canvas/path/struct.Path.html#method.rectangle) fairly recently. Maybe it could be used to simplify some of this code.

core/src/lib.rs
```diff
@@ -35,3 +35,6 @@ pub use point::Point;
 pub use rectangle::Rectangle;
 pub use size::Size;
 pub use vector::Vector;
+
+#[cfg(feature = "palette")]
+pub use palette;
```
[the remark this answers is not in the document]
clarkmoody: Removed in 27fadad3246d555f52b991230a0352353d6700b4

examples/color_palette/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "color_palette"
+version = "0.1.0"
+authors = ["Clark Moody <clark@clarkmoody.com>"]
+edition = "2018"
+publish = false
+
+[features]
+palette = []
```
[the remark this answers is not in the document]
clarkmoody: Removed in 27fadad3246d555f52b991230a0352353d6700b4

examples/color_palette/src/main.rs
```diff
@@ -0,0 +1,643 @@
...
+                    .push(
+                        TextInput::new(
+                            &mut self.rgb_text_state,
+                            "",
+                            &mut self.rgb_text_value,
+                            |_s| Message::TextInput,
+                        )
+                        .width(Length::Units(150))
+                        .size(14)
+                        .padding(2),
```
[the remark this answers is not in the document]
clarkmoody: At the time I thought we could copy from the text field. The clipboard button will be a nice addition.

Maybe later this becomes a tool that can accept a pasted color definition (CSS hex, rgb, etc) and use it for the palette generation.

Updated in 758a444d7f11809959aa73d7da32f06e98ecc89b

examples/color_palette/src/main.rs
clarkmoody: Switched to use `rectangle` in 3e71eaee37bc3aea85feb0f643dcbd4ecc11d0c4

examples/color_palette/src/main.rs
clarkmoody: Good idea! I've made a bit of progress toward this design on my local copy. Need to figure out a couple things.

clarkmoody: Converted into something like this in 430f78a. Total lines of code actually increased 😆 , but we don't have the repetitive view code any longer.

clarkmoody:
Very cool changes @hecrj!

clarkmoody:
Aww yeah :shipit:

Issue #184, opened 2020-02-09
Purpose of `text_input::State::is_pasting`


Finnerale:
While trying to understand the architecture of Iced I started implementing selecting text for the native `TextInput`.
When I wanted to implement copying to the clipboard I got irritated about what the purpose of [`is_pasting`](https://github.com/hecrj/iced/blob/95880ca74bddb6a23774621ef766b91956d40a61/native/src/widget/text_input.rs#L503) is as it seems to get cleared right after being set.
Does it have some purpose or can it be ignored / removed?

hecrj:
Currently, it's used to ignore any received characters while a user keeps pasting text (for instance, if a user keeps `Ctrl + V` pressed for a while). However, I think we should just cancel the pasting completely in that case.

That said, text selection is tricky and may involve changing the `State` considerably. For instance, the scrolling offset of a `TextInput` is currently determined from the cursor position alone. However, most toolkits implement a stateful offset that only changes when the cursor is close to the edges of the clipping area. Text selection will force us to implement this and unify it somehow.

Additionally, if we are going to add text selection complexity it may be good idea to have in mind a multi-line text input or `TextArea` and implement a general solution instead.

hecrj:
Oh, I forgot it is also used to cache the clipboard contents. Accessing the clipboard can be quite slow.

I took a look at your branch and looks very promising! :tada:

hecrj:
Closing as #202 landed.

PR #202, opened 2020-02-23
Text Selection for text_input widget

FabianLars:
TL;DR Based on Finnerales fork, still WIP but looking for tips and/or wishes.

First of all, this is based on work done by @Finnerale (#184). It's the biggest and imo most difficult part of what has been done in this pull request (as of now).
My problem was, that I wasn't patient enough to wait any longer and i wanted this functionality for my side project. So although I'm an absolute beginner regarding Rust/GUI/Collab Work, I thought I'd give it a go.

What's implemented:
* Selection via Mouse dragging (by Finnerale)
* Ctrl + A or 3 clicks to select everything
* double click to select a word

What's left afaik:
* Shift + Left/Right = Expand selection by one character
* Ctrl + Shift + Left/Right = Expand selection by one "word"
* Shift + Home/End = Expand selection to Start/End
(* I really need to work on Comments/Documention huh -> should be added tomorrow)

So most importantly I'd like to know if i did something wrong or inappropriate (regarding the code(style)) and if there is something i forgot to add.
Oh and also tell me if this is a waste of time for you (or me).

I'm gonna post this as a draft pull request, because it's obviously still WIP :)

P.S. Yea, i noticed the failing builds :P


FabianLars:
Hey, thanks for your response.

To be honest that's already slightly over my head, but i really want to get it done, so I'll need a little time to really get into it (as I mentioned I don't have much experience). Especially the type opacity thingy is something I need to learn more about... (I'd appreciate any hints or something in the code base :P)

Btw, I'm also on Discord (+rust-lang server) and your Zulip server, if you want to reach out to me somewhere else.

hecrj:
> To be honest that's already slightly over my head, but i really want to get it done, so I'll need a little time to really get into it

Sorry, I didn't mean to overwhelm you! There is no rush. I think we have done the easy part here, which is code. The hard part is figuring out how the new code fits with the older code: the design.

This is the hardest part because we need to be able to see the big picture and the details of the problem at the same time. However, the changes here are very local, which should make our lives easier when it comes to simplifying things.

> Especially the type opacity thingy is something I need to learn more about

Think about it like encapsulation. A type where all its internals are opaque and can only be built and mutated in a controlled manner through its exposed methods. Ideally, these methods should be meaningful (no setters!) and intuitive, representing clear use cases (`move_left`, `move_right`, `select`...). In case you may find it useful, I wrote an article about [opaque types in Elm](https://dev.to/hecrj/use-opaque-types-in-elm-3oal).

> (I'd appreciate any hints or something in the code base :P)

#95 is a good example of how some logic can be simplified by introducing an additional type. I think something similar may apply here. It is hard to give you any more hints because every problem is unique! Exploration is part of the solution.

---

I know I am very nitpicky when it comes to design and simplicity. But I really believe that investing extra effort and time in making things simple is worth it in the long run.

I am also aware that I may be asking too much, so no pressure! This is already great! I should be able to eventually take it to the finish line. It may take some time, though, as there are other features  and PRs I want to tackle first.

That said, if you still want to give it a shot, I will be glad to help and discuss the details on Discord or Zulip. In fact, [this is the first thing I encourage contributors to do](https://github.com/hecrj/iced/blob/master/CONTRIBUTING.md)!

Finnerale:
I currently don't have the time to work on it, so thanks for continuing this :+1: 

It might be interesting to take a look at how [druid](https://github.com/xi-editor/druid/tree/master/druid/src/text) deals with cursors and editable text, as they did it in a way more abstract / general approach than I did.

Especially it could be worth to allow backing text edit components by something like [ropey](https://github.com/cessen/ropey) as it would likely perform much better than a String for long text.

FabianLars:
So this all more or less happened while trying out different stuff to further get into this part of the codebase and the stuff you told me about, so it's still far from done (more or less the same features as previous commit but i guess it needs a huge cleanup), but I think it's already better than the previous version.

My question is if this approach is any better in your eyes any (or actually somewhat worth doing).

(And yes, i wanted to reach out to you about it before investing that much time, but i just got lost or sth, idk ¯\_(ツ)_/¯ )

yaskhan:
Maby add some shortcut keys util class? Like: [https://docs.sencha.com/extjs/6.5.3/classic/Ext.util.KeyMap.html](https://docs.sencha.com/extjs/6.5.3/classic/Ext.util.KeyMap.html)

native/src/widget/text_input.rs
```diff
@@ -1041,3 +841,182 @@ mod platform {
...
+mod cursor {
+    use crate::widget::text_input::Value;
+    use iced_core::Point;
+    use std::time::{Duration, SystemTime};
+
+    /// Even the compiler bullies me for not writing documentation
+    #[derive(Debug, Copy, Clone)]
+    pub struct Cursor {
+        start: usize,
+        end: usize,
```
hecrj: I liked the idea of an enum to represent the cursor state. We should be to create a `cursor::State` enum and maybe expose it through a `Cursor::state` method. I think this should simplify the `Cursor` API a bit and improve/remove some of its methods.

native/src/widget/text_input.rs
```diff
@@ -1041,3 +841,182 @@ mod platform {
...
+            let (a, b) = self.cursor_position(value);
+            a.min(b)
+        }
+
+        pub fn cursor_position_right(&self, value: &Value) -> usize {
+            let (a, b) = self.cursor_position(value);
+            a.max(b)
+        }
+
+        pub fn draw_position(&self, value: &Value) -> usize {
```
hecrj: It doesn't make too much sense to talk about drawing in this context. We should find a better name.

native/src/widget/text_input.rs
```diff
@@ -347,19 +308,18 @@ where
...
+                    self.value.remove_many(
+                        self.state.cursor.left(),
+                        self.state.cursor.right(),
+                    );
+                    self.state.cursor.move_left();
+                    self.value.insert(self.state.cursor.end(), c);
                 }
-                self.state.move_cursor_right(&self.value);
+
+                self.state.cursor.move_right(&self.value);
```
hecrj: `Cursor` and `Value` are very coupled when we perform text edition. They always change at the same time.

I think it may be a good idea to play with the concept of an `Editor` (name subject to change). Maybe something like:

```rust
struct Editor<'a> {
    value: &'a mut Value,
    cursor: &'a mut Cursor,
}
```

And then implement all the text editing methods in this context. Then, whenever we need to edit a value, we will only be able to do it through an `Editor`, which will take care of the `Cursor` for us too.

native/src/widget/text_input.rs
```diff
@@ -650,77 +596,11 @@ pub struct State {
     is_focused: bool,
     is_pressed: bool,
     is_pasting: Option<Value>,
-    cursor_position: Cursor,
-    /// Double- / Tripleclick
-    click_count: usize,
-    last_position: Option<Point>,
-    last_timestamp: Option<std::time::SystemTime>,
+    /// TODO: Compiler wants documentation here
+    pub cursor: cursor::Cursor,
```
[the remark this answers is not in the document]
FabianLars: Well that's what i tried at first but somehow fucked up. This is one of those things i meant with cleanup. I'll just try again.

native/src/widget/text_input.rs
```diff
@@ -1041,3 +841,182 @@ mod platform {
         }
     }
 }
+
+mod cursor {
```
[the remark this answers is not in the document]
FabianLars: where would i want to place it? It surely feels wrong at natve/src/widget :P

native/src/widget/text_input.rs
```diff
@@ -1041,3 +841,182 @@ mod platform {
...
+        pub fn select_all(&mut self, len: usize) {
+            self.start = 0;
+            self.end = len;
+        }
+        /* Selection section end */
+
+        /* Double/Triple click section */
+        // returns the amount of clicks on the same position in specific timeframe
+        // (1=double click, 2=triple click)
+        pub fn process_click(&mut self, position: Point) -> usize {
```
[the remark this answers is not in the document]
FabianLars: Yea i tried to do something with the Interaction enum you mentioned in your first post, but it didn't work out that well. But I like your state idea.

native/src/widget/text_input.rs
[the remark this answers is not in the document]
hecrj: I would place it in: `native/src/widget/text_input/cursor.rs`

native/src/widget/text_input.rs
FabianLars: cleanup section again. I chose this name just to understand what the renderer wants (based on Finnerales changes). So yes, i want to rename it too or maybe even replace/remove it somehow (because its just used by renderer/widget/text_input).

native/src/widget/text_input.rs
FabianLars: Hm, i guess that's a good idea huh. I'll look into it.

native/src/widget/text_input.rs
FabianLars: So more or less like the first try (Cursor::Index and Cursor::Selection)? Because that sometimes felt a little bit awkward to work with and ended up with more code to use it (always used with match).
Because of that I just ditched it and always use start & end positions and is_selection to check if they are both on the same position.

So i don't know about the simpler API, but you're right about methods needing improvements (->cleanup :P) and that's needed even without your state changes.

But you're the more experienced so I guess I should give it a try

hecrj: > So more or less like the first try (Cursor::Index and Cursor::Selection)?

Yes, but nested here without exposing it directly.

> Because that sometimes felt a little bit awkward to work with and ended up with more code to use it (always used with match).

It may be more verbose, but it made things more explicit. It doesn't make sense to have a `start` and `end` when you do not have an active selection.

native/src/widget/text_input.rs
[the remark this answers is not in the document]
FabianLars: Follow up question. Would I add a new field to the State in text_input (I called it "click_state" for now) with the created crate::input::mouse::State? So we got something like this:
`pub struct State {
    is_focused: bool,
    is_pressed: bool,
    is_pasting: Option<Value>,
    cursor: Cursor,
    click_state: crate::input::mouse::State,
}`
with mouse::State(mouse.rs) looking something like this:
`mod button;
mod event;

pub use button::Button;
pub use event::{Event, ScrollDelta};
use std::time::{Duration, SystemTime};

#[derive(Debug, Copy, Clone)]
pub struct State {
    // Maybe we should track which button (left/right) was clicked. don't know if needed
    click_count: usize,
    last_click_position: Option<crate::Point>,
    last_click_timestamp: Option<SystemTime>,
}

// same as before so I'll leave it empty for this comment
impl Default for State {[...]}

impl State {
    // same as before so I'll leave it empty for this comment
    pub fn process_left_click(&mut self, position: crate::Point) -> usize {
        [...]
    }
}`

This feels kinda imperfect right?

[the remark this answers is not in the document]
hecrj: I would call the field `mouse`.

Then, maybe rename `process_left_click` to `update` and make it:

```rust
pub fn update(&mut self, event: Event) -> Option<Interaction>
```

`Interaction` can be an enum:

```rust
pub enum Interaction {
    Click(Point),
    DoubleClick(Point),
    TripleClick(Point),
}
```

Just brain-storming, this may need a bit more thinking. Feel free to fine-tune it to your liking. I will change any small details before merging.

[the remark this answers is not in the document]
hecrj: Also, don't feel forced to support more use cases for now. Let's keep it coupled to the `TextInput` needs. We can iterate further in the future.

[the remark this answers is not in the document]
FabianLars: Yeah I keep it coupled to TextInputs needs for now.
I implemented update as `pub fn update(&mut self, position: Point) -> Option<Interaction>` for the next iteration instead of using `event: Event` because I didn't know how else to get the position.

FabianLars:
I played around with the things you suggested, but left out the Editor struct for now (or at least for this iteration). -> See Commit Message

let me know if you see something standing out. (I really need to work on my english skills lol)

Thanks for being so patient with me btw :)

FabianLars:
I changed it from Draft to normal pull request because I implement the last missing select actions (tell me if there are more), expect for copying into Clipboard but as far as i can see this is impossible with current Clipboard lib? (edit: i just ran cargo update and it mentioned a clipboard repo so i'll look into it again).

I'll see if i can clean up things further and finally take a look at your Editor suggestion if still needed (wanted to implement the missing keyboard actions first to see what the Editor needs to handle).

native/src/widget/text_input.rs
FabianLars: What exactly should be the scope of the Editor?
Should it be used only where we edit a value (delete/insert), so that text_input still has a value field and a cursor in its state.
Or should it handle the text selection too, so that we could replace the cursor with the editor in the text_input state?
Or does it make sense to somehow move the Value completely into the editor (if this is even possible) so that we use the editor for everything related to the value, for example just getting the actual value of it (on top of editing/setting it)

FabianLars: Alright, so I wasn't really able to get something done that either just works or actually makes sense. Because of that I'll close this Conversation thingy here for now. But I'm still open for suggestions or "bigger" hints in the right direction if you still want to have such thing implemented (maybe my problem here is the notation of the struct in your example idk). I mean we can still talk about it or you can ofc implement it yourself :P

hecrj:
No worries! I think I can take it to the finish line from here.

I will let you review the changes once I get a chance to work on this.

Thank you for all the work so far!

FabianLars:
Thank you for actually giving me a chance to do this and helping me along the way! If you change your mind and see something you want me to do for this pr, just lemme know. (if this would be the editor thingy, we had to talk some more in depth about it i guess :P)

I hope there's something more to do for me in the future. (Hopefully I'm a more experienced rust user by then, so it'll be easier to work with me :P)

P.S. I'll push a really small commit with changes i forgot.

Restioson:
An assertion fails if the `value` passed to the textbox doesnt change and some text is typed. See the following example:

```rust
use iced::{Element, Sandbox, Settings, TextInput, text_input, Column};

pub fn main() {
    App::run(Settings::default())
}

#[derive(Default)]
struct App {
    state: text_input::State,
    value: String,
}

impl Sandbox for App {
    type Message = String;

    fn new() -> Self {
        Self::default()
    }

    fn title(&self) -> String {
        String::from("Reproduction")
    }

    fn update(&mut self, new: String) {
        // self.value = new;
    }

    fn view(&mut self) -> Element<String> {
        Column::new()
            .spacing(20)
            .push(
                TextInput::new(&mut self.state, "a", &self.value, |v| v)
                    .padding(10)
                    .size(30)
            )
            .into()
    }
}
```

After uncommenting this line, it works as expected. The panic is as follows:

```
thread 'main' panicked at 'assertion failed: index <= len', <::core::macros::panic macros>:3:10
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
...
  15: iced_native::widget::text_input::Value::insert
             at native/src/widget/text_input.rs:769
  16: <iced_native::widget::text_input::TextInput<Message,Renderer> as iced_native::widget::Widget<Message,Renderer>>::on_event
             at native/src/widget/text_input.rs:327
  17: <iced_native::widget::column::Column<Message,Renderer> as iced_native::widget::Widget<Message,Renderer>>::on_event::{{closure}}
             at native/src/widget/column.rs:161
```

Restioson:
Found another issue (definitely bug) with the same example (two for one, heh). If you type a character and press backspace twice (or two and backspace twice), it will panic with `attempted to subtract with overflow`:

The relevant code appears to be here:
```rust
keyboard::KeyCode::Backspace => {
    match self.state.cursor.selection_position() {
        ...
        None => {
            if self.state.cursor.start() > 0 {
                self.state.cursor.move_left();
                let _ = self
                    .value
                    .remove(self.state.cursor.start() - 1); // <--- HERE
            }
        }
    }
    ...
}
```

I think that that that line should be changed to `.remove(self.state.cursor.start() - 1)`, as the `move_left` will already remove 1 from the `start`. Changing this line to that appears to fix the problem. I'm not sure if it causes any other side effects though.

FabianLars:
Hey, thanks for reporting these bugs! I'll push 2 small changes which should fix these issues, hoping that hecrj didnt work on this pr locally already lol

Restioson:
Np. What do you think about the first issue (not a bug per se)?

FabianLars:
Well, the crashes in 1) itself are fixed with these changes (although there might be a better solution), so that it can't panic if you're not updating the value. This should more or less represent the behaviour in the current master. And manually updating values after a Message is a requirement for every widget, not just for text_input.

Edit: Oh well, you meant the last part about programatically setting the value via `self.value = "blabla".to_string()`, huh? Then the cursor would be at 0 or 1 or something, which would be kinda meh indeed, but I'm not sure if there are actually use cases for this where you would update the value this way while it's focused/being used.
Tbh, I'll just wait for hecrj's opinion :P

P.S. Oh man, I'm too tired to write understandable sentences in english D:

hecrj:
The value provided to a `TextInput` can change at any time. There is no consistency guarantee that we can rely on. The `State` should be completely fault-tolerant and __never__ panic, independently of what the user/application code does.

Therefore, clamping with `min` and `max` is the way to go here.

FabianLars:
Absolutely, that's what the commit was about (although the pr might need more improvements regarding this). So it shouldn't panic anymore but the cursor behaves not as expected (from the users POV) when you change the value programatically in the apps update function. But I'm not sure if that's really needed and if so how to implement a fix.

I hope I actually understood both of you correctly...

hecrj:
No need to worry about that use case for the time being, although I would expect the cursor position to stay in the same place even if the value of the input changes. Isn't this how it behaves currently?

FabianLars:
Yes, that's the current behaviour. But with the last commit if you start with an empty input and later change it, the cursor might be at position 1 instead of 0 because of the way I implement the fix. (If you do what caused the program to panic before == typing into an input that doesn't update its value)

FabianLars:
Looks great!
I have nothing to add at this point, so I'm ready to see it getting merged.

Most commits seem pretty straight forward (or better: were easily understandable to me). This should help me doing better PRs in the future.

And the editor implementation looks good and actually way easier than I expected it to look like :D

hecrj:
Awesome! Let's merge this! Thanks again, everyone. :beers:

PR #210, opened 2020-02-28
Mention OpenGL/Vulkan as install depedencies

SyrianSpock:


hecrj:
Thanks for the suggestion. I haven't added this yet for a couple of reasons.

Many dependencies differ based on the platform you are targetting. For instance, on macOS you need Metal while on Windows you can use DirectX 12. There is also a bunch of additional dependencies needed for Linux like fontconfig, X11, and more. Also, we do not use OpenGL for rendering yet, so it shouldn't be necessary.

Because of this, I think a proper solution would be an additional `INSTALLATION.md` document with an exhaustive list and then link to it from the `README`. However, the library is currently in an experimental stage and is meant to be used by advanced users. I believe this works as an entry barrier of sorts.

I would like to wait before we commit to maintaining a dependency list.

SyrianSpock:
I can remove the install instructions and simply warn about OpenGL/Vulkan/Metal/DirectX

hecrj:
I think that is already mentioned in the features section.

SyrianSpock:
True, I missed it and wasted 2 days on a non-issue, thanks for pointing that out :+1:

hecrj:
Sorry to hear that. We have [a Zulip server](https://iced.zulipchat.com/) where any questions and feedback are welcome. Of course, opening an issue is also an option (don't worry about duplicates!).

PR #211, opened 2020-02-29
Add support for loading already-decoded image pixels

mrkgnao:
This solves a use-case I have (an app with an image gallery) where I decode images asynchronously using `Commands`.

### DIFF
--- a/native/src/widget/image.rs
+++ b/native/src/widget/image.rs
@@ -125,7 +125,9 @@ impl Handle {
         Self::from_data(Data::Path(path.into()))
     }
 
-    /// Creates an image [`Handle`] containing the image pixels directly.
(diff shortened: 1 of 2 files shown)


native/src/widget/image.rs
```diff
@@ -125,6 +125,19 @@ impl Handle {
         Self::from_data(Data::Path(path.into()))
     }
 
+    /// Creates an image [`Handle`] containing the image pixels directly.
+    ///
+    /// This is useful if you have already decoded your image.
+    ///
+    /// [`Handle`]: struct.Handle.html
```
hecrj: We should probably mention the BGRA format requirement here too.

mrkgnao: Done!

PR #214, opened 2020-03-07
Implement text support in canvas widget

artursapek:
This PR implements a new method `Frame.text` which allows a canvas user to render text to the canvas frame.

### DIFF
--- a/wgpu/src/primitive.rs
+++ b/wgpu/src/primitive.rs
@@ -78,7 +78,18 @@ pub enum Primitive {
         origin: Point,
 
         /// The vertex and index buffers of the mesh
-        buffers: Arc<triangle::Mesh2D>,
+        buffers: triangle::Mesh2D,
+    },
+    /// A cached primitive.
+    ///
+    /// This can be useful if you are implementing a widget where primitive
+    /// generation is expensive.
+    Cached {
+        /// The origin of the coordinate system of the cached primitives
+        origin: Point,
+
+        /// The cached primitive
+        cache: Arc<Primitive>,
     },
 }
 
--- a/wgpu/src/renderer.rs
+++ b/wgpu/src/renderer.rs
@@ -10,7 +10,6 @@ use iced_native::{
     layout, Background, Color, Layout, MouseCursor, Point, Rectangle, Vector,
     Widget,
 };
-use std::sync::Arc;
 
 mod widget;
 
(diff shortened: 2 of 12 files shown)


wgpu/src/primitive.rs
```diff
@@ -78,7 +78,18 @@ pub enum Primitive {
...
+    /// A cached primitive.
+    ///
+    /// This can be useful if you are implementing a widget where primitive
+    /// generation is expensive.
+    Cached {
+        /// The origin of the coordinate system of the cached primitives
+        origin: Point,
+
+        /// The cached primitive
+        cache: Arc<Primitive>,
```
hecrj: I ended up implementing the `Primitive::Cached` as I think it may be useful for other widgets if we want to start saving allocations.

wgpu/src/renderer.rs
```diff
@@ -137,17 +131,16 @@ impl Renderer {
 
     fn draw_primitive<'a>(
         &mut self,
+        translation: Vector,
```
hecrj: It turns out that most of the plumbing was already there.

I basically removed the `offset` field from `Layer` and added an explicit translation for each `Primitive`. This allows us to control the translation of children without worrying about unintended side-effects.

wgpu/src/renderer.rs
```diff
@@ -222,11 +214,13 @@ impl Renderer {
...
-                // TODO: Move some of this computations to the GPU (?)
+                let layer = layers.last_mut().unwrap();
+
+                // TODO: Move some of these computations to the GPU (?)
                 layer.quads.push(Quad {
                     position: [
-                        bounds.x - layer.offset.x as f32,
-                        bounds.y - layer.offset.y as f32,
+                        bounds.x + translation.x,
+                        bounds.y + translation.y,
```
hecrj: We were already applying `layer.offset` manually anyways for most primitives, except `Image` and `Svg`.

I think this is nicer. We also remove a bunch of conversions, so we may be saving some cycles.

wgpu/src/widget/canvas/frame.rs
```diff
@@ -154,6 +156,52 @@ impl Frame {
...
+
+        let position = if self.transforms.current.is_identity {
+            text.position
+        } else {
+            let transformed = self.transforms.current.raw.transform_point(
+                lyon::math::Point::new(text.position.x, text.position.y),
+            );
+
+            Point::new(transformed.x, transformed.y)
+        };
```
hecrj: I have made the text position obey the current transform of the `Frame` and added a warning of the current limitations of text rendering.

artursapek:
:+1:

PR #220, opened 2020-03-16
Instantiate Column and Row with children

rowungiles:
Instantiate a Column or a Row with children. A compliment to the existing `push` functions.

Duplicates the default values of the `new` function, save for `children`.


rowungiles:
Great changes. Thank you 🎉

PR #224, opened 2020-03-18
Pane grid widget

hecrj:
[Pane grid - Iced](https://gfycat.com/mixedflatjellyfish)

This PR introduces a new widget: `PaneGrid`.

A `PaneGrid` consists of a collection of panes, which are distributed using either vertical or horizontal splits to completely fill the space available. This distribution of space is common in tiling window managers (like [`awesome`](https://awesomewm.org/), [`i3`](https://i3wm.org/), or even [`tmux`](https://github.com/tmux/tmux)).

The new `PaneGrid` features:

* Vertical and horizontal splits
* Tracking of the last active pane
* Mouse-based resizing
* Drag and drop to reorganize panes
* Hotkey support
* Configurable modifier keys
* API to perform actions programmatically (`split`, `swap`, `resize`, etc.)

I believe this widget can be very useful if you want to allow your users to dynamically customize the layout of the application. For instance, it could be used to build a [three-pane interface](https://en.wikipedia.org/wiki/Three-pane_interface), a dashboard with different kinds of data visualization, etc.

Additionally, I have:

* Added a couple of new variants to `MouseCursor` for the new resizing interactions.
* Moved some reusable `keyboard` types to `iced_core`, as we will need them in `iced_web` soon enough.
* Implemented `Subscription::with`, which can be used to move external data into a subscription.

As always, I have created an example (`pane_grid`) to showcase all the current features of the new widget.


Songtronix:
Holy moly! This is awesome!

core/src/point.rs
```diff
@@ -22,6 +22,16 @@ impl Point {
...
     }
+
+    /// Computes the distance to another [`Point`].
+    ///
+    /// [`Point`]: struct.Point.html
+    pub fn distance(&self, to: Point) -> f32 {
+        let a = self.x - to.x;
+        let b = self.y - to.y;
+
+        f32::sqrt(a * a + b * b)
```
clarkmoody: Can also use the built-in [`hypot` method](https://doc.rust-lang.org/std/primitive.f32.html#method.hypot)

Return value would be `a.hypot(b)`

hecrj: Oh, nice one. Thank you!

PR #228, opened 2020-03-20
Remove excess whitespace from end of line to comply with `cargo fmt`.

maaku:
My trivial first commit... I noticed that we are *almost* `cargo fmt` compliant. There's just two lines with extra whitespace at the end. This PR removes that whitespace.

### DIFF
--- a/native/src/widget/container.rs
+++ b/native/src/widget/container.rs
@@ -77,7 +77,7 @@ where
         self.max_height = max_height;
         self
     }
-    
+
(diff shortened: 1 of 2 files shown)


hecrj:
Oh, good catch. Thank you!

My editor runs `cargo fmt` automatically on save, so these were probably introduced by some contributors.

I think it's probably a good idea to check this in CI, at least as a warning for future contributions. Let me see if I can set it up quickly.

PR #235, opened 2020-03-26
Layers

JCapucho:
This PR aims to implement layers.
The way I did is to have the layering be relative to the last widget on the same level so for example a column with two images overlapping, if the `Depth` is set `None` on both then with will be order dependent, if the `Depth` is set as `Above` on the first then it will be order dependent, if it's set on the second one then the second image will be displayed on top, if the first is set to `Below` then the first will be displayed, if it's set on the second then it's order dependent. An exception is the `Topmost` which is global.
It might not be the proper way to implement so I'm available to change it


hecrj:
Hey, thanks for taking a shot at this! I appreciate the effort!

I see you have tackled layering at the rendering level, but the main challenge with layers is related to event processing! For instance, when displaying a button on top of another interactive element the events should be captured captured by the button and not by the element behind it.

This is the main challenge that we need to solve elegantly. I talked about this topic in #30 and [discussed it in our Zulip server very recently](https://iced.zulipchat.com/#narrow/stream/213316-general/topic/layers). We are not ready to write code for this yet, but we are getting close!

Before tackling an important feature like this one, I'd really prefer to hear from you first! As the [`CONTRIBUTING` guide](https://github.com/hecrj/iced/blob/master/CONTRIBUTING.md) mentions, it helps us to coordinate work, get on the same page, and start building trust.

It is very hard to accept a solution to a problem that we have been discussing for a while without much context. Specifically, there are two very important questions that we'd need to answer (extracted from [`rust-lang/rfcs`](https://github.com/rust-lang/rfcs)):

- Why is this design the best in the space of possible designs?
- What other designs have been considered and what is the rationale for not choosing them?

These questions tend to be answered organically with collaboration and discussion. Ideally before a full implementation!

JCapucho:
Sorry hadn't noticed the Contributing guide and didn't think of event processing. Maybe I should close this PR what is your opinion?

hecrj:
> Sorry hadn't noticed the Contributing guide and didn't think of event processing.

No worries! Thank you for spending time working on this :bowing_man: 
> Maybe I should close this PR what is your opinion?

I don't mind keeping it open until we settle on a design. I think there are some interesting ideas here (I like the relative `Depth` enum, reminds me of [`elm-ui`](https://package.elm-lang.org/packages/mdgriffith/elm-ui/latest/Element#above)) that may be useful depending on what we end up doing.

Please, feel free to join [the discussion](https://iced.zulipchat.com/#narrow/stream/213316-general/topic/layers)!

PR #242, opened 2020-03-28
Some more code cleanups.

kaimast:
It might sense to actually run cargo clippy in the future. There's a bunch of other stuff that would require larger code changes so I did not touch that...


hecrj:
> It might sense to actually run cargo clippy in the future.

I like to run it every once in a while. I feel it helps in many situations but sometimes it's a bit too pedantic for my tastes.

I have fixed some of the lints left. I believe the ones remaining are either false positives or very subjective.

Thank you for taking a look! We can merge this :clinking_glasses:

PR #245, opened 2020-03-29
refactor: remove not mandatories 'static lifetimes

tirz:
Some Widgets have intensive use of 'static lifetime (for the `From` trait) which does not seems to be necessary.


hecrj:
Awesome. Thank you!

Did you run into any particular issues because of the stricter-than-necessary lifetimes?

tirz:
`Did you run into any particular issues because of the stricter-than-necessary lifetimes?`
Not really, I just checked out how was implemented the currents widgets and I was curious about the `'static`.

PR #247, opened 2020-03-30
Implement `svg::Handle::from_memory`

### LINKED ISSUE #206: svg api explicit from bytes
This ticket just some clarification about svg, and maybe more explicit api.

So backstory, I noticed svg render was empty sometimes; i then realized i was passing path with typo, and svg (presumbly) wasn't loaded, but no error.

Ideally, I handle loading raw bytes of svg once with regular file into `vec<u8>`, and pass this to svg api constructor. Maybe similar to #76 ?

I only see path api in svg though, unless I'm missing it? This also makes me wonder, is svg loaded from disk by iced and cached somewhere? I prefer to be in control of resource loading, so:

1. can we have explicit `from_bytes` (or whatever) api for handling resource loading ourselves
2. less ideal: can we add error (or panic at least) if file not found for svg api (probably error is harder because the actual loading is done by something else in iced?)

and thanks for great library :))))

hecrj:
Fixes #206.

### DIFF
--- a/examples/svg/Cargo.toml
+++ b/examples/svg/Cargo.toml
@@ -7,4 +7,3 @@ publish = false
 
 [dependencies]
 iced = { path = "../..", features = ["svg"] }
-env_logger = "0.7"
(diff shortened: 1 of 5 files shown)


dyedgreen:
The api looks exactly like what I would have wanted. 🤩

PR #249, opened 2020-03-31
Add Ctrl+Backspace text field behaviour

michael-swan:
Self explanatory.


michael-swan:
Oh one minor mistake, I should probably make it select to the beginning from the current cursor, not select all in the private case.

michael-swan:
Unless there is a selection of any text (secure or not), in which case Ctrl+Backspace should perform the same function as backspace.

michael-swan:
It should now be all set. @hecrj

PR #259, opened 2020-04-04
reference feature flags in docs

robjtede:
Makes it clearer when browsing docs which feature flags are needed for certain features using a docsrs config flag.

Test locally using:
```sh
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --workspace --lib --no-deps --all-features --open
```

eg: 
2020-04-04_02 26 46-e99bc6aa

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,3 +65,7 @@ iced_wgpu = { version = "0.2", path = "wgpu" }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 iced_web = { version = "0.2", path = "web" }
+
+[package.metadata.docs.rs]
+rustdoc-args = ["--cfg", "docsrs"]
+features = ["image", "svg", "canvas"]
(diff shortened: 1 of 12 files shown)


hecrj:
I like the idea!

I believe it's a bit noisy when marking whole modules, as mentioned in https://github.com/rust-lang/rust/issues/43781#issuecomment-568829501.

Maybe we could settle with a note in the module documentation for the time being, while the feature matures.

robjtede:
Yeah, honestly the reason I didnt do everything yet is incase you said that. Project like Tokio has gone ahead with this approach: https://docs.rs/tokio/0.2.16/tokio/fs/index.html, https://docs.rs/tokio/0.2.16/tokio/fs/struct.File.html.

hecrj:
> Should do something to make the features clearer though. Did trip me up initially when starting with the canvas module. Note in the module doc would be fine, too.

Agreed. I believe it's better to be too noisy than failing to warn our users properly. I also feel the warning blends in nicely once you get used to it. And if it's too distracting, you can always run `cargo doc` yourself.

So let's keep it! I added the message to the widgets in the root crate and enabled the feature flags for `docs.rs`. Let me know if I screwed something up!

robjtede:
LGTM

hecrj:
Awesome. Thank you! :clinking_glasses:

PR #260, opened 2020-04-04
Checkbox label is now `Into<String>`

0x7CFE:
This PR allows checkbox labels to be dynamically modified depending on the business logic.

For example:
image

```rust
use iced::{Sandbox, Settings, Text, Align, Column};

fn main() {
    CheckUi::run(Settings::default());
}

#[derive(Default)]
struct CheckUi {
    checks: Vec<bool>,
}

#[derive(Debug, Copy, Clone)]
enum CheckMessage {
    InputChecked(usize, bool),
}

impl Sandbox for CheckUi {
    type Message = CheckMessage;

    fn new() -> Self {
        Self {
            checks: (0..10).map(|_| false).collect(),
            ..Default::default()
        }
    }
    
    fn title(&self) -> String { "Dynamic checkbox label".into() }
    
    fn update(&mut self, message: Self::Message) {
        match message {
            CheckMessage::InputChecked(index, value) => self.checks[index] = value,
        }
    }

    fn view(&mut self) -> iced::Element<'_, Self::Message> {
        let mut inputs = Column::new()
            .height(iced::Length::Fill)
            .width(iced::Length::Fill)
            .align_items(Align::Start);

        for (index, checked) in self.checks.iter().enumerate() {
            inputs = inputs.push(
                iced::Checkbox::new(
                    *checked,
                    format!("Checkbox {}", index),
                    move |state| CheckMessage::InputChecked(index, state)
                )
                .width(iced::Length::Fill)
            );
        }

        inputs.into()
    }
}
```

### DIFF
--- a/examples/tour/src/main.rs
+++ b/examples/tour/src/main.rs
@@ -528,10 +528,9 @@ impl<'a> Step {
             .push(Language::all().iter().cloned().fold(
                 Column::new().padding(10).spacing(20),
                 |choices, language| {
-                    let label: &str = language.into();
                     choices.push(Radio::new(
                         language,
-                        label,
+                        language,
                         selection,
                         StepMessage::LanguageSelected,
                     ))
@@ -730,16 +729,16 @@ impl Language {
     }
 }
 
-impl From<Language> for &str {
-    fn from(language: Language) -> &'static str {
-        match language {
+impl From<Language> for String {
+    fn from(language: Language) -> String {
+        String::from(match language {
             Language::Rust => "Rust",
             Language::Elm => "Elm",
             Language::Ruby => "Ruby",
             Language::Haskell => "Haskell",
             Language::C => "C",
             Language::Other => "Other",
-        }
+        })
     }
 }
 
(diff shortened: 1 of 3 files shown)


0x7CFE:
> Could you do the same for the `Radio` widget?

Done. 

By the way, would you mind refactoring the text in the same manner as it is done for normal push button? I mean so that Checkbox would accept `content` instead of `label`. For example that would allow editing text styles without need to delegate a ton of methods.

When working on my project I already hacked an option to set the color of the text.

examples/tour/src/main.rs
```diff
@@ -730,16 +729,16 @@ impl Language {
     }
 }
 
-impl From<Language> for &str {
-    fn from(language: Language) -> &'static str {
-        match language {
+impl From<Language> for String {
+    fn from(language: Language) -> String {
+        String::from(match language {
```
0x7CFE, line 734: If I'm not mistaken, such variant could be implemented with just `#[derive(Display)]` and `language.to_string()`.

hecrj: Yes, that'd be cleaner. I'll change the example. Thanks!

PR #267, opened 2020-04-06
impl default for canvas cache

robjtede:
Previously needed to write the full Default implementation:

```rust
#[derive(Debug)]
pub struct AppState {
    all_data: State,
    current_frame_data: D,
    current_frame_canvas_cache: canvas::layer::Cache<D>,
}

impl Default for AppState {
    fn default() -> Self {
        Self {
            all_data: Default::default(),
            current_frame_data: Default::default(),
            current_frame_canvas_cache: iced::widget::canvas::layer::Cache::new(),
        }
    }
}
```

Now deriving a default application state is possible:
```rust
#[derive(Debug, Default)]
pub struct AppState {
    all_data: State,
    current_frame_data: D,
    current_frame_canvas_cache: canvas::layer::Cache<D>,
}
```

### DIFF
--- a/wgpu/src/widget/canvas/layer/cache.rs
+++ b/wgpu/src/widget/canvas/layer/cache.rs
@@ -6,6 +6,19 @@ use crate::{
 use iced_native::Size;
 use std::{cell::RefCell, marker::PhantomData, sync::Arc};
 
+enum State {
+    Empty,
+    Filled {
+        bounds: Size,
+        primitive: Arc<Primitive>,
+    },
+}
+
+impl Default for State {
+    fn default() -> Self {
+        State::Empty
+    }
+}
 /// A simple cache that stores generated geometry to avoid recomputation.
 ///
 /// A [`Cache`] will not redraw its geometry unless the dimensions of its layer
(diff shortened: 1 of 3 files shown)


wgpu/src/widget/canvas/layer/cache.rs
```diff
@@ -13,7 +13,7 @@ use std::{cell::RefCell, marker::PhantomData, sync::Arc};
 ///
 /// [`Layer`]: ../trait.Layer.html
 /// [`Cache`]: struct.Cache.html
-#[derive(Debug)]
+#[derive(Debug, Default)]
```
hecrj, line 16: I thinks this adds a `Default` bound to `T`, unnecessarily. It may be better to implement the trait manually.

robjtede: I thought this too and checked using cargo expand. The impl for default has the trait bound, it imposes no additional traits on the struct itself.

https://gist.github.com/robjtede/316a2cde764e28a78f24b73ef7477fcb

hecrj: Sorry, my point is that the `Default` implementation for `Cache` doesn't need a `Default` bound on `T`, as we do not need it to initialize the `PhantomData`.

image

robjtede: Excellent idea.

2020-04-08_23 11 35-7f0b1bc9


Updated the canvas examples too.

PR #281, opened 2020-04-09
Align text in `iced_wgpu` on a case-by-case basis

### LINKED ISSUE #280: Support Centered Text in Canvas
When using `frame.fill_text` with a `canvas::Text` with `HorizontalAlignment::Center`, we get a thread panic down in `rusttype`. This might be due to the infinite bounding box send down the pipeline. The center of infinity is still infinity.

hecrj:
Fixes #280.

### DIFF
--- a/core/src/rectangle.rs
+++ b/core/src/rectangle.rs
@@ -17,6 +17,32 @@ pub struct Rectangle<T = f32> {
 }
 
 impl Rectangle<f32> {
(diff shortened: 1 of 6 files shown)


clarkmoody:
Tested and works. I've got both `HorizontalAlignment::Center` and `VerticalAlignment::Bottom` working for `frame.fill_text`. Thanks!

PR #283, opened 2020-04-10
Improve download_progress example

Folyd:
It seems like we have no example to demonstrate the powerful strength of `Subscription.batch()` API. So this is why I add this example. I know we already have a nice download example that has some duplicate code with this one. I still think we deserve such an example to instruct the new user to handle multiple asynchronous commands with `Subscription.batch()`. Maybe we can merge two download examples into a single one and distinct with different modes (_simple_ and _advanced_)?

image


hecrj:
Looks cool! :tada: 

I think we should add progress bars and merge the example with the `download_progress` one.

Folyd:
@hecrj I have added progress bar and replaced the old download example with this one. Is it ok to be merged?

examples/download_progress/src/download.rs
```diff
@@ -1,89 +1,105 @@
-use iced_futures::futures;
+use std::hash::Hash;
 
-// Just a little utility function
-pub fn file<T: ToString>(url: T) -> iced::Subscription<Progress> {
-    iced::Subscription::from_recipe(Download {
-        url: url.to_string(),
-    })
-}
```
hecrj, line 8: I believe exposing a function to create a `Subscription` is a simpler API than exposing the `Download` internals.

examples/download_progress/src/download.rs
```diff
@@ -1,89 +1,105 @@
...
+    Finished,
 }
 
 // Make sure iced can use our download stream
 impl<H, I> iced_native::subscription::Recipe<H, I> for Download
 where
     H: std::hash::Hasher,
 {
-    type Output = Progress;
+    type Output = (String, Progress);
```
hecrj, line 27: We should probably consider using [`url`](https://docs.rs/url/2.1.1/url/).

examples/download_progress/src/download.rs
```diff
@@ -1,89 +1,105 @@
...
     fn stream(
         self: Box<Self>,
-        _input: futures::stream::BoxStream<'static, I>,
-    ) -> futures::stream::BoxStream<'static, Self::Output> {
-        Box::pin(futures::stream::unfold(
-            State::Ready(self.url),
+        _input: BoxStream<'static, I>,
+    ) -> BoxStream<'static, Self::Output> {
+        Box::pin(stream::unfold(
+            State::Ready(self.url.to_string()),
```
hecrj, line 39: I don't believe cloning is necessary here.

examples/download_progress/src/download.rs
```diff
@@ -1,89 +1,105 @@
...
-                                    Some((Progress::Errored, State::Finished))
-                                }
-                            }
-                            Err(_) => {
-                                Some((Progress::Errored, State::Finished))
+                    State::Ready(url) => match reqwest::get(&url).await {
+                        Ok(response) => {
+                            if let Some(total) = response.content_length() {
+                                Some((
+                                    (url.to_string(), Progress::Started),
```
hecrj, line 46: We probably should be using `clone` instead of `to_string`, as it better conveys the intent.

examples/download_progress/src/main.rs
```diff
@@ -1,144 +1,277 @@
...
+        window: window::Settings {
+            size: (800, 600),
+            resizable: true,
+            decorations: true,
+        },
+        flags: (),
+        default_font: None,
+        antialiasing: false,
+    });
+}
```
hecrj, line 277: All the examples have `main` at the beginning, followed by the application state and trait implementation. We should also use the default `Settings`.

examples/download_progress/src/main.rs
```diff
@@ -1,144 +1,277 @@
...
     type Flags = ();
 
     fn new(_flags: ()) -> (Example, Command<Message>) {
         (
-            Example::Idle {
-                button: button::State::new(),
-            },
-            Command::none(),
+            Example::Loading,
+            Command::perform(Example::load_data(), Message::Loaded),
```
hecrj, line 133: For simplicity, I think we should remove the `Loading` variant and `load_data` altogether.

examples/download_progress/src/main.rs
```diff
@@ -1,144 +1,277 @@
...
+    name: String,
+    url: String,
+    button: button::State,
+    state: TaskState,
+}
+
+#[derive(Debug, Clone, Default)]
+struct State {
+    tasks: Vec<Task>,
+    downloading_urls: Vec<String>,
```
hecrj, line 29: I think we should be able to remove this list, and simply compute it when necessary by traversing the `tasks` and checking their `state`.

examples/download_progress/src/main.rs
```diff
@@ -1,144 +1,277 @@
...
-                    download::Progress::Advanced(percentage) => {
-                        *progress = percentage;
+            }
+            Example::Loaded(State {
+                tasks,
+                downloading_urls,
+                ..
+            }) => match message {
+                Message::StartDownload(url) => {
+                    if let Some(task) = tasks.iter_mut().find(|t| t.url == url)
```
hecrj, line 154: We could define `Message` as a struct with an `Url` and a `DownloadMessage`. This would allow us to easily find the index of the corresponding `Task` before matching on the particular `DownloadMessage`, removing duplication considerably as a consequence.

Folyd:
@hecrj Thanks for reviewing. :)

hecrj (approved):
Thank you for this :tada:

I have rebased and simplified the example a bit while keeping the `ProgressBar` widget. I think we can merge this!

PR #285, opened 2020-04-11
Add support for setting window icon

### LINKED ISSUE #433: Window icon
[winit seems to have this option](https://docs.rs/winit/0.22.2/winit/window/struct.Icon.html), but it isn't `Copy`able.

frapa:
This PR is just to start the discussion around this topic, the code is currently a hack, the naming is not final and I just added it somewhere, not even in the right place. But since this is my first contribution to the library and my experience with Rust is not extensive, I would like if somebody more experienced can have a look an tell me if I'm on the right track. 

This adds a new property from `Settings::window::icon` and an `Icon` struct which can be converted to `winit::window::Icon`.

It also adds code to display this icon in `Application::run`.

Due to the fact that the Icon struct cannot be copied, I had to remove the Copy trait from all Settings,
both in `iced` and `iced_winit`. This is an API breaking change.

Currently, I'm able to set an icon with the code:
image
 
Thanks to `winit` it gets correctly displayed also on the taskbar:
image

And of course in the window switcher (alt+tab).

Fixes #433.

### DIFF
--- a/winit/src/settings.rs
+++ b/winit/src/settings.rs
@@ -14,7 +14,7 @@ use winit::monitor::MonitorHandle;
 use winit::window::WindowBuilder;
 
 /// The settings of an application.
-#[derive(Debug, Clone, Copy, PartialEq, Default)]
+#[derive(Debug, Clone, PartialEq, Default)]
 pub struct Settings<Flags> {
     /// The [`Window`] settings
     ///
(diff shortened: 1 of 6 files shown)


winit/src/settings.rs
```diff
@@ -8,8 +8,34 @@ mod platform;
 
 pub use platform::PlatformSpecific;
 
+/// An Icon
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Icon {
+    ///
+    rgba: Box<[u8]>,
```
hecrj, line 15: I would use a `Vec<u8>` here.

frapa: Fixed

frapa:
@hecrj Do you mean that no code should be added to the main iced package? As far as I can see, the web package is very experimental, but I could try implementing the favicon there, too. I don't really get the build process, tough...

PR #290, opened 2020-04-11
Remove unnecessary 'static lifetimes on `Renderer` traits

### LINKED ISSUE #288: Unnecessary 'static requirement for Renderers
Hey there, 

I'm currently writing a backend for `iced`, to glue it together with the Amethyst game engine. 

As you may know, resource management in the framework is done by having `AssetStorage<T>` stored inside the ECS World, and borrow from it when you need to compute things. In my case I need the actual font in order to compute the size of a text widget, and I need the SpriteSheet information to know the size of an Image widget, for example.  

Sadly, that means that **my Renderer has to borrow the ECS world** somehow (or **at least borrow from it**, the problem here remains the same), effectively making it invalid for the 'static lifetime.    

But, `iced_native` requires the Renderer to be 'static on some implementation of widget. I thought it was strange, so I had a look at `iced_native`. 

As it turns out, the only reason why a renderer should be 'static is because of the implementation of `hash_layout` that uses e.g. `std::any::TypeId::of::<Column<'_, (), Renderer>>().hash(state);`, which requires the type to be 'static. 

AIUI the only reason this `.hash` is called is so that every widget leaves a different "footprint" for the hash_layout result. This can also be implemented with a marker type, to avoid applying a constraint on the renderer, unless I completely misunderstood the point. Example for Column widget : 

https://github.com/hecrj/iced/blob/master/native/src/widget/column.rs

```rust
#[derive(Debug, Hash)]
/// Marker type for layout-hashing of Column Widget
pub struct ColumnWidget;

impl<'a, Message, Renderer> Widget<Message, Renderer>
    for Column<'a, Message, Renderer>
where
    Renderer: self::Renderer, // Notice how the renderer no longer needs to be 'static
{
    // ... omitted 
    fn hash_layout(&self, state: &mut Hasher) {
        std::any::TypeId::of::<ColumnWidget>().hash(state);

        self.width.hash(state);
        self.height.hash(state);
        self.max_width.hash(state);
        self.max_height.hash(state);
        self.align_items.hash(state);
        self.spacing.hash(state);

        for child in &self.children {
            child.widget.hash_layout(state);
        }
    }
}
```

I applied this change on a local version, everything looks good.

Are there any other motivation as to why a Renderer should be 'static ? The requirement looks a bit unnecessary to me. 

Happy easter! 
Cheers

AlisCode:
Fixes #288 .

Removes 'static lifetime bound on Renderer implementors by defining marker types where applicable.

Thanks a lot for your reactivity on this issue :)

### DIFF
--- a/native/src/widget/button.rs
+++ b/native/src/widget/button.rs
@@ -234,8 +234,8 @@ where
     }
 
     fn hash_layout(&self, state: &mut Hasher) {
-        struct ButtonWidget;
-        std::any::TypeId::of::<ButtonWidget>().hash(state);
+        struct Marker;
+        std::any::TypeId::of::<Marker>().hash(state);
 
         self.width.hash(state);
         self.content.hash_layout(state);
--- a/native/src/widget/button.rs
+++ b/native/src/widget/button.rs
@@ -143,7 +143,7 @@ impl State {
 impl<'a, Message, Renderer> Widget<Message, Renderer>
     for Button<'a, Message, Renderer>
 where
-    Renderer: 'static + self::Renderer,
+    Renderer: self::Renderer,
     Message: Clone,
 {
     fn width(&self) -> Length {
@@ -234,7 +234,8 @@ where
     }
 
     fn hash_layout(&self, state: &mut Hasher) {
-        std::any::TypeId::of::<Button<'_, (), Renderer>>().hash(state);
+        struct ButtonWidget;
+        std::any::TypeId::of::<ButtonWidget>().hash(state);
 
         self.width.hash(state);
         self.content.hash_layout(state);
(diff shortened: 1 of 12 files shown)


native/src/widget/button.rs
```diff
@@ -234,7 +234,8 @@ where
     }
 
     fn hash_layout(&self, state: &mut Hasher) {
-        std::any::TypeId::of::<Button<'_, (), Renderer>>().hash(state);
+        struct ButtonWidget;
+        std::any::TypeId::of::<ButtonWidget>().hash(state);
```
hecrj, line 238: Maybe we could rename this to simply `Marker`, as the module already works as a scope. This should also help us copy-paste these two lines when implementing other widgets.

AlisCode: Done :)

Issue #282, opened 2020-04-10
Question: Accessibility Features


porglezomp:
Iced looks extremely nice in terms of visual quality, but I'm concerned about accessibility. I don't see any mention of it elsewhere, so I'm curious if there's any support for accessibility features currently implemented, and if there are future plans for explicitly supporting it? Thanks.

hecrj:
>  I'm curious if there's any support for accessibility features currently implemented

There is no explicit support for accessibility features yet. This is missing, like many other features. 
The library is a few months old!

> and if there are future plans for explicitly supporting it

I think this is an important feature, and I agree we should at least mention it in the roadmap soon.

I believe the modular architecture could help us greatly here. The `iced` public API is meant to be high-level, and `iced_native` is shell-agnostic and renderer-agnostic. The same way we can target the DOM with `iced_web`, I think we should have enough room to target accessibility APIs, build screen readers, specific shells, etc.

porglezomp:
Thanks for the answer! I wasn't trying to rush anything, just glad to see see that it's an issue being considered.

Restioson:
Hopefully the default widgets can have automatic implementations, like with GTK/ATK (but cross platform).

PR #292, opened 2020-04-12
add some accessibility features to web widgets

### LINKED ISSUE #71: Radio buttons need to be grouped for proper interaction
Radio buttons should *always* belong to a group, so that things like the arrow keys work to select a different value, and so that accessibility tools can understand what’s going on with them.

In HTML, this means that grouped radio buttons need to share a `name` attribute.

TomPridham:
starts on #282, at least for web
adds some basic accessibility functionality to radio buttons, checkboxes, text inputs, and images.

### DIFF
--- a/web/src/widget/checkbox.rs
+++ b/web/src/widget/checkbox.rs
@@ -28,7 +28,7 @@ pub struct Checkbox<Message> {
     is_checked: bool,
     on_toggle: Rc<dyn Fn(bool) -> Message>,
     label: String,
-    id: String,
+    id: Option<String>,
     width: Length,
     style: Box<dyn StyleSheet>,
 }
@@ -52,7 +52,7 @@ impl<Message> Checkbox<Message> {
             is_checked,
             on_toggle: Rc::new(f),
             label: label.into(),
-            id: Default::default(),
+            id: None,
             width: Length::Shrink,
             style: Default::default(),
         }
@@ -78,7 +78,7 @@ impl<Message> Checkbox<Message> {
     ///
     /// [`Checkbox`]: struct.Checkbox.html
     pub fn id(mut self, id: impl Into<String>) -> Self {
-        self.id = id.into();
+        self.id = Some(id.into());
         self
     }
 }
@@ -97,8 +97,6 @@ where
 
         let checkbox_label =
             bumpalo::format!(in bump, "{}", self.label).into_bump_str();
-        let checkbox_id =
-            bumpalo::format!(in bump, "{}", self.id).into_bump_str();
 
         let event_bus = bus.clone();
         let on_toggle = self.on_toggle.clone();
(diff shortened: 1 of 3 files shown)


web/src/widget/checkbox.rs
```diff
@@ -118,8 +131,7 @@ where
                         vdom.schedule_render();
                     })
                     .finish(),
-                span(bump).children(vec![
```
TomPridham, line 121: i don't think this span is necessary, and the other inputs don't have it. so i removed it for consistency

web/src/widget/image.rs
```diff
@@ -15,13 +15,16 @@ use std::{
...
-/// let image = Image::new("resources/ferris.png");
+/// let image = Image::new("resources/ferris.png", "Ferris the crab");
 /// ```
 #[derive(Debug)]
 pub struct Image {
     /// The image path
     pub handle: Handle,
 
+    /// The alt text of the image
+    pub alt: String,
```
TomPridham, line 26: images should always have alt text unless they are purely presentational, e.g. a background image. im not sure the best way to go about making this not necessary for those sorts of images, and this feels like a good default

web/src/widget/text_input.rs
```diff
@@ -50,15 +54,19 @@ impl<'a, Message> TextInput<'a, Message> {
     ///
     /// It expects:
     /// - some [`State`]
-    /// - a placeholder
+    /// - an optional placeholder
+    /// - a label describing the input
+    /// - an id
```
TomPridham, line 59: placeholders aren't replacements for labels. i made the placeholder optional so that people can decide whether they want one or not, but made label text required

web/src/widget/text_input.rs
```diff
@@ -67,7 +75,13 @@ impl<'a, Message> TextInput<'a, Message> {
     {
         Self {
             _state: state,
-            placeholder: String::from(placeholder),
+            placeholder: if let Some(p) = placeholder {
```
TomPridham, line 78: i'm not sure if there is a better way to do this. it seems like there should be, but i don't know enough about dodrio

web/src/widget/text_input.rs
```diff
@@ -160,61 +174,72 @@ where
...
+        let text_input_label =
+            bumpalo::format!(in bump, "{}", self.label).into_bump_str();
+        let text_input_id =
+            bumpalo::format!(in bump, "{}", self.id).into_bump_str();
+
+        label(bump)
+            .attr("for", text_input_id)
+            .children(vec![
+                text(text_input_label),
+                input(bump)
```
TomPridham, line 186: most of this was whitespace because i nested the input inside of a label

web/src/widget/image.rs
Restioson: Could always be passed an empty string? Not sure how that works on screen readers and such though.

TomPridham: it could. i wasn't sure that implementing some sort of length requirement would be the correct thing to do since it would end up just being arbitrary. and i feel like most people ignore accessibility out of ignorance rather than malice, so having a reminder that `alt` text is required would be enough for most cases

TomPridham:
okay, that makes sense i'll rework this into builder methods

web/src/widget/image.rs
TomPridham: but it seems like adding them as builder methods as @hecrj suggested will result in less empty alt tags since people will have to opt in

web/src/widget/text_input.rs
Restioson: placeholder.map(Into::into).unwrap_or_default() could work

TomPridham:
@hecrj, i updated this to use builder methods. i ended up just reverting the text input changes since it would require a lot more work to implement those changes

web/src/widget/image.rs
hecrj: We can make `alt` text mandatory once we tackle accessibility in the native side of the ecosystem.

PR #293, opened 2020-04-14
Improve `Canvas` ergonomics

hecrj:
This PR makes some arguments in the methods of `canvas::Frame` generic, improving usability and reducing verbosity in many common scenarios.

Additionally, helper methods have been added to `Path` for creating lines, rectangles, and circles directly without the need of a closure.

As a result, the `clock` and `solar_system` examples are simpler and easier to read now!

### DIFF
--- a/wgpu/src/widget/canvas/path/builder.rs
+++ b/wgpu/src/widget/canvas/path/builder.rs
@@ -133,11 +133,14 @@ impl Builder {
     ///
     /// [`Path`]: struct.Path.html
     #[inline]
-    pub fn rectangle(&mut self, p: Point, size: Size) {
-        self.move_to(p);
-        self.line_to(Point::new(p.x + size.width, p.y));
-        self.line_to(Point::new(p.x + size.width, p.y + size.height));
-        self.line_to(Point::new(p.x, p.y + size.height));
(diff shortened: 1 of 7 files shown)


wgpu/src/widget/canvas/path/builder.rs
```diff
@@ -133,11 +133,14 @@ impl Builder {
     ///
     /// [`Path`]: struct.Path.html
     #[inline]
-    pub fn rectangle(&mut self, p: Point, size: Size) {
-        self.move_to(p);
-        self.line_to(Point::new(p.x + size.width, p.y));
-        self.line_to(Point::new(p.x + size.width, p.y + size.height));
-        self.line_to(Point::new(p.x, p.y + size.height));
+    pub fn rectangle(&mut self, top_left: Point, size: Size) {
```
artursapek, line 136: this is totally pedantic, but generally code like this refers to the top left point as `origin`

hecrj: I have seen graphics APIs where rectangles are defined using the center or the bottom-left corner as the starting point. This is just an explicit way of documenting the API. The same way the `circle` method takes a `center` point.

This way, you can tell how to use this method by just looking at the method signature.

PR #294, opened 2020-04-14
Fix tour example

MrMonotone:


hecrj:
Oh, is this to fix compilation when targgeting Wasm?

I believe we should fix the API instead. Let me do that.

hecrj:
Thank you for pointing this out!

MrMonotone:
@hecrj sorry I didnt mention that. 🙃 I am targeting wasm for my project so I will mention it in future if anything else comes up.

PR #300, opened 2020-04-16
Process ScaleFactorChanged events

Matthias-Fauconneau:
This is necessary for HiDPI support on Wayland (Sway).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,3 +69,7 @@ iced_web = { version = "0.2", path = "web" }
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
 features = ["image", "svg", "canvas"]
+
+[patch.'crates-io']
+winit = { git='https://github.com/Matthias-Fauconneau/winit' }
+#winit = { path = "../winit" }


winit/src/application.rs
```diff
@@ -358,6 +358,13 @@ pub trait Application: Sized {
                         size = Size::new(new_size, window.scale_factor());
                         resized = true;
                     }
+                    WindowEvent::ScaleFactorChanged {
+                        scale_factor,
+                        ref new_inner_size,
+                    } => {
+                        size = Size::new(**new_inner_size, scale_factor);
+                        resized = true;
+                    }
```
hecrj, line 367: Isn't this event followed by a `Resized` event? If not, I believe this may be an inconsistency on `winit`.

Matthias-Fauconneau: Actually it sends the ScaleFactorChanged event after the Resized event.
The actual issue is that iced relies on Window::inner_size to be current and for some reason the last guy to mess up with winit wayland thought it would be a good idea to first trigger the Resized event callback before updating the field returned by inner_size... >< 
https://github.com/rust-windowing/winit/pull/1538

kchibisov: `ScaleFactorChange` doesn't cause resize on Wayland, however it's indeed inconsistency, since Resize is in physical pixels, so when Wayland changes scale factor, it changes the size of the window in physical pixels, however winit is looking at logical size for whatever reason. So winit should be patched to do that as well as report events in a mentioned order.

Matthias-Fauconneau:
Until winit gets fixed (0.22.1), wayland hidpi users can patch iced Cargo.toml :
[patch.'crates-io']
winit = { git='https://github.com/Matthias-Fauconneau/winit' }

PR #311, opened 2020-04-21
Sctkshm

Matthias-Fauconneau:
smithay-client-toolkit ^0.9 backend.
Can be used for better Wayland support than iced_winit/winit_sctk. e.g layer-shell
with a shared memory renderer, can be useful for dev and lightweightness.
Currently only display an empty window, renderer in progress.


hecrj:
This comes out of nowhere. Could you elaborate about the context of this? What is the purpose of this PR? Expectations? Goals?

Matthias-Fauconneau:
> This comes out of nowhere. What is the purpose of this PR? 
Indeed >< This is still very WIP but as you said, it's useful to talk early, so this PR is just a coordination point.

> Could you elaborate about the context of this? 
Of course! There are special use cases for which winit is not a useful abstraction, and cannot be poked through, or even made through-pokeable.
For example, layer-shell :)
Also winit_sctk is currently relying on an out of date possibly unsound version of sctk.

> Expectations? Goals?
I'm looking for the most pleasing way to do Rust UI on wayland.
The options I see are :
* Rework the winit backend => 
 - no layer-shell bonus goal
 - unnecessary constraints 
 + useful for other wayland winit users (but I don't know any)
* direct sctk without iced =>
 + what i've been doing so far. Freedom !
 - No code reuse. Less inspiration
* direct sctk with iced =>
 - lot of initial work understanding and piecing together interfaces
 + should translate easily back to winit-like platform abstraction only code
+ forge the most exquisite event loop using pure async (select socket|channel)
 ++ i.e removes the need for such inelegant kludge as iced_winit::EventProxy->winit::send_event

[https://github.com/Smithay/client-toolkit/issues/115]
[https://github.com/rust-windowing/winit/pull/1544]

hecrj:
Thanks for elaborating a bit!

If I understand correctly, you want to create:

- `iced_sctk`, a windowing shell with an opinionated API to target Wayland nicely.
- `iced_shm`, a software renderer to draw on Wayland windows.

That sounds great! The next step here is deciding how we will collaborate.

Personally, I don't think I can commit to maintaining/reviewing any Wayland-specific code. Therefore, you should take the lead on the `iced_sctk` and `iced_shm` crates and own that code yourself. If you want to, I could give you ownership of a repository under the [`iced-rs` organization](https://github.com/iced-rs).

So, how should we collaborate? Your crates will need to depend on `iced_native` and `iced_core`. Therefore, I am very interested in hearing about how they work for you and discuss any changes that may be necessary to accomodate your Wayland-specific use cases.

Additionally, I believe your work on Wayland may give us some insights about how to improve `iced_winit` too. I'd love to hear about design decisions that may be applicable to other parts of the ecosystem!

Matthias-Fauconneau:
That's the spirit.

Sounds like a plan if it comes to that.

I'll probably have more changes once I get a bit further.
Now I just have feature gates updated from !wasm to either iced_winit/sctk
gates.
Since that's really your side of the design, I just went mechanical there,
but there is clearly duplication which could be factorized.
That very much depend what is important to preserve in the interface, I
cannot really determine that.

I think it'd be interesting to explore composing with more async patterns.
Though that might be difficult on the winit side.

myfreeweb:
ha, I'm actually currently building something with sctk for layer-shell too, but with wgpu.

Would be nice to have the conversions from raw events (`wl_pointer::Event` etc.) to iced events in a shared place.

Matthias-Fauconneau:
I'd be interested in following what you're doing if possible.

An sctk platform is not constrained to a specific choice of shm/GL/wgpu, so
a crate implementing an sctk platform would also, be the main coordination
point for anything iced-sctk.

While this PR contains both an sctk platform and an shm backend, they would
become 2 crates.
This PR also changed the iced flags a bit with the idea to allow any
combination of winit|sctk with shm|wgpu (and now gl).

So if I understand it correctly, instead of forking the sctk platform to
suit your needs (glib?), you are using iced-native with your own custom
platform support, without going through the additional plumbing of the sctk
platform making it all go through the shared iced-app abstraction.

In that case, I suggest you just put the shared code in a crate. Hecrj can
open an iced/sctk repository for you to push to.

We could merge the shared code with the backend again once it is working.

myfreeweb:
Yeah, I actually published the initial code: https://github.com/myfreeweb/waysmoke
currently only a counter demo on layer shell, and it renders on each input event, w/o throttling yet.

I don't want an abstracted platform just yet, I'd like to implement actual desktop widgets first – I want the widgets to react to e.g. pointer enter events and manipulate their layer shell config, so I want to see what plumbing I'll need to do for that. Also I'll add a way for widgets to react to arbitrary async channels so that custom Wayland and D-Bus protocols could be connected through them.

Also in general the platform glue is really small, I'm not sure we even need a whole shared sctk platform that would somehow support both desktop widgets and xdg shell apps. Event conversion is what I want to share, but we'll have to think about how e.g. pointer enter/leave would be handled (for me it toggles state right now).

Actually would be nice to have pointer enter/leave as an iced event I guess.. and apps should have a way to react to raw iced events, not just messages.

Matthias-Fauconneau:
Thanks.

I suggest you maintain the iced event conversion in your toolkit.
With regular apps using the winit backend, either interface already fill
every needs for an iced sctk backend.

Don't hesitate to ask for any clarification concerning my implementation.
There is indeed some state tracking per surface and across the
application/connection.
I think the aspect which needs most consideration is key repeat (client
side on Wayland): sctk supports it only if you use their 'calloop' event
loop. I'm quite proud of my implementation using an async stream, hopefully
this can be reused ;D.

Actually what would be more interesting is to share this common state
tracking code between winit-sctk and your toolkit. I'm still using calloop
in winit but it could be switched to a nice shared idiomatic async
implementation.

https://github.com/rust-windowing/winit/pull/1544

Also, more long term, wayland-client will become more convenient to use :
https://github.com/Smithay/wayland-rs/issues/317
And might be the more natural dependency for a layer shell toolkit: no need
for sctk xdg shell decorations.

PR #315, opened 2020-04-23
Add `children_size()` method for `Row`, `Column` and `Scrollable`

Folyd:
The end-user cannot access the children's size of `Row`, `Column`, and `Scrollable` widget as the `children` field is private. So I think adding this method should idea to fix this.

### DIFF
--- a/native/src/widget/column.rs
+++ b/native/src/widget/column.rs
@@ -119,6 +119,13 @@ impl<'a, Message, Renderer> Column<'a, Message, Renderer> {
         self.children.push(child.into());
         self
     }
+
+    /// Gets the children size of the [`Column`].
+    ///
+    /// [`Column`]: struct.Column.html
+    pub fn children_size(&self) -> usize {
+        self.children.len()
+    }
 }
 
 impl<'a, Message, Renderer> Widget<Message, Renderer>
--- a/native/src/widget/row.rs
+++ b/native/src/widget/row.rs
@@ -120,6 +120,13 @@ impl<'a, Message, Renderer> Row<'a, Message, Renderer> {
         self.children.push(child.into());
         self
     }
+
+    /// Gets the children size of the [`Row`].
+    ///
+    /// [`Row`]: struct.Row.html
+    pub fn children_size(&self) -> usize {
+        self.children.len()
+    }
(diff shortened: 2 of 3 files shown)


hecrj:
Hey, thanks!

What would be the use case for this? I think `view` code can figure out the amount of elements in a container without these methods. In the end, `view` code is building these elements.

Folyd:
I think the use cases in the user end are diverse. For example, in my project, I need to display two columns of data, both columns of data are derived from a single `Vec` and distinct from a field `environment` (`production` and `test`). Then I use a `fold` method to assemble those two columns. However, I should give the first row of each column a check flag which only possible by checking the current children's index of that column.

```rust
    let (production, test) = versions
        .iter_mut()
        .filter(|version| version.app_name == *current_app)
        .fold(
            (
                Column::new().spacing(40).width(Length::Fill).push(
                    Text::new("Production")
                        .size(50)
                        .color(Color::from_rgb8(239, 41, 28)),
                ),
                Column::new().spacing(40).width(Length::Fill).push(
                    Text::new("Test")
                        .size(50)
                        .color(Color::from_rgb8(249, 187, 45)),
                ),
            ),
            |(production, test), version| match version.environment {
                Environment::Production => {
                    let index = production.children_size();
                    (production.push(version.view(index)), test)
                }
                Environment::Test => {
                    let index = test.children_size();
                    (production, test.push(version.view(index)))
                }
            },
        );
```
Here is the screenshot what I described:
image


I know we can use two variables to record the current children's size of each column, this could be better if we have `children_size()` API though.

hecrj:
> I know we can use two variables to record the current children's size of each column, this could be better if we have children_size() API though.

In this case, I believe you could use [`Iterator::enumerate`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.enumerate) instead.

This is what I meant by:

> view code can figure out the amount of elements in a container without these methods

`view` should be a straightforward mapping method. I don't believe elements themselves should affect view logic.

You could also generate a `Vec` and then use [`Column::with_children`](https://docs.rs/iced_native/0.2.1/iced_native/widget/column/struct.Column.html#method.with_children).

PR #319, opened 2020-04-23
Add `ImagePane` widget

### LINKED ISSUE #306: Image Zoom & Pan
I've searched here and Zulip and couldn't find any relevant discussion.

Are there plans to implement a widget to allow zooming in on an Image (past the dimensions of the parent element / application window) and panning?

tarkah:
This isn't complete, but the zooming and panning features "work". If an image handle's dimensions are larger than the dimensions of the `ImagePane`, it is scaled down to fit inside the pane. Images smaller than the pane won't get rescaled until the mouse wheel is scrolled.

Some improvements needed before merging:

- Image should be centered inside the `ImagePane`. Can't implement this until negative offsets are allowed inside `Clip` primitive
- Clipping offset should get updated when scaling the image to "zoom" around wherever the cursor is placed. Right now offset doesn't change on scaling, so image just zooms out from top left corner.
- Add builder methods for `State` to allow specifying the min / max scale factor and step size

Let me know what you think and if there are any other features / defaults that should be added!

resolves #306

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,5 +29,5 @@ jobs:
       run: cargo check --package iced --target wasm32-unknown-unknown
     - name: Check compilation of `tour` example
       run: cargo build --package tour --target wasm32-unknown-unknown
-    - name: Check compilation of `pokedex` example
-      run: cargo build --package pokedex --target wasm32-unknown-unknown
+    - name: Check compilation of `todos` example
+      run: cargo build --package todos --target wasm32-unknown-unknown
--- a/examples/pokedex/src/main.rs
+++ b/examples/pokedex/src/main.rs
@@ -1,6 +1,6 @@
 use iced::{
     button, futures, image, Align, Application, Button, Column, Command,
-    Container, Element, Image, Length, Row, Settings, Text,
+    Container, Element, Length, Row, Settings, Text,
 };
 
 pub fn main() -> iced::Result {
@@ -112,16 +112,20 @@ struct Pokemon {
     name: String,
     description: String,
     image: image::Handle,
+    image_viewer: image::viewer::State,
 }
 
 impl Pokemon {
     const TOTAL: u16 = 807;
 
-    fn view(&self) -> Element<Message> {
+    fn view(&mut self) -> Element<Message> {
         Row::new()
             .spacing(20)
             .align_items(Align::Center)
-            .push(Image::new(self.image.clone()))
+            .push(image::Viewer::new(
+                &mut self.image_viewer,
+                self.image.clone(),
+            ))
             .push(
                 Column::new()
                     .spacing(20)
@@ -200,6 +204,7 @@ impl Pokemon {
                 .map(|c| if c.is_control() { ' ' } else { c })
                 .collect(),
             image,
+            image_viewer: image::viewer::State::new(),
         })
     }
 
--- a/graphics/src/widget/image.rs
+++ b/graphics/src/widget/image.rs
@@ -1,11 +1,14 @@
 //! Display images in your user interface.
+pub mod viewer;
+
 use crate::backend::{self, Backend};
+
 use crate::{Primitive, Renderer};
 use iced_native::image;
 use iced_native::mouse;
 use iced_native::Layout;
 
-pub use iced_native::image::{Handle, Image};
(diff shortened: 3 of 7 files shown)


hecrj:
This looks great! Thank you.

I think we could put this new widget inside the current `image` module, as it seems to be an `Image` variation of some sort. Maybe we could call it `Viewer` or something similar. That way, users would use it as `image::Viewer`.

> Image should be centered inside the ImagePane. Can't implement this until negative offsets are allowed inside Clip primitive

I will try to get this ready soon!

> Clipping offset should get updated when scaling the image to "zoom" around wherever the cursor is placed. Right now offset doesn't change on scaling, so image just zooms out from top left corner.

I implemented [some logic to do this](https://github.com/hecrj/iced/blob/7dc02a5e16a3143b7c3ba9270207e3ebda71d567/examples/game_of_life/src/main.rs#L361) in the new [`game_of_life` example](https://github.com/hecrj/iced/tree/7dc02a5e16a3143b7c3ba9270207e3ebda71d567/examples/game_of_life). We probably shouldn't copy the behavior verbatim, but it may be useful for inspiration.

> Add builder methods for State to allow specifying the min / max scale factor and step size

I'd probably put these in `ImagePane` itself. This way, we let users decide how to derive these values in `view` logic.

tarkah:
Thanks for taking a look! I'll definitely move this into the `image` module and look to incorporate your other suggestions.

tarkah:
I went ahead and renamed it `ImageViewer` since I've made it available in the root of Iced's widget module, and `Viewer` seems a little generic for it there. Unless you don't want it available in the root of that module.

I'm going to hold off on implementing logic for zooming to the cursor until we can center the image inside the Clipping primitive, as to not have to rewrite anything. I'll implement the builder methods then as well.

hecrj:
> Unless you don't want it re-exported to the root.

Yeah, I tend to only re-export the types that have the same name as the module.

> I'm going to hold off on implementing logic for zooming to the cursor until we can center the image inside the Clipping primitive

Since #325, there is a new `Translate` primitive that we may be able to use. I forgot about this! It's very likely we will end up dropping the `offset` from `Clipping` and just use this new primitive instead.

tarkah:
> Yeah, I tend to only re-export the types that have the same name as the module.

Ok, I renamed to `image::Viewer` and removed the re-export.


> Since #325, there is a new `Translate` primitive that we should be able to use here. I forgot about this! It's very likely we will end up dropping the `offset` from `Clipping` and just use this new primitive instead.

Awesome! Looking forward to this change

hecrj:
> Awesome! Looking forward to this change

The new primitive already landed in `master`. We should be able to use to move the `Image` primitive even with a clipping offset of `(0, 0)`.

tarkah:
Oh nice! I moved to `Primitive::Translate` and it works, just need to inverse my current offset. I'll try to wrap up this PR by end of week, thanks!

tarkah:
Nice, I wrapped the `Image` primitive in a `Translate` and applied the inverse of my original offset to this, then left the `Clip` primitive offset as `(0, 0)` and it worked! I'll get this PR hopefully wrapped up by end of week. Thanks!

tarkah:
Ok, everything should be implemented! After performing mental gymnastics all morning, I finally got the calculations worked out to translate the image so it's centred with the proper adjustments made for smooth scaling in / out from cursor location.

Note that currently image panning is bounded to the edges of the image, so you cannot infinitely pan past the image and therefore panning is only enabled when the image becomes clipped by it's width and/or height.

hecrj:
Awesome! Thanks. I will review the changes soon.

Merging #354 seems to have caused some conflicts. But don't worry about it, I should be able to fix them.

hecrj (approved):
I think I've brought this up to speed! :tada: 

I have replace the old `Image` with this new widget in the `pokedex` example. I tried to describe all the other changes in the commit history.

I think this should be ready to merge. Hopefully I didn't break anything. Let me know!

tarkah:
This is awesome! I'll test it out today and let you know :)

tarkah:
@hecrj I've tested on my old project and it works great. Thanks for getting master back-merged into this properly!

hecrj:
Awesome! We can finally merge this then!

Thank you again :clinking_glasses:

PR #323, opened 2020-04-26
Programmatic scrolling

### LINKED ISSUE #307: Scrolling programmatically
The scroll_to method on the State in Scrollable is useful for scrolling programmatically but it requires knowledge of the layout bounds which isn't public.

Either this information should be exposed, probably for all widgets, or the State should contain the bounds of the Scrollable so that scroll_to can be used as public api without needing to create a custom widget.

What do you think?

EDIT: zulip links
https://iced.zulipchat.com/#narrow/stream/213316-general/topic/infinite.20list
https://iced.zulipchat.com/#narrow/stream/213445-help/topic/Scrollable.3A.3Apush_back
https://iced.zulipchat.com/#narrow/stream/213445-help/topic/how.20to.20scroll.20programmatically.3F
https://iced.zulipchat.com/#narrow/stream/213445-help/topic/scroll_to

krooq:
Fixes #307 

It's now possible to call `scroll` and `scroll_to` in `update` to scroll a `Scrollable` programmatically.

Changed `offset` to `position` to remove the dependence on the layout from the state of `Scrollable`.
I renamed this member to differentiate it from the `offset` method which can be called still by the `draw` method to render a `Scrollable` at some `position` based on its `bounds` and `content_bounds`.

### DIFF
--- a/native/src/widget/scrollable.rs
+++ b/native/src/widget/scrollable.rs
@@ -168,10 +168,16 @@ where
                     match delta {
                         mouse::ScrollDelta::Lines { y, .. } => {
                             // TODO: Configurable speed (?)
-                            self.state.scroll(y * 60.0, bounds, content_bounds);
+                            let speed = 60.0;
+                            self.state.scroll(
+                                speed * -y
+                                    / (content_bounds.height - bounds.height),
+                            );
                         }
                         mouse::ScrollDelta::Pixels { y, .. } => {
-                            self.state.scroll(y, bounds, content_bounds);
+                            self.state.scroll(
+                                -y / (content_bounds.height - bounds.height),
+                            );
                         }
                     }
                 }
@@ -198,14 +204,10 @@ where
                     if let (Some(scrollbar), Some(scroller_grabbed_at)) =
                         (scrollbar, self.state.scroller_grabbed_at)
                     {
-                        self.state.scroll_to(
-                            scrollbar.scroll_percentage(
-                                scroller_grabbed_at,
-                                cursor_position,
-                            ),
-                            bounds,
-                            content_bounds,
-                        );
+                        self.state.scroll_to(scrollbar.scroll_percentage(
+                            scroller_grabbed_at,
+                            cursor_position,
+                        ));
                     }
                 }
                 _ => {}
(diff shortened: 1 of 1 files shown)


hecrj:
Using a relative value to store scroll position won't work well with dynamic content or layout. The scrolling offset will jump around instead of staying at a fixed position from the top.

As I said in #307, this feature needs a bit more thinking. We are not ready to write code for this yet.

krooq:
> Using a relative value to store scroll position won't work well with dynamic content or layout. The scrolling offset will jump around instead of staying at a fixed position from the top

Ah ok I understand now.
I didn't consider this case, my use case is very similar to a terminal so I always scroll to the bottom after adding new content.  
I think using the offset will still have this issue if the user adds content above the current offset.  
We will need some extra logic to handle these cases and the hidden content height both above and below the `offset` may need to be present in the state.

I think separating out the `position` into an independent value is still the way to go.
It is the responsiblity of the `layout` function to recompute the new `position` based on the previous `position` and change in the size of the content.

> As I said in #307, this feature needs a bit more thinking. We are not ready to write code for this yet.

I agree it needs more thinking. Writing code helps me think, learn how Iced works and provokes good discussion.  If you mean "we are not ready to merge yet" then yeah I 100% agree, lets get it right.
I live in Australia, litterally the other side of the world to you so conversing can be hard when best case scenario it takes a full day to get a reply due to time difference (I work for Thales, a French company this is a daily struggle for me :P).
I just want to have as much info as possible for you (and others) to work with.
If you want you can decline this PR and I can just add the code to the issue as a comment?

hecrj:
> I think using the offset will still have this issue if the user adds content above the current offset.

Yes, but I believe this is the behavior users will expect by default.

> We will need some extra logic to handle these cases and the hidden content height both above and below the offset may need to be present in the state.

I don't think we can consistently handle all the cases at once. We need the user to tell us what kind of behavior they want.

I think we are mising the concept of a scrolling `Anchor` of some sorts.

> I agree it needs more thinking. Writing code helps me think, learn how Iced works and provokes good discussion. If you mean "we are not ready to merge yet" then yeah I 100% agree, lets get it right.

By all means, write all the code you want if it helps you!

What I mean is that we have not settled for a particular API design and/or solution yet. This is the hard part! I shouldn't have to read code to learn why a particular approach was taken (i.e. how the hard part was solved).

> I live in Australia, litterally the other side of the world to you so conversing can be hard when best case scenario it takes a full day to get a reply due to time difference

No worries. There is no hurry. On the contrary, I believe taking time to think things through will help us make better decisions in the end.

hecrj:
Let's close this and keep the discussion going in #307.

PR #361, opened 2020-05-26
Proper Pane Grid spacing

clarkmoody:
On low-DPI screens, the rounding order of operations made it impossible to produce an odd-pixel spacing. Specifying 1, for instance, produced zero space between panes.

This approach subtracts half the spacing from the first pane prior to rounding and uses the whole spacing for the second pane size and coordinate.

We're no longer pre-computing the `halved_spacing` and passing that around. Rather `Axis::split` operates on the full spacing.

### DIFF
--- a/native/src/widget/pane_grid/axis.rs
+++ b/native/src/widget/pane_grid/axis.rs
@@ -14,37 +14,39 @@ impl Axis {
         &self,
         rectangle: &Rectangle,
         ratio: f32,
-        halved_spacing: f32,
+        spacing: f32,
     ) -> (Rectangle, Rectangle) {
         match self {
             Axis::Horizontal => {
-                let height_top = (rectangle.height * ratio).round();
-                let height_bottom = rectangle.height - height_top;
+                let height_top =
+                    (rectangle.height * ratio - spacing / 2.0).round();
+                let height_bottom = rectangle.height - height_top - spacing;
 
                 (
                     Rectangle {
-                        height: height_top - halved_spacing,
+                        height: height_top,
                         ..*rectangle
                     },
                     Rectangle {
-                        y: rectangle.y + height_top + halved_spacing,
-                        height: height_bottom - halved_spacing,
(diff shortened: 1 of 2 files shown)


clarkmoody:
> Thus, the real underlying problem remains unsolved: targetting physical pixels when the rendering primitives do not align perfectly with the pixel grid (clamping?). The fix should probably happen in the quad rendering pipeline in iced_wgpu. I will try to play with it.

Might be a configuration option. SVG has the [`shape-rendering`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering) attribute, which may take the value `crispEdges`. There we find:

> To achieve crisp edges, the user agent might turn off anti-aliasing for all lines and curves or possibly just for straight lines which are close to vertical or horizontal. Also, the user agent might adjust line positions and line widths to align edges with device pixels.

hecrj:
@clarkmoody I believe quads should always be pixel-aligned, independently of the anti-aliasing strategy. They are GUI rendering primitives that should always be crisp.

The `Mesh` primitive, together with the `antialiasing`  flag in `Settings`, can be leveraged to render geometry with subpixel precision.

I have opened #362 which should fix the main issue here. I still believe we should merge this, though!

native/src/widget/pane_grid/axis.rs
```diff
@@ -14,41 +14,130 @@ impl Axis {
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn split_horizontal() {
+        let a = Axis::Horizontal;
+        // rectangle height, spacing, top height, bottom y, bottom height
+        let cases = vec![
+            // Even height, even spacing
+            (10.0, 2.0, 4.0, 6.0, 4.0),
```
hecrj, line 68: Maybe we could create a `Case` struct for this? I don't mind verbosity!

clarkmoody: Updated in 858eafe

PR #363, opened 2020-05-27
Make the Text widget Clone even if the Renderer isn't

Imberflur:


native/src/widget/text.rs
```diff
@@ -31,6 +31,21 @@ pub struct Text<Renderer: self::Renderer> {
...
+            size: self.size,
+            color: self.color,
+            font: self.font,
+            width: self.width,
+            height: self.height,
+            horizontal_alignment: self.horizontal_alignment,
+            vertical_alignment: self.vertical_alignment,
+        }
+    }
+}
```
hecrj, line 47: Maybe we could move this implementation to the end of the file, as it's mostly boilerplate.

Imberflur: I considered that but I wasn't sure if that would be preferred at the time. Will do!

Issue #351, opened 2020-05-19
Very slow startup due to font-kit crate asking fontconfig to scan all system fonts


nyanpasu64:
Trying out iced 0.1.1 "clock" example on Linux OpenSUSE Tumbleweed.

On every startup, the app hangs on a blank window for around 31 seconds while loading fonts. The slowness occurs in fontconfig C functions, called by the the font-kit crate. Incidentally, reloading `fc-cache` is also a fairly slow process taking the same "order of magnitude" time.

This happens both with and without gdb. Other iced examples (not just clock) take a similarly long time to start, and get stuck in fontconfig functions.

```
(gdb) bt
#0  0x000055555600bc7c in FT_Stream_ReadFields ()
#1  0x0000555556029b9c in pcf_load_font ()
#2  0x000055555602b193 in PCF_Face_Init ()
#3  0x000055555600cd14 in open_face ()
#4  0x000055555600db15 in FT_Open_Face ()
#5  0x000055555600e086 in FT_New_Face ()
#6  0x0000555555fcb13b in FcFreeTypeQuery ()
#7  0x0000555555fc8646 in FcFileScanFontConfig ()
#8  0x0000555555fc87a1 in FcFileScanConfig ()
#9  0x0000555555fc8a56 in FcDirScanConfig ()
#10 0x0000555555fc8bb8 in FcDirCacheScan ()
#11 0x0000555555fc8da1 in FcDirCacheRead ()
#12 0x0000555555fc0092 in FcConfigAddDirList ()
#13 0x0000555555fc0162 in FcConfigBuildFonts ()
#14 0x0000555555fcdcec in FcInitLoadOwnConfigAndFonts ()
#15 0x0000555555fcdd17 in FcInitLoadConfigAndFonts ()
#16 0x0000555555f885ba in font_kit::sources::fontconfig::fc::Config::new ()
    at /home/nyanpasu64/.cargo/registry/src/github.com-1ecc6299db9ec823/font-kit-0.6.0/src/sources/fontconfig.rs:300
#17 0x0000555555f8ced9 in font_kit::sources::fontconfig::FontconfigSource::new ()
    at /home/nyanpasu64/.cargo/registry/src/github.com-1ecc6299db9ec823/font-kit-0.6.0/src/sources/fontconfig.rs:40
#18 0x0000555555b4e2fa in iced_wgpu::text::font::Source::new () at wgpu/src/text/font.rs:12
[...]

Issue #270, opened 2020-04-07
cargo run with the sample hello world application fails


kszlim:
Running `cargo run` on the sample code https://docs.rs/iced/0.1.0/iced/trait.Application.html creates runtime errors:

```
Fontconfig warning: line 5: unknown element "its:rules"
Fontconfig warning: line 6: unknown element "its:translateRule"
Fontconfig error: line 6: invalid attribute 'translate'
Fontconfig error: line 6: invalid attribute 'selector'
Fontconfig error: line 7: invalid attribute 'xmlns:its'
Fontconfig error: line 7: invalid attribute 'version'
Fontconfig warning: line 9: unknown element "description"
Fontconfig warning: "/etc/fonts/conf.d/10-hinting-slight.conf", line 4: unknown element "its:rules"
Fontconfig warning: "/etc/fonts/conf.d/10-hinting-slight.conf", line 5: unknown element "its:translateRule"
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 5: invalid attribute 'translate'
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 5: invalid attribute 'selector'
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 6: invalid attribute 'xmlns:its'
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 6: invalid attribute 'version'
Fontconfig warning: "/etc/fonts/conf.d/10-hinting-slight.conf", line 8: unknown element "description"
Fontconfig warning: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 4: unknown element "its:rules"
Fontconfig warning: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 5: unknown element "its:translateRule"
Fontconfig error: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 5: invalid attribute 'translate'
Fontconfig error: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 5: invalid attribute 'selector'
Fontconfig error: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 6: invalid attribute 'xmlns:its'
Fontconfig error: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 6: invalid attribute 'version'
[...]

Issue #199, opened 2020-02-21
`iced` hangs on boot-up and shows nothing for several minutes


Evrey:
- **OS:** Ubuntu-based Linux
- **GPU:** Radeon RX 5700
- **iced:** https://github.com/hecrj/iced/commit/9758c121764a6e8e5a45b8a66915127b31e61df7 , but the `0.1.0-beta` on https://crates.io/ has the same problem.
- **Code that I ran:** All the examples and an own simple test UI. Same result for all.
- **Compiler flags used:** `--release` and normal debug build, same issue.

<details>
<summary>Output to stdout:</summary>

```txt
Fontconfig warning: line 5: unknown element "its:rules"
Fontconfig warning: line 6: unknown element "its:translateRule"
Fontconfig error: line 6: invalid attribute 'translate'
Fontconfig error: line 6: invalid attribute 'selector'
Fontconfig error: line 7: invalid attribute 'xmlns:its'
Fontconfig error: line 7: invalid attribute 'version'
Fontconfig warning: line 9: unknown element "description"
Fontconfig warning: "/etc/fonts/conf.d/10-hinting-slight.conf", line 4: unknown element "its:rules"
Fontconfig warning: "/etc/fonts/conf.d/10-hinting-slight.conf", line 5: unknown element "its:translateRule"
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 5: invalid attribute 'translate'
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 5: invalid attribute 'selector'
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 6: invalid attribute 'xmlns:its'
Fontconfig error: "/etc/fonts/conf.d/10-hinting-slight.conf", line 6: invalid attribute 'version'
Fontconfig warning: "/etc/fonts/conf.d/10-hinting-slight.conf", line 8: unknown element "description"
Fontconfig warning: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 4: unknown element "its:rules"
Fontconfig warning: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 5: unknown element "its:translateRule"
Fontconfig error: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 5: invalid attribute 'translate'
[...]

PR #354, opened 2020-05-23
OpenGL renderer and backend-agnostic graphics subcrate

hecrj:
This PR implements an OpenGL renderer on top of [`glow`], introduces a [`glutin`] shell, and abstracts a bunch of backend-agnostic rendering logic in a new graphics subcrate.

# `iced_glow`
`iced_glow` is an OpenGL renderer powered by [`glow`] (GL on Whatever). This renderer is meant to be an alternative (not a replacement!) to [`iced_wgpu`] while [`wgpu`] and its ecosystem mature.

As `iced_glow` uses OpenGL, it should offer us better compatibility, stability, and maintainability for the time being. Furthermore, users that currently have OpenGL-based applications should be able to integrate them directly with `iced`.

On a related note, my work on this renderer also prompted me to write and open-source [`glow_glyph`], a direct port of [`wgpu_glyph`] for [`glow`].

Please, feel free to give the renderer a try by switching to the `feature/glow-renderer` branch and running any of the examples! Keep in mind that `Image` and `Svg` widgets are not supported yet.

# `iced_glutin`
`iced_glutin` is a shell on top of [`glutin`]. It can be used to easily initialize OpenGL-based renderers, as it will create and properly configure an OpenGL context for them.

Given that [`glutin`] is built on top of [`winit`] itself, [`iced_winit`] and `iced_glutin` both share a good amount of logic. However, while [`iced_winit`] provides a [`raw-window-handle`] to initialize the chosen renderer, `iced_glutin` provides an OpenGL loader function instead.

The work on this crate also introduced a couple of new concepts in `iced_native` that should ease the integration of `iced` in existing projects (#313). Specifically, a `Program` trait has been added which can be used together with some `program::State` to feed events to a user interface and obtain its drawing primitives. [The `integration` example] has been updated accordingly to showcase how to use them!

# `iced_graphics`
The first draft of `iced_glow` was implemented by copy-pasting a lot of code from [`iced_wgpu`]. As most of this code was backend-agnostic and shared between renderers, it was abstracted away and moved to an intermediate crate: `iced_graphics`.

`iced_graphics` contains a bunch of backend-agnostic types that can be leveraged to build a renderer. The most important concepts it contains are:

- The current [`Primitive`] type from [`iced_wgpu`].
- A [`Layer`] type that flattens a [`Primitive`] tree for rendering.
- A default renderer that can be customized with a generic backend. It implements the `Renderer` trait for all the built-in widgets.
[...]

PR #364, opened 2020-05-27
Feature gates `font-kit` behind `"system_font"` feature.

### LINKED ISSUE #58: Use something other than `font-kit` and/or put it behind feature gate
`font-kit` seems like a nice lib, however it's very much *not* pure Rust, requiring not only a full c++ build system but also several C libs and their associated header files to be installed on a system (at least on Linux), and as such presents a fairly significant bar to entry. It would be nice not to depend on it. Also, even after getting it to compile on Fedora, I stlil get this output in my terminal when runing the `todos` example, though it does seem to work at least.

*edit*

error output was fixed by updating outdated `fontconfig-devel`.

Just for reference, on Fedora 30, I had to install `cmake`, `g++`, `freetype`, `freetype-devel`, `expat-devel`, and `fontconfig-devel` for it to work.

azriel91:
Closes #58.

I thought `iced` wasn't working when nothing was showing up in the examples, and it happened to be #199 / #270 / #351. Should've searched issues earlier =/.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `"system_font"` feature gates reading system fonts. [#364]
+
+https://github.com/hecrj/iced/pull/364
 
 ## [0.1.1] - 2020-04-15
 ### Added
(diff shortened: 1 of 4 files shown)


hecrj:
Hey, thanks!

I have some related work in #354 that will have some conflicts with this. I think it'd be better to wait until that PR lands before tackling this. I will merge it very soon!

azriel91:
Closing as this is already feature gated in `master` :v:

hecrj:
It's not completely feature-gated yet!

`iced_wgpu` and `iced_glow` both always enable the `font-source` feature of `iced_graphics`. We should probably add a feature flag `default-system-font` to both like you did in this PR,

azriel91:
oh! shall quickly do that now

PR #365, opened 2020-05-28
Make `Font` an associated type of `text_input::Renderer`

Imberflur:


Imberflur:
Updated to resolve conflicts

graphics/src/widget/text_input.rs
```diff
@@ -27,6 +27,8 @@ impl<B> text_input::Renderer for Renderer<B>
 where
     B: Backend + backend::Text,
 {
+    type Font = Font;
+
```
hecrj, line 31: For consistency, we should remove this blank line between associated types.

native/src/renderer/null.rs
```diff
@@ -103,6 +103,8 @@ impl scrollable::Renderer for Null {
 }
 
 impl text_input::Renderer for Null {
+    type Font = Font;
+
```
hecrj, line 107: Same here!

Imberflur:
@hecrj Thanks for the review. I removed the blank lines :slightly_smiling_face:

PR #370, opened 2020-05-30
Improvement/58/feature gate font kit

### LINKED ISSUE #58: Use something other than `font-kit` and/or put it behind feature gate

azriel91:
Closes #58.

I thought `iced` wasn't working when nothing was showing up in the examples, and it happened to be #199 / #270 / #351. Should've searched issues earlier.

Successor to #364 -- couldn't reopen it after deleting the branch unfortunately.

I'm not sure whether to choose `-` or `_` between feature words, since there's a `"glow_canvas"` feature in the top level `Cargo.toml`.

I've tested this using the `custom_widget` example for both `iced_wgpu` and `iced_glow`. For `iced_glow` I disabled default features from `iced` in `examples/custom_widget/Cargo.toml`, and then running:

```bash
cargo run --package custom_widget --no-default-features --features "glow"
cargo run --package custom_widget --no-default-features --features "glow glow_default_system_font"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `"system_font"` feature gates reading system fonts. [#370]
+
+[#370]: https://github.com/hecrj/iced/pull/370
 
 ## [0.1.1] - 2020-04-15
 ### Added
--- a/glow/src/text.rs
+++ b/glow/src/text.rs
@@ -12,15 +12,20 @@ pub struct Pipeline {
 
 impl Pipeline {
     pub fn new(gl: &glow::Context, default_font: Option<&[u8]>) -> Self {
+        let default_font = default_font.map(|slice| slice.to_vec());
+
         // TODO: Font customization
-        let font_source = font::Source::new();
+        #[cfg(feature = "default_system_font")]
+        let default_font = {
+            default_font.or_else(|| {
+                font::Source::new()
+                    .load(&[font::Family::SansSerif, font::Family::Serif])
+                    .ok()
+            })
+        };
 
         let default_font =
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `"system_font"` feature gates reading system fonts. [#370]
+
+https://github.com/hecrj/iced/pull/370
```
hecrj, line 11: Thanks for updating the `CHANGELOG`! I believe we are missing a `[#370]:` prefix here.

glow/src/text.rs
```diff
@@ -12,15 +12,21 @@ pub struct Pipeline {
...
-        // TODO: Font customization
-        let font_source = font::Source::new();
+        let default_font = default_font.map(|slice| slice.to_vec());
 
-        let default_font =
-            default_font.map(|slice| slice.to_vec()).unwrap_or_else(|| {
+        // TODO: Font customization
+        #[cfg(feature = "default_system_font")]
+        let default_font = {
+            let font_source = font::Source::new();
```
hecrj, line 20: We may be able to move this line inside the `or_else` closure. This way we avoid creating a `Source` when the user has already provided a `default_font`.

CHANGELOG.md
azriel91: wow, hard to believe I missed that, fixing

PR #371, opened 2020-05-31
Add custom window background/clear color (incl. transparency) support, fixes #272

### LINKED ISSUE #272: Is it currently possible to create a (semi-)transparent window on native platforms?
I am currently working on a fullscreen overlay on Windows.
There are a few things I am still trying to figure out. That includes:
- most importantly: making a window transparent and / or semi-transparent while keeping it's widgets visible (maybe through iced_native or iced_winit?)
- starting a window in fullscreen mode
- getting the window handle (that is specifically the hwnd in Windows)

If anybody could lend a hand, thank you very much!

myfreeweb:
wgpu would currently ignore the alpha: https://github.com/gfx-rs/wgpu/issues/687

glow (and naively patched wgpu) requires premultiplied alpha, so if you don't multiply the RGB by the A right now, the semi-transparent color would be wrong (too bright).

winit with_transparent doesn't seem necessary — on wayland at least.

---

for testing:

```diff
--- i/examples/counter/src/main.rs
+++ w/examples/counter/src/main.rs
@@ -1,7 +1,12 @@
-use iced::{button, Align, Button, Column, Element, Sandbox, Settings, Text};
+use iced::{
+    button, Align, Button, Color, Column, Element, Sandbox, Settings, Text,
+};

 pub fn main() {
-    Counter::run(Settings::default())
+    Counter::run(Settings {
+        background_color: Color::new(0.1, 0.2, 0.3, 0.3),
+        ..Settings::default()
+    })
 }

 #[derive(Default)]
```

Screenshot showing both patched wgpu and glow with the same too-bright color (this is not how 0.1,0.2,0.3 is supposed to look like) as a result of not premultiplying alpha:

Screen_2020-05-31-16:46:20-fs8

Of course for `Color::TRANSPARENT` everything is fine.


hecrj:
Hmm... After thinking a bit more about it, I believe the background color should be able to change at runtime.

I imagine we could add it as an additional `Application` method. Thus, I am reverting this change for now.

Issue #348, opened 2020-05-18
Centering TextInput


imjasonmiller:
Hi! 

I wondered whether it is possible to center (_x-_ and _y_) the contents of a [`TextInput`](https://docs.rs/iced_native/0.2.2/iced_native/widget/text_input/struct.TextInput.html)? 

I've currently placed the [`TextInput`](https://docs.rs/iced_native/0.2.2/iced_native/widget/text_input/struct.TextInput.html) inside of a [`Container`](https://docs.rs/iced_native/0.2.2/iced_native/widget/container/struct.Container.html) with the [`align_x`](https://docs.rs/iced_native/0.2.2/iced_native/widget/container/struct.Container.html#method.align_x) and [`align_y`](https://docs.rs/iced_native/0.2.2/iced_native/widget/container/struct.Container.html#method.align_y) methods, but while this seems to work for [`Text`](https://docs.rs/iced_native/0.2.2/iced_native/widget/text/struct.Text.html), with [`.horizontal_alignment(HorizontalAlignment::Center)`](https://docs.rs/iced_native/0.2.2/iced_native/widget/text/struct.Text.html#method.horizontal_alignment), it fails for [`TextInput`](https://docs.rs/iced_native/0.2.2/iced_native/widget/text_input/struct.TextInput.html).

I'm trying to build the UI below:

![image](https://user-images.githubusercontent.com/12190184/82203678-e5f1c480-9903-11ea-91b3-4387c22b544e.png)

This is where I'm currently at:

![image](https://user-images.githubusercontent.com/12190184/82204033-77f9cd00-9904-11ea-8785-8b663f33245a.png)

The content for each [`TextInput`](https://docs.rs/iced_native/0.2.2/iced_native/widget/text_input/struct.TextInput.html) is aligned to the left. Can this be centered in some way? I tried applying [`.width(Length::Shrink)`](https://docs.rs/iced_native/0.2.2/iced_native/widget/text_input/struct.TextInput.html#method.width), thinking it might size it according to the value/placeholder length, but the element seems to disappear in its entirety in that case.

Thank you for creating this library. It's easy to use for a Rust beginner so far.

hecrj:
That looks great!

It's not currently possible to center the contents of a `TextInput`, but it's definitely something we should implement eventually.

I imagine an `horizontal_alignment` method would work well and be consistent with the `Text` widget. The implementation will be a bit tricky because the alignment will affect the logic of text measurements for mouse interaction and text selection.

[...]

PR #373, opened 2020-05-31
[Feat] Add Horizontal Text alignment support in TextInput

phoenisx:
## Details
Add support for Horizontally aligning `Text` primitive inside `TextInput` widget.

This PR also takes care of Cursor position and selection respective to Horizontal Alignment of the Text inside TextInput.

This closes issue #348 

### Left Aligned
left-aligned

### Center Aligned
center-aligned

### Right Aligned
right-aligned

## Added functionalities

* [x] Horizontal alignment of `Text` inside `TextInput` widget.
* [x] Cursor movement and selection around the `Text`, with different alignments
* [x] Clipping`Text` for different alignments
* [x] Mouse Click and Move events to select `Text` for different alignments


native/src/widget/text_input.rs
```diff
@@ -193,8 +209,38 @@ where
...
+            self.font,
+        );
+
         let mut text = layout::Node::new(limits.resolve(Size::ZERO));
-        text.move_to(Point::new(padding, padding));
+
+        let x = {
+            let width = limits.max().width;
+            let center = (width / 2.0) - (text_visual_size / 2.0);
+            let right = width - text_visual_size;
```
Imberflur, line 223: I think if you added the `padding` to `center` and `right` here that would fix the padding issue. However, it seems like text alignment should be handled the same way as it is in the `Text` widget where the actual alignment is deferred to the renderer / whatever text layout lib the renderer is using.

native/src/widget/text_input.rs
```diff
@@ -193,8 +209,38 @@ where
...
+                    if right < padding {
+                        padding
+                    } else {
+                        right
+                    }
+                }
+            }
+        };
+
+        text.move_to(Point::new(x, padding));
```
hecrj, line 243: I am not convinced the `layout` logic needs to change to accomodate the alignment. Instead, I think we should change rendering and measuring functions.

Is there any particular reason you went with this approach?

graphics/src/widget/text_input.rs
```diff
@@ -107,6 +108,34 @@ where
 
         let text = value.to_string();
 
+        let current_text_bounds = {
+            if text.is_empty() {
+                let placeholder_visual_width =
+                    self.measure_value(&placeholder.to_string(), size, font);
```
hecrj, line 114: Do we need to measure again here?

There is a `text_width` measurement in line 250 we may be able to reuse to position things properly.

examples/text_input_aligned/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+name = "text_input_aligned"
+version = "0.1.0"
+authors = ["Subroto Biswas <sub14biswas@gmail.com>"]
+edition = "2018"
+publish = false
+
+[dependencies]
+iced = { path = "../..", features = ["async-std", "debug"] }
+log = "0.4.0"
+log4rs = "0.11.0"
```
hecrj, line 11: We use `env_logger`. We should avoid adding additional development dependencies unless necessary.

examples/text_input_aligned/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "text_input_aligned"
+version = "0.1.0"
+authors = ["Subroto Biswas <sub14biswas@gmail.com>"]
+edition = "2018"
+publish = false
+
+[dependencies]
+iced = { path = "../..", features = ["async-std", "debug"] }
```
hecrj, line 9: I believe `async-std` is not necessary.

examples/text_input_aligned/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+log4rs = "0.11.0"
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+async-std = "1.0"
+
+[package.metadata.deb]
+assets = [
+    ["target/release/todos", "usr/bin/iced-todos", "755"],
+    ["iced-todos.desktop", "usr/share/applications/", "644"],
+]
```
hecrj, line 20: This section can be removed. Probably a leftover from copy-pasting :sweat_smile:

examples/text_input_aligned/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+            .into()
+    }
+}
+
+fn main() {
+    log4rs::init_file("examples/hello_world/log4rs.yml", Default::default())
+        .unwrap();
+
+    State::run(Settings::default());
+}
```
hecrj, line 79: Maybe it'd be interesting to merge this example in the `tour`, as it's only showcasing a single method of a `TextInput`. For now, it's very useful for testing, though!

graphics/src/widget/text_input.rs
Shub1427: `text_width` is for user typed text, while this one is for `placeholder_text` width.

This is required to place the `placeholder` text at the center/left-aligned when `horizontal_alignment` is `Center`/`Right`. This will only be calculated when user input is empty.

image


I couldn't think of any other way to make placeholder center/left-aligned.

hecrj: `text_width` measures the placeholder if the value is empty:

https://github.com/hecrj/iced/blob/718c8fd393987e9c4e1c486e03ce31175b5f94a3/graphics/src/widget/text_input.rs#L250-L254

native/src/widget/text_input.rs
Shub1427: I initially tried with the `draw` logic itself. I was successful to align text and cursor, but Clip primitive was also bound to `text_bounds`.

I figured out that I was changing every x position for Text, Cursor Quad, and Clip, using kind of similar code every place for these measurements.

They all were dependent on the same `text_bounds` variable and I had to change the `x` position in `text_bounds`
for each of them, with similar kind of code (not exactly the same, thus the duplicity in code). Thus, I moved the logic to the layout call where `text_bounds` was initially created.

Even I am not sure if this is right, but while going through the code I figured out that `widget.draw` is basically called during the layout phase, i.e before every `RequestRedraw`, so thought this might be ok.

native/src/widget/text_input.rs
Shub1427: Oh!! Got it. But shouldn't padding be applied after the position is figured out. Not sure!! Anyways Thanks I think your suggestion will fix the issue, will try it out tonight! 👍

graphics/src/widget/text_input.rs
Shub1427: My Bad. I will move this measurement to the top then. 😅

native/src/widget/text_input.rs
Shub1427: So I tried implementing this on `draw` calls and have successfully implemented the cursor, cursor selection, and text alignments as per the `horizontal_alignment` set.

But now I am stuck with some wrong offset set for horizontal_alignment `Center` and `Right`.

issue

As you can see in the above gif, for Left alignment things work, but for Center/Right alignment the moment Clipping starts, it starts moving to the left, due to some wrong offset.

I tried to change the `measure_cursor_and_scroll_offset` functions call signature, but it is being used some other `offset` call, where I don't have access to `horizontal_alignment`.

Any suggestions for this situation would help me a lot. 🙇

Shub1427:
Thanks for your suggestions. I will make the changes accordingly.

Also, I do get your point @hecrj about moving the logic to `draw` calls, as it allows any developer using `iced` more control over `layout` and `draw` calls. 👍

native/src/widget/text_input.rs
Shub1427: I fixed the above issue, but there are some other issues still.

I will try to figure it out before posting it here 👍

Shub1427:
Hi @hecrj @Imberflur 

I need some help understanding the Clip primitive.

issue

As you can see in the above gif, using a positive `offset` which can have a min value of `0.0`, does not allow to move any further to left, for `Center` and `Right` Horizontal alignments.

I tried using a negative `offset` as well, but it breaks the Clip Primitive. Now to make things work, I need to fiddle with Clip primitive's `layer` implementation modifying its translation, which will affect other widgets as well.

My question is how should I handle negative offsets in Clip Primitive.

imjasonmiller:
This looks great @phoenisx! Is this something you're still working on?

phoenisx:
No! I have done the changes. Waiting for the PR to get reviewed. 

cc: @hecrj 

If changes look good, will rebase and push (or the other way around, whatever works @hecrj ).

Imberflur:
What is the status of this?

phoenisx:
@Imberflur I am done with my changes and awaiting for PR review!!

If everything works fine, will rebase the branch and update the PR (resolving conflicts).

imjasonmiller:
I'm not sure if you were notified @hecrj with the request for review, but have the requested changes for this pull request been met?

hecrj (commented):
I don't think it's a good idea to merge this at this point.

The changes here seem to affect the logic of the old behavior and introduce overhead unconditionally. The alignment logic also seems to be spread everywhere, affecting many different parts of graphics and native code.

Additionally, there are important changes coming to our text rendering pipeline and all this code will most likely need to change.

PR #377, opened 2020-06-01
[WIP] Add damage tracking, fixes #367

### LINKED ISSUE #367: Damage tracking / partial present
[Damage tracking](https://emersion.fr/blog/2019/intro-to-damage-tracking/) allows for power savings by not rerendering regions unchanged between frames. There's actually two optimizations made possible by tracking damage: in the desktop compositor (you just tell it the region) and in the application (a bit more tricky, but e.g. [I've implemented support in WebRender+EGL](https://phabricator.services.mozilla.com/D61062) recently).

Doesn't look like wgpu supports passing the `VkPresentRegionKHR` yet, so I've raised an issue for this: https://github.com/gfx-rs/wgpu/issues/682

On the new GL backend though, it should be possible to use `EGL_KHR_swap_buffers_with_damage` ([e.g.](https://phabricator.services.mozilla.com/D51517)).

But first, iced should do the actual tracking :)

unrelentingtech:
fixes #367

Experimenting with implementing damage by "diffing" `Primitive`s.

- [x] primitives' `bounds` is not what I expected, e.g. the `x` of a centered text box under the slider in the tour example has its `x` in the *middle* of the screen o_0
- [x] be more precise by not joining everything into one rect
  - but for debugging one rect is easier..
- [x] don't use unstable `fold_first` (`iterator_fold_self`)
- [ ] wgpu for now should at least skip rendering when there's no damage
- [ ] why do mouse clicks always cause whole screen damage?
- [x] with the compositor hints (`swap_buffers_with_damage`) we lose the winit border hover effects (we don't damage for them).. fix this somehow?
- [ ] mouse interaction when skipping rendering due to zero damage ?? why is mouse cursor determined by actual pixel rendering?

Demo: [Wayfire `-d` damage debug view, couple examples, joined rect](https://gfycat.com/immaculateorganicblackmamba); [now with many rects!](https://gfycat.com/joyfulfragranthog)

### DIFF
--- a/core/src/font.rs
+++ b/core/src/font.rs
@@ -1,5 +1,5 @@
 /// A font.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub enum Font {
     /// The default font.
     ///
--- a/core/src/rectangle.rs
+++ b/core/src/rectangle.rs
@@ -126,6 +126,27 @@ impl Rectangle<f32> {
         }
     }
 
+    /// Computes the union with the given [`Rectangle`].
+    ///
+    /// [`Rectangle`]: struct.Rectangle.html
+    pub fn union(&self, other: &Rectangle<f32>) -> Rectangle<f32> {
+        let x = self.x.min(other.x);
+        let y = self.y.min(other.y);
+
+        let lower_right_x = (self.x + self.width).max(other.x + other.width);
+        let lower_right_y = (self.y + self.height).max(other.y + other.height);
+
+        let width = lower_right_x - x;
+        let height = lower_right_y - y;
+
+        Rectangle {
+            x,
+            y,
+            width,
+            height,
+        }
+    }
+
     /// Snaps the [`Rectangle`] to __unsigned__ integer coordinates.
     ///
     /// [`Rectangle`]: struct.Rectangle.html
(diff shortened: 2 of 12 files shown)


Imberflur:
@myfreeweb I think the text bounds are an artifact of preparing the bounds for the behavior of [`glyph_brush` alignment](https://docs.rs/glyph_brush/0.7.0/glyph_brush/enum.HorizontalAlign.html). I ran into this when implementing a renderer for `iced`. It is pretty wacky imo...

myfreeweb:
@Imberflur yeah. This should at least be documented in the `Primitive::Text` doc string. Or, indeed, pushed out closer to interacting with that crate, since we have the alignments stored in the `Primitive::Text` anyway.

Imberflur:
> I intentionally implemented this behavior here: #281
> 
> It allows us to represent unbounded text with alignment without measuring it first. We should probably change the `Rectangle` type to make this distinction explicit. If you have any other suggestions, I'd love to hear them.

Thanks for the clarification! That makes a lot of sense. My main annoyance with this was how a seemingly unrelated parameter in `glyph_brush` affected the meaning of the position. I think clearer docs for [`Section`](https://docs.rs/glyph_brush/0.7.0/glyph_brush/struct.Section.html#fields) could help with that. Not really related to `iced` though.

myfreeweb:
huh, looks like the winit header bar fixed itself after rebase

hecrj:
As we are not ready to write code for this yet, I believe we can close this for now.

This should probably land some time after #553. We can start additional discussions over Zulip or in an issue.

PR #378, opened 2020-06-01
Add on_release message to Slider

Masterchef365:
Sometimes, it's useful to have a message fire only when the user has finished changing a parameter. Other widgets like text inputs may also benefit from this functionality, and I'd be happy to implement those as well!

### DIFF
--- a/native/src/widget/slider.rs
+++ b/native/src/widget/slider.rs
@@ -187,10 +187,12 @@ where
                     }
                 }
                 mouse::Event::ButtonReleased(mouse::Button::Left) => {
-                    if let Some(on_release) = self.on_release.clone() {
-                        messages.push(on_release);
+                    if self.state.is_dragging {
+                        if let Some(on_release) = self.on_release.clone() {
+                            messages.push(on_release);
+                        }
+                        self.state.is_dragging = false;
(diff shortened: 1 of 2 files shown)


Masterchef365:
Incorporated the changes you suggested into it; had to make the message `impl Clone` but I assume that's not a big deal since `Button` requires it too. Added a bit of an explanation for the on_release message as well; the use case I have is that a long running task (actually a whole thread) is spawned to perform an update to a large amount of data every time the slider is changed. `on_change` would fire that event every time a mouse position update was detected, and the result was that my app suddenly used 100% of all cores every time the slider was moved. I could've solved this by limiting when the thread could be spawned, but that wasn't exactly the behavior I wanted.

native/src/widget/slider.rs
```diff
@@ -171,6 +187,9 @@ where
                     }
                 }
                 mouse::Event::ButtonReleased(mouse::Button::Left) => {
```
hecrj, line 189: I just noticed we may need to check `self.state.is_dragging` is `true` in this branch.

Masterchef365: Good catch!

PR #380, opened 2020-06-02
Removes `0.5` addition when using `round()`.

azriel91:
Yo, saw a `(f32 + 0.5).round()`, and [`f32::round()`](https://doc.rust-lang.org/std/primitive.f32.html#method.round) already rounds to the nearest integer ([playpen](https://play.rust-lang.org/)).

```rust
3.8.round(): 4
3.3.round(): 3
```

### DIFF
--- a/core/src/rectangle.rs
+++ b/core/src/rectangle.rs
@@ -133,8 +133,8 @@ impl Rectangle<f32> {
         Rectangle {
             x: self.x as u32,
             y: self.y as u32,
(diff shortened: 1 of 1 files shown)


hecrj:
This method should have actually been named `snap` or something similar.

It's not meant to round, but instead create a new `Rectangle` with unsigned integer coordinates that at least contains the original one. We may move this logic to `iced_graphics` at some point.

Issue #313, opened 2020-04-22
Less frameworky way to run the application


vorner:
Hello

First, I believe Iced is actually the first good-looking GUI API out there, so thanks!

Nevertheless, I have some small suggestion/feature request. There are two extremes how to run the application. One of them is the `Application::run` method that just takes care of everything and insists on managing the lifetime of the whole application, including setting up of a futures executor. On the other hand, there's the very low-level and manual way shown in https://github.com/hecrj/iced/blob/master/examples/integration/src/main.rs.

I'd like to have something in between. Something less verbose and more high level than the integration example, but something that's still flexible enough to let me setup (or reuse) eg. tokio threadpool the way I want, or to start the application, let it run until someone closes the window and then run some code afterwards. I'm not sure how exactly it looks inside, but for example `hyper` went from high-level „we manage everything“ serve method in previous versions to manual set up of server and converting it into a future one can spawn on the tokio executor.

I've tried scanning through the issues here and didn't find one similar, but maybe I'm not looking well enough.

hecrj:
I agree we need to offer further abstractions to ease the integration process.

The `integration` example showcases one approach. It is there to invite users to build their own custom event loop to satisfy their use cases. As I said when I wrote the example in #183:

> This process, although very flexible, is currently not as straightforward as I would like it to be. However, I prefer to wait, see how folks implement it and gather more use cases before designing further abstractions.

I believe the building blocks are there and now it's just a matter of gathering use cases, exploring different approaches, and finding patterns.

I use the library through my own opinionated approach. Therefore, I don't believe I should be the one building or exploring these abstractions.

Overall, I think this should happen as a result of a bunch of community efforts and collaboration. I placed the low-level building blocks there for this reason!

Finally, I'd like to remind anyone willing to help that, when it comes to collaboration, writing code is a very small part of the equation! It should be the last part of the process.  In my experiencce, the best kind of collaboration happens when a user:

- Socializes (gets in touch!)
- Uses the library
- Gets inspired
- Explores an approach
- Reports results (not just code!)

Code is the easy part. Before that, a considerable amount of communication, inspiration, and exploration is necessary. Please, keep this in mind if you want to take a shot at this!

krooq:
This would be a great way to sandbox ideas that may not fit into the current architecture/lifecycle.
I'd like to see how we can abstract out the lifecycle :D

krooq:
[...]

PR #384, opened 2020-06-03
Spike/313/less frameworky integration

azriel91:
Initial approach for #313. This allows users to control the `event_loop`:

```rust
pub fn main() {
    let mut event_loop = EventLoop::with_user_event();
    let mut context =
        Events::initialize(&mut event_loop, Settings::default());

    event_loop.run(move |event, _, control_flow| {
        // users have control over `control_flow`
        // and can run other code as well

        context.handle_winit_event(event, control_flow);
    })
}
```

Todo:

* [ ] Re-export `EventLoop` from `iced_winit` / `iced_glutin`?
* [ ] Figure out what `Context` looks like for `wasm32`.
* [x] Cut down example?
* [ ] Update example screenshot

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,7 @@ members = [
     "examples/game_of_life",
     "examples/geometry",
     "examples/integration",
+    "examples/library_usage",
     "examples/pane_grid",
     "examples/pokedex",
     "examples/progress_bar",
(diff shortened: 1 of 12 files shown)


hecrj:
Hey, thanks! I think it's a bit soon to start reviewing code here.

As I said in https://github.com/hecrj/iced/issues/313#issuecomment-617911332, I really think code should be the last part of the process and we have not settled on a particular design yet.

If this draft is meant to showcase a design, then I believe a write-up listing the most important API changes here and the reasoning behind them would help me (and others) digest and discuss the ideas behind it.

hecrj:
Closing this for now in favor of more discussion in #313.

PR #387, opened 2020-06-03
Add a comment of how to clear the display to `integration` example

hatoo:
When I use `integration` example, it took some time to figure out how to clear display without own `scene`.
So I think it's good to share the code by adding this comment.

Feel free to correct me if my code is wrong or my english comment is grammertically bad.

### DIFF
--- a/examples/integration/src/main.rs
+++ b/examples/integration/src/main.rs
@@ -154,37 +154,19 @@ pub fn main() {
                     &wgpu::CommandEncoderDescriptor { label: None },
                 );
 
-                // We draw the scene first
                 let program = state.program();
 
-                scene.draw(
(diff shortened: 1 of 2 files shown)


hecrj:
The comment feels a bit out of place to me.

I think we can simply change the code to clear the screen right there, instead of hiding the side-effect inside `Scene`.

hecrj:
I just gave it a shot but I am not convinced by the result.

The example is meant to showcase how values from `iced` can be injected in a `Scene`. Moving the clearing logic defeats this purpose.

However, the comment adds noise to the integration code, which is the focus of the example.

In short, I am not sure there is a problem worth fixing here.

Issue #388, opened 2020-06-04
Cursor Manipulation


bansheerubber:
Hello! I'm new to Rust and I've decided to use this library for my first project. I'm trying to make a tab auto-complete for a text input. When you press tab, the text inside the input is updated. What I'm wondering is if it is possible to manipulate the cursor so it snaps to the end of the text input when you press tab. Right now, when I update the text input's value when tab is pressed the cursor's position remains unchanged, and as far as I can tell I there is no way make it snap to the end (are the cursor's move_to commands private? is there a way to get around this?).

Also not exactly sure how issues work, this is my first time opening one on an open source project. I noticed there's other questions in here with the question label, so I hope I'm not screwing this up too bad. If I am, I'm sorry!

Loving the library! Thank you for doing a great job maintaining it!

hecrj:
> are the cursor's move_to commands private? is there a way to get around this?

Yes, I tend to make everything private by default unless I have a clear use case in mind. There is no way to currently move the cursor of a `TextInput` programmatically.

However, I believe we should be able to add a `move_cursor_to_end` to `text_input::State` easily.

> Also not exactly sure how issues work, this is my first time opening one on an open source project. I noticed there's other questions in here with the question label, so I hope I'm not screwing this up too bad. If I am, I'm sorry!

No worries! I do not personally have any rules for issues, as long as they are on topic. They are just another channel of communication. In any case, feature requests are completely valid issues!

bansheerubber:
Thank you for the response! I'll try taking a crack at move_cursor_to_end since it doesn't seem too complicated.

PR #391, opened 2020-06-05
Cursor Manipulation

bansheerubber:
This adds some cursor commands to text_input::State, allowing the text_input's cursor's location to be manipulated (as discussed in #388). It should be noted that this only manipulates the cursor's index, and not it's selection.

3 new commands are added:
move_cursor_to_front
move_cursor_to_end
move_cursor_to

The first two should be pretty self explanatory. The last one allows you to supply an arbitrary index, allowing the cursor to be moved into the middle of a word if needed.

One of the things I think could use some improvement is move_cursor_to_end. Right now, you have to supply a string in order to use it (it gets the length of that string, and then sets the cursor's index to the length). I would like it so no argument is required. I can't see a way around this unless we refactor text_input::State to include the text its currently editing, and I'm not sure if this change is wanted or even needed. Would like some feedback on that. If the way I've implemented is good enough, then that's ok too.

### DIFF
--- a/native/src/widget/text_input.rs
+++ b/native/src/widget/text_input.rs
@@ -696,8 +696,8 @@ impl State {
     /// 
     /// [`Cursor`]: struct.Cursor.html
     /// [`TextInput`]: struct.TextInput.html
-    pub fn move_cursor_to_end(&mut self) {
-        self.cursor.move_to(5000);
+    pub fn move_cursor_to_end(&mut self, value: &String) {
+        self.cursor.move_to(value.len());
     }
 }
 
--- a/native/src/widget/text_input.rs
+++ b/native/src/widget/text_input.rs
@@ -696,8 +696,8 @@ impl State {
     ///
     /// [`Cursor`]: struct.Cursor.html
     /// [`TextInput`]: struct.TextInput.html
-    pub fn move_cursor_to_end(&mut self, value: &String) {
-        self.cursor.move_to(value.len());
+    pub fn move_cursor_to_end(&mut self) {
+        self.cursor.move_to(usize::MAX);
     }
 
     /// Moves the [`Cursor`] of the [`TextInput`] to an arbitrary location.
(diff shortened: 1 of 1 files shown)


native/src/widget/text_input.rs
```diff
@@ -683,6 +683,30 @@ impl State {
     pub fn cursor(&self) -> Cursor {
         self.cursor
     }
+
+    /// Moves the [`Cursor`] of the [`TextInput`] to the front of the input text.
+    ///
+    /// [`Cursor`]: struct.Cursor.html
+    /// [`TextInput`]: struct.TextInput.html
+    pub fn move_cursor_to_front(&mut self) {
```
teymour-aldridge, line 691: Given that this could be achieved with `move_cursor_to(0)`, is this method necessary?

native/src/widget/text_input.rs
```diff
@@ -683,6 +683,30 @@ impl State {
...
+        self.cursor.move_to(0);
+    }
+
+    /// Moves the [`Cursor`] of the [`TextInput`] to the end of the input text.
+    ///
+    /// [`Cursor`]: struct.Cursor.html
+    /// [`TextInput`]: struct.TextInput.html
+    pub fn move_cursor_to_end(&mut self, value: &String) {
+        self.cursor.move_to(value.len());
```
hecrj, line 700: Instead of asking for a string reference (which should be `&str`), I believe we should be able to use `usize::MAX` here. The `Cursor` API is already fault-tolerant (i.e. the read methods ask for a `Value` and clamp the indices accordingly).

native/src/widget/text_input.rs
bansheerubber: I implemented the front/end methods because they are self documenting.

native/src/widget/text_input.rs
bansheerubber: Sounds good, will do that.

PR #392, opened 2020-06-05
Add image format options to reduce code bloat

### LINKED ISSUE #376: Add image format options to reduce code bloat
The `image` crate with default options is the absolute leader in code size:

```
 File  .text     Size Crate
 5.7%  14.6% 364.0KiB image
 4.6%  12.0% 297.3KiB wgpu_native
 3.8%   9.9% 246.5KiB std
 3.5%   8.9% 222.1KiB usvg
 2.6%   6.8% 170.0KiB iced_wgpu
```

I don't need to support `ico`, `bmp`, `tga`, `hdr`, `dxt`, `dds`, `farbfeld` and so on in a program that would only ever decode PNGs :)

It's pretty straightforward to add

```toml
[features]
png = ["image/png"]
# and so on
```

but for backwards compat, the `image` feature would have to enable some reasonable default set of formats.

Looks like this can be accomplished using cargo's dependency rename thing (e.g. `image = { … package = "image_rs" }`). What do you think?

---

Ideally, there should be an option to not use `image` at all and just provide raw pixel data. (And with explicit gpu memory management instead of a magical cache :D)

unrelentingtech:
Fixes #376.

### DIFF
--- a/glow/src/backend.rs
+++ b/glow/src/backend.rs
@@ -208,7 +208,7 @@ impl backend::Text for Backend {
     }
 }
 
-#[cfg(feature = "image_rs")]
+#[cfg(feature = "image")]
 impl backend::Image for Backend {
     fn dimensions(&self, _handle: &iced_native::image::Handle) -> (u32, u32) {
         (50, 50)
--- a/glow/src/backend.rs
+++ b/glow/src/backend.rs
@@ -208,7 +208,7 @@ impl backend::Text for Backend {
     }
 }
 
-#[cfg(feature = "image")]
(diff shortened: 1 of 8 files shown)


hecrj (approved):
Thank you! :partying_face: I have rebased onto `master` without issues.

And sorry for the delay! I wanted to find a way to expose the individual features in the root `iced` crate too, but it'd be quite cumbersome to do so currently.

In any case, I think this is still very valuable for those that want to integrate `iced_wgpu` in their own event loop. Also, relying on `iced_winit` directly is always an option.

I think we can go ahead and merge!

PR #397, opened 2020-06-08
Intuitive `PaneGrid` resize

hecrj:
This PR allows the panes in a `PaneGrid` to be resized by simply dragging the splits, instead of using a modifier key + right click.

[Pane grid resize - Iced](https://gfycat.com/unconsciousperfectindianpalmsquirrel)

### DIFF
--- a/native/src/widget/pane_grid.rs
+++ b/native/src/widget/pane_grid.rs
@@ -26,7 +26,7 @@ pub use state::{Focus, State};
 
 use crate::{
     keyboard, layout, mouse, Clipboard, Element, Event, Hasher, Layout, Length,
-    Point, Size, Widget,
+    Point, Rectangle, Size, Widget,
 };
 
 /// A collection of panes distributed using either vertical or horizontal splits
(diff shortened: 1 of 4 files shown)


native/src/widget/pane_grid.rs
```diff
@@ -208,14 +208,15 @@ impl<'a, Message, Renderer> PaneGrid<'a, Message, Renderer> {
     /// Enables the resize interactions of the [`PaneGrid`], which will
     /// use the provided function to produce messages.
     ///
-    /// Panes can be resized using `Modifier keys + Right click`.
+    /// The `leeway` describes the amount of space around a split that can be
```
clarkmoody, line 211: Maybe clarification on how `leeway` is applied. Is it

- Space *on either side* of the split?
- Total space straddling the split, so you have `leeway / 2` on either side?

hecrj: I tried to clarify it in https://github.com/hecrj/iced/pull/397/commits/8b93c9cb6a623d8f936e7e7135a2598e3ce4e4ef.

clarkmoody: That's better :-)

PR #406, opened 2020-06-13
Add `background_color` to `Application` and `Sandbox`

hecrj:
Supersedes #371.

This PR adds a `background_color` method to the `Application` and `Sandbox` traits to control the default clear color of the underlying renderer. Like `title` and `mode`, the returned value can change at runtime.

/cc @myfreeweb


myfreeweb:
Works great, thanks.

PR #412, opened 2020-06-18
Color.mix method for mixing two colors together

artursapek:
This PR adds a simple method for mixing colors together! :^)


artursapek:
nevermind lol!

Issue #443, opened 2020-07-09
What is the intended use of text_input::State::cursor()?


mtkennerly:
I'm trying to get the selection of a `TextInput`. I see that there is a public `text_input::State::cursor()` returning a `Cursor`, which has a public `state()` method. However, I can't call `Cursor::state()` because it takes a `text_input::value::Value`, which doesn't seem to be exposed, and even if I could call it, I don't think `text_input::cursor::State` is exposed in order to match on the return value. I can understand why `cursor::State` and `Value` would be kept private if users are supposed to go through `text_input::State` as a wrapper around them, but then I'm not sure why `text_input::State::cursor()` would ever be used as a public method, since everything else on `Cursor` is private. Am I missing something?

hecrj:
`iced_graphics` uses the `Cursor` API to draw a `TextInput`:

https://github.com/hecrj/iced/blob/5c4f5ae5ecb36703a95cafb2cd58692529c9466d/graphics/src/widget/text_input.rs#L127-L132

mtkennerly:
Ah, so it supports cases where you're working directly with the `iced_native` crate rather than the `iced` crate. Thanks for the quick response!

Would you be open to adding methods to `text_input::State` for accessing the cursor selection (similar to how `move_cursor_to_end` works)?

hecrj:
> Would you be open to adding methods to text_input::State for accessing the cursor selection (similar to how move_cursor_to_end works)?

Maybe! What are the use cases you have in mind?

Imberflur:
Might be unrelated, but when working with input field tabbing I noticed a `select_all` method would be useful.

mtkennerly:
My particular use case is that I'm using a Subscription to implement my own copy/cut/undo/redo shortcuts (I'd like to figure out how to contribute them back if possible, but for now I just need to get something working). For copy/cut, I need to know what's currently selected, and for cut, I also need to delete the selected part of the text and clear the selection:

```rust
// in update()
match shortcut {
    // backup_target_input is a text_input::State, and backup_target_history is my own TextHistory
    Shortcut::ClipboardCut => {
        self.config.backup.path = crate::shortcuts::cut_to_clipboard_from_iced(&self.config.backup.path, &self.backup_screen.backup_target_input.cursor());
        self.backup_screen.backup_target_history.push(&self.config.backup.path);
    },
}
```

```rust
[...]

PR #445, opened 2020-07-10
Fix panic on paste in TextInput after programmatic modification of contents

mtkennerly:
This fixes the panic mentioned in #443, where the end of the cursor selection might be larger than the length of the current input value.


hecrj:
Thanks for catching this!

The main issue here was that `Cursor::selection` was not using the `Cursor::state` method, which clamps the indices accordingly.

I made some changes to fix this. Let me know what you think!

mtkennerly:
Even better \:D I tested your changes with my simplified example, and it's working just fine.

Issue #460, opened 2020-07-27
Tooltip widget


yusdacra:
Now that we can use `Overlay`s, how would one add a "tooltip widget"? As far as I understand to display an `Overlay` a widget needs to implement the `overlay()` function, but that means if I want to display a tooltip when I hover, say, a radio button I wouldn't be able to use iced's implementation, because I would need to implement the function myself (which I can't, since the trait is already implemented).

So I guess a way would be to wrap the iced radio button in a custom widget and implement the `overlay()` function so that it displays a tooltip when the radio button is hovered? Could also be made generic by making a `Tooltip` widget that takes any widget as `content`, and shows a tooltip over that `content` widget when it's being hovered? Would love to see something like this in.

hecrj:
Fixed by #465.

PR #465, opened 2020-07-28
Tooltip widget

yusdacra:
This PR implements #460 as described in the issue. I have added a very simple example (definitely needs to be improved) to showcase the widget. I'm not sure about the widget name, since this acts as kind of a Container too. Or maybe the API could be done in a different way? I'd like to hear ideas about this.

Stuff to do:
- [x] Improve the example
- [ ] Make the tooltip change position if it overflows a window edge to keep it visible

### DIFF
--- a/native/src/widget/tooltip.rs
+++ b/native/src/widget/tooltip.rs
@@ -3,46 +3,43 @@ use std::hash::Hash;
 
 use iced_core::Rectangle;
 
+use crate::widget::text::{self, Text};
 use crate::{
-    event, layout, overlay, Clipboard, Element, Event, Hasher, Layout, Length,
-    Point, Size, Vector, Widget,
+    event, layout, Clipboard, Element, Event, Hasher, Layout, Length, Point,
+    Widget,
 };
 
 /// An element to display a widget over another.
 #[allow(missing_debug_implementations)]
-pub struct Tooltip<'a, Message, Renderer: self::Renderer> {
+pub struct Tooltip<'a, Message, Renderer: self::Renderer + text::Renderer> {
     content: Element<'a, Message, Renderer>,
-    tooltip: Element<'a, Message, Renderer>,
-    tooltip_position: TooltipPosition,
+    tooltip: Text<Renderer>,
+    position: Position,
 }
 
 impl<'a, Message, Renderer> Tooltip<'a, Message, Renderer>
 where
-    Renderer: self::Renderer,
+    Renderer: self::Renderer + text::Renderer,
 {
     /// Creates an empty [`Tooltip`].
     ///
     /// [`Tooltip`]: struct.Tooltip.html
-    pub fn new<T, H>(
-        content: T,
-        tooltip: H,
-        tooltip_position: TooltipPosition,
-    ) -> Self
-    where
-        T: Into<Element<'a, Message, Renderer>>,
-        H: Into<Element<'a, Message, Renderer>>,
-    {
+    pub fn new(
+        content: impl Into<Element<'a, Message, Renderer>>,
+        tooltip: Text<Renderer>,
+        position: Position,
+    ) -> Self {
         Tooltip {
             content: content.into(),
-            tooltip: tooltip.into(),
-            tooltip_position,
+            tooltip,
+            position,
         }
     }
 }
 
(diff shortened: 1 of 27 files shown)


native/src/widget/tooltip.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// Sets the hover_content of the [`Tooltip`].
+    ///
+    /// [`Tooltip`]: struct.Tooltip.html
+    pub fn hover_content<T>(mut self, hover_content: T) -> Self
+    where
+        T: Into<Element<'a, Message, Renderer>>,
+    {
+        self.hover_content = Some(hover_content.into());
+        self
+    }
```
hecrj, line 44: I think we should make empty tooltips impossible. In other words, we should ask for both the `target` and the `content` of the `Tooltip` in `new`.

native/src/widget/tooltip.rs
```diff
@@ -0,0 +1,239 @@
...
+    fn hash_layout(&self, state: &mut Hasher) {
+        struct Marker;
+        std::any::TypeId::of::<Marker>().hash(state);
+
+        self.content.hash_layout(state);
+    }
+
+    fn overlay(
+        &mut self,
+        layout: Layout<'_>,
```
hecrj, line 124: We could consider extending `Widget::overlay` to also provide a `cursor_position`. This should allow us to remove the local `State` of the `Tooltip`.

I'll think about it!

native/src/widget/tooltip.rs
yusdacra: 19fb8caee81bf8856f11772ed47196cbda7b2e3d solves that.

clarkmoody:
I would love for this to have an `overlay_mode` enum that allows me to place it relative to the anchor element rather than flying with the cursor. This would mimic the `data-placement` directive for [Bootstrap tooltip](https://getbootstrap.com/docs/4.1/components/tooltips/).

```rust
enum OverlayMode {
    WithCursor, // current behavior
    Top,
    Right,
    Bottom,
    Left,
}
```

yusdacra:
> I would love for this to have an `overlay_mode` enum that allows me to place it relative to the anchor element rather than flying with the cursor. This would mimic the `data-placement` directive for [Bootstrap tooltip](https://getbootstrap.com/docs/4.1/components/tooltips/).
> 
> ```rust
> enum OverlayMode {
>     WithCursor, // current behavior
>     Top,
>     Right,
>     Bottom,
>     Left,
> }
> ```
> 
> Edit: Great job so far! The example program works for me.

Yep, that sounds like a good idea! I'll try implementing that when I have time.

yusdacra:
> > I would love for this to have an `overlay_mode` enum that allows me to place it relative to the anchor element rather than flying with the cursor. This would mimic the `data-placement` directive for [Bootstrap tooltip](https://getbootstrap.com/docs/4.1/components/tooltips/).
> > ```rust
> > enum OverlayMode {
> >     WithCursor, // current behavior
> >     Top,
> >     Right,
> >     Bottom,
> >     Left,
> > }
> > ```
> > 
> > 
> > Edit: Great job so far! The example program works for me.
> 
> Yep, that sounds like a good idea! I'll try implementing that when I have time.

[https://streamable.com/3rsu4r](url) video showcasing this! @clarkmoody

clarkmoody:
> https://streamable.com/3rsu4r video showcasing this! @clarkmoody

Oh yeah!

native/src/widget/tooltip.rs
yusdacra: Did that in b8c5ebca04d90c99a3fd887a525aca399e575613 and added a `overlay_content_bounds` field. I need it to calculate where to put the tooltip (if `tooltip_position` is different than `TooltipPosition::FollowCursor` in `Tooltip`). There is probably a better way to do it (?), but that's what I do currently.

hecrj (approved):
Thank you for working on this! :tada: 

I made some changes to restrict the API to text-only tooltips for the time being. As they are not interactive, we can avoid using the `Widget::overlay` method completely and simply use a `Clip` primitive in the renderer to produce a new layer.

I have also introduced styling support (same as `Container`) and simplified the example a bit.

I think this is ready to go! We can improve it further in future iterations.

PR #479, opened 2020-08-13
Added Rule widget that creates horizontal and vertical rules for dividing content

BillyDM:
Run the `styling` example to see it in action.

Screenshot from 2020-08-13 13-20-55

In short, a rule can be created simply with:
```rust
// Add a horizontal rule with a vertical spacing of 38.
.push(Rule::horizontal(38))

// Add a vertical rule with a horizontal spacing of 38.
.push(Rule::vertical(38))
```

This is the style sheet:
```rust
/// The appearance of a rule.
#[derive(Debug, Clone, Copy)]
pub struct Style {
    /// The color of the rule.
    pub color: Color,
    /// The width (thickness) of the rule line.
    pub width: u16,
    /// The radius of the rectangle corners.
    pub radius: u16,
    /// The percent from [0, 100] of the filled space the rule
    /// will be drawn.
    pub fill_percent: u16,
}
```

I have not added anything to the `iced_web` crate since I don't have any experience with it.


style/src/rule.rs
```diff
@@ -0,0 +1,51 @@
...
+pub struct Style {
+    /// The color of the rule.
+    pub color: Color,
+    /// The width (thickness) of the rule line.
+    pub width: u16,
+    /// The radius of the rectangle corners.
+    pub radius: u16,
+    /// The percent from [0, 100] of the filled space the rule
+    /// will be drawn.
+    pub fill_percent: u16,
```
clarkmoody, line 16: A few things here:

- Why not either `f32` or `u8`? The upper byte of a `u16` will never be used for values on `[0, 100]`.
- What about a `FillMode` enum?

```rust
enum FillMode {
    // Current behavior
    Percent(f32),
    // Uniform offset from each end, length units
    Padded(f32),
    // Different offset on each end of the rule. Start = top  or left
    AsymmetricPadding { start: f32, end: f32 }, 
}
```

BillyDM: I like the idea of the `FillMode` enum. I'll go ahead and implement that.

As for the u16, all the other styles in iced use u16, so I thought I would keep it consistent. It would make sense for `Percent` to be floating point.

In iced's other stylesheets, anything that deals with units (pixels) is always u16. Maybe this could be changed once iced gets a better sub-pixel anti-aliasing system.

hecrj: What are the use cases for `AsymetricPadding`? I would try to keep it simple for now.

graphics/src/widget/rule.rs
```diff
@@ -0,0 +1,152 @@
...
+                border_radius: style.radius,
+                border_width: 0,
+                border_color: Color::TRANSPARENT,
+            }
+        } else {
+            let line_x = (bounds.x + (bounds.width / 2.0)
+                - (style.width as f32 / 2.0))
+                .round();
+
+            let (line_y, line_height) = match style.fill_mode {
```
hecrj, line 93: It looks like this logic can be abstracted away. Maybe a `FillMode::fill(space: f32) -> (f32, f32)` method returning the starting `offset` and the `length` of the `Rule`.

We should be able to use it for both horizontal and vertical rules.

style/src/rule.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// Produces the style of a rule.
+    fn style(&self) -> Style;
+}
+
+struct Default;
+
+impl StyleSheet for Default {
+    fn style(&self) -> Style {
+        Style {
+            color: [0.6, 0.6, 0.6, 0.51].into(),
```
hecrj, line 48: Just curious, why `0.51`? :eyes:

style/src/rule.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/// The appearance of a rule.
+#[derive(Debug, Clone, Copy)]
+pub struct Style {
+    /// The color of the rule.
+    pub color: Color,
+    /// The width (thickness) of the rule line.
+    pub width: u16,
+    /// The radius of the rectangle corners.
```
hecrj, line 29: Maybe "line corners" is clearer?

style/src/rule.rs
BillyDM: Lol, I thought I would increase the contrast just a tiny bit. I have a color-accurate monitor, so I put whatever looked just right to me.

graphics/src/widget/rule.rs
BillyDM: Good catch! Sometimes I forget you can add member methods to enums. I could probably clean up a lot of code in `iced_audio` with them as well :).

BillyDM:
I just wondered since I haven't added anything to the `iced_web` crate, that it may break if someone tries to use the Rule widget in a web application?

style/src/rule.rs
clarkmoody: Plenty of aesthetic reasons to want asymmetry. You might want the start of the line to be padded and the end to run-out the full container. I'm sure you could achieve the same effect by wrapping a rule in other containers with spacers, padding, etc, but that is more complex than first-class support in a single element.

PR #482, opened 2020-08-17
Upgrade glam to 0.9 & use glam's Mat4 ortho rh gl.

bitshifter:
Hi there, glam 0.8 to 0.9 was a potentially breaking change so I've been trying to upgrade people. In Iced though there was no breakage.


bitshifter:
Ah whoops, I had it working then inverted it to check I was actually changing the right transform then forgot to change it back before I made the PR. I'm on my phone but I think that looks right if you've just switched top and bottom.

hecrj:
Awesome. Thank you again for the PR and also for your work on `glam`!

PR #488, opened 2020-08-20
doc: Add example for Image widget to API docs

andy5995:


hecrj:
I don't think this new example is very appropriate.

Specifically, it references a `Container` first and has conditional compilation flags for Wasm support (which should eventually be an implementation detail).

Maybe we could think of a better one? What prompted you to add this example?

andy5995:
> I don't think this new example is very appropriate.
> 
> Specifically, it references a `Container` first and has conditional compilation flags for Wasm support (which should eventually be an implementation detail).
> 
> Maybe we could think of a better one? What prompted you to add this example?

I might be able to think of a better example as I learn more about iced. I get what you're saying about it not being a good example for the API docs. Maybe best just to close this PR and if I get any ideas as I move forward, I'll ask about them in the iced Zulip server first. Sound good?

hecrj:
@andy5995 Sounds good! Thank you.

Issue #408, opened 2020-06-14
How to integrate input events with existing engine, and use iced as ui library?


DzmitryFil:
I looked at the integration example, and from what i gather all winit events are simply queued to iced, and later processed in a batch. What i would like to do however, is handle event with iced if it can (mouse over ui, text input, etc), and if user input is not handled by iced ui elements, send it to controls of my engine (in the same frame). How can this be done?

Nice work on the library, web/native combination is especially impressive.

kaimast:
Could this be exposed to the `Program` trait? It seems that right now `program::State` just discards the vector of `event::Status`s.

PR #495, opened 2020-08-24
Feature/keyboard handling

dmweis:
Hey!
I am using Iced for a small applicaiton and I wanted to add the abaility to exit by pressing Escape. 
But it seems that there isn't a straight forward way to subscribe to keyboard events as far as I can tell.

There is a way to do it in PaneGrids but the handling is a bit more complicated and it seems you can't handle different combinations of moddifiers (For example handle `ctrl + d` differently from `alt + d`)

I've implemented a small keyboard handler for `container` and I updated the stopwatch exmaple to show it I would use this.
If this is something you like, I'd be interested in extending it to other widgets or including mouse events.

Also Iced is really cool!

### DIFF
--- a/examples/stopwatch/src/main.rs
+++ b/examples/stopwatch/src/main.rs
@@ -1,7 +1,7 @@
 use iced::{
-    button, executor, time, Align, Application, Button, Column, Command,
-    Container, Element, HorizontalAlignment, Length, Row, Settings,
-    Subscription, Text, keyboard,
+    button, executor, keyboard, time, Align, Application, Button, Column,
+    Command, Container, Element, HorizontalAlignment, Length, Row, Settings,
+    Subscription, Text,
 };
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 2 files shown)


dmweis:
Hmmm. I just found out that I could have solved this using a `subscription`. 
It doesn't offer the abaility to bubble events as far as I can tell. But it would be enough for my needs

hecrj:
Thanks!

I think we need to be careful with adding too much functionality to simple widgets. In this case, I believe a better approach would be to create a new widget extending a `Container` by using composition.

But as you mention, using a `Subscription` is the recommended way to go until some kind of event capturing/bubbling lands. It is not completely clear yet how this feature will shape up. Therefore, we should probably wait until then before introducing more logic that could conflict with it.

There are a bunch of related issues and discussions that we should keep in mind:

- [Trying to integrate with Kiss3d](https://iced.zulipchat.com/#narrow/stream/213316-general/topic/Trying.20to.20integrate.20with.20Kiss3d)
- [Layers](https://iced.zulipchat.com/#narrow/stream/213316-general/topic/layers)
- #408 
- #455

PR #500, opened 2020-08-27
Update `wgpu` to `0.6` in `iced_wgpu`

### LINKED ISSUE #438: iced_wgpu should not pin gfx-memory to 0.1.1
The iced-wgpu crate should not pin the gfx-memory dependency to 0.1.1. The issue of performance degradation is because 0.1.1 doesn't actually deallocate memory.

### LINKED ISSUE #486: update to wgpu 0.6
wgpu 0.6 was released, will be nice to update :)

hecrj:
Updates `iced_wgpu` to use the latest release of `wgpu` :tada:

The new release of `wgpu` features a new `StagingBelt` type that we can leverage to reduce reallocations of staging buffers. As a consequence, we no longer pin `gfx-memory` to an old version and, therefore, `iced_wgpu` should no longer leak video memory!

The `integration` example has also been updated accordingly.

Fixes #438 and closes #486.

### DIFF
--- a/examples/integration/src/scene.rs
+++ b/examples/integration/src/scene.rs
@@ -22,17 +22,18 @@ impl Scene {
             color_attachments: &[wgpu::RenderPassColorAttachmentDescriptor {
                 attachment: target,
                 resolve_target: None,
-                load_op: wgpu::LoadOp::Clear,
-                store_op: wgpu::StoreOp::Store,
-                clear_color: {
-                    let [r, g, b, a] = background_color.into_linear();
+                ops: wgpu::Operations {
+                    load: wgpu::LoadOp::Clear({
+                        let [r, g, b, a] = background_color.into_linear();
 
-                    wgpu::Color {
-                        r: r as f64,
-                        g: g as f64,
-                        b: b as f64,
-                        a: a as f64,
-                    }
+                        wgpu::Color {
+                            r: r as f64,
+                            g: g as f64,
+                            b: b as f64,
+                            a: a as f64,
+                        }
+                    }),
+                    store: true,
                 },
             }],
             depth_stencil_attachment: None,
@@ -46,25 +47,23 @@ impl Scene {
 }
 
 fn build_pipeline(device: &wgpu::Device) -> wgpu::RenderPipeline {
-    let vs = include_bytes!("shader/vert.spv");
-    let fs = include_bytes!("shader/frag.spv");
+    let vs_module =
+        device.create_shader_module(wgpu::include_spirv!("shader/vert.spv"));
 
-    let vs_module = device.create_shader_module(
-        &wgpu::read_spirv(std::io::Cursor::new(&vs[..])).unwrap(),
-    );
-
-    let fs_module = device.create_shader_module(
-        &wgpu::read_spirv(std::io::Cursor::new(&fs[..])).unwrap(),
-    );
+    let fs_module =
+        device.create_shader_module(wgpu::include_spirv!("shader/frag.spv"));
 
     let pipeline_layout =
         device.create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
+            label: None,
+            push_constant_ranges: &[],
             bind_group_layouts: &[],
         });
 
     let pipeline =
         device.create_render_pipeline(&wgpu::RenderPipelineDescriptor {
-            layout: &pipeline_layout,
+            label: None,
+            layout: Some(&pipeline_layout),
             vertex_stage: wgpu::ProgrammableStageDescriptor {
                 module: &vs_module,
                 entry_point: "main",
@@ -79,6 +78,7 @@ fn build_pipeline(device: &wgpu::Device) -> wgpu::RenderPipeline {
                 depth_bias: 0,
                 depth_bias_slope_scale: 0.0,
                 depth_bias_clamp: 0.0,
+                ..Default::default()
             }),
             primitive_topology: wgpu::PrimitiveTopology::TriangleList,
             color_states: &[wgpu::ColorStateDescriptor {
--- a/examples/integration/src/scene.rs
+++ b/examples/integration/src/scene.rs
@@ -75,9 +75,6 @@ fn build_pipeline(device: &wgpu::Device) -> wgpu::RenderPipeline {
             rasterization_state: Some(wgpu::RasterizationStateDescriptor {
                 front_face: wgpu::FrontFace::Ccw,
                 cull_mode: wgpu::CullMode::None,
-                depth_bias: 0,
-                depth_bias_slope_scale: 0.0,
-                depth_bias_clamp: 0.0,
                 ..Default::default()
             }),
             primitive_topology: wgpu::PrimitiveTopology::TriangleList,
--- a/wgpu/src/backend.rs
+++ b/wgpu/src/backend.rs
@@ -152,6 +152,7 @@ impl Backend {
 
                 self.image_pipeline.draw(
                     device,
+                    staging_belt,
                     encoder,
                     &layer.images,
                     scaled,
--- a/wgpu/src/backend.rs
+++ b/wgpu/src/backend.rs
@@ -64,6 +64,7 @@ impl Backend {
     pub fn draw<T: AsRef<str>>(
         &mut self,
         device: &wgpu::Device,
+        staging_belt: &mut wgpu::util::StagingBelt,
         encoder: &mut wgpu::CommandEncoder,
         frame: &wgpu::TextureView,
         viewport: &Viewport,
@@ -85,6 +86,7 @@ impl Backend {
                 scale_factor,
                 transformation,
                 &layer,
+                staging_belt,
                 encoder,
                 &frame,
                 target_size.width,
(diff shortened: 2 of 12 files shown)


kvark:
Did you consider `write_buffer`/`write_texture` instead of the staging belt?

examples/integration/src/scene.rs
```diff
@@ -79,6 +78,7 @@ fn build_pipeline(device: &wgpu::Device) -> wgpu::RenderPipeline {
                 depth_bias: 0,
```
kvark, line 78: may omit the depth stuff since it's default as well

wgpu/src/image.rs
```diff
@@ -75,24 +82,25 @@ impl Pipeline {
...
 
-        let uniforms_buffer = device.create_buffer_with_data(
-            uniforms.as_bytes(),
-            wgpu::BufferUsage::UNIFORM | wgpu::BufferUsage::COPY_DST,
-        );
+        let uniforms_buffer =
+            device.create_buffer_init(&wgpu::util::BufferInitDescriptor {
+                label: None,
+                contents: uniforms.as_bytes(),
+                usage: wgpu::BufferUsage::UNIFORM | wgpu::BufferUsage::COPY_DST,
```
kvark, line 89: It seems a bit strange to create this one with contents, and with COPY_DST at the same time.
Since it's for uniforms, and they can change, wouldn't it be easier to create it *without* contents but then mark a flag somewhere that the initial contents are dirty, so that the rendering code will re-upload them?

wgpu/src/backend.rs
```diff
@@ -64,6 +64,7 @@ impl Backend {
     pub fn draw<T: AsRef<str>>(
         &mut self,
         device: &wgpu::Device,
+        staging_belt: &mut wgpu::util::StagingBelt,
```
kvark, line 67: reiterating what I asked in another comment - any strong reason to use this versus `write_buffer`/`write_texture`?

wgpu/src/image.rs
```diff
@@ -75,24 +82,25 @@ impl Pipeline {
             transform: Transformation::identity().into(),
         };
 
-        let uniforms_buffer = device.create_buffer_with_data(
-            uniforms.as_bytes(),
-            wgpu::BufferUsage::UNIFORM | wgpu::BufferUsage::COPY_DST,
-        );
+        let uniforms_buffer =
+            device.create_buffer_init(&wgpu::util::BufferInitDescriptor {
+                label: None,
```
kvark, line 87: nit: would be nice to sprinkle meaningful labels to the internal objects. Users who debug this would appreciate!

wgpu/src/image.rs
```diff
@@ -147,6 +155,7 @@ impl Pipeline {
                     depth_bias: 0,
```
kvark, line 155: similarly, can omit the depth stuff if it's zero

wgpu/src/image/atlas.rs
```diff
@@ -78,15 +82,39 @@ impl Atlas {
 
         log::info!("Allocated atlas entry: {:?}", entry);
 
+        // It is a webgpu requirement that:
```
kvark, line 85: if you use `write_texture`, your data doesn't have to be aligned to this, fwiw

wgpu/src/quad.rs
```diff
@@ -20,50 +21,54 @@ impl Pipeline {
...
+                    },
+                    count: None,
                 }],
             });
 
-        let constants_buffer = device.create_buffer_with_data(
-            Uniforms::default().as_bytes(),
-            wgpu::BufferUsage::UNIFORM | wgpu::BufferUsage::COPY_DST,
-        );
+        let constants_buffer =
```
kvark, line 37: same here: if it's a buffer that's sometimes updated, it would be better to create it without contents and mark for the next update

wgpu/src/triangle.rs
```diff
@@ -137,6 +145,7 @@ impl Pipeline {
                     depth_bias: 0,
```
kvark, line 145: can omit this

wgpu/src/triangle.rs
```diff
@@ -261,65 +272,85 @@ impl Pipeline {
...
-                wgpu::BufferUsage::COPY_SRC,
-            );
+            let vertices = bytemuck::cast_slice(&mesh.buffers.vertices);
+            let indices = bytemuck::cast_slice(&mesh.buffers.indices);
 
-            let index_buffer = device.create_buffer_with_data(
-                mesh.buffers.indices.as_bytes(),
-                wgpu::BufferUsage::COPY_SRC,
-            );
+            if let Some(vertices_size) =
```
kvark, line 278: nit: combine two pattern matches into one?

wgpu/src/triangle.rs
```diff
@@ -261,65 +272,85 @@ impl Pipeline {
...
+            let (attachment, resolve_target, load) =
                 if let Some(blit) = &mut self.blit {
                     let (attachment, resolve_target) =
                         blit.targets(device, target_width, target_height);
 
-                    (attachment, Some(resolve_target), wgpu::LoadOp::Clear)
+                    (
+                        attachment,
+                        Some(resolve_target),
+                        wgpu::LoadOp::Clear(wgpu::Color {
```
kvark, line 347: `wgpu::Color::TRANSPARENT`

wgpu/src/triangle.rs
```diff
@@ -363,20 +387,17 @@ impl Pipeline {
...
-                    &self.vertex_buffer.raw,
-                    vertex_offset * std::mem::size_of::<Vertex2D>() as u64,
-                    0,
-                );
+                render_pass
+                    .set_vertex_buffer(0, self.vertex_buffer.raw.slice(..));
 
-                render_pass.draw_indexed(0..indices as u32, 0, 0..1);
+                render_pass.draw_indexed(
+                    index_offset as u32
```
kvark, line 396: perhaps it's better to pass down the offset in `index_buffer.raw.slice(index_offset * size_of::<u32>()..)`?

hecrj:
@kvark Hey, thanks for the review!

> Did you consider write_buffer/write_texture instead of the staging belt?

Yes, but we are kind of stuck with a `StagingBelt` and multiple render passes for now until we can lift some limitations in `wgpu_glyph` (see https://github.com/hecrj/wgpu_glyph/pull/46). Once those are dealt with, we will change the architecture to upload everything upfront and record all the draw calls in a single render pass.

While we could partially use `write_buffer`/`write_texture` right now, I think it's better to batch all the work once a full transition is possible. This is just the first step! I plan to work on this very soon.

examples/integration/src/scene.rs
hecrj: Done in b689778ed94b34706bf97e5994a721a8648386a5.

wgpu/src/triangle.rs
hecrj: Done in b689778ed94b34706bf97e5994a721a8648386a5.

wgpu/src/quad.rs
hecrj: Done in 3eb63762c72ccd91717d4feb98128aa12cc1b126.

wgpu/src/image.rs
hecrj: Fixed in 3eb63762c72ccd91717d4feb98128aa12cc1b126.

wgpu/src/image.rs
hecrj: Done in b689778ed94b34706bf97e5994a721a8648386a5.

wgpu/src/triangle.rs
hecrj: Done in 8d605be4e3e0b678c7563708ea941d6188f45678.

wgpu/src/triangle.rs
hecrj: I actually tried before resorting to this, but I forgot to multiply by `size_of` as usual and got a driver crash :sweat_smile: 

Changed in 7559e4fb30a282271717f700cb6e420cdf6123f0.

wgpu/src/triangle.rs
hecrj: Done in 07880c392c865d700cbaff7601c1fffd3354dd4f.

PolyMeilex:
Great work, I updated my project to wgpu 6.0 and new iced and did some initial tests and everything works as expected so far :ok_hand:

But I have to say that as a user that integrates iced into existing wgpu project `StagingBelt` is quite annoying, I have to keep track of belt that I don't even use and I don't think I need one in my engine (as it is a quite simple one) I have to add it into my abstractions solely for use in `wgpu_glyph` and `iced`.
tldr: I'm glad you're considering to move away from `StagingBelt` in future, and I just wanted to voice my opinion :smiley:

dhardy:
I agree with @PolyMeilex; still at least in KAS (https://github.com/kas-gui/kas/pull/121) the `StagingBelt` doesn't leak into any of the examples, so it is less of an impact.

hecrj:
> I'm glad you're considering to move away from StagingBelt in future, and I just wanted to voice my opinion

Once we can get rid of the `StagingBelt`, we should be able to also follow [the middleware pattern](https://github.com/gfx-rs/wgpu-rs/wiki/Encapsulating-Graphics-Work).

> still at least in KAS (kas-gui/kas#121) the StagingBelt doesn't leak into any of the examples, so it is less of an impact.

Glad to know! It doesn't leak into any of the `iced` examples either. `iced_wgpu` is the only subcrate directly leaking it, given that the user should be able to choose the chunk size.

myfreeweb:
grrr, another local futures executor, breaking my GLib based project again >_< patched to spawn on GLib for now. Hopefully the "move away from StagingBelt" would delete the executor.

PR #505, opened 2020-09-04
Support glow backend on wasm target

tedsta:
Hello!

This PR adds proof-of-concept support for the glow backend on web. I did this because I wanted the `Canvas` widget for the web in my application. I had to modify modify the version in the shaders to be `#version 300 es`, and they still work on the native target for me so maybe it's fine to just leave them that way? Otherwise I can put logic in to have it be `#version 330` for native and `#version 300 es` for web.

This allows you to build the solar system example if you use the following features:
`iced = { path = "../..", default-features = false, features = ["glow", "glow_canvas", "winit"] }`
I also had trouble with the `rand` crate on wasm so had to remove the random star position generation.

I had to do some questionable things to make things `!Send` for the `wasm32` target...

I have to do more tests so definitely not merge-able as-is. One thing I know is broken is text boxes on the web with the glow backend.

iced-solar-system

image

### DIFF
--- a/winit/src/clipboard.rs
+++ b/winit/src/clipboard.rs
@@ -1,19 +1,47 @@
-/// A buffer for short-term storage and transfer within and between
-/// applications.
-#[allow(missing_debug_implementations)]
-pub struct Clipboard(window_clipboard::Clipboard);
+pub use inner::*;
 
-impl Clipboard {
-    /// Creates a new [`Clipboard`] for the given window.
-    ///
-    /// [`Clipboard`]: struct.Clipboard.html
-    pub fn new(window: &winit::window::Window) -> Option<Clipboard> {
-        window_clipboard::Clipboard::new(window).map(Clipboard).ok()
+#[cfg(target_arch = "wasm32-unknown-unknown")]
+mod inner {
+    /// A buffer for short-term storage and transfer within and between
+    /// applications.
+    #[allow(missing_debug_implementations)]
+    pub struct Clipboard(window_clipboard::Clipboard);
+
+    impl Clipboard {
+        /// Creates a new [`Clipboard`] for the given window.
+        ///
+        /// [`Clipboard`]: struct.Clipboard.html
+        pub fn new(window: &winit::window::Window) -> Option<Clipboard> {
+            window_clipboard::Clipboard::new(window).map(Clipboard).ok()
+        }
+    }
+
+    impl iced_native::Clipboard for Clipboard {
+        fn content(&self) -> Option<String> {
+            self.0.read().ok()
+        }
     }
 }
 
-impl iced_native::Clipboard for Clipboard {
-    fn content(&self) -> Option<String> {
-        self.0.read().ok()
+#[cfg(not(target_arch = "wasm32-unknown-unknown"))]
+mod inner {
+    /// A buffer for short-term storage and transfer within and between
+    /// applications.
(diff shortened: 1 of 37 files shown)


tedsta:
Alright I think I've gotten this into a fairly good state. Summary of changes:

- Make `iced_web_winit` crate that is analogous to `iced_glutin` (in that it builds on top of `iced_winit` for a specific renderer). Eventually I imagine with some clever use of feature items these 3 crates could be consolidated if `GLCompositor` and `Compositor` can be consolidated. I'd be willing to take a stab at it if you'd like.
- Use `wasm_timer::Instant` in place of `std::time::Instant` everywhere for the `wasm32` target.
- Don't require runtime's `Message` `Stream`/`Sync` to be `Send` on `wasm32` target. Ultimately this was required because something didn't implement `Send` under `wasm32` target... for the life of me I can't remember... maybe it was `EventStream` or `iced_futures::subscription::Tracker`? I'll try to figure it out again so we know for sure these changes are required.
- Use `crate::BoxFuture` instead of getting it from `futures` crate in `iced_futures::time`. Related to the `Send` / `!Send` discrepancy between `wasm32` and non-wasm targets.
- Rename `distance` function to `quad_distance` in `iced_glow` quad fragment shader. It was failing to compile under WebGL, guessing there's a builtin function named `distance`.
- Add wasm / non-wasm specific instantiation of `glow` Context in `iced_glow` crate.
- Work around lack of `draw_elements_base_vertex` for `wasm32` target in `glow` crate.
- Disable `font-kit` dependency under `wasm32` target.

tedsta:
@hecrj Do you have any thoughts on this? I think using WebGL and eventually WebGPU to render Iced UIs in browsers is valuable to have native widgets work as-is in the browser. I understand if you don't think it's the way to go. This work has served its purpose for me, I have an app that uses it that I don't expect to change much. I am willing to put some more work in to get this merge-able if you want it though - I would very much like to create custom widgets for myself without writing both native and web versions, and I don't mind limiting myself to places where WebGL is supported. No rush though.

hecrj:
Thanks for working on this! I think this is really cool and very promising.

However, there are a many steps you took here that involve somewhat important changes. I'd be more comfortable if we could take on step at a time.

I believe we should start by making a [`glow_glyph`](https://github.com/hecrj/glow_glyph) release that officially works on Wasm/WebGL. This alone would already be great! Maybe you could open a PR there with your changes?

Once `glow_glyph` is ready, we can keep unravelling the necessary changes to make WebGL rendering happen.

winit/src/clipboard.rs
```diff
@@ -1,19 +1,47 @@
...
+        ///
+        /// [`Clipboard`]: struct.Clipboard.html
+        pub fn new(_window: &winit::window::Window) -> Option<Clipboard> {
+            Some(Clipboard)
+        }
+    }
+
+    impl iced_native::Clipboard for Clipboard {
+        fn content(&self) -> Option<String> {
+            Some("".to_string())
```
ccope, line 44: Were you trying to avoid interacting with the clipboard for wasm32 here? Looks like you accidentally disabled the clipboard on non-wasm32? (This is old and conflicts with master now, I'm trying to rebase)

hecrj:
Given that this has been inactive for almost a year since my last comment, I am closing this for the time being.

I really like the idea of leveraging `glow` and `wgpu` to target Wasm! We will circle back to this eventually.

PR #514, opened 2020-09-07
Error propagation

hecrj:
Introduces `Error` types in `iced`, `iced_winit`, and `iced_graphics` while propagating them all the way to the `run` method in both `Application` and `Sandbox`.

In other words, `Application::run` and `Sandbox::run` return a `Result` now, instead of panicking. Therefore, it is now possible to handle startup errors and properly display them to the user.


artursapek:
How Rust-like of you

Issue #491, opened 2020-08-23
iced_web: Add an option to select the element


MarcAntoine-Arnaud:
On running the main application, it can be awesome to be able to choose something else than the `body` [here](https://github.com/hecrj/iced/blob/master/web/src/lib.rs#L182).
Is it possible to add a parameter to the `run` with an `Option<&str>` to allow to select an element by it's ID where we can put the renderer.

It can be awesome to do that to allow integrated view inside other frameworks like React, AngularJS, Vue etc.

Marc-Antoine

Kaiden42:
I would not like to consider adding parameters to the `run` function as this would cause a breaking change for everyone currently using iced (web). 🤔

Instead I would prefer to add this to the application settings instead. So for those who already use `Settings::default()` or similar nothing would change. 🙂

But there is one thing I don't quite understand here. `iced_web` has no settings, only flags. In the documentation of iced's `Settings` it is mentioned that, for example, the [`window`](https://github.com/hecrj/iced/blob/master/src/settings.rs#L9) field is ignored on the web, but as mentioned before there are no settings for web. 🤨

Maybe someone who knows a bit more about this project can help me here. Is there a reason why iced_web has no `Settings`?

MarcAntoine-Arnaud:
Hi @Kaiden42,

Thank for sharing your point of view.
It sounds better to pass this king of parameter via [`Settings`](https://github.com/hecrj/iced/blob/master/src/settings.rs#L6) structure, I'm 100% agreed.

The comment on [`window`](https://github.com/hecrj/iced/blob/master/src/settings.rs#L9) is just there to said the [`window::Setting`](https://github.com/hecrj/iced/blob/master/src/window/settings.rs#L5) parameters are not used in the web case.

So to improve that I suggest to add a field in `window::Settings` - only for wasm32 target to add an `HtmlElement`.
So it can look like:
```rust
#[derive(Debug, Clone)]
pub struct Settings {
    /// The initial size of the window.
    pub size: (u32, u32),

    ....

    /// The icon of the window.
    pub icon: Option<Icon>,

    /// Optional HtmlElement to put the "window" into it.
    #[cfg(target_arch = "wasm32")]
    pub html_element: Option<HtmlElement>
}
```

Marc-Antoine

hecrj:
An `Application` is meant to take control of the full document, as it can control the `title`.

The best way to satisfy this use case would be to create a different trait without a `title` method and an additional argument for `run`.

[...]

PR #517, opened 2020-09-09
iced_web: Add an option to select the element

Kaiden42:
This PR implements the feature requested by #491.

I named the new trait `Embedded` since it is an embedded web application. Further name suggestions are welcome. :)

For now it is placed in the same file as the `Application` trait. If you wish, I can swap the new trait to another file.

It works similar to the Application trait but does not take control of the title of the document and the ID of a HTML element can be specified at `run` as desired.

Example:
```Rust
use iced_web::{executor, Embedded, Command, Element, Text};

pub fn main() {
    Hello::run((), Some(String::from("container")));
}

struct Hello;

impl Embedded for Hello {
    type Executor = executor::Null;
    type Message = ();
    type Flags = ();

    fn new(_flags: ()) -> (Hello, Command<Self::Message>) {
        (Hello, Command::none())
    }

    fn update(&mut self, _message: Self::Message) -> Command<Self::Message> {
        Command::none()
    }

    fn view(&mut self) -> Element<Self::Message> {
        Text::new("Hello, world!").into()
    }
}
```

### DIFF
--- a/web/src/lib.rs
+++ b/web/src/lib.rs
@@ -262,12 +262,12 @@ where
 }
 
 /// An interactive embedded web application.
-/// 
+///
 /// This trait is the main entrypoint of Iced. Once implemented, you can run
 /// your GUI application by simply calling [`run`](#method.run). It will either
 /// take control of the `<body>' or of an HTML element of the document specified
 /// by `container_id`.
-/// 
+///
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Looks great! Thank you 🙇

PR #527, opened 2020-09-15
Account for empty ranges in `Slider` and `ProgressBar`

### LINKED ISSUE #526: Perenthenses problem in progress bar draw formula
I think there is a perenthenses problem in this formula

https://github.com/hecrj/iced/blob/4f2962d73f3bdeeca8a11817e404c45e91e2c2cc/graphics/src/widget/progress_bar.rs#L39

May be : 
**((value - range_start) / (range_end - range_start))**.max(1.0)
Instead of :
((value - range_start) / **(range_end - range_start)**.max(1.0))

rubik83:
fixes #526

### DIFF
--- a/graphics/src/widget/progress_bar.rs
+++ b/graphics/src/widget/progress_bar.rs
@@ -34,7 +34,7 @@ where
 
         let (range_start, range_end) = range.into_inner();
         let active_progress_width = bounds.width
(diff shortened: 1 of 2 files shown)


hecrj (approved):
Thank you!

I ended up with simply adding a check for `range_start >= range_end` and shortcircuiting to `0.0` in that case.

PR #533, opened 2020-09-18
Fix: ignoring all `target` directories

Kaiden42:
The `.gitignore` file contained the line `/target` which ignores the `target` directory on the root but not in subdirectories like `core/target/`, `examples/custom_widget/target/`, ... . For me, this directories should also be ignored. This PR fixes this by changing the line to `target/`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-/target
+target/
 pkg/
 **/*.rs.bk
 Cargo.lock


hecrj:
Interesting. I thought Cargo shared the `target` directory for all of the crates in a workspace. How could I reproduce the issue?

Kaiden42:
For me, this is cause by the [Rust Language Server (RLS)](https://github.com/rust-lang/rls) wich is used by many IDEs/Editors (including the one I use). It is (currently) the official IDE project of Rust for providing code analysation, completion, etc. RLS is based on the "save-analysis" data from `rustc` wich are stored in the `target` directory of each crate (and subcrate). This results in having a lot of `target` directories when using RLS.

I hope this will change in the future once the [rust-analyzer](https://rust-lang.github.io/rfcs/2912-rust-analyzer.html) will become stable and will replace RLS.

hecrj (approved):
Thanks for the explanation.

Let's go ahead then!

PR #545, opened 2020-10-01
Remove outdated `Fill` comment for `Column` and `Row`

### LINKED ISSUE #544: Column.width = Shrink by default
https://docs.rs/iced_native/0.2.2/src/iced_native/widget/column.rs.html#17-26
> A Column will try to fill the horizontal space of its container.

However `Column::new` uses `Shrink` by default.
```rs
Column {
            spacing: 0,
            padding: 0,
            width: Length::Shrink,
...
```

MonliH:
This PR makes the default width `Length::Fill` for the column widget.
Fixes #544.

### DIFF
--- a/native/src/widget/column.rs
+++ b/native/src/widget/column.rs
@@ -42,7 +42,7 @@ impl<'a, Message, Renderer> Column<'a, Message, Renderer> {
         Column {
             spacing: 0,
             padding: 0,
-            width: Length::Shrink,
+            width: Length::Fill,
             height: Length::Shrink,
             max_width: u32::MAX,
             max_height: u32::MAX,
--- a/native/src/widget/column.rs
+++ b/native/src/widget/column.rs
@@ -10,9 +10,10 @@ use std::u32;
 
 /// A container that distributes its contents vertically.
 ///
-/// A [`Column`] will try to fill the horizontal space of its container.
+/// A [`Column`] will not fill the horizontal space of its container. (set to [`Length::shrink`] by defualt)
 ///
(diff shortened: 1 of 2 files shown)


hecrj:
Hey, thanks!

I changed the behavior from `Fill` to `Shrink` in #139, but I forgot to update the documentation 😅 We should fix that instead!

MonliH:
Oh, awesome! I'll document it then :smile:.

MonliH:
I've documented this behaviour in the `Column` and `Row` widget, please let me know if I missed any other widgets that are out of date.

native/src/widget/column.rs
```diff
@@ -10,9 +10,10 @@ use std::u32;
 
 /// A container that distributes its contents vertically.
 ///
-/// A [`Column`] will try to fill the horizontal space of its container.
+/// A [`Column`] will not fill the horizontal space of its container. (set to [`Length::shrink`] by defualt)
```
hecrj, line 13: Maybe we could simply remove the comment, as the `Shrink` behavior should be considered the default for all widgets.

MonliH: Yes, that's actually a good idea. Maybe we could document `Shrink` behaviour in the `widgets` module doc?

hecrj: Let's just remove the comments for now. I plan to review and improve all the documentation soon.

MonliH: Sure.

hecrj (approved):
Awesome! Thank you 🎉

PR #547, opened 2020-10-03
Strawman implementation of blinking cursor in text input + animation 'WaitUntil'

twitchyliquid64:
This is just a demo, I agree we need to wait until we have a better idea of the API before we tackle animations

Anyway, this is what it would look like to hack it in XD

A few lessons here:

1. Adding stuffing into `draw()` is a pain in the ass, theres over 80 instances to update XD
2. Adding stuff into `draw()` is probably a bad idea - the current event loop in `application.rs` seems to assume state changes only happen on `update()`, and `draw()` just, draws. The concept of 'next draw me at X moment/event' doesnt neatly fit into either.

### DIFF
--- a/examples/custom_widget/src/main.rs
+++ b/examples/custom_widget/src/main.rs
@@ -60,6 +60,7 @@ mod circle {
             _defaults: &Defaults,
             layout: Layout<'_>,
             _cursor_position: Point,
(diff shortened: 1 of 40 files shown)


twitchyliquid64:
Closing this as its obsoleted by #560

PR #550, opened 2020-10-04
Fixes #549: Implements preload() method on images to help users avoid jank

### LINKED ISSUE #549: Rendering janks when large images are added to the layout
EG: The `tour` example, when you click _Next_ to go to the page with the resizable rust logo.

As far as I can tell, this is because upload of the raster data is synchronous in calls to `iced_wgpu::image::Pipeline::draw()`, and I think it takes a while (and hence blocks the event loop).

A few ideas for improving this:

1. Use a real cache (with a better eviction policy) in the raster cache. At a guess, I would use 2-queue or LRU. This won't fix the root cause, but will greatly improve cache efficiency in situations where images are re-used, but are not in every layout tree (for example, you experience jank on the `tour` example when going beyond the image page and then back again).

2. Parallelize raster uploads across `draw()`. This would involve:

    a. Making `raster::Cache` Send+Sync or otherwise workable in an async context
    b. Track uploads-in-progress, and render a fallback primitive (or none at all) for images which are still being uploaded when they are needed (this behavior should probably be configurable).
    c. To power the parallelism, either creating a thread for the image pipeline / raster cache, or plumbing through async + the executor from `Application` or something like that?

twitchyliquid64:
* Implements a `preload()` method to make it easier for users to setup an `Image` they can share about, which has the BGRA data loaded.
 * Update the `tour` example to use this technique.

### DIFF
--- a/examples/tour/Cargo.toml
+++ b/examples/tour/Cargo.toml
@@ -8,3 +8,4 @@ publish = false
 [dependencies]
 iced = { path = "../..", features = ["image", "debug"] }
 env_logger = "0.7"
+lazy_static = "1.4.0"
--- a/examples/tour/src/main.rs
+++ b/examples/tour/src/main.rs
@@ -4,8 +4,16 @@ use iced::{
     Sandbox, Scrollable, Settings, Slider, Space, Text, TextInput,
 };
 
+#[macro_use]
+extern crate lazy_static;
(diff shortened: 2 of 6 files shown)


twitchyliquid64:
Closing this as discussion should preclude code

PR #558, opened 2020-10-08
Adds From<Point> and From<Size> for [f32; 2]

Azorlogh:
Example use case:
An `Axis { X, Y }` enum with useful generic methods such as:
`with_major<T: Into<[f32; 2]> + From<[f32; 2]>>(&self, coords: T, value: f32) -> T`
(replace the corresponding coordinate in coords with value)


hecrj (approved):
Looks great! Thank you :tada:

PR #568, opened 2020-10-17
Require `Clone` for `Message` early when needed

hecrj:
Prior to this change, the widgets that needed a `Clone` bound on `Message` to implement the `Widget` trait could be created with a non-cloneable `Message`.

As a consequence, the compiler complained only when actually trying to use the `Widget` trait. Normally, this happens when trying to `push` the widget in a container or turn it into an `Element`.

Furthermore, the compiler error in this case does not mention `Message` nor the `Clone` bound, but instead complains about a missing `From` implementation. Thus, it can easily cause confusion!

This change introduces `Clone` bounds in the main implementation of the widgets that need it to properly implement the `Widget` trait. As a result, the compiler complains early when trying to create one of these widgets with a non-cloneable `Message` and explicitly mentions that the `Message` needs to implement `Clone`.

# Example

Given this code:

```rust
use iced::{button, Button, Container, Element, Text};

enum Message {
    ButtonPressed,
}

fn view(button: &mut button::State) -> Element<Message> {
    let button = Button::new(button, Text::new("Press me!"))
        .on_press(Message::ButtonPressed);

    Container::new(button).into()
}
```

Before these changes, the compiler error was:

```
error[E0277]: the trait bound `Message: std::convert::From<iced_native::widget::button::Button<'_, Message, iced_graphics::renderer::Renderer<iced_wgpu::backend::Backend>>>` is not satisfied
  --> examples/clone_bound/src/main.rs:12:20
   |
12 |     Container::new(button).into()
   |                    ^^^^^^
   |                    |
   |                    expected an implementor of trait `std::convert::From<iced_native::widget::button::Button<'_, Message, iced_graphics::renderer::Renderer<iced_wgpu::backend::Backend>>>`
   |                    help: consider borrowing here: `&button`
   |
   = note: required because of the requirements on the impl of `std::convert::From<iced_native::widget::button::Button<'_, Message, iced_graphics::renderer::Renderer<iced_wgpu::backend::Backend>>>` for `iced_native::element::Element<'_, Message, iced_graphics::renderer::Renderer<iced_wgpu::backend::Backend>>`
   = note: required because of the requirements on the impl of `std::convert::Into<iced_native::element::Element<'_, Message, iced_graphics::renderer::Renderer<iced_wgpu::backend::Backend>>>` for `iced_native::widget::button::Button<'_, Message, iced_graphics::renderer::Renderer<iced_wgpu::backend::Backend>>`
   = note: required by `iced_native::widget::container::Container::<'a, Message, Renderer>::new`
```

After these changes, the compiler now complains earlier:

```
error[E0277]: the trait bound `Message: std::clone::Clone` is not satisfied
 --> examples/clone_bound/src/main.rs:9:18
  |
9 |     let button = Button::new(button, Text::new("Press me!"))
  |                  ^^^^^^^^^^^ the trait `std::clone::Clone` is not implemented for `Message`
  |
  = note: required by `iced_native::widget::button::Button::<'a, Message, Renderer>::new`
```

### DIFF
--- a/examples/color_palette/src/main.rs
+++ b/examples/color_palette/src/main.rs
@@ -284,7 +284,7 @@ impl<C: 'static + ColorSpace + Copy> ColorPicker<C> {
         let [s1, s2, s3] = &mut self.sliders;
         let [cr1, cr2, cr3] = C::COMPONENT_RANGES;
 
-        fn slider<C>(
+        fn slider<C: Clone>(
             state: &mut slider::State,
             range: RangeInclusive<f64>,
             component: f32,
--- a/examples/tour/src/main.rs
+++ b/examples/tour/src/main.rs
@@ -689,7 +689,7 @@ fn ferris<'a>(width: u16) -> Container<'a, StepMessage> {
     .center_x()
 }
 
-fn button<'a, Message>(
+fn button<'a, Message: Clone>(
     state: &'a mut button::State,
     label: &str,
 ) -> Button<'a, Message> {
--- a/native/src/widget/button.rs
+++ b/native/src/widget/button.rs
@@ -18,6 +18,7 @@ use std::hash::Hash;
 /// # type Button<'a, Message> =
 /// #     iced_native::Button<'a, Message, iced_native::renderer::Null>;
 /// #
+/// #[derive(Clone)]
 /// enum Message {
 ///     ButtonPressed,
 /// }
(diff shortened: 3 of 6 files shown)


clarkmoody:
I've been bitten by this obscure compiler error a time or two! Thanks for the update.

PR #583, opened 2020-10-28
Add conversion functions to Size and Vector

Limeth:
This PR adds conversion functions to `Size` and `Vector` for convenience.

### DIFF
--- a/core/src/size.rs
+++ b/core/src/size.rs
@@ -1,3 +1,4 @@
+use super::vector::Vector;
 use std::f32;
 
 /// An amount of space in 2 dimensions.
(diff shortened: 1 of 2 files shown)


core/src/vector.rs
```diff
@@ -65,3 +67,24 @@ where
...
+    fn from(other: Vector<T>) -> Self {
+        [other.x, other.y]
+    }
+}
+
+impl From<Size> for Vector<f32> {
+    fn from(size: Size) -> Self {
+        Vector::new(size.width, size.height)
+    }
+}
```
hecrj, line 90: I think we should move this implementation to the `size` module to avoid a dependency cycle.

core/src/size.rs
```diff
@@ -1,3 +1,4 @@
+use super::vector::Vector;
```
hecrj, line 1: For consistency with the rest of the codebase, we should use `crate::Vector`.

hecrj (changes requested):
Thank you!

I think we need to address just a couple of details before merging.

Limeth:
Thanks for the review, I applied the requested changes.

hecrj (approved):
Awesome! Thank you :tada:

PR #590, opened 2020-10-31
Fix lifetimes in `Layout::children`

Limeth:
This change allows for chaining of `children` calls without having to bind the iterator to a local variable, and consequently also allows for the result to be returned from the current scope.

```rust
    pub fn get_content_layout_from_child_layout(child_layout: Layout<'_>) -> Layout<'_> {
        child_layout.children().nth(0).unwrap().children().nth(1).unwrap()
    }
```

### DIFF
--- a/native/src/layout.rs
+++ b/native/src/layout.rs
@@ -64,7 +64,7 @@ impl<'a> Layout<'a> {
     ///
     /// [`Layout`]: struct.Layout.html
     /// [`Node`]: struct.Node.html
-    pub fn children(&'a self) -> impl Iterator<Item = Layout<'a>> {
+    pub fn children(&self) -> impl Iterator<Item = Layout<'a>> + '_ {
         self.node.children().iter().map(move |node| {
             Layout::with_offset(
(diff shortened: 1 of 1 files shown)


native/src/layout.rs
```diff
@@ -64,7 +64,7 @@ impl<'a> Layout<'a> {
     ///
     /// [`Layout`]: struct.Layout.html
     /// [`Node`]: struct.Node.html
-    pub fn children(&'a self) -> impl Iterator<Item = Layout<'a>> {
+    pub fn children(&self) -> impl Iterator<Item = Layout<'a>> + '_ {
```
hecrj, line 67: It's probably better to simply take `self` by value here and drop the anonymous iterator lifetime, given that `Layout` implements `Copy`.

hecrj (changes requested):

Limeth:
Thanks for the review, I applied the suggestions. Should be good to go now. :+1:

hecrj (approved):
Awesome. Thank you!

PR #592, opened 2020-11-01
Fixes #591 DimensionsMismatch error message

### LINKED ISSUE #591: Wrong error message for DimensionsMismatch error
The message given when creating an `Icon` using `from_rgba` is:

> The number of RGBA pixels (3) does not match the provided dimensions (7x100)

When it should read

> The number of RGBA pixels (100) does not match the provided dimensions (3x7)

sum-elier:
The values passed to write! were shifted 1 to the left.

Fixes #591


hecrj (approved):
Good catch! Thanks 👍

PR #595, opened 2020-11-05
upgrade tokio to latest version(v0.3)

valbendan:


src/executor.rs
```diff
@@ -40,7 +40,8 @@ mod platform {
         }
 
         fn enter<R>(&self, f: impl FnOnce() -> R) -> R {
-            self.0.enter(f)
+            let _guard = self.0.enter();
+            f()
```
hecrj, line 44: I think this will not work with other executor implementations. Instead, it should probably be:

```suggestion
            Executor::enter(&self.0, f)
```

hecrj (changes requested):

hecrj:
Before merging, I think it may be better to wait a bit until most of the async ecosystem upgrades to the latest `tokio` release.

A bunch of examples are using `reqwest` and causes duplicated dependencies, as it has not been upgraded yet: https://github.com/seanmonstar/reqwest/pull/1076

clarkmoody:
Tested locally, and this works for the Game of Life and Solar System example programs.

hecrj:
As I mentioned, `reqwest` has not yet migrated, so examples like `pokedex` and `download_progress` do not work.

We could use the compatibility layer, but we can just wait.

hecrj (approved):
I have introduced a new `tokio_old` feature flag to allow users to enable the old `tokio` and transition once the ecosystem is ready.

I think we can merge this. Let me know what you think!

valbendan:
I think we can merge it

hecrj:
I have rolled back the latest changes to `.gitignore`.

We should try to keep the repository decoupeld from specific dev environments. For that purpose, you should change the `.git/info/exclude` instead.

valbendan:
It seems I forget checkout a new branch, sorry for the inconvenient

PR #598, opened 2020-11-06
Add textinput focus method

oknozor:
This is just a small addition : 
- Add a`focus` method` to control `TextInput` focus state. 

A simple use case  would be to force the focus on every call to the `Application::update`, but this could be used 
to move the focus around whenever a specific event occurs.

### DIFF
--- a/native/src/widget/text_input.rs
+++ b/native/src/widget/text_input.rs
@@ -693,6 +693,20 @@ impl State {
         self.cursor
     }
 
+    /// Focuses the [`TextInput`].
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks!

I have split the `focus` method into `focus` and `unfocus`, instead of asking for a boolean.

I also removed the `.gitignore` changes. You should instead update the `.git/info/exclude` file in your local copy of the repository to ignore any files related to your particular environment.

PR #599, opened 2020-11-06
Add wayland app_id

oknozor:
This add  wayland `app_id` for linux targets

**Links to the spec** : 

-  https://github.com/wayland-project/wayland-protocols/blob/4ed0cafeefd9f81b974465ca495cbd9118508cdb/stable/xdg-shell/xdg-shell.xml#L640
- https://wayland-book.com/xdg-shell-basics/xdg-toplevel.html

**Why ?** : 

It allow to send specific commands to the app via the wayland compositor. 

For example this will tell [sway](https://github.com/swaywm/sway) to always start the app in floating mode : 
```
for_window [app_id="my-iced-app"] floating enable
```

**Note** : 

I have set the default value for wayland `app_id` to `Application::title` in `iced_winit::settings::Window` but it might be interesting to expose a public setter for this to build apps that conforms to freedesktop's [desktop-file-id specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id)

**Suggestion on X11** : 

The following X11 window properties can also be set via winit : `visual_infos`, `screen_id`, `resize_increments`, `base_size`, `class`, `override_redirect`, `gtk_theme_variant`. I am not sure if they are relevant for iced apps but from my experience, exposing at least  `class`  (or setting a default value) would be useful for i3 user (and maybe for other X11 wm) .

### DIFF
--- a/winit/src/settings.rs
+++ b/winit/src/settings.rs
@@ -16,6 +16,12 @@ use winit::window::WindowBuilder;
 /// The settings of an application.
 #[derive(Debug, Clone, Default)]
 pub struct Settings<Flags> {
+    /// The identifier of the application.
+    ///
+    /// If provided, this identifier may be used to identify the application or
+    /// communicate with it through the windowing system.
+    pub id: Option<String>,
+
     /// The [`Window`] settings
     pub window: Window,
 
@@ -70,6 +76,7 @@ impl Window {
         title: &str,
         mode: Mode,
         primary_monitor: Option<MonitorHandle>,
+        _id: Option<String>,
     ) -> WindowBuilder {
         let mut window_builder = WindowBuilder::new();
 
@@ -112,7 +119,10 @@ impl Window {
         ))]
         {
             use ::winit::platform::unix::WindowBuilderExtUnix;
-            window_builder = window_builder.with_app_id(title.to_string());
+
+            if let Some(id) = _id {
+                window_builder = window_builder.with_app_id(id);
+            }
         }
 
         #[cfg(target_os = "windows")]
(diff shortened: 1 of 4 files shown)


winit/src/settings.rs
```diff
@@ -90,6 +90,12 @@ impl Window {
                 .with_max_inner_size(winit::dpi::LogicalSize { width, height });
         }
 
+        #[cfg(target_os = "linux")]
```
myfreeweb, line 93: Please match the cfg winit uses for `winit::platform::unix`:

```rust

#![cfg(any(
    target_os = "linux",
    target_os = "dragonfly",
    target_os = "freebsd",
    target_os = "netbsd",
    target_os = "openbsd"
))]
```

oknozor: just did it and rebased master

twitchyliquid64 (approved):

NooneXe:
So.. whats stopping this from getting merged ?

hecrj (approved):
Relying on the `title` of the `Application` for the application identifier didn't feel right, so I have introduced an explicit `id` field to the `Settings` struct.

Thank you 🙇 Let's merge this! 🎉

unrelentingtech:
Might be nice to document in the docstring where this is used ("On Wayland, this is the xdg-shell `app_id`.") and the expected conventions ("For Unix-like/Freedesktop platforms, this is usually expected to be a reverse domain name identifier like com.example.MyApp, matching the `.desktop` file ID and D-Bus name. See https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html for the specification.")

hecrj:
@unrelentingtech Agreed. PRs welcome!

PR #605, opened 2020-11-09
This PR fixes a bug with select all (CMD + A on MacOS) when using a text_input.

ZakisM:
This PR fixes a bug with select all (CMD + A on MacOS) when using a text_input.

Previous behaviour: when selecting all (CMD + A) would delete the current text inside the input and replace the content with just the letter 'a'.

Now we check if the logo key (modifier key) has been pressed before checking any other key and save it to the state level. This way we can prevent any text being deleted when using the select all shortcut or text being entered at all when a modifier key is pressed (this behaviour matches other text input behaviour i.e text inputs in the browser etc...).

### DIFF
--- a/native/src/widget/text_input.rs
+++ b/native/src/widget/text_input.rs
@@ -382,6 +382,7 @@ where
             Event::Keyboard(keyboard::Event::CharacterReceived(c))
                 if self.state.is_focused
                     && self.state.is_pasting.is_none()
+                    && !self.state.is_logo_pressed
                     && !c.is_control() =>
             {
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks!

I have rebased and replaced the `is_logo_pressed` with a `keyboard::ModifiersState` instead. This should be more future-proof, as the `modifiers` field in the key events will eventually be removed.

Let me know what you think, and we can merge this!

ZakisM:
Yep looks good. Happy for it to be merged, thanks.

PR #607, opened 2020-11-09
Add methods to control `Scrollable` programmatically

### LINKED ISSUE #307: Scrolling programmatically

yusdacra:
This PR adds methods:
- `snap_to_bottom` and `on_scroll` to `Scrollable`
  - `snap_to_bottom` takes a bool and if true makes the `Scrollable` snap to bottom when the user scrolls to bottom. If the contents of the `Scrollable` change, it will automatically snap to bottom. Whenever the user scrolls up it will stop snapping to bottom. By default a `Scrollable` won't snap to bottom.
  - `on_scroll` takes a function whose arguments are two `f32`s and produces a `Message`. First argument is to pass the current `offset` of the `Scrollable` to provide the user with "where is this `Scrollable` at right now" and second argument is to pass the *previous* `offset` of the `Scrollable` to provide the user with "where was this `Scrollable` before this scroll event".
- `scroll_to_percentage` and `scroll_to_bottom` to `scrollable::State`
  - `scroll_to_percentage` takes a value between `0.0` and `1.0`. It marks the `Scrollable` to scroll to the specified percentage in the next `draw` call.
  - `scroll_to_bottom` is just a shorthand for `scroll_to_percentage(1.0)`.

This should fix #307 

I wonder what people think about this so please post your thoughts!

### DIFF
--- a/native/src/widget/scrollable.rs
+++ b/native/src/widget/scrollable.rs
@@ -23,6 +23,7 @@ pub struct Scrollable<'a, Message, Renderer: self::Renderer> {
     scrollbar_margin: u16,
     scroller_width: u16,
     content: Column<'a, Message, Renderer>,
+    on_scroll: Option<Box<dyn Fn(f32) -> Message>>,
     style: Renderer::Style,
 }
 
@@ -37,6 +38,7 @@ impl<'a, Message, Renderer: self::Renderer> Scrollable<'a, Message, Renderer> {
             scrollbar_margin: 0,
             scroller_width: 10,
             content: Column::new(),
+            on_scroll: None,
             style: Renderer::Style::default(),
         }
     }
@@ -101,9 +103,19 @@ impl<'a, Message, Renderer: self::Renderer> Scrollable<'a, Message, Renderer> {
     }
 
     /// Sets the scroller width of the [`Scrollable`] .
-    /// Silently enforces a minimum value of 1.
+    ///
+    /// It silently enforces a minimum value of 1.
     pub fn scroller_width(mut self, scroller_width: u16) -> Self {
         self.scroller_width = scroller_width.max(1);
+        self
+    }
+
+    /// Sets a function to call when the [`Scrollable`] is scrolled.
+    ///
+    /// The function takes the new relative offset of the [`Scrollable`]
+    /// (e.g. `0` means top, while `1` means bottom).
+    pub fn on_scroll(mut self, f: impl Fn(f32) -> Message + 'static) -> Self {
+        self.on_scroll = Some(Box::new(f));
         self
     }
 
@@ -120,6 +132,24 @@ impl<'a, Message, Renderer: self::Renderer> Scrollable<'a, Message, Renderer> {
     {
         self.content = self.content.push(child);
         self
+    }
+
+    fn notify_on_scroll(
+        &self,
+        bounds: Rectangle,
+        content_bounds: Rectangle,
+        messages: &mut Vec<Message>,
+    ) {
+        if content_bounds.height <= bounds.height {
+            return;
+        }
+
+        if let Some(on_scroll) = &self.on_scroll {
+            messages.push(on_scroll(
+                self.state.offset.absolute(bounds, content_bounds)
+                    / (content_bounds.height - bounds.height),
+            ));
+        }
     }
 }
 
@@ -228,6 +258,8 @@ where
                         }
                     }
 
+                    self.notify_on_scroll(bounds, content_bounds, messages);
+
                     return event::Status::Captured;
                 }
                 Event::Touch(event) => {
@@ -251,6 +283,12 @@ where
 
                                 self.state.scroll_box_touched_at =
                                     Some(cursor_position);
+
+                                self.notify_on_scroll(
+                                    bounds,
+                                    content_bounds,
+                                    messages,
+                                );
                             }
                         }
                         touch::Event::FingerLifted { .. }
@@ -290,6 +328,8 @@ where
                             content_bounds,
                         );
 
+                        self.notify_on_scroll(bounds, content_bounds, messages);
+
                         return event::Status::Captured;
                     }
                 }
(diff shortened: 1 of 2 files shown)


native/src/widget/scrollable.rs
```diff
@@ -35,9 +37,36 @@ impl<'a, Message, Renderer: self::Renderer> Scrollable<'a, Message, Renderer> {
...
 
+    /// Set the [`Scrollable`] to snap to bottom when the user
+    /// scrolls to bottom. This will keep the scrollable at the bottom
+    /// even if new content is added to the scrollable.
+    ///
+    /// [`Scrollable`]: struct.Scrollable.html
+    pub fn snap_to_bottom(mut self) -> Self {
+        self.snap_to_bottom = true;
+        self
+    }
```
clarkmoody, line 53: I'm envisioning a chat app with a checkbox for "Scroll Automatically." With this pattern, the implementation needs to do a conditional builder call:

```rust
let mut scrollable = Scrollable::new();
if settings.snap {
    scrollable = scrollable.snap_to_bottom();
}
```

It might be nicer to simply pass a bool:

```rust
let scrollable = Scrollable::new()
    .snap_to_bottom(settings.snap);
```

yusdacra: Yeah, that would indeed look nicer. And I suppose `snap_to_bottom(true / false)` shows the intent better than `snap_to_bottom()`. I'll change that.

native/src/widget/scrollable.rs
```diff
@@ -18,6 +18,8 @@ pub struct Scrollable<'a, Message, Renderer: self::Renderer> {
     scroller_width: u16,
     content: Column<'a, Message, Renderer>,
     style: Renderer::Style,
+    on_scroll: Option<Box<dyn Fn(f32, f32) -> Message>>,
```
twitchyliquid64, line 21: It might be better to have `Fn(f32, f32) -> Message` to be a generic param rather than a box/dynamic dispatch?

I guess this is up to @hecrj, but ive only seen `Box<dyn xxx>` used in `Element`, and generics used everywhere else.

native/src/widget/scrollable.rs
```diff
@@ -389,10 +460,13 @@ where
...
 ///
 /// [`Scrollable`]: struct.Scrollable.html
-#[derive(Debug, Clone, Copy, Default)]
+#[derive(Debug, Default, Clone)]
 pub struct State {
     scroller_grabbed_at: Option<f32>,
-    offset: f32,
+    prev_offset: u32,
+    snap_to_bottom: bool,
+    offset: RefCell<f32>,
```
twitchyliquid64, line 468: Oooof, I think interior mutability is frowned upon in the Elm/iced ethos. Would it be plausible to refactor this into messages instead?

(Again @hecrj is the expert here)

yusdacra: @hecrj mentioned using interior mutability in https://github.com/hecrj/iced/issues/307#issuecomment-626245954 and I myself didn't see a better way, so that's why I chose to do that.

twitchyliquid64: Ahhh yepp, nvm! ^^'

native/src/widget/scrollable.rs
yusdacra: Yeah I'm not sure about that, just did the first thing that came into my mind. Will wait for @hecrj's review then.

agraven:
`snap_to_bottom` is a feature I need to make the chat client I'm building work right, it being merged would be strong cause for celebration for me.

yusdacra:
> `snap_to_bottom` is a feature I need to make the chat client I'm building work right, it being merged would be strong cause for celebration for me.

Haha, same. It's why I implemented this feature in the first place :D

twitchyliquid64:
I've also found myself needing this, as I am implementing a search results list and need to scroll so that the result focused after a `<tab>` is visible. Plz! :crossed_fingers: :crossed_fingers: :crossed_fingers:

twitchyliquid64:
In the meantime here's a disgusting hack to `scroll_to_bottom` on iced 0.2/0.3, while we wait for this to land.

```rust
self.scroll = unsafe {
    let mut tmp = std::mem::transmute::<_, (Option<f32>, f32)>(self.scroll);
    tmp.1 = 999999.0;
    std::mem::transmute::<_, scrollable::State>(tmp)
};
```

(Assuming `self.scroll` is your `scrollable::State`)

Caveat: Dirty, disgusting, unsafe, brittle, use at your own risk.

pleshevskiy:
О! That's just what I need! Thank you!

hecrj:
I made some changes that I believe simplify everything overall.

Specifically:

- We avoid the `RefCell` by introducing a new `Offset` enum and deferring the absolute computation to `on_event` instead of `draw`.
- Instead of a `snap_to_bottom`, we introduce a `snap_to` method that allows snapping to any given relative position.
- The `scroll_to` API remains unchanged.

I have also updated the `scrollable` example to showcase the new API:

image

Let me know what you think!

yusdacra:
> I made some changes that I believe simplify everything overall.
> 
> Specifically:
> 
> * We avoid the `RefCell` by introducing a new `Offset` enum and deferring the absolute computation until the next scrolling event.
> * Instead of a `snap_to_bottom`, we introduce a `snap_to` method that allows snapping to any given relative position.
> * The `scroll_to` API remains unchanged.

Looks nice! Deferring the computation is nice, I like that. Can it be merged soon then?

hecrj:
@yusdacra Yes! Let's merge and fix any potential issues and implement new use cases as they come up.

hecrj:
Thank you everyone! 🎉

PR #608, opened 2020-11-10
Improve flexibility of `PaneGrid`

hecrj:
The current `PaneGrid` widget implicitly manages and exposes pane focus state without producing application messages.

This is generally forbidden. Meaningful application state changes that can trigger differences in the widget tree should always happen as a result of message processing and its respective call to `Application::update`.

After merging #597, this design flaw has become very apparent and [the `pane_grid` example](https://github.com/hecrj/iced/tree/master/examples/pane_grid) does not work as expected anymore.

This PR fixes the design of the `PaneGrid` widget while, at the same time, simplifying its API and making it more flexible overall. Specifically:

- The built-in focus handling is completely gone. A `PaneGrid` does not know about any kind of pane focus.
- A new `on_click` event handler has been added to `PaneGrid`. This handler can be used to reproduce and extend the old focus functionality on top of application code.
- The `on_key_press` handler has been removed. Instead, it is recommended to use `subscription::events` or the brand new `subscription::events_with` in order to listen to keyboard events.

As usual, [the `pane_grid` example](https://github.com/hecrj/iced/tree/master/examples/pane_grid) has been updated to showcase how to use this new API. It should reproduce the old behavior (prior to merging #597).

### DIFF
--- a/examples/pane_grid/src/main.rs
+++ b/examples/pane_grid/src/main.rs
@@ -123,7 +123,7 @@ impl Application for Example {
             Event::Keyboard(keyboard::Event::KeyPressed {
                 modifiers,
                 key_code,
-            }) if modifiers.control => handle_hotkey(key_code),
+            }) if modifiers.is_command_pressed() => handle_hotkey(key_code),
             _ => None,
         })
     }
--- a/examples/pane_grid/src/main.rs
+++ b/examples/pane_grid/src/main.rs
@@ -19,6 +19,7 @@ enum Message {
     Split(pane_grid::Axis, pane_grid::Pane),
     SplitFocused(pane_grid::Axis),
     FocusAdjacent(pane_grid::Direction),
+    Clicked(pane_grid::Pane),
     Dragged(pane_grid::DragEvent),
     Resized(pane_grid::ResizeEvent),
     Close(pane_grid::Pane),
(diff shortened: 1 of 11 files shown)


examples/pane_grid/src/main.rs
```diff
@@ -82,37 +102,54 @@ impl Sandbox for Example {
...
+        Command::none()
+    }
+
+    fn subscription(&self) -> Subscription<Message> {
+        subscription::events_with(|event| match event {
+            Event::Keyboard(keyboard::Event::KeyPressed {
+                modifiers,
+                key_code,
+            }) if modifiers.control => handle_hotkey(key_code),
+            _ => None,
```
clarkmoody, line 127: May I propose a platform-specific modifier key?

```rust
Event::Keyboard(keyboard::Event::KeyPressed {
    modifiers,
    key_code,
}) => {
    #[cfg(target_os = "macos")]
    let modifier = modifiers.logo;

    #[cfg(not(target_os = "macos"))]
    let modifier = modifiers.control;

    if modifier {
        handle_hotkey(key_code)
    } else {
        None
    }
}
_ => None,
```

hecrj: Done in 86fa12229e7117306b8297a09aa22c99802600c8. I have introduced an additional method to `ModifiersState` to help us avoid hardcoding this kind of logic.

clarkmoody: Nice! Love taking care of that at the source rather than at the edge 👌

PR #617, opened 2020-11-16
Add integration_gl example

sztomi:
This PR adds an integration example for the glow backend. Being new to iced and not being familiar with wgpu, I had a hard time dissecting the wgpu integration example. Since using the library like this removes a good chunk of the initialization that iced performs otherwise, I think it's a good idea to add this. I tried to follow the structure of the other integration example as close as possible.

Peek 2020-11-16 17-55


hecrj (approved):
Thank you! Looks great 🎉

Sorry for the delay 🙇

sztomi:
Hey, thanks! Glad to see this being picked up. Anything else I can do? Though I see you updated it already.

hecrj:
@sztomi All good! I just forgot to press the button 😅

PR #620, opened 2020-11-19
Disable dragging in `TextInput` after double click

### LINKED ISSUE #616: Text_input double-click selection issue
When I double or triple click on text in `text_input` it first selects a whole word (or whole line for triple click) then reselects from the beginning up to the current cursor position. I can see the correct selection quickly flash for a frame or so.

The bug is present in all input boxes I've tried, like `examples/todos`.
Tested with both mouse and trackpad.
Platform: macos / wgpu
Branch: master

Video of the issue:
[TextInput Bug.mov.zip](https://github.com/hecrj/iced/files/5545686/TextInput.Bug.mov.zip)

hecrj:
Fixes #616.

When using a trackpad, mouse move events may happen between the press/release events. This was incorrectly triggering selection dragging in the `TextInput` widget.

Eventually, we should implement proper word-aware selection dragging.

### DIFF
--- a/native/src/widget/text_input.rs
+++ b/native/src/widget/text_input.rs
@@ -278,7 +278,6 @@ where
             Event::Mouse(mouse::Event::ButtonPressed(mouse::Button::Left)) => {
                 let is_clicked = layout.bounds().contains(cursor_position);
 
-                self.state.is_dragging = is_clicked;
                 self.state.is_focused = is_clicked;
 
                 if is_clicked {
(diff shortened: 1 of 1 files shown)


dhardy:
I already did; feel free to steal code from [here](https://github.com/kas-gui/kas/blob/master/src/widget/editbox.rs#L901) and [here](https://github.com/kas-gui/kas/blob/master/src/text/selection.rs#L95).

hecrj:
@dhardy Thanks, but code is not the issue in this case.

adamwern:
This commit fixed the bug for me. Thanks!

Note: A regular mouse also triggered the bug here, so it was not limited to trackpads.

PR #610, opened 2020-11-10
Update dependencies

hecrj:
This PR updates a bunch of outdated dependencies in the workspace.

After this patch, there are 3 outdated dependencies left:

- `tokio`, which should be up-to-date after #595 is merged.
- `font-kit`, which has only been updated to `0.8`. Updating to `0.10` is blocked by https://github.com/manuel-rhdt/harfbuzz_rs/issues/27.
- `resvg`, which now depends on Skia (see #573). We should wait until [`tiny-skia`](https://github.com/RazrFalcon/tiny-skia) matures!

PR #623, opened 2020-11-20
change graphics fontkit to v0.10.0

e0gs:
Builds fine and helpful for version miss-match with Servo


hecrj:
As you can see in CI, updating is not possible because it causes a linking conflict when using the `svg` feature.

I mentioned this in #610. Hopefully, a new version of `resvg` is released soon on top of the latest `tiny-skia`.

Issue #519, opened 2020-09-12
Horizontal + vertical padding support


raphtlw:
This library is awesome, will there ever be a horizontal/vertical padding support instead of just the normal padding?

hecrj:
Thanks!

Yes, but I'd like to gather a bit more details about the most common use cases for non-uniform padding. Maybe, instead of introducing additional methods in existing widgets, we can come up with better abstractions to tackle them directly!

raphtlw:
I'd like to use the horizontal padding on buttons, cuz they look a little weird with the default 2 axis padding applied, just my input on this :)

PR #630, opened 2020-11-23
Add support for asymmetrical padding

blefevre:
This adds support for asymmetrical padding allowing each of the four sides to have a different padding value if desired.

The `padding` function now accepts a type of `Into<Padding>` which is implement by `u16`, `[u16;2]` and `[u16;4]`.
This allows paddings to specified in the following ways:
```rust
// Non breaking, keeps current behavior of 20px on all sides
Row::new().padding(20);

// 10px on the top and bottom, 20px on the left and right, 
Row::new().padding([10, 20]);

// 5px on the top, 10px on the right, 15px on the bottom, 20px on the left
Row::new().padding([5, 10, 15, 20]);
```

Closes: #519


hecrj (approved):
This is great! Thank you 🎉 

I made some small changes here and there and... I believe we can finally ship this! 🚢

PR #635, opened 2020-11-25
Create Tab Widget, and tweak Rule styling

BillyDM:
I've created a `Tab` widget. It functions like the `Radio` widget, but can hold content like the `Button` widget. I also added a modern "indicator line" to the style.

Screenshot from 2020-11-27 10-28-53

I also tweaked the styling of the `Rule` widget. It makes more sense to have the default style fill the whole length. I also tweaked the contrast a little.


BillyDM:
It would also be nice if bold/font family was added to defaults::Text. There is currently no simple way to make the active tab have bold text while keeping the rest have normal text.

PR #646, opened 2020-12-01
Disabled button docs and consistent behavior

mtsr:
This PR:
* Adds doc comment for disabled button
* Makes disabled button behavior consistent in iced_web

I also added a new style to the tour, because disabled buttons were indistinguishable from normal buttons and changed the Secondary style color, to make plain gray available for Disabled.

### DIFF
--- a/examples/tour/src/main.rs
+++ b/examples/tour/src/main.rs
@@ -70,15 +70,13 @@ impl Sandbox for Tour {
 
         controls = controls.push(Space::with_width(Length::Fill));
 
-        let mut button = button(next_button, "Next");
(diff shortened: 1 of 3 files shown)


mtsr:
I did not foresee having to add that type annotation. On the other hand it will probably not be needed, because the `on_press` is only conditionally not set so the type can still be inferred.

cryptoquick:
Thank you for this, it wasn't quite clear how that worked from the docs I was reading. @hecrj This should be merged.

hecrj (approved):
Thank you! 🎉

PR #647, opened 2020-12-02
Add pass_event field to Button

cossonleo:
Add pass_event field to Button to control whether pass event to its content or not


twitchyliquid64:
Instead of turning on / off this behavior with `pass_event()`, another option could be to always pass through events except button clicks.

hecrj:
Opened #668, which should supersede this.

PR #652, opened 2020-12-05
text_input: Do not capture Tab/Up/Down keys

### LINKED ISSUE #651: `TextInput` captures all `keyboard::KeyPressed` events while focused, breaking subscription-based hotkeys
I'm using `subscription::events_with()` to implement hotkey functionality. This works well, except when a text input in focused. I've tracked it down to

https://github.com/hecrj/iced/blob/d16b9cf7cd98a3d65ea5408ac9b72298cb267e85/native/src/widget/text_input.rs#L370-L373

`<snipped>`

https://github.com/hecrj/iced/blob/d16b9cf7cd98a3d65ea5408ac9b72298cb267e85/native/src/widget/text_input.rs#L547-L551

From what I can gather, widgets should return `event::Status::Ignored` for events that don't concern them, so other parts of the UI can react. `text_input` doesnt seem to do that.

twitchyliquid64:
Fixes #651

This PR has been updated based on the conversation in #651 to ignore three keys that are never part of the text input's characters, and are commonly used for widget navigation.

This is in lieu of an exhaustive mapping of virtual key codes which will never be part of a sequence which results in a character, as I could not find such a canonical mapping.

### DIFF
--- a/native/src/widget/text_input.rs
+++ b/native/src/widget/text_input.rs
@@ -549,6 +549,11 @@ where
                         self.state.keyboard_modifiers =
                             keyboard::Modifiers::default();
                     }
+                    keyboard::KeyCode::Tab
+                    | keyboard::KeyCode::Up
+                    | keyboard::KeyCode::Down => {
+                        return event::Status::Ignored;
+                    }
                     _ => {}
                 }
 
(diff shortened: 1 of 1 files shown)


drmason13:
What about web buttons? I'm not very experienced with iced but I think the same change would work there as well.

hecrj (approved):
Thank you! 🎉

PR #654, opened 2020-12-06
Hide titlebar on macos

### LINKED ISSUE #653: Hide titlebar on macos (but keep other decorations)
Let me use some images to explain what I want to achieve:

Screen Shot 2020-12-06 at 10 26 12 am

This is the current default:
Screen Shot 2020-12-06 at 10 26 51 am

This is the defaults with settings `decorations: false` on window settings:
Screen Shot 2020-12-06 at 10 32 59 am

In order to achieve the result in the first screenshot the following needs to be run on the window builder:

```
window_builder = window_builder
    .with_title_hidden(true)
    .with_titlebar_transparent(true)
    .with_fullsize_content_view(true);
```

If no one is currently working on also exposing these options I would open a PR that:

1. Create a iced_winit `PlatformSpecific` settings struct explicitly for macos and expose flags for `title_hidden`, `titlebar_transparent` and `fullsize_content_view`.
2. Extend the window_builder to use those new settings when on macos
3. Somehow expose the `platform_specific` settings. Currently these cannot be set and only the default value is used (see https://github.com/hecrj/iced/blob/master/src/window/settings.rs#L58). I would like some guidance on the best approach to expose this without breaking some of the design goals here.

atbentley:
Fixes #653 

Allows creating windows on macos that looks like this:

Screen Shot 2020-12-06 at 10 26 12 am


hecrj (approved):
Looks great! Thank you! :bow:

Let's merge 🎉

PR #657, opened 2020-12-10
Generic Element Content in Pane Grid TitleBar

clarkmoody:
Makes the `TitleBar` more flexible by replacing the basic string + size with a generic `Element`.

Not quite sure if I've connected all the dots, especially regarding events/interactions on the elements within the title content.

Upgraded the example program to have a colored Pane ID, showing off a `Row` containing two `Text` elements.

Screen Shot 2020-12-10 at 1 47 13 PM


hecrj (approved):
Looks great! It simplifies a bunch of code, as users can leverage the `Text` widget and, therefore, we can avoid rendering the title manually.

Thanks :tada:

PR #666, opened 2020-12-16
Add Tab widget and tweak Rule style

BillyDM:
I have re-created this PR because I deleted my previous fork to be even for another PR.

I added a tab widget with a modern indicator-line style. I tweaked the Rule style to have a bit more contrast. It also makes more sense to have the rule span its full width by default.

Screenshot from 2020-12-15 18-25-23

Again, could there be a way to have `default::Text` in `iced_graphics` to also include the font weight/family? This would allow the selected tab to have bold text while the unselected ones do not.


hecrj:
It may be better to contribute the new widget to [`iced_aw`](https://github.com/iced-rs/iced_aw) first.

Could you create a PR only with the `Rule` style changes?

PR #672, opened 2020-12-23
Update tokio to v1.0

yusdacra:
https://docs.rs/tokio/1.0.0/tokio/stream/index.html is why `async-stream` is included, so that we can turn `Interval` ticks to a `impl Stream`.

### DIFF
--- a/futures/src/time.rs
+++ b/futures/src/time.rs
@@ -81,6 +81,6 @@ where
             tokio::time::interval_at(start, self.0)
         };
 
-        stream.map(|_| std::time::Instant::now()).boxed()
+        stream.map(tokio::time::Instant::into_std).boxed()
     }
 }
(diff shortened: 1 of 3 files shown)


futures/src/time.rs
```diff
@@ -62,11 +62,25 @@ where
...
+        {
+            let stream = async_stream::stream! {
+               let mut interval = tokio::time::interval_at(start, self.0);
+               loop {
+                   yield interval.tick().await;
+               }
+            };
+
+            stream.map(|_| std::time::Instant::now()).boxed()
+        }
```
hecrj, line 82: Instead of relying on `async-stream`, could we leverage [`stream::unfold`](https://docs.rs/futures/0.3.8/futures/stream/fn.unfold.html) instead?

hecrj (changes requested):

futures/src/time.rs
yusdacra: Alright, did that in daa9dbfccd1fc54010469e9d4d0d06eb02cb262b.

hecrj (approved):
Thank you! :tada:

I think this should be good to go!

PR #683, opened 2021-01-04
Update `lyon` to `0.17` in `iced_graphics`

hecrj:
Updates `lyon` to the latest release (`0.17`).

I have not reviewed all of the API changes yet, so there may be opportunities for further improvements.

### DIFF
--- a/graphics/src/widget/canvas/frame.rs
+++ b/graphics/src/widget/canvas/frame.rs
@@ -108,7 +108,10 @@ impl Frame {
         size: Size,
         fill: impl Into<Fill>,
     ) {
-        use lyon::tessellation::{BuffersBuilder, FillOptions};
+        use lyon::path::builder::PathBuilder;
+        use lyon::tessellation::{
+            BuffersBuilder, FillOptions, FillTessellator,
+        };
 
         let Fill { color, rule } = fill.into();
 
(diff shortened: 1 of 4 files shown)


graphics/src/widget/canvas/frame.rs
```diff
@@ -127,12 +130,17 @@ impl Frame {
...
+        builder.add_rectangle(
             &lyon::math::Rect::new(top_left, size.into()),
-            &FillOptions::default().with_fill_rule(rule.into()),
-            &mut buffers,
-        )
-        .expect("Fill rectangle");
+            lyon::path::Winding::Positive,
+        );
+
+        let _ = builder.build().expect("Fill rectangle");
```
hecrj, line 143: I'm noticing a performance regression of around an order of magnitude in the `game_of_life` example.

I imagine the new `Frame::fill_rectangle` implementation is the culprit. We should try to address this before merging.

yamadapc:
One thing to try is `fill_options.with_intersections(false)` (though it's not new to 0.17).

I got around 2x improvement in `fill` with that path, I suppose there're places where it'd be safe to assume there're no intersections or to allow users to provide a hint to canvas that it's the case.

graphics/src/widget/canvas/frame.rs
```diff
@@ -127,12 +130,17 @@ impl Frame {
                 lyon::math::Vector::new(size.width, size.height),
             );
 
-        let _ = lyon::tessellation::basic_shapes::fill_rectangle(
+        let mut tessellator = FillTessellator::new();
```
yamadapc, line 133: This would also be retained in #988

graphics/src/widget/canvas/frame.rs
hecrj: The author of `lyon` addressed the issue in https://github.com/nical/lyon/pull/696 🎉

PR #686, opened 2021-01-07
Set IME's window position on TextInput

hatoo:
# Before
before

# After
after

I've created a very early implementation for setting IME's window position on TextInput ~~which works only on Windows~~.
Currently, I extended the `Clipboard` struct also handles the Input Method to avoid large diff. And set IME's window position in the `on_event` method.

## TODO

- [ ] Rename `Clipboard` and make also have responsibility of Input Method, or Separate struct and pass `Clipboard` and `InputMethod` to `on_event` arguments. Do you think which is better?
- [x] ~~Separate codes related to Input Method to another crate, like `window_clipboard `.~~ I've found that `winit` has the method for this situation.
- [x] Support MacOS
- [x] Support Linux

### DIFF
--- a/winit/src/clipboard.rs
+++ b/winit/src/clipboard.rs
@@ -46,7 +46,7 @@ impl Clipboard {
     }
 }
 
-impl<'a> iced_native::Clipboard for Clipboard {
+impl iced_native::Clipboard for Clipboard {
     fn read(&self) -> Option<String> {
         self.read()
     }
--- a/winit/src/clipboard.rs
+++ b/winit/src/clipboard.rs
@@ -56,6 +56,10 @@ impl iced_native::Clipboard for Clipboard {
                         &mut composition_form as _,
                     )
                 };
+
+                let _ = unsafe {
+                    winapi::um::imm::ImmReleaseContext(handle.hwnd as _, himc)
+                };
             }
         }
     }
--- a/winit/src/clipboard.rs
+++ b/winit/src/clipboard.rs
@@ -1,66 +1,28 @@
-use raw_window_handle::HasRawWindowHandle;
-
 /// A buffer for short-term storage and transfer within and between
 /// applications.
 #[allow(missing_debug_implementations)]
-pub struct Clipboard(
+pub struct Clipboard<'a>(
     window_clipboard::Clipboard,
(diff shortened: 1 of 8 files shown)


hatoo:
To follow the master branch, I've decided to create a new crate [window_ime](https://github.com/hatoo/window_ime) that handles IME features like [window_clipboard](https://crates.io/crates/window_clipboard) to avoid borrow-checker error.

You can check the status of my progress on https://github.com/hatoo/iced/tree/integrate_window_ime and it works only in Windows now.

 Any support for [window_ime](https://github.com/hatoo/window_ime) is appreciated.

hatoo:
Finally, I've successfully merged the master branch with a minimum interface change.

winit/src/clipboard.rs
```diff
@@ -51,4 +54,11 @@ impl iced_native::Clipboard for Clipboard {
     fn write(&mut self, contents: String) {
         self.write(contents)
     }
+
+    fn set_ime_position(&self, position: iced_core::Point) {
```
pickfire, line 58: I don't understand why is this method on clipboard? Shoudn't it be with window?

hatoo: > I don't understand why is this method on clipboard?

I thought it's natural that `TextInput` set IME window's position in [`on_event`](https://docs.rs/iced_native/0.4.0/iced_native/widget/trait.Widget.html#method.on_event) method and I chose to extend `Clipboard` struct to also handles IME to avoid larger diff. That's why this PR is WIP.

As I wrote above, I think there are 2 options.
1. Rename `Clipboard` to something and make also handles IME.
2. Pass another variable that handles IME to `on_event`.

> Shoudn't it be with window?

Is there a `window` that accessible from `Widget` trait? If so, maybe it's better to use that, but I also think that extend `Clipboard` is better still.

Piping:
for winit windows to handle xim input method, i think this PR can be referenced

https://github.com/alacritty/alacritty/pull/691

hecrj (changes requested):
We still need to figure out a proper API for IME support.

I don't think `Clipboard` should be concerned with IME positioning. Let's think about it before writing code!

PR #695, opened 2021-01-11
WIP: Horizontal Scrollable via duplication

GTimothy:
This is the implementation of a Horizontal version of the `Scrollable` widget.

Instead of being based on a `Column`, it is based on a `Row`.
The associated `Renderer` has to render the `Scrollbar` and `Scroller` at the bottom rather than on the right, and has to render it sideways rather than upright.

The `stylesheet` used are the same.

I renamed every height/width related field and function to be coherent with new orientations.

I added an example based on the original *scrollable* example.

This work is related to [this other branch of mine](https://github.com/GTimothy/iced/tree/row_column_unification), where instead of duplicating the `Scrollbar` widget, I merged the `Row` and `Column` widgets into one single `Block` widget (name subject to change if this work is one day used) :
https://github.com/GTimothy/iced/tree/row_column_unification
On this other branch I find it easier to see what code is shared a vertical and a horizontal version of a scrollbar. You can see this here : ee13eda806513dce5b4b0f30dcae4e7c237ee29f


hecrj:
I think it's better to move this PR to [`iced_aw`](https://github.com/iced-rs/iced_aw) and start a discussion there before we decide to make it a built-in `iced` widget.

Let me know if you need any help with that!

PR #696, opened 2021-01-12
Lighter dual axis scrollable

GTimothy:
After some discussions, here is a new version of the existing Scrollable widget.
 * This widget becomes much simpler: it exists to wrap a widget that needs a scrollbar.
 * This widget becomes more *flex*ible: it can be configured to be a horizontal or a vertical Scrollable using the .axis(...) method. For horizontal: `.axis(Axis::Horizontal)` and for vertical: `.axis(Axis::Vertical)`.
 * This widget has a default axis: vertical. Large content is usually displayed vertically (web pages for instance) so the default axis is `Axis::Vertical` to save you from having to press more keys.

Additional info:
* `spacing` `padding` `align_item` `push` are gone: wrap the contents in `Container` and `Column/Row` widgets to obtain the desired properties.
 * `width: Length`, `height: Length`, `max_width: u32`, and `max_height: u32` remain. 
 * `style()` setter remains as well.
 * `scrollable` and `scroller` properties: their *thickness* is called `width`, regardless of orientation.  
 

Any feedback is very welcome :) 
For instance, should scrollbar/scroller thickness be called something else than `width` (since `width` is usually a horizontal measurement but `scrollbar/scroller` width can be vertical when they are sideway?

### DIFF
--- a/examples/geometry/src/main.rs
+++ b/examples/geometry/src/main.rs
@@ -211,8 +211,8 @@ impl Sandbox for Example {
                  geometry for your widget.",
             ));
 
-        let scrollable = Scrollable::new(&mut self.scroll)
-            .push(Container::new(content).width(Length::Fill).center_x());
+        let scrollable = Scrollable::new(&mut self.scroll,
+            Container::new(content).width(Length::Fill).center_x());
 
         Container::new(scrollable)
             .width(Length::Fill)
--- a/examples/pane_grid/src/main.rs
+++ b/examples/pane_grid/src/main.rs
@@ -277,17 +277,15 @@ impl Content {
             ));
         }
 
-        let content = Scrollable::new(scroll)
-            .width(Length::Fill)
-            .spacing(10)
-            .align_items(Align::Center)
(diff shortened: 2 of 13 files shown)


cossonleo:
I think it support  vertical  and  horizontal at  the  same time  in  a  instance

GTimothy:
@cossonleo I am looking into that. I think I figured out which event capture to remove to allow a composition of vertical and horizontal scrollables to obtain the desired result. Stay tuned :)

GTimothy:
c7671eb removes an event capture, allowing nested scrollable to scroll simultaneously, but it may very well have unexpected side effects in a more complex widget tree.

Also, a nested scrollable's scrollbar is nested as well and can therefore invisible. A better solution is probably to have one scrollable do the 2 directions as once.

hecrj:
I think it's better to move this PR to [`iced_aw`](https://github.com/iced-rs/iced_aw) and start a discussion there before we decide to make it a built-in `iced` widget.

Let me know if you need any help with that!

PR #699, opened 2021-01-13
Add `smol` async runtime support

JayceFayne:
This pull adds support for the `smol` async runtime to iced!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - `"system_font"` feature gates reading system fonts. [#370]
+- Support for the [`smol`] async runtime. [#699]
 
 [#370]: https://github.com/hecrj/iced/pull/370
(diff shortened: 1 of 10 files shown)


hecrj (approved):
Thanks for this! Glad to see that supporting `smol` wasn't too much of a hassle.

I made some changes to fix the `time::Every` implementation, as `smol::Timer::interval` doesn't seem to exist.

I think this should be good to go! :beers:

JayceFayne:
> Thanks for this! Glad to see that supporting smol wasn't too much of a hassle.
Yeah, thanks to your Executor wrapper it was actually just a few lines of code!
> I made some changes to fix the time::Every implementation, as smol::Timer::interval doesn't seem to exist.
Hmm, it should exist, smol just reexports the async-io `Timer` implemented [here](https://github.com/smol-rs/async-io/blob/master/src/lib.rs#L192) :thinking:

hecrj:
> Hmm, it should exist, smol just reexports the async-io Timer implemented here

Yeah, my bad. Looks like I needed to run `cargo update`. Restored in 9da0a3de54e8e7161d4a3405633bc574ce43eb2d.

PR #720, opened 2021-01-29
impl Clone and Copy for all Style types in iced_style

taiki-e:
Many of the `Style` types in `iced_style` implement these traits, which is useful. It would be nice if the rest of the `Style` types in `iced_style` also have implementations of these traits.


hecrj (approved):
Good catch! We should always try to implement these traits.

Thanks :partying_face:

PR #760, opened 2021-03-03
Update: rand in solar_system example

TriedAngle:
gen_range() api from rand has changed from 0.7 to 0.8

this pull requests updates the use of rand in the solar_system example.


hecrj (approved):
Thank you!

Issue #560, opened 2020-10-10
[RFC] Blinking a `TextInput` cursor using periodic animation


twitchyliquid64:
# Introduction

As discussed in #89, users expect a _blinking_ cursor to signify the current insert position, of a focused text input.

However, implementing this is more complicated than simply updating the `TextInput` widget, because currently iced widgets can only update in response to user events, or application messages (via the Subscription system).

This document aims to walk through this problem space and propose a minimal solution to enable a blinking `TextInput`.

# Guide-level explanation

#### `TextInput` widget

Currently, the cursor is always shown (ie: it does not blink) for a focused text input. No cursor is shown for a `TextInput` which does not have focus.

All changes to the text-input's appearance are driven by user-interaction events.

#### The runtime

Underlying `iced` widgets is an [Event Loop](https://en.wikipedia.org/wiki/Event_loop), whose role is to bring together all the layers of `iced`. Think of this as the 'dispatch' or traffic light of an `iced` GUI - it controls the flow. The 'runtime' is composed of a number of separate components (`Application`, `user_interface` etc), but these details can be omitted for the sake of this explanation.

The runtime's job is to perform initial setup and then perform the following operations in sequence:

1. _Receive_ application events and user-interaction (mouse movement, clicks, keyboard etc)
2. _Deliver_ these events to the widgets, so they can update their internal state to reflect the happenings of the application or the user
3. _Layout_ the widgets. As the relative size and position of widgets may have changed in response to an event, _layout_ resolves the current position and size of every widget.
4. _Draw_ the widgets. Drawing involves updating the screen with what the widgets want to show.
5. _Wait_. In the current implementation, the runtime waits for a user-interaction or application event before doing it all again.

#### The widgets

Widgets communicate integrate with the rest of `iced` (the 'runtime') by implementing the [Widget](https://docs.rs/iced_native/0.2.2/iced_native/widget/trait.Widget.html) trait. This trait defines:

 * A way to communicate the size needs of the rendered widget (`width()`, `height()`), and resolve its position (`layout()`)
 * A way to provide a [display list](https://en.wikipedia.org/wiki/Display_list) (a list of drawing commands) to the visual backend (`draw()`)
 * A way to update the internal state of the widget based on user interaction or application defined events (`on_event()`)

Notably absent from this trait is any means to indicate to the underlying runtime that an update needs to occur at some time in the future: This current design assumes all updates are the result of user interaction or application events.

### Whats missing

[...]

Issue #31, opened 2019-10-23
Animations


hecrj:
Allow widgets to request a redraw at a specific time.

This is a necessary feature to render loading spinners, a blinking text cursor, GIF images, etc.

[`winit`] allows controlling the event loop in a flexible way. We may be able to use [`ControlFlow::WaitUntil`](https://docs.rs/winit/0.20.0-alpha3/winit/event_loop/enum.ControlFlow.html#variant.WaitUntil) for this purpose.

[`winit`]: https://github.com/rust-windowing/winit

michael-hart:
I'm not sure why a redraw at a specific time is necessary. Is this so that a component can render at e.g. a specific framerate? It seems like a timer or a separate scheduler to call update functions might be a way to go about this, followed by requesting a window redraw?

MichaelMcDonnell:
I need this too. I'm trying to write a program for the web that does custom animation. I looked at the `solar_system` and `stopwatch` example but they both use `async_std` to drive the animation with a timer. The `async_std` crate does not support WASM yet. See https://github.com/async-rs/async-std/issues/220

ron-wolf:
Hey @hecrj! I’m a noob but would love to take a shot at helping out with this. Are there any blockers for this issue?; and if not, which place(s) in the codebase would the change be made?

zyansheep:
I have an idea for a simple but versatile Animation widget where different functional methods could be used to modify a child widget.
Perhaps something like this:
```rust
fn update(&mut self, message: Message){
    match message {
        Message::TriggerAnimation => self.animation_state.forward(), // If the animation has not been run yet, run it forward
    }
}
fn view(&mut self) {
    Button::new(&mut self.increment_button, Text::new("Trigger")).on_press(Message::TriggerAnimation)
    Animation::new(&mut self.animation_state, child: Text::new("I just faded in!"))
    .fade_in(Curve::linear()) // Fade in linearly
    .translate(Curve::ease(), (-10, 0), (0,0)) // Move in from relative left
}
```

twitchyliquid64:
Following on from a conversation on the zulip chat, @hecrj mentioned a good start would be enumerating use-cases, and think through how animations/draw state will interact with other features and API design decisions (of particular note, async rendering, and layout/widget state).

In the interest of getting started, he's my attempt at rounding up the use-cases:

[...]

PR #762, opened 2021-03-05
Number input widget

hangleang:
Hi there, I would like to PR that implement number input widget in #209 issue. I have also added a simple example for show case this widget. I'm look forward to see any idea, feel free to improve it or leave a comment here.


examples/number_input/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+};
+
+#[derive(Default)]
+pub struct NumberInputDemo {
+    state: number_input::State,
+    value: u8,
+}
+
+#[derive(Debug, Clone)]
+pub enum NumInpMessage {
```
13r0ck, line 13: pub enum Message

examples/number_input/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+        window: window::Settings {
+            size: (250, 200),
+            ..Default::default()
+        },
+        ..Settings::default()
+    })
+}
+
+impl Sandbox for NumberInputDemo {
+    type Message = NumInpMessage;
```
13r0ck, line 29: type Message - Message;

examples/number_input/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+            value: 27,
+            ..Self::default()
+        }
+    }
+
+    fn title(&self) -> String {
+        String::from("Number Input Demo")
+    }
+
+    fn update(&mut self, message: Self::Message) {
```
13r0ck, line 42: fn update(&mut self, message: Message)

examples/number_input/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+        match message {
+            Self::Message::NumInpChanged(val) => {
```
13r0ck, line 44: Message::NumInpChanged

examples/number_input/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+                self.value = val;
+            }
+        }
+    }
+
+    fn view(&mut self) -> Element<Self::Message> {
```
13r0ck, line 50: fn view(&mut self) -> Element<Message>

13r0ck:
It looks good so far, there are some comments about how to remove the `Self` from your messages. I really wish that it accepted scroll inputs to increase and decrease the values, as well if I hold on the up/down arrow it would make sense for the value to increase while held. 

There is a bug with your implementation of the max value. Lets say I clear the input to `0` then i put my cursor infront of the `|0` (like that) then i press, for example 9, then the input is `9|0` with my cursor in the middle. Though if I press 9 again the input is blocked rather than overwriting the 0. If would feel more natural to go from `9|0` -> `99|` which is still under 255 (the max in your example), though your code checks the input as `99|0` which is above 255 and blocks the input. Hopefully that makes sense?

hecrj (commented):
Thank you so much for this! :smile: 

It may be interesting to contribute this widget to [`iced_aw`] first. This is, since very recently, the official community crate for additional widgets (see #761). I think it will be easier for users to contribute and improve the widget further there and then, once it eventually matures enough, we could migrate it to this repository.

Ideally, we should leave this repository for important fixes and core changes. This way, I can stop bottlenecking interesting contributions while focusing on the core parts of the library!

[`iced_aw`]: https://github.com/iced-rs/iced_aw

Kaiden42:
> It may be interesting to contribute this widget to iced_aw first.

Sounds like a good idea 😉. If you have problems with moving your widget to iced_aw, please let me know.

>  I really wish that it accepted scroll inputs to increase and decrease the values,

This can be easily done with `mouse::Event::WheelScrolled { delta }`. You can take a look at the [Color Picker](https://github.com/iced-rs/iced_aw/blob/main/src/native/overlay/color_picker.rs#L273) implementation for this.

> as well if I hold on the up/down arrow it would make sense for the value to increase while held.

As far as I know, this is currently not possible. It would require the widget to send an update/message by itself after some period, wich is currently not implemented. This is slightly related to #31 and #560 .

> There is a bug with your implementation of the max value. Lets say I clear the input to 0 then i put my cursor infront of the |0 (like that) then i press, for example 9, then the input is 9|0 with my cursor in the middle. Though if I press 9 again the input is blocked rather than overwriting the 0. If would feel more natural to go from 9|0 -> 99| which is still under 255 (the max in your example), though your code checks the input as 99|0 which is above 255 and blocks the input. Hopefully I makes sense?

For comparison, I looked at how the number-input field of [HTML](https://wiki.selfhtml.org/wiki/HTML/Tutorials/Formulare/input/number) handles this (chromium). The maximum value there, is not considered at all with the text input. With a max value of 220 you can easily enter 9999990 or similar. I find persöhnlich also not just intuitive.

However, with your suggestion I would rather go for `99|0` -> 255 instead of `99|0` -> 99. Maybe we can look at how other GUI libraries handle this case and pick out the best behavior for us.

hangleang:
> It looks good so far, there are some comments about how to remove the `Self` from your messages. I really wish that it accepted scroll inputs to increase and decrease the values, as well if I hold on the up/down arrow it would make sense for the value to increase while held.

Thank for review @13r0ck. I've format the example for convenient as you told. I also merge with your branch(more_scroll) and in the integration, I've change on scroll event to increase or decrease the value directly inside widget logic because I don't want user to handle those two messages by themselves, but the logic its self is still the same. I was given you credit on that section, Hope you don't mind on that.

> There is a bug with your implementation of the max value. Lets say I clear the input to `0` then i put my cursor infront of the `|0` (like that) then i press, for example 9, then the input is `9|0` with my cursor in the middle. Though if I press 9 again the input is blocked rather than overwriting the 0. If would feel more natural to go from `9|0` -> `99|` which is still under 255 (the max in your example), though your code checks the input as `99|0` which is above 255 and blocks the input. Hopefully I makes sense?

This update is fixed those bugs if I do understand you well. So, please check on that again if there is some point I missing. Leave it a comment. Thank again for spending your time on review.

hangleang:
> Thank you so much for this! smile
> It may be interesting to contribute this widget to [`iced_aw`](https://github.com/iced-rs/iced_aw) first. This is, since very recently, the official community crate for additional widgets (see #761). I think it will be easier for users to contribute and improve the widget further there and then, once it eventually matures enough, we could migrate it to this repository.

Yeah, you are welcome. I'll put this widget in that repo when the time allowed. I got a lot of ideas from the repo, hahaa.

> Ideally, we should leave this repository for important fixes and core changes. This way, I can stop bottlenecking interesting contributions while focusing on the core parts of the library!

Ohh I see. Thank for warm welcome.

hangleang:
> > It may be interesting to contribute this widget to iced_aw first.

Thank for interested on the widget. Good if we can works together.

> Sounds like a good idea wink. If you have problems with moving your widget to iced_aw, please let me know.

Not now, but soon hahaa. 

> > I really wish that it accepted scroll inputs to increase and decrease the values,
> This can be easily done with `mouse::Event::WheelScrolled { delta }`. You can take a look at the [Color Picker](https://github.com/iced-rs/iced_aw/blob/main/src/native/overlay/color_picker.rs#L273) implementation for this.

This update has implementation on that event which I merge from @13r0ck. Thank to @13r0ck.

> > There is a bug with your implementation of the max value. Lets say I clear the input to 0 then i put my cursor infront of the |0 (like that) then i press, for example 9, then the input is 9|0 with my cursor in the middle. Though if I press 9 again the input is blocked rather than overwriting the 0. If would feel more natural to go from 9|0 -> 99| which is still under 255 (the max in your example), though your code checks the input as 99|0 which is above 255 and blocks the input. Hopefully I makes sense?
> 
> For comparison, I looked at how the number-input field of [HTML](https://wiki.selfhtml.org/wiki/HTML/Tutorials/Formulare/input/number) handles this (chromium). The maximum value there, is not considered at all with the text input. With a max value of 220 you can easily enter 9999990 or similar. I find persöhnlich also not just intuitive.
> 
> However, with your suggestion I would rather go for `99|0` -> 255 instead of `99|0` -> 99. Maybe we can look at how other GUI libraries handle this case and pick out the best behavior for us.

This widget is required max value for checking valid number, not like the number input in HTML.

13r0ck:
I am going to hold out on more input until this PR is moved over to [Iced aw](https://github.com/iced-rs/iced_aw). This is going to be a useful widget!

Kaiden42:
Since this is now merged into `iced_aw` this PR can be closed now. :slightly_smiling_face:

PR #776, opened 2021-03-13
Added `select_all` method to `TextInput`.

AldoMX:
Hi, and thank you for creating iced.

I was tinkering with the `todos` example and I wanted to pre-select the text when you clicked the edit button.

But when I tried to update the example I noticed that the public methods from `TextInput` only allowed to move the cursor, so I added a `select_all` method and updated the `todos` example to achieve the desired outcome.

### DIFF
--- a/native/src/widget/text_input.rs
+++ b/native/src/widget/text_input.rs
@@ -792,6 +792,11 @@ impl State {
     pub fn move_cursor_to(&mut self, position: usize) {
         self.cursor.move_to(position);
     }
+
+    /// Selects all the content of the [`TextInput`].
+    pub fn select_all(&mut self) {
+        self.cursor.select_range(0, usize::MAX);
+    }
 }
 
 // TODO: Reduce allocations
--- a/examples/todos/src/main.rs
+++ b/examples/todos/src/main.rs
@@ -265,8 +265,10 @@ impl Task {
                 self.completed = completed;
             }
             TaskMessage::Edit => {
+                let mut text_input = text_input::State::focused();
+                text_input.select_all();
                 self.state = TaskState::Editing {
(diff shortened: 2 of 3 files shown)


native/src/widget/text_input.rs
```diff
@@ -792,6 +792,11 @@ impl State {
     pub fn move_cursor_to(&mut self, position: usize) {
         self.cursor.move_to(position);
     }
+
+    /// Selects all the content of the [`TextInput`].
+    pub fn select_all(&mut self) {
+        self.cursor.select_range(0, usize::MAX);
+    }
```
derezzedex, line 799: I think it would make more sense to expose a more general `select` function, accepting both the start and the end, then in case you needed to select all you would use it just like you currently call `cursor.select_range`.

I'm not sure if there's a way to clip the range to the actual text size, since the `Value` is not available for the `State`, but right now this is done anyway on the Renderer: https://github.com/hecrj/iced/blob/f076649fbb575ee036ab0b3f4511690e3379c115/graphics/src/widget/text_input.rs#L157-L158

hecrj (approved):
Works great! Thank you 🎉 

I have added the method to `iced_web` too, although we can't properly implement it yet! This way, the `todos` example can still be built for web.

derezzedex (changes requested):
I've left a comment regarding the proposed function. There's also the need to implement this on the `web` backend, but I'm not sure how that works because I'm not familiar enough with it.

native/src/widget/text_input.rs
hecrj: I think `select_all` satisfies a clear use case.

We can expose a `select_range` or similar and focus on those use cases in a different unit of work.

PR #788, opened 2021-03-21
feat(web): Support in-memory image data

Liamolucko:
I had to create two methods which basically do the same thing, `from_memory` and `from_slice`, because `from_memory` takes ownership of the bytes to be compatible with `iced_native`.

Also, `Data` is incompatible, because if I stored the bytes in `Data` and created a new object URL every render, it would have caused a memory leak because bumpalo doesn't call destructors and there'd be no way to call URL.revokeObjectUrl on it. It's also more efficient this way.

### DIFF
--- a/web/Cargo.toml
+++ b/web/Cargo.toml
@@ -20,7 +20,7 @@ wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
 url = "2.0"
 num-traits = "0.2"
-js-sys = "0.3"
+base64 = "0.13"
 
 [dependencies.iced_core]
 version = "0.3"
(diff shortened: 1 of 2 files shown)


Liamolucko:
Whoops, this isn't actually working - I think when I was testing if the mime type needed to be specified it must have still been in browser cache or something.

Liamolucko:
I didn't realise widgets were dropped immediately after being constructed, which meant that the image handle was being dropped and the object url invalidated. So I've changed `Data` to work the same as in `iced_native`, and create the object url when rendering, which means this now works but causes a memory leak every time an image is rendered :/.

I'm not sure how this can be done without causing a memory leak.

hecrj (commented):
We could force users to keep the `Handle` in their `Application` state and clone it when creating the `Image` widget, instead of creating the `Handle` every time while rendering.

hecrj (approved):
We can merge this for now and improve the API / performance in another PR.

Thank you! 🙇

Issue #207, opened 2020-02-25
How can I can use dx11 or dx12 backend?


cossonleo:
@hecrj Thanks for your work
It is vulkan backend when I run wgpu-rs's example describe.
Is Iced alaways use vulkan?
If not, how can I check the backend of iced.
If it is not default dx11 or dx12 on windows, How Can I switch to dx11 or dx12?

hecrj:
`wgpu` should choose the best option depending on the OS. If I remember correctly, it should choose DirectX 12 on Windows if available before Vulkan.

I've been meaning to add this information to the debug overlay. I will try to open a PR today!

cossonleo:
@hecrj  When I assign Application's setting.antialiasing true, it use dx12 and example will run fine.
But default, wgpu will use PowerPreference default, and it will use vulkan backend on my win10 laptop, but win10's driver does not fully support vulkan. #172

hecrj:
Glad to help!

Just for the record, this is caused by `wgpu` choosing your integrated GPU by default. Once you enable anti-aliasing, the `PowerPreference` changes to `HighPerformance` and `wgpu` chooses your dedicated GPU instead.

PR #789, opened 2021-03-21
Add WGPU_BACKEND environment variable

### LINKED ISSUE #787: How to explicitly specify the rendering backend as DX12
The `WGPU` has `Vulkan` enabled on the default backend of the Windows platform, and there will be some Cottonwood when the windows platform is resizing, but if you replace it with a `DX12` backend, you don’t have Cottonwood. I don’t know how to simply specify the render backend in `iced`. Can anyone help me?

Cupnfish:
Fixes #787. Related to #207.
Inspired by Bevy’s source code.
You can explicitly specify the render backend when you run your app by setting the WGPU_BACKEND environment variable. If this environment variable does not exist, the default is the wgpu::BackendBit::PRIMARY.

If you use cargo with version 1.52 or later, you can easily set your environment variables.

First create a folder called `.cargo` in your project directory, then create a `config.toml` file in the `.cargo` folder:
before:
```shell
.
├── Cargo.lock
├── Cargo.toml
└── src
    └── main.rs
```
after: 
```shell
.
├── Cargo.lock
├── .cargo
│   └── config.toml
├── Cargo.toml
└── src
    └── main.rs
```
`.cargo/config.toml`:
```toml
[env]
WGPU_BACKEND = "dx12"
```
Specific configuration method can be seen [here](https://github.com/rust-lang/cargo/pull/8839#issuecomment-725678657).

### DIFF
--- a/glow/src/settings.rs
+++ b/glow/src/settings.rs
@@ -29,3 +29,12 @@ impl Default for Settings {
         }
     }
 }
+
+impl Settings {
+    /// Creates new [`Settings`] using environment configuration.
+    ///
+    /// Currently, this is equivalent to calling [`Settings::default`].
+    pub fn from_env() -> Self {
+        Self::default()
+    }
+}
--- a/src/application.rs
+++ b/src/application.rs
@@ -206,7 +206,7 @@ pub trait Application: Sized {
                 } else {
                     None
                 },
-                ..crate::renderer::Settings::default()
+                ..crate::renderer::Settings::from_env()
             };
 
             Ok(crate::runtime::application::run::<
--- a/wgpu/src/settings.rs
+++ b/wgpu/src/settings.rs
@@ -16,6 +16,9 @@ pub struct Settings {
     /// [`Backend`]: crate::Backend
     pub present_mode: wgpu::PresentMode,
 
+    /// The internal graphics backend to use.
+    pub internal_backend: wgpu::BackendBit,
+
     /// The bytes of the font that will be used by default.
     ///
     /// If `None` is provided, a default system font will be chosen.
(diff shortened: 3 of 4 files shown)


Cupnfish:
Running after setting through `.cargo/config.toml` requires `cargo run -Zconfigurable-env` instead of `cargo run`.If you set the environment variable on your system, rather than through `.cargo/config.toml`, you do not need to add this `-Zconfigurable-env`.

hecrj:
Is an environment variable the proper way to introduce this kind of configuration? Should it always be a user choice?

Cupnfish:
I think the ideal situation is the compiled iced program, you can choose their own internal rendering back-end, just need to restart the program can switch back-end, but it seems difficult to achieve. Environment variables make it easy for developers to specify which backend should be used for the compiled version, but users have no choice.

dhardy:
> Is an environment variable the proper way to introduce this kind of configuration? Should it always be a user choice?

I would say yes: sometimes because of a driver issue with some backend, sometimes just for testing.

At any rate I did something similar (though `WGPU_BACKEND` is a better env var name): https://github.com/kas-gui/kas/blob/master/kas-wgpu/src/options.rs#L50

derezzedex:
> Should it always be a user choice?

If by user you mean the end-user of an `iced` application, I don't think so. I think it should be up to the developer using `iced` to decide if they want support `X` backend.
I'd include a new field under `iced::Settings` which exposes an enum to currently supported `Backend`s.

In doing so, we would allow developers to create a "backend fallback strategy",
where they can request for some backend and fallback if that backend is not supported
(e.g. using `Application::run`s `Error::GraphicsAdapterNotFound` to try using other `Backend`).

This would allow any developer to add a `WGPU_BACKEND` flag if they wish to do so (we could also add to `iced`s example if convinient)
and would help me on my current work to add support for previous OpenGl versions on the `glow` backend :wink: (which would fallback on runtime).
This is certainly requires more work, but in my opinion, would make backend selection more flexible.

```rust
pub enum GlowBackend{
    Modern,
    Compatibility,
    Legacy,
}

pub enum Backend{
    #[cfg(feature="wgpu")]
    Wgpu(wgpu::BackendBit),
    #[cfg(feature="glow")]
    Glow(GlowBackend)
}
```

Maybe we should move this to a discussion?

hecrj (approved):
I have changed the implementation details a bit, but the result for users should be the same.

Specifically, I have introduced a `from_env` method to the `Settings` of our renderers, instead of coupling the environment variable to `Compositor::new`.

hecrj:
@derezzedex Backend fallback is in the roadmap, but I believe solves a different kind of use case.

PR #815, opened 2021-04-05
Enable qr_code feature on docs.rs

taiki-e:
The `QRCode` widget does not currently appear in the document. https://docs.rs/iced/0.3.0/iced/widget/index.html


hecrj (approved):
Thank you!

PR #825, opened 2021-04-08
add window visibility

### LINKED ISSUE #824: Add visible window setting
Would be great to set initial visibility of window with a new window setting that gets passed to https://docs.rs/winit/0.24.0/winit/window/struct.WindowBuilder.html#method.with_visible.

Related to #822 and implementing a sys tray for our application. We want to have the option to launch it to the tray. We have a workaround but we get a very brief "flash" of the iced window before it gets hidden.

tarkah:
resolves #824

### DIFF
--- a/glutin/src/application.rs
+++ b/glutin/src/application.rs
@@ -55,6 +55,7 @@ where
         let builder = settings.window.into_builder(
             &application.title(),
             application.mode(),
+            application.visible(),
             event_loop.primary_monitor(),
         );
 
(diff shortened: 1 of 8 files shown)


hecrj (commented):
I wonder if this could be a new variant in `window::Mode` instead?

tarkah:
I had the same thought as well, but I went with this implementation as it's more 1:1 with winit.

Let me push a new commit with that change and see if you prefer it that way.

tarkah:
@hecrj Yeah, I think `Mode::Hidden` is cleaner. A simpler implementation and we don't have to add a new Application method.

hecrj (approved):
Agreed. Looks great!

Thank you 🎉

PR #831, opened 2021-04-14
Added text color and font options for native radio and checkbox

zdevwu:


13r0ck:
I believe these can be set by putting a radio / checkbox in a container.

native/src/widget/radio.rs
```diff
@@ -196,8 +212,8 @@ where
             label_layout.bounds(),
             &self.label,
             self.text_size.unwrap_or(renderer.default_size()),
-            Default::default(),
-            None,
+            self.font,
+            self.text_color.clone(),
```
hecrj, line 216: Is calling `clone` necessary here? `Color` implements `Copy`.

hecrj (changes requested):
Thank you! This is great 😄

native/src/widget/radio.rs
zdevwu: nope, I just typed unconsciously :)

hecrj (approved):
Looks great! Thank you 🎉

PR #851, opened 2021-05-03
Fix `Scrollable` scrollbar being rendered behind contents

### LINKED ISSUE #765: Scrollbar rendered behind the scrollable content
I've recently updated to the latest head of the master branch and noticed some change in the rendering of the scrollbar.
It looks like it is now rendered behind the content of the scrollable.

Something between [before](https://github.com/hecrj/iced/commit/12c0c18d662d2b817b559b94c71d18e122c76990) and [now](https://github.com/hecrj/iced/commit/a74974a8e4d8afc72c33434a22df74c3582b78a8) changed this behavior. Anyone else noticed this?

Before: 
scrollbar_before

Now:
scrollbar_now

### LINKED ISSUE #821: Scrollbar appears behind buttons + Broken 'explain UI' layout
To summaries, I've noticed that placing buttons in columns then putting columns into a scroll-view seems to lead to the scrollbar being forced behind the buttons, even though the columns and scrollbars both have padding and spacing.

image

In the image above, you can see the scrollbar (Right) appears to be behind some buttons. 

Interestingly, with the UI wireframe enabled, it shows the UI is even more broken!
image

It appears like the wireframe for my custom table (Which is made out of buttons) is rendered correctly, but the wireframe for it is placed incorrectly at the bottom of the screen behind the Freeze frame data table at the bottom.

The code to generate a table of buttons is located here:
https://github.com/rnd-ash/OpenVehicleDiag/blob/main/app_rust/src/widgets/table.rs

The table emits a Message showing which x and y element was selected by the user

hecrj:
... by issuing a new clip layer just for the scrollbar itself.

Fixes #765.
Fixes #821.

### DIFF
--- a/examples/scrollable/Cargo.toml
+++ b/examples/scrollable/Cargo.toml
@@ -6,4 +6,4 @@ edition = "2018"
 publish = false
 
 [dependencies]
-iced = { path = "../.." }
+iced = { path = "../..", features = ["debug"] }
--- a/examples/scrollable/src/main.rs
+++ b/examples/scrollable/src/main.rs
@@ -1,8 +1,8 @@
 mod style;
 
 use iced::{
-    scrollable, Column, Container, Element, Length, Radio, Row, Rule, Sandbox,
-    Scrollable, Settings, Space, Text,
+    button, scrollable, Button, Column, Container, Element, Length, Radio, Row,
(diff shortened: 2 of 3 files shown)


clarkmoody:
Tested and working for me on Mac 👍  59c2500

PR #729, opened 2021-02-05
Some Scrolling tweaks

13r0ck:
While all of the changes work well, I am quite new to Rust, and I imagine that many changes will have to be made for the code quality to meet the rest of the Iced code base.

3 things in this pull request:
1. Adding scroll to `pick_list`s
    I think that this functionality for pick_lists should be expected. Just allows for selecting from a `pick_list` by scrolling over a closed `pick_list`. Though I am not confident with my code quality. It works reliably, but code is a mess. Please let me know how to improve it.

2. Adding `on_scroll` functions to `text_input`
    I added `on_scroll_up` and `on_scroll_down` to `Text_Input`s. I am not confident on implementation nor code quality on this one. I needed it on my project specifically for `Text_Input`s that the user enters integers, and then scroll for tweaks if wanted (think a timer input that allows to typing in the time as well as scrolling to the time). These functions could also be useful for other widgets (`on_scroll_left` could be useful for a slider for example). Though I am open to any code implementation ideas.

3. Adding convenience functions for `scrollable`s
    Simple functions just for convenience, rather than having to play with `Rectangles` and content, just one function call to scroll to the bottom. (I use this on a GUI that creates new elements in a list above the create button. The create button is always at the bottom of a `Scrollable` thus scrolling to the bottom after pressing the create button is convenient).

PR #872, opened 2021-05-10
add scrolling to pick_lists

13r0ck:
This is a PR to clean up #729 as that PR had 3 features, 2 of which are handled by [iced_aw/#11](https://github.com/iced-rs/iced_aw/pull/11) and #607.

All this PR does is create a message with a new Pick_List value when a scroll event is detected on said Pick_List.

### DIFF
--- a/native/src/widget/pick_list.rs
+++ b/native/src/widget/pick_list.rs
@@ -265,19 +265,11 @@ where
                 };
 
                 if y.is_sign_negative() {
+                    let mut options_iter = self.options.iter();
                     if let Some(selected) = self.selected.as_ref() {
-                        let i = self
-                            .options
-                            .iter()
-                            .position(|option| option == selected)
-                            .unwrap_or(0)
-                            + 1;
-                        if i < self.options.len() {
-                            messages.push((self.on_selected)(
-                                self.options[i].clone(),
-                            ));
-                            if let Some(msg) = self.on_change.take() {
-                                messages.push(msg)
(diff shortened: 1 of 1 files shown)


native/src/widget/pick_list.rs
```diff
@@ -114,6 +116,12 @@ where
         self
     }
 
+    /// Sets the message sent when [`PickList`] selection changes
+    pub fn on_change(mut self, msg: Message) -> Self {
```
hecrj, line 120: Why is the `on_change` trigger necessary? Shouldn't we be able to use the `on_selected` one instead? It seems a bit redundant.

hecrj (changes requested):

native/src/widget/pick_list.rs
13r0ck: Yep totally redundant. It's now removed. Thanks!

hecrj (approved):
Looks great! Thanks! 🎉 

I have abstracted and simplified the logic a bit in 6069e90 so there's not a much code duplication and we avoid edge cases when the list of options is empty.

Additionally, I have disabled the scrolling behavior for the `Pixels` variant of `ScrollDelta`. This way, the `PickList` will react to discrete scroll events (e.g. an actual mouse wheel) while dismissing continuous scroll events (e.g. scrolling with a trackpad).

I believe this ship is ready to sail! 🚢

13r0ck:
No, thank you!!!!!! 😄 
Your changes are a great learning lesson for me too. I hope you have a great weekend. Thanks!

PR #875, opened 2021-05-11
This commit optimizes the algorithm used to convert rgba pixels into bgra pixels for SVG's.

ZakisM:
This commit optimizes the algorithm used to convert rgba pixels into bgra pixels for rendering SVG's. Instead of creating a new Vec<u8> with the pixels reordered we re-order the existing pixels in place.

The difference in speed between the algorithms can be seen below (it is much much much faster):

Screenshot 2021-05-11 at 22 43 57

This results in resizing the window to be much smoother when you have a large svg (such as the one used in the SVG example).


hecrj (approved):
Awesome! 🎉 

Thank you for the detailed description 🙇

PR #876, opened 2021-05-13
Add a primary backend that can be set

Cupnfish:
I’ve found that some machines don’t have some backend ( For example, win10, but there is no DX12 ) , and that setting it to a fixed backend can cause problems, so it makes sense to add the option to support fallback within the configurable backend.

Setting an environment variable does make me feel good during development, but the need to set such an environment variable later on when loading the application to the user makes things weird.

I now use dotenv to help me set up environment variables so that I don’t have to deal with the environment variables inside the user system, but it still requires some cumbersome setup.

Therefore, while submitting this PR, I also hope that someone can think about whether there is a more elegant way to deal with this issue. 

The current idea of ​​optional backend is only somewhat useful under the windows platform, because wgpu is currently only supported by vulkan and metal on linux and mac, so this issue is not very important.

### DIFF
--- a/wgpu/src/settings.rs
+++ b/wgpu/src/settings.rs
@@ -78,6 +78,7 @@ fn backend_from_env() -> Option<wgpu::BackendBit> {
             "dx11" => wgpu::BackendBit::DX11,
             "gl" => wgpu::BackendBit::GL,
             "webgpu" => wgpu::BackendBit::BROWSER_WEBGPU,
+            "primary" => wgpu::BackendBit::PRIMARY,
             other => panic!("Unknown backend: {}", other),
         }
     })
--- a/wgpu/src/settings.rs
+++ b/wgpu/src/settings.rs
@@ -47,6 +47,7 @@ impl Settings {
     ///     - `dx11`
     ///     - `gl`
     ///     - `webgpu`
+    ///     - `primary`
     pub fn from_env() -> Self {
         Settings {
             internal_backend: backend_from_env()


Cupnfish:
I have an another  requirement that if I set the default backend to PRIMARY, I want to get what the current backend is for display on the UI.
And I found this on the WGPU:
```rust
// https://docs.rs/wgpu/0.8.1/wgpu/struct.Adapter.html#method.get_info
pub fn get_info(&self) -> AdapterInfo
```
I wonder if it is possible to expose a similar method in iced？

wgpu/src/settings.rs
```diff
@@ -78,6 +78,7 @@ fn backend_from_env() -> Option<wgpu::BackendBit> {
             "dx11" => wgpu::BackendBit::DX11,
             "gl" => wgpu::BackendBit::GL,
             "webgpu" => wgpu::BackendBit::BROWSER_WEBGPU,
+            "primary" => wgpu::BackendBit::PRIMARY,
```
hecrj, line 81: Seems reasonable. Could you update the documentation comment too?

hecrj (changes requested):

wgpu/src/settings.rs
Cupnfish: I think I’ve updated the document. In addition to this, is it possible to implement another requirement, which is to get the AdapterInfo?Although it's not really a necessary feature, I'll see if there's a way to expose this method, and if not, forget it.

hecrj: >  In addition to this, is it possible to implement another requirement, which is to get the AdapterInfo?

Since `iced` is backend agnostic, we will need to figure out a way to do it without exposing the backend internals.

hecrj (approved):

Issue #880, opened 2021-05-15
render svg into graphics canvas


aentity:
Hi i want to render svg i have into graphics canvas, is it possible?

For example, i use graphics to draw things, but occasionally i want to draw/render an svg to somewhere in this canvas.

I can't see how to do it, is there help or example for such thing? or do we need new api?

xd009642:
Just wondering if there's any progress on this or a solution on how to do it in iced currently?

aquirdTurtle:
Also came up for me recently. I'll give the change in the referenced #881 a try. For example I have a little chess-related app and want to draw a chessboard and then place chess piece SVGs on top of the board. Also related, issue #30; if that were implemented maybe I wouldn't care as much, although I will mention that svg support seems very much in the spirit of a canvas object. Things to prevent the user from having to draw everything manually.

xd009642:
@aquirdTurtle  what a coincidence a chess-related app is what I'm working on. If you want something that practically works I've had the most luck with Tauri and leaning on JS for the UI via react-chessboard. I say most luck because the javascript/rust workflow isn't the most ergonomic but at least the javascript ecosystem has a lot more hands working on it

PR #881, opened 2021-05-15
WIP allow svg render in canvas

aentity:
@hecrj for #880 i have made simple PR to see how svg can render to canvas. It works for my purpose!

For example, i can do:

```rust
self.cache.draw(bounds.size(), |frame| {
  frame.fill_svg(self.svg, bounds);
});
```

and this works!!!

however, some things:
1. I am not sure if it is good how i pass svg primitive in the `fill_svg` method
2. private handle exposed in PR, but not sure best way to expose
3. there appears to be some kind of gpu memory leak, as it is very slow and render time goes up and up (i do not invalidate cache in my example, and show that fill svg only called once, so i'm not sure what is going on, hopefully you can advise)

thank you!!

I would very like to see svg rendering in canvas, very cool, hope we can fix this and get it in iced sometime

### DIFF
--- a/native/src/program/state.rs
+++ b/native/src/program/state.rs
@@ -57,6 +57,11 @@ where
         &self.program
     }
 
+    /// Returns a reference to the [`Program`] of the [`State`].
+    pub fn program_mut(&mut self) -> &mut P {
+        &mut self.program
+    }
+
     /// Returns a reference to the current rendering primitive of the [`State`].
     pub fn primitive(&self) -> &<P::Renderer as Renderer>::Output {
         &self.primitive
--- a/graphics/src/widget/canvas/frame.rs
+++ b/graphics/src/widget/canvas/frame.rs
@@ -215,6 +215,14 @@ impl Frame {
         });
     }
 
+    /// Fill svg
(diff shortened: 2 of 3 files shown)


native/src/program/state.rs
```diff
@@ -57,6 +57,11 @@ where
         &self.program
     }
 
+    /// Returns a reference to the [`Program`] of the [`State`].
+    pub fn program_mut(&mut self) -> &mut P {
```
aentity, line 61: sorry this was accident; will remove if this PR is good idea. i will think i will need a `program_mut` though, for unrelated reason :)

hecrj:
While the changes are straightforward, the main concern here is whether we should make the promise that the `Canvas` API will support SVG rendering or not.

SVG support is a huge promise to make! For instance, `iced_glow` supports `Canvas` but does not support `Svg` yet!

Furthermore, I'd like to eventually port the `Canvas` widget completely to the `iced_web` runtime and other renderers as well.

Because of these reasons, I don't think we should make this promise at this point!

aentity:
ok good points to make.

some alternative: can we allow behind feature "svg-canvas" or something?

And the changes I made are how to do this, or is there some extra? I felt it was too simple, but maybe it means library is good :)

PR #885, opened 2021-05-17
image viewer: Only calculate viewport based width/length for Length::Unit|Shrink

### LINKED ISSUE #878: Image Viewer bounds expansion when zooming
This is both a question and depending on if I missed something a feature request.

Right now, when using the viewer as follows:
```rust
image::viewer::Viewer::new(&mut self.viewer_state, img.clone())
    .width(Length::Fill)
    .height(Length::Fill)
```
the result is an image constrained by it's viewport. 

This creates an in my opinion somewhat unexpected result when zooming into the image:
On a blank screen with just the viewer, the image can never exceed the original bounds even when zoomed in even if space would be available.

Image displayed in full window, no zooming  (container has centerx, centery hence in middle):

Slightly zoomed and moved with explain turned on, image now cut off on the right:

As width/height are specified as `Length::Fill` my expectation would have been that the image expands as far as it can within those constraints, not stay bound to it's original size.

Did I miss something or is this currently not possible? 

From what I can see this is due to the `layout` function in `viewer.rs` always receiving the same width / height, not letting them be bigger than the initially displayed image. If I do an ugly hack and replace `renderer.dimensions(&self.handle)` with hardcoded values, I get my desired effect, but I didn't see an obvious way to do it cleanly.

Chiheisen:
Currently image::viewer will not expand it's image into empty space even when `Length::Fill` or `Length::FillPortion` are used and space is available. 

This is because height/width is always calculated with viewport. In my eyes, viewport shouldn't have an influence if the layout size isn't constrained to the size that the viewport is limiting to.
Instead of always, only calculate viewport based length/width when the image has Length::Unit or Length::Shrink.

I might be missing something, so corrections are very welcome.
 
Fixes #878

### DIFF
--- a/native/src/widget/image/viewer.rs
+++ b/native/src/widget/image/viewer.rs
@@ -132,19 +132,30 @@ where
     ) -> layout::Node {
         let (width, height) = renderer.dimensions(&self.handle);
 
-        let aspect_ratio = width as f32 / height as f32;
-
         let mut size = limits
             .width(self.width)
             .height(self.height)
             .resolve(Size::new(width as f32, height as f32));
 
-        let viewport_aspect_ratio = size.width / size.height;
-
-        if viewport_aspect_ratio > aspect_ratio {
-            size.width = width as f32 * size.height / height as f32;
+        let expansion_size = if height > width {
+            self.width
         } else {
-            size.height = height as f32 * size.width / width as f32;
+            self.height
+        };
+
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thank you!

I am not the main author of this widget, so I am not completely sure of the original use case. However, I believe the changes seem reasonable.

Let's merge! 🎉

PR #886, opened 2021-05-18
pick_list: fix layouting not respecting fonts

### LINKED ISSUE #879: PickList triangle icon overlaps text
19 Thursday-slop0209

19 Thursday-slop0212

I don't know if it has to do with the font, which is NerdFont (Agave). In the case of two words, it seems like it's too late to break into two lines:
19 Thursday-slop0742
19 Thursday-slop0734

Chiheisen:
pick_list layouting always uses default font when calculating width, resulting in overlaps when using differently spaced fonts.

Fixes #879

### DIFF
--- a/native/src/widget/pick_list.rs
+++ b/native/src/widget/pick_list.rs
@@ -163,7 +163,7 @@ where
                         let (width, _) = renderer.measure(
                             &label,
                             text_size,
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thank you!

PR #725, opened 2021-02-03
Update to wgpu 0.7

PolyMeilex:
Notes:
- `wgpu::PowerPreference::Default` no longer exists, maybe there is another way to replicate its behavior.
- Not sure if `filtering sampler` should be used. (so it is not used)
- Anti-aliasig fix had to be removed because of wgpu ScissorRect validation error [link](https://github.com/hecrj/iced/pull/725/commits/bd6b8304bd940c5519fdf705698974994d5d1ca1)

What is working:
- [x] bezier_tool
- [x] clock
- [x] color_palette
- [x] counter
- [x] custom_widget
- [x] download_progress
- [x] events
- [x] game_of_life
- [x] geometry
- [x] integration
- [x] pane_grid
- [x] pick_list
- [x] pokedex
- [x] progress_bar
- [x] qr_code
- [x] scrollable
- [x] solar_system
- [x] stopwatch
- [x] styling
- [x] svg
- [x] todos
- [x] tour

Closes #723.

PR #830, opened 2021-04-13
Upgrade wgpu

Dispersia:
Draft PR to upgrade wgpu and use wgsl so cross can stop being compiled for android, includes temporary patches until fix for staging belt goes out.

Fixes #511.
Closes #523.

PR #887, opened 2021-05-22
Update to wgpu 0.8

mvilim:
Most of these are simple changes due to upstream refactorings, but there are two exceptions:

### 1. Alignment of structs based on largest field size in `wgpu/src/quad.rs` and `wgpu/src/triangle.rs`

Error I get in e.g. `integration` with wgpu 0.8 without the alignment changes:

```
Caused by:
    In Device::create_render_pipeline
      note: label = `iced_wgpu::quad pipeline`
    error matching VERTEX shader requirements against the pipeline
    shader global ResourceBinding { group: 0, binding: 0 } is not available in the layout pipeline layout
    buffer structure size 80, added to one element of an unbound array, if it's the last field, ended up greater than the given `min_binding_size`
```

The upstream source of these changes is `naga` https://github.com/gfx-rs/naga/pull/663 and the web gpu spec itself (which now requires this alignment, see https://gpuweb.github.io/gpuweb/wgsl/#structure-layout-rules).

In `triangle.rs` I rely on the fact that the structure is already aligned and simply perform a no-op alignment (so if someone else comes along to change the padding it's clearer that they need to consider the alignment). If you'd like something done different done here let me now. See https://github.com/hecrj/iced/compare/master...mvilim:wgpu-0.8?expand=1#diff-e1f0045458bcc8906b9880928a65c95b5a5d49dbdb59058f9b4973e528808bfbR436

### 2. Consistent bind group and sampler filtering in `wgpu/src/triangle/msaa.rs`

In the wgpu 0.7 update (#725), the filtering was disabled in the bind group `filtering: false` but left as `Linear` in the sampler. That combination causes me to receive this error with wgpu 0.8:

```
wgpu error: Validation Error

Caused by:
    In Device::create_bind_group
      note: label = `iced_wgpu::triangle::msaa uniforms bind group`
    sampler binding 0 expects filtering = false, but given a sampler with filtering = true
```

I wasn't sure why this change was made in the 0.7 update (`filtering: true` works for me in wgpu 0.7.0 but perhaps a transitive version has changed), so I erred on the side of caution and simply changed the sampler to match with `Nearest`. If you would rather just change `filtering` back to `true`, let me know (it works for me in wgpu 0.8 as well).

See https://github.com/hecrj/iced/compare/master...mvilim:wgpu-0.8?expand=1#diff-e25ec11fdbfd4ad5ba3acbaf97caabc9cf532e735e50076912edbc5966815c76R23

I have verified that all the examples work for me:

- [x] bezier_tool
- [x] clock
- [x] color_palette
- [x] counter
- [x] custom_widget
- [x] download_progress
- [x] events
- [x] game_of_life
- [x] geometry
- [x] integration
- [x] pane_grid
- [x] pick_list
- [x] pokedex
- [x] progress_bar
- [x] qr_code
- [x] scrollable
- [x] solar_system
- [x] stopwatch
- [x] styling
- [x] svg
- [x] todos
- [x] tour

### DIFF
--- a/examples/integration/src/scene.rs
+++ b/examples/integration/src/scene.rs
@@ -20,8 +20,8 @@ impl Scene {
     ) -> wgpu::RenderPass<'a> {
         encoder.begin_render_pass(&wgpu::RenderPassDescriptor {
             label: None,
-            color_attachments: &[wgpu::RenderPassColorAttachmentDescriptor {
-                attachment: target,
+            color_attachments: &[wgpu::RenderPassColorAttachment {
+                view: target,
                 resolve_target: None,
                 ops: wgpu::Operations {
                     load: wgpu::LoadOp::Clear({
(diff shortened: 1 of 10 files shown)


mvilim:
Whoops, looked like I somehow missed that #830 already existed for this (thanks to #889 for noticing that). Closing this in favor of the official version; sorry for the noise!

PR #888, opened 2021-05-22
Add Placeholders to PickList

### LINKED ISSUE #726: Picklist with placeholder
Hi sir, I would like to request _placeholder_ on existing Picklist widget. It should be displayed with a placeholder text whenever no option is selected. Thank

Ace4896:
Hi! This PR adds support for displaying a placeholder label when no value is currently selected in a PickList.

Closes #726.


hecrj (approved):
This is great! Thank you 🙇 

I have made some changes to remove some duplication and fix a minor issue with the `hash_layout` implementation.

Additionally, I have introduced a `placeholder_color` field to the `Style` of a `PickList`. This way, we can help the user differentiate between a selected option and a placeholder.

I think this should be ready to go! 🎉

Ace4896:
Thanks for making those changes and catching the issue with the hash layout! The separate colour for the placeholder definitely makes sense, it might be difficult to tell otherwise.

PR #889, opened 2021-05-22
Update wgpu to 0.8

PolyMeilex:
Simple update to wgpu 0.8.
Everything seams to works as expected so far.

TODO:
- [ ] Find a source of those validation warnings 
```
Unexpected varying type: Array { base: [1], size: Constant([7]), stride: 4 }
Unexpected varying type: Array { base: [1], size: Constant([5]), stride: 4 }
Unexpected varying type: Array { base: [1], size: Constant([7]), stride: 4 }
```


PolyMeilex:
My bad, it's a duplicate of #887 and #830

PR #893, opened 2021-05-27
Allow disabling drag and drop on windows

traxys:
I am using `rodio` and so `cpal` in an `iced` application, that has the unfortunate effect of not working at all because of an interraction between `cpal` and `winit` (see  RustAudio/cpal#504). There is a fix available in `winit` by disabling drag and drop: rust-windowing/winit#1524, this PR adds this switch to `iced`. 
I don't know much about iced internals, so I don't know if this change breaks anything


hecrj (approved):
Looks reasonable! Thank you 🥂

PR #907, opened 2021-06-08
Selection Scroll List

genusistimelord:
I have Added a Selection Scroll List. Please let me know if there is anything i can do better to improve upon this add to allow it to get submitted into the System. I made a Change the Menu's Overlay so it could be Set as float-able or static positioning.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,7 @@ members = [
     "examples/progress_bar",
     "examples/qr_code",
     "examples/scrollable",
+    "examples/selection_list",
     "examples/solar_system",
     "examples/stopwatch",
     "examples/styling",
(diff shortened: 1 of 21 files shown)


13r0ck:
What is the point of this, how is it different than the pick list?

genusistimelord:
It is just a different View of a list generally found in Other GUI's. Rather than being a Drop down select only. It could be further expanded later to support a more dynamic approach to adding, sorting and removing the internals. This could also have drag and drop added to it for sorting elements or inserting removing elements from one list to another.

13r0ck:
Oh gotcha, for widgets that can be expanded upon further we like to have them committed to [iced_aw](https://github.com/iced-rs/iced_aw) first.

Could you do that please. Then as the scroll list gets more features it can be moved over into iced

genusistimelord:
Ahh, yes ill close this and add it there.

13r0ck:
Thanks!

genusistimelord:
the only thing is without the change made to the Overlay for Menu my thing wont be able to render properly as the Menu likes to change its position which will mess up the rendering. So ill have to remake menu in the widget OR push just that change here which is adding a Floating: bool to it to allow it to switch between floating and static rendering.

PR #914, opened 2021-06-14
feat: expose draw_cache_multithread as a feature

yusdacra:


hecrj (approved):
Thank you!

Instead of using a feature flag, I have made some changes to provide the flag at runtime as part of the `Settings` struct.

PR #916, opened 2021-06-14
refactor: make clippy almost silent, some optimizations

yusdacra:
I was annoyed by the many clippy warnings so I tried to make it shut up as much as possible.

This PR also reduces allocations in `text_input`. It adds one new dependency, [smol_str](https://lib.rs/crates/smol_str).

Commit 9be7a368030c11560fd8de6bd2d7e6fda8cc2128 changes some public API (especially `Toggler`) to make them take `Into<SmolStr>` instead of `Into<String>`. This causes some minimal breakages (easily fixable, see tour example changes + i think it looks nicer), but if it's not okay I can leave it out.

### DIFF
--- a/native/src/overlay/menu.rs
+++ b/native/src/overlay/menu.rs
@@ -158,9 +158,9 @@ where
 
         Self {
             container,
-            width: width,
+            width,
             target_height,
-            style: style,
+            style,
         }
     }
 }
@@ -224,7 +224,7 @@ where
         messages: &mut Vec<Message>,
     ) -> event::Status {
         self.container.on_event(
-            event.clone(),
+            event,
             layout,
             cursor_position,
             renderer,
@@ -289,13 +289,15 @@ where
         use std::f32;
 
         let limits = limits.width(Length::Fill).height(Length::Shrink);
-        let text_size = self.text_size.unwrap_or(renderer.default_size());
 
         let size = {
+            #[allow(clippy::or_fun_call)]
             let intrinsic = Size::new(
                 0.0,
-                f32::from(text_size + self.padding.vertical())
-                    * self.options.len() as f32,
+                f32::from(
+                    self.text_size.unwrap_or(renderer.default_size())
+                        + self.padding.vertical(),
+                ) * self.options.len() as f32,
             );
 
             limits.resolve(intrinsic)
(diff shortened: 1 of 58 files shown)


native/src/overlay/menu.rs
```diff
@@ -289,13 +289,15 @@ where
...
-        let text_size = self.text_size.unwrap_or(renderer.default_size());
 
         let size = {
+            #[allow(clippy::or_fun_call)]
             let intrinsic = Size::new(
                 0.0,
-                f32::from(text_size + self.padding.vertical())
-                    * self.options.len() as f32,
+                f32::from(
+                    self.text_size.unwrap_or(renderer.default_size())
```
pleshevskiy, line 298: `self.text_size.unwrap_or_else(|| renderer.default_size())` better.
A little memory optimization here and not have to make an exclusion for the linter.

I would even make a separate structure with Default and Deref into u16. Then you could do `.unwrap_or_default()`.

Especially since such an expression is very often used in the code.

native/src/widget/text_input/value.rs
```diff
@@ -58,9 +61,10 @@ impl Value {
     pub fn next_end_of_word(&self, index: usize) -> usize {
         let next_string = &self.graphemes[index..].concat();
 
-        UnicodeSegmentation::split_word_bound_indices(&next_string as &str)
-            .filter(|(_, word)| !word.trim_start().is_empty())
-            .next()
+        #[allow(clippy::or_fun_call)]
+        // self.len() is very cheap, we don't need lazy eval
```
pleshevskiy, line 65: But why allocate memory every time it is not used?

pleshevskiy (changes requested):

native/src/widget/text_input/value.rs
yusdacra: `self.len()` doesn't allocate any memory; it returns a usize which is stored on the stack. It will be stored on the stack even if we put it in a closure; so the closure is needless indirection.

native/src/overlay/menu.rs
yusdacra: This also returns a u16 which is stack allocated. Since it will be allocated in the stack we don't need lazy eval, and it's very cheap anyways (ideally `default_size` would be marked as `const`, but trait const fn's aren't finished yet), so we don't need the indirection.

yusdacra: On second thought; Rust should figure out and just inline the value, bypassing the lazyness.

native/src/widget/text_input/value.rs
yusdacra: Same goes for here; Rust should inline the value.

hecrj:
Superceded by #1379.

PR #917, opened 2021-06-14
Enable receiving URLs on MacOS

derezzedex:
This PR uses the newly created `ReceivedUrl` to allow receiving URLs from the system on MacOS.  For this to work the executable needs to be properly bundled (as stated on the documentation of the event). It also adds a simple example showcasing the event.

Note: This PR makes use of the new organization repositories ([iced-rs/winit](https://github.com/iced-rs/winit) and [iced-rs/glutin](https://github.com/iced-rs/glutin)).

Depends on the merging of https://github.com/iced-rs/winit/pull/1.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,9 +92,6 @@ iced_core = { version = "0.4", path = "core" }
 iced_futures = { version = "0.3", path = "futures" }
 thiserror = "1.0"
 
-[patch.crates-io]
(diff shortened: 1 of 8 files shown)


hecrj (approved):
Thank you! :+1: 

I have updated the `winit` and `glutin` dependencies to properly point to specific commits of the forks in the `iced-rs` organization.

We should consider contributing back the URL changes to `winit` or, at the very least, start a related discussion.

I believe we can ship this! :ship:

PR #941, opened 2021-07-05
add text Input filter

Cupnfish:
You can filter your input like this：
```rust
                let input = TextInput::new(
                    input,
                    "What needs to be done?",
                    input_value,
                    Message::InputChanged,
                )
                .input_filter(|c, value| {
                    let value = value.to_string();
                    value.chars().last().map(|last| last != c).unwrap_or(true)
                })
                .padding(15)
                .size(30)
                .on_submit(Message::CreateTask);
```


Cupnfish:
I found that the same effect could be achieved using the existing update method 😂

PR #945, opened 2021-07-06
feat: add menus

derezzedex:
This PR does a couple of things to add menus for Windows & MacOS.
- Introduces a menu struct that should be used by shells to implement a menu (native or 'emulated').
- Implements the menu for `iced_winit` (and so `iced_glutin`) for Windows & MacOS.
- Creates a 'copy' of the `Hotkey` struct found on `winit` to better map key combinations.

Relies on https://github.com/iced-rs/winit/pull/3, closes https://github.com/hecrj/iced/issues/114 and https://github.com/hecrj/iced/issues/704.


Cargo.toml
```diff
@@ -85,13 +85,17 @@ members = [
...
+    "examples/menu",
 ]
 
 [dependencies]
 iced_core = { version = "0.4", path = "core" }
 iced_futures = { version = "0.3", path = "futures" }
 thiserror = "1.0"
 
+[patch.crates-io]
+winit = { git = "https://github.com/iced-rs/winit", rev = "e351421a32bf01b428325dde44dea39ee2656153"}
```
derezzedex, line 97: Use a patch so that we don't need to modify `iced-rs/glutin` while developing this branch, should be removed before merging.

examples/menu/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+            control: false,
+            logo: false,
+            shift: false,
+        };
+        let ctrl_shift = Modifiers {
+            control: true,
+            shift: true,
+            logo: false,
+            alt: false,
+        };
```
derezzedex, line 56: This is somewhat awkward to work with, maybe we could use something like: `Modifiers::control().and(Modifiers::shift())`?

native/src/menu.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub fn new() -> Self {
+        Menu { items: Vec::new() }
+    }
+
+    /// Adds an item to the [`Menu`].
+    pub fn item<S: Into<String>>(
+        mut self,
+        content: S,
+        hotkey: impl Into<Option<Hotkey>>,
+        on_activation: Message,
```
derezzedex, line 23: Currently there's no implementation of the menu activation because we don't want to add `Message: Clone` to `Application`, maybe we should use `fn() -> Message` (even though the user would have to do `Menu::item(.., || Message::Example)`)?

native/src/menu.rs
```diff
@@ -0,0 +1,81 @@
...
+        submenu: Menu<Message>,
+    ) -> Self {
+        let content = content.into();
+
+        self.items.push(MenuEntry::Dropdown { content, submenu });
+        self
+    }
+
+    /// Returns a [`MenuEntry`] iterator.
+    pub fn iter(self) -> std::vec::IntoIter<MenuEntry<Message>> {
```
derezzedex, line 55: Ideally I don't want to expose this method and `MenuEntry`, allowing the end-user to only use our constructor methods (`new`, `item`, `dropdown`, `separator`), but then how would the shell implementations `iter` through the `Menu` to actually build it? Maybe there's a better solution here?

winit/src/application/state.rs
```diff
@@ -36,6 +36,8 @@ impl<A: Application> State<A> {
             )
         };
 
+        window.set_menu(Some(conversion::menu(application.menu())));
```
derezzedex, line 39: At this point we want to store the `Menu` in the `State`, should the `State` hold a map for the `Message`s produced by the `Menu` or should the `Menu` do it, how do we actually produce those without using `Clone`?

examples/menu/src/main.rs
hecrj: I have changed `Modifiers` to use `bitflags` like `winit` does now in b57d567981bb7ef5f9ff397c210778f211d2de5b.

It should be a bit more ergonomic.

native/src/menu.rs
hecrj: I just went ahead and added the `Clone` bound to `Message` in 5df2a92f28be1d53d32d5b42a6645459b7d78efe.

I know some users will be bitten by this, but I really think all messages should be cloneable and they should always represent and contain pure data.

native/src/menu.rs
hecrj: I think it's fine for `menu::Entry` to be completely public. I don't see what kinds of guarantees we would be capturing with opacity here. The methods can be seen as just convenient helpers with some generics.

hecrj (commented):
I made some changes to bring this to the finish line.

The `menu` example should now work as expected for both Windows and macOS. I tried to make the commits self-contained and somewhat descriptive. Let me know what you think!

However, before merging, I think we may need to extend the `Menu` API a bit more to allow for better composition. Specifically, I believe `Menu` should implement a `map` operation like `Element::map`, and only apply it when a `Message` is produced.

In order to achieve that, I believe we will need to use trait objects and allocate a `Box` for each menu entry... The upside of it is that we may be able to remove the `Clone` bound for `Message` at the `Application` level.

derezzedex (commented):
> I made some changes to bring this to the finish line.
> 
> The `menu` example should now work as expected for both Windows and macOS. I tried to make the commits self-contained and somewhat descriptive. Let me know what you think!

Thank you for the help! Looks awesome!
I still want to improve the menu example though, because right now it doesn't show that we can create context-dependent menus, where the runtime rebuilds it when necessary.

>However, before merging, I think we may need to extend the Menu API a bit more to allow for better composition. Specifically, I believe Menu should implement a map operation like Element::map, and only apply it when a Message is produced.

You mean mapping `Event::MenuEntryActivated(usize)` to the provided message?

> In order to achieve that, I believe we will need to use trait objects and allocate a Box for each menu entry... The upside of it is that we may be able to remove the Clone bound for Message at the Application level.

I hadn't thought of that, I'll look into it.

hecrj:
@derezzedex

> I still want to improve the menu example though, because right now it doesn't show that we can create context-dependent menus, where the runtime rebuilds it when necessary.

Maybe instead of adding a `menu` example, it'd be easier to extend some of the existing examples.

> You mean mapping Event::MenuEntryActivated(usize) to the provided message?

Not really. This is already implemented in `conversion::menu_message`, which recursively iterates and finds the proper entry based on the id provided by `winit`.

What I meant here is a `Menu::map` method, which would allow users to turn a `Menu<A>` to a `Menu<B>` given a function `Fn(A) -> B`. This way, `Menu` becomes very easy to compose, as you can have `Menu` instances at different abstraction levels and plug them together. `Element::map` works exactly in the same way.

winit/src/conversion.rs
```diff
@@ -156,6 +157,119 @@ pub fn visible(mode: Mode) -> bool {
...
+                    id += 1;
+                }
+                menu::Entry::Dropdown { submenu, .. } => {
+                    match find_message(target, id, submenu) {
+                        Ok(message) => {
+                            return Ok(message);
+                        }
+                        Err(n_children) => {
+                            id += n_children;
+                        }
```
clarkmoody, line 255: While this pattern of "send in IDs -> get out ID and find it" is not my favorite, it seems simple enough to follow the logic. I would prefer there be some sort of `EmplacedEntry` or something that has this ID auto-enumerated by the generating function. But that adds context and state (not to mention winit IDs) out into the Iced layer, which is less desirable than a little bit of ID math.

`/ ramble`

clarkmoody (changes requested):
Seems like this is ready to ship. API looks fine, and the example works for me on Mac.

Requesting a couple docstrings be filled out before merge.

src/application.rs
```diff
@@ -191,6 +191,11 @@ pub trait Application: Sized {
         false
     }
 
+    /// TODO
```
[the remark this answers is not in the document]
hecrj: Done in c8ac77e4e99414746adedf38cf69ac8dcd1601a4.

winit/src/application.rs
```diff
@@ -98,6 +99,11 @@ pub trait Application: Program<Clipboard = Clipboard> {
     fn should_exit(&self) -> bool {
         false
     }
+
+    /// TODO
```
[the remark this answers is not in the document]
hecrj: Done in c8ac77e4e99414746adedf38cf69ac8dcd1601a4.

hecrj (approved):
I have implemented `Menu::map` (somewhat naively, but I don't want to optimize prematurely) in 3099f3610003f513a386125d3cb81bfbf0ffe887.

Also, I have introduced a presets `Menu` in the `game_of_life` example to showcase how composition can be achieved with it in b97954a1ee3ec7bc85d1d41b397e994752ff1831.

I think this should be ready to go 🎉

Issue #841, opened 2021-04-21
Random memory leak + crash with SVGs


Jammyjamjamman:
Iced 0.3.0 sometimes freezes + memory leaks on Linux when rendering a Container with SVGs + other widgets.

From debugging. I narrowed down the bug to an infinite loop happening here: https://github.com/RazrFalcon/tiny-skia/blob/2e6aa247308867f2715399d12dc2afb454b925c7/src/dash.rs#L429-L446 .

Here's steps which will reproduce it on a project I'm working on (Caution: may crash your computer):

1. Clone ~https://github.com/TelluricDeckay/telluricdeckay~ TelluricDeckay/telluricdeckay@4d56dcc (Apologies for my horrible code xD.)
2. Inside the repo, copy config_h.rs.in' to config_h.rs.
3. cargo run
4. Select "start" -> "5 player"
5. If the window freezes, you've reproduced the bug. If the page loads, close the program and repeat steps 3 and 4 until it does freeze. (Usually freezes after up to 5 goes. But it can take more than that.)

Removing SVGs stops the freezes.

andy5995:
I'm the co-maintainer of the project @Jammyjamjamman mentioned above (Telluric Deckay).

The last time I ran it, all my RAM was used, then my swap file. My computer slowed to a crawl becoming unusable (unable to kill the app or switch to a terminal). I was running the app through gdb and waited for the the app to crash.

![image](https://user-images.githubusercontent.com/16764864/115625545-d3290200-a2c1-11eb-970a-21d6057b776d.png)

I was able to kill the app at that point.

Me and Jammy are both on the iced Zulip server, so if you'd like to discuss this a little there for more info, just give us a ping.

andy5995:
> Here's steps which will reproduce it on a project I'm working on (Caution: may crash your computer):
> 
>     1. Clone https://github.com/TelluricDeckay/telluricdeckay (Apologies for my horrible code xD.)
> 

The trunk (master) branch now uses pngs not svgs, so anyone wanting to try to reproduce should clone using the commit before the change

https://github.com/TelluricDeckay/telluricdeckay/commit/4d56dccdd7f63f4aa8b301944ee75487eddd4c3b

hecrj:
I think this may be related to #786 and #820.

aentity:
I have also encountered this, and after much investigation, I think I have found source (but do not understand exact cause).

[...]

PR #952, opened 2021-07-14
Use ceil on svg dimensions, fix svg memory usage ref #841

aentity:
Calls ceil() on dimension bounds as this appears fix svg memory
unbounded usage because no longer cache miss.

The height and width return from resvg seem to always be ceiling
of float dimensions, so we try to match.

### DIFF
--- a/wgpu/src/image/vector.rs
+++ b/wgpu/src/image/vector.rs
@@ -75,8 +75,8 @@ impl Cache {
         let id = handle.id();
 
         let (width, height) = (
-            (scale * width).round() as u32,
-            (scale * height).round() as u32,
+            (scale * width).ceil() as u32,
+            (scale * height).ceil() as u32,
         );
 
         // TODO: Optimize!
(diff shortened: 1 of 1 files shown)


wgpu/src/image/vector.rs
```diff
@@ -122,6 +122,7 @@ impl Cache {
                     device,
                     encoder,
                 )?;
+                log::warn!("allocating {} {}x{}", id, width, height);
```
hecrj, line 125: I think this should be `debug` or `trace` level.

hecrj (changes requested):

wgpu/src/image/vector.rs
aentity: ok i changed to debug. i do think warn is useful to see how much allocating, since it should really be never once application is running, but i do see it reallocating if for example i change svg icon / button depending on some state, but it's ok for me to make debug

hecrj (approved):
Thank you! 🙇

PR #965, opened 2021-07-26
Add window::Event::Moved

tarkah:
Creates a new `window::Event` when the window is moved. These coordinates can be persisted and used with the new window settings `Position::Specific` to restore an application to it's previous location.


hecrj (approved):
Awesome! Thank you 🎉

PR #966, opened 2021-07-27
Allow `PickList` selection with mouse scroll only if command is pressed

hecrj:
This PR changes the scrolling behavior implemented in #872 to only trigger when the "command" keyboard modifier is active.

This should avoid unintended interactions when using a `PickList` inside a `Scrollable`.


13r0ck:
In my case it was better to always change values on scroll
I see there usefull implementations, 
1. Default: Scroll requires command
2. Always scroll
3. Never scroll

Would you like a PR that makes this setting toggleable?
If so, how would you like for me to implement it?

PR #971, opened 2021-07-27
fix: allow titlebar overlays to close when dragging pane

tarkah:
Currently, if you have a widget that produces an overlay in the `TitleBar` and you try to drag that pane, the overlay won't close (due to the button press outside the overlay bounds not getting processed by the overlay) and will "hang" in place while the pane is dragged.

This fixes the `pane_grid` widget by allowing `TitleBar` content to process events when "dragging" the pane.

When a pane is being dragged, only the `TitleBar` will get to process events, not the pane body.

### DIFF
--- a/native/src/widget/pane_grid.rs
+++ b/native/src/widget/pane_grid.rs
@@ -452,24 +452,20 @@ where
             _ => {}
         }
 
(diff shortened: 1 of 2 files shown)


hecrj (approved):
Thanks! 🙇

Issue #891, opened 2021-05-24
 Android Tracking Issue


Dispersia:
As a follow up to https://github.com/hecrj/iced/issues/302, but for android specifically, this is a tracking issue to address the issues highlighted in my inital post. Now that https://github.com/hecrj/iced/pull/830 is merged can start working on these individually.
 
 Required to run on android:
- [ ] Move off of Bgra8UnormSrgb, should be requesting swapchains preferred format
- [ ] Handle Suspend/Resumed events from winit
- [ ] Redraw Requested is not implemented on android, temporarily just needs to draw on the spot until https://github.com/rust-windowing/winit/pull/1822 is completed.

 Better Experience:
- [ ] Font loading doesn't working, as font-kit relies on things that don't exist on android
- [ ] Sizing is quite a bit messed up, not sure if iced currently handles different dpi's, but need to look into that and implement that here.

 Will update with prs as I work through it.

ColonelThirtyTwo:
Looks like winit has the redraw request stuff, via https://github.com/rust-windowing/winit/pull/1953

hecrj:
As I mentioned in https://github.com/iced-rs/iced/pull/631#issuecomment-897369162, this is out of scope for now!

PR #978, opened 2021-08-01
wgpu: Use the preferred texture format of the surface

ATiltedTree:
Use the preferred texture format of the surface instead of `Bgra8UnormSrgb`.

Related to: #891

### DIFF
--- a/examples/integration/src/main.rs
+++ b/examples/integration/src/main.rs
@@ -34,7 +34,7 @@ pub fn main() {
     let instance = wgpu::Instance::new(wgpu::BackendBit::PRIMARY);
     let surface = unsafe { instance.create_surface(&window) };
 
-    let (mut device, queue) = futures::executor::block_on(async {
+    let (format, (mut device, queue)) = futures::executor::block_on(async {
         let adapter = instance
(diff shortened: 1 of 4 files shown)


wgpu/src/image/atlas.rs
```diff
@@ -341,7 +343,7 @@ impl Atlas {
             mip_level_count: 1,
             sample_count: 1,
             dimension: wgpu::TextureDimension::D2,
-            format: wgpu::TextureFormat::Bgra8UnormSrgb,
+            format: self.format,
```
hecrj, line 346: I don't think we can change the `format` of the `Atlas` easily like this because we are uploading the image data in BGRA format.

hecrj (changes requested):
Thanks for giving this a shot! I think this should improve compatibility considerably 🎉 

There is just one concern I have. Let me know what you think!

wgpu/src/image/atlas.rs
ATiltedTree: I didn't think about that. Just checked with `Rgba8UnormSrgb` which turned ferris blue and thus reverted those changes.

hecrj (approved):
Looks great! Thank you 🙇

PR #980, opened 2021-08-02
Fix glow and multiple windows usage

yamadapc:
Hey there 👋 

I'm writing a VST plugin host, which might have another `iced` app as a guest.

In that case, when the guest window opens, the glow context might not be current when redrawing, causing the UI to be frozen.

This commit adds a check prior to redrawing for whether the context is current. It forces the context to become current if it's not. This fixes hosting issues with multiple windows for me. I suppose it might be a necessary change if `iced` itself had more than one window open at a time.

This was not a problem using the wgpu back-end.


hecrj (approved):
Looks reasonable.

I am interested in learning more about running `iced` as a guest application. How are you doing that? The `run` method takes control of the main thread and does not return.

yamadapc:
The guest is using https://github.com/BillyDM/iced_baseview/

That library is changing how `iced` runs so that it can be embedded.

PR #988, opened 2021-08-09
Allow canvas' "Frame"s to be retained

yamadapc:
This allows users to improve performance of canvas by retaining the
`Frame` struct between view updates. This avoids
allocations/deallocations that happen internally in lyon. See
https://github.com/nical/lyon/pull/683.

Full example usage: https://github.com/yamadapc/augmented-audio/blob/master/crates/plugin-host-gui2/src/ui/main_content_view/volume_meter/mod.rs

Minimal (non-working) example:
```rust
struct Visualisation {
    frame: RefCell<Frame>, // <- RefCell is to hack around the current `draw(&self, ...)` API (no mutable ref. to self)
}

impl Visualisation {
    fn new() -> Self {
        Self {
            // The size does not matter at this point
            frame: RefCell::new(Frame::new(Size::new(100., 100.))),
        }
    }

   fn view(&mut self) -> Element<Message> {
        Canvas::new(self).width(Length::Fill).height(Length::Fill)
            .into()
    }
}

impl Program<Message> for Visualisation {
    fn draw(&self, bounds: Rectangle, _cursor: Cursor) -> Vec<Geometry> {
        let mut frame = self.frame.borrow_mut();
        // Because frame is retained internal vecs/structures already have proper capacity
        frame.resize(bounds.size());

        // Draw on the frame

        geometry.push(frame.geometry()); // `Frame::geometry` doesn't consume the Frame
        geometry
    }
}
```

~~I should update the other tessellator usages as well, since they probably would also benefit~~. Potentially it'd be good that https://github.com/hecrj/iced/pull/683 is merged first.

### DIFF
--- a/graphics/src/widget/canvas/frame.rs
+++ b/graphics/src/widget/canvas/frame.rs
@@ -4,16 +4,18 @@ use crate::{
     canvas::{Fill, Geometry, Path, Stroke, Text},
     triangle, Primitive,
 };
+use lyon::tessellation::FillTessellator;
 
 /// The frame of a [`Canvas`].
 ///
 /// [`Canvas`]: crate::widget::Canvas
-#[derive(Debug)]
+#[allow(missing_debug_implementations)]
 pub struct Frame {
     size: Size,
     buffers: lyon::tessellation::VertexBuffers<triangle::Vertex2D, u32>,
     primitives: Vec<Primitive>,
     transforms: Transforms,
+    fill_tessellator: FillTessellator,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


graphics/src/widget/canvas/frame.rs
```diff
@@ -262,6 +268,30 @@ impl Frame {
...
+    /// primitives.
+    pub fn geometry(&mut self) -> Geometry {
+        if !self.buffers.indices.is_empty() {
+            self.primitives.push(Primitive::Mesh2D {
+                buffers: triangle::Mesh2D {
+                    vertices: self.buffers.vertices.clone(),
+                    indices: self.buffers.indices.clone(),
+                },
+                size: self.size,
+            });
```
hecrj, line 281: Cloning all the vertices and indices just to avoid reallocating the tessellators feels counter-productive.

hecrj (commented):
I have implemented a middle ground solution in #683.

Let's keep the discussion going there!

PR #993, opened 2021-08-11
feat(pick_list): seperate menu width, text font & color

### LINKED ISSUE #918: Add ability to set width of picklist widget and it's child dropdown/overlay independently
I'm not sure how feasible this is, but I think this would provide slightly nicer styling and allow the insertion of a much wider array of elements in the picklist. Alternatively, add automatic line breaks to elements in the dropdown, if possible, when the overall width of the widget is less than a given element.

An example of my current situation:
image
What I'd prefer:
image
What actually results:
image

yusdacra:
This fixes #918, and also adds the ability to change menu entries' font, text size and menu width seperately.

### DIFF
--- a/native/src/widget/pick_list.rs
+++ b/native/src/widget/pick_list.rs
@@ -32,7 +32,9 @@ where
     width: Length,
     padding: Padding,
     text_size: Option<u16>,
+    menu_text_size: Option<u16>,
     font: Renderer::Font,
+    menu_font: Option<Renderer::Font>,
     style: <Renderer as self::Renderer>::Style,
 }
 
@@ -93,8 +95,10 @@ where
             selected,
             width: Length::Shrink,
             text_size: None,
+            menu_text_size: None,
             padding: Renderer::DEFAULT_PADDING,
             font: Default::default(),
+            menu_font: None,
             style: Default::default(),
         }
     }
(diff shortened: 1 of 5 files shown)


native/src/widget/pick_list.rs
```diff
@@ -168,50 +184,63 @@ where
...
+        let max_width = {
+            let labels = self.options.iter().map(ToString::to_string);
+
+            let labels_width = labels
+                .map(|label| measure(&label, menu_text_size, menu_font))
+                .max()
+                .unwrap_or(100);
+
+            labels_width.max(placeholder_width)
+        };
```
hecrj, line 227: This change will make `layout` hit the worst case and measure all the labels independently of the `PickList` configuration. Thus, a `PickList` with thousands of labels would cause a considerable slow down.

Instead, we should probably measure the labels when opening the `Menu` while also making sure the bounds don't overflow the screen edges.

Also, maybe we should keep the current behavior as the default one and make the new one opt-in with an additional configuration enum?

hecrj (changes requested):

native/src/widget/pick_list.rs
yusdacra: Yeah, this should probably be opt-in behaviour. As for measuring the labels when opening the `Menu`, that should be done in `Menu` layout itself then I guess?

hecrj: Yes, but only when strictly necessary (i.e. when the `PickList` does not provide a predefined `width` for the `Menu`).

hecrj:
I think it may be better to retackle this from the ground up since the codebase has changed considerably since this PR was originally opened.

PR #994, opened 2021-08-12
tweak default rule style

BillyDM:
(Reuploaded from https://github.com/hecrj/iced/pull/666)

I tweaked the Rule style to have just a bit more contrast. It also makes more sense to have the rule span its full width by default.


hecrj (approved):
Thank you 🎉

PR #997, opened 2021-08-13
Added reexport for iced_web

madmaxio:
This allows to create custom widgets via using dodrio, and other stuff.


hecrj (commented):
We want to avoid exposing implementation details in the root crate: `iced`.

If users want to implement custom widgets, they should explicitly add `iced_native` or `iced_web` to their dependencies.

Issue #903, opened 2021-06-02
Proposals for custom shaders/rendering


mvilim:
Hi! Thanks for the wonderful UI library.

I would like to add support for custom rendering of a widget (i.e. not just a custom widget made of existing primitives, but a widget that uses a custom shader to draw itself on the screen). I have experimented with a couple ways to achieve this, but would appreciate some feedback on the potential approaches before I commit to fully implementing any of them, as you may have an opinion on which is the best fit for this project.

The only discussion of custom rendering that I have been able to find is here: https://github.com/hecrj/iced/issues/32#issuecomment-588637930 but perhaps I've missed some discussion elsewhere.

I plan to do this custom rendering using `wgpu` (and therefore `iced_wgpu`), but as you will see, some of the changes would reach into the lower level crates.

The general approaches I have been able to come up with, in order of most complete/invasive to least are:

## 1. Genericize `iced_graphics::Primitive`

Genericize `iced_graphics::Primitive` with a type (used for a new `Custom` primitive variant) which can be specified by the downstream application in its creation of a custom `iced_graphics::Renderer`. When no custom rendering is required, this type would be set to the unit type. This would have a very far reaching impact on iced, as the generic parameter would cascade into pretty much everything in `iced_graphics` and downstream (as well as probably breaking (in a relatively shallow way) existing custom widgets). It's possible some of the breakage could be minimized by extracting a lower-level type e.g. `GenericPrimitive<CustomType>` and leaving `Primitive` as is. The internal APIs could would be then be written in terms of `GenericPrimitive<CustomType>` but the external API would use the non-generic `Primitive` (which would map to `GenericPrimitive<()>`).

The `wgpu::Backend` would then have a generic `CustomPipeline` (with would have `draw` function which takes a command encoder, output target, staging belt, etc.) which it would invoke with the `Custom` primitives. Alternatively, a completely separate backend could be implemented that delegates to `wgpu::Backend` (though likely that would require making more of the `wgpu::Backend` public to allow the custom primitives to be drawn in the correct layers.

## 2. Genericize via `Renderer::Output`

[...]

Issue #343, opened 2020-05-14
Implement a 3D widget


Yamakaky:
The current Canvas widget supports only 2D graphics. It would be cool to have a 3D counterpart, maybe limited to wgpu at first. I see it basically as a boilerplate-free version of https://github.com/hecrj/iced/blob/0.1/examples/integration/src/main.rs.
My personal use-case is to have a window with a 3D view on one side (only when the option is enabled) and the other side with configuration options.

hecrj:
Yes, there has been some discussion about this in #32.

codec-abc:
I would be also interested in the ability to 3D (in order to render a point cloud inside a GUI app). Is 3D rendering somewhat doable for people with little experience in Iced? #32 seems to indicate that it might be feasible if someone dare to implement a 3D Canvas for him/herself but this seems out of reach for me.

dhardy:
If you want WGPU support in a Rust-native GUI, this is (somewhat) possible in KAS: see the [mandlebrot example](https://github.com/kas-gui/kas/blob/master/kas-wgpu/examples/mandlebrot.rs). I'd be interested in getting some feedback on this (sorry @hecrj to jump in here)!

codec-abc:
@dhardy: I try it really quick but I was unable to build it. This is only because of my weird setup (a Ubuntu 16.04 without sudo group and missing dependencies (for example ``cmake``)). I also tried to cross compile it using mingw into a docker container but ran into issues. It seems like the C++ code need a C++ 11 compiler and mingw doesn't support it. Anyway, KAS looks great and I will give it another shot but if somehow KAS could be pure Rust it would make things somewhat easier in my case.

dhardy:
I'm a little surprised, the non-Rust dependencies are basically the same as Iced. Open an issue over on the KAS repo and let me know what fails. You could also try `--no-default-features`.

NicTanghe:
Hello,
I`m just here to make sure that the format used is USD and not old ones.
https://github.com/PixarAnimationStudios/USD

THis does mean a MaterialX /hydra deligate should be provided.
https://github.com/AcademySoftwareFoundation/MaterialX

IT`s also always a good idea in anyhing that outputs color to implement. (best wait until 2.0 is released  ? )
ACES:
https://github.com/AcademySoftwareFoundation/OpenColorIO-Config-ACES


the entire industry is switching over and gltf is basically depricated.
this does mean a materialX renderer is supposed to be implemented.

obj ects are just intermediates these days.

[...]

PR #1003, opened 2021-08-20
Add wgpu custom rendering example

mvilim:
# wgpu custom rendered widget example

Hi! I am returning to #903, and I realized that the existing `layout` information is sufficient to appropriately locate a custom render pass with the rest of the UI. I have put together this PR with an example that demonstrates a custom rendered widget.

This solution is effectively option 4 from #903; it's just that necessary information was already mostly exposed. It's a bit hacky, since it captures the contents of the `layout` argument to the widget `draw` call. It's only possible in projects that directly integrate with `wgpu` (such as the `integration_wgpu` example, from which my example is derived), and such projects already want to have fine control over their rendering.

Despite its slight hackiness, this seems like a perfectly sufficient solution for my use case. Do you think this example is worth including in the repo? If the overall approach is appropriate, but the particular method of extracting the `layout` from the `draw` call is too hacky, then perhaps I can instead make some extensions to `iced` so that the `layout` of a given widget can be queried.


examples/custom_rendering_wgpu/src/controls.rs
```diff
@@ -0,0 +1,368 @@
...
+        &self,
+        _renderer: &mut Renderer<B>,
+        _defaults: &Defaults,
+        layout: Layout<'_>,
+        _cursor_position: Point,
+        _viewport: &Rectangle,
+    ) -> (Primitive, mouse::Interaction) {
+        // We update the current bounds of this custom widget, so that the custom rendering
+        // step will render it in the correct spot
+        self.bounds.replace(layout.bounds());
```
mvilim, line 76: Although the full example is pretty long, this is the key point, where the `layout` is extracted to locate the custom render pass.

hecrj:
While using a completely new render pass on top of the GUI can work for most cases, I don't think we should officially recommend it since it can easily break overlays.

For this reason, I think it's best to wait until a proper 3D widget lands (see #343)!

mvilim:
Okay, no problem! Thanks for taking the time to review.

PR #1007, opened 2021-08-23
Fix winit conversion on ios

### LINKED ISSUE #1005: iced winit conversion broken on ios target
example:

```
iced/winit/src/conversion.rs:240:40
240 |                     converted.add_item(id, title, hotkey.map(self::hotkey));
    |                                        ^^
    |                                        |
    |                                        expected `usize`, found `u16`
    |                                        help: you can convert a `u16` to a `usize`: `id.into()`
```

aentity:
fix #1005


hecrj:
Instead of a patch here, I fixed the API discrepancy in https://github.com/iced-rs/winit :+1:

PR #1019, opened 2021-09-02
Platform-specific `Command` implementations

### LINKED ISSUE #493: Is any possible to resize the whole window whithout outer Event?
How to  dynamically change the size of  a iced app at runtime?

### LINKED ISSUE #644: How to minimal window and resize window
How to minimal window and resize window

### LINKED ISSUE #767: How can you resize the application at runtime?
I am trying to grow/shrink the window whenever the content changes. Is this currently possible? Also I tried to look through the source code to fix #756, but I can't find where you are actually rendering anything? Any pointers?

hecrj:
This PR introduces a platform-specific `Action` as an underlying implementation of `Command` to both `iced_native` and `iced_web`.

This allows us to remove the `Clipboard` from `Application::update` and keep it side-effect free (i.e. pure). Additionally, this should let us implement further actions to let users query and modify the shell environment (e.g. window, clipboard, and more!) without introducing more arguments to `Application::update`, but instead leveraging the `Command` API.

As a consequence, the clipboard should be accessed by using the new `read` and `write` functions in the new `clipboard` module.

Furthermore, I have added actions for interacting with the application window at runtime. The new `move_to` and `resize` functions in the `window` module can be leveraged for this purpose.

More actions coming soon!

Fixes #493.
Fixes #644.
Fixes #767.
Closes #930.


native/src/command/action.rs
```diff
@@ -0,0 +1,24 @@
+use crate::clipboard;
+use crate::window;
+
+pub enum Action<T> {
+    Future(iced_futures::BoxFuture<T>),
```
alex13sh, line 5: for the future, I suggest adding `Action::Stream(futures:: stream:: Box Stream<T>)` for animation and for download_progress

alex13sh: but what if I want to use tokio's async-stream? async-stream generates a state structure. how to hash them?

it seems to me that it would be easier to create streams this way
```
Command::stream(stream! {
... 
yield Message::Progress(..) ;
})
```

hecrj: Commands cannot be canceled while subscriptions can.

I know it'd be more convenient, but convenience has a cost in this case.

PR #1020, opened 2021-09-02
Fix window holes behind transparent objects

### LINKED ISSUE #1018: Glow canvas handles transparency differently from wgpu
Here is the moon's shadow that becomes transparent such you can see text from window below.
image

All I did is changed backend to glow:
```
diff --git a/examples/solar_system/Cargo.toml b/examples/solar_system/Cargo.toml
index 327fe0aa..bd355c9a 100644
--- a/examples/solar_system/Cargo.toml
+++ b/examples/solar_system/Cargo.toml
@@ -6,5 +6,5 @@ edition = "2018"
 publish = false
 
 [dependencies]
-iced = { path = "../..", features = ["canvas", "tokio", "debug"] }
+iced = { path = "../..", features = ["glow", "glow_canvas", "tokio", "debug"] }
 rand = "0.8.3"

```

akhilman:
Without this change, the transparent objects composed over the opaque
background creates the transparency of the entire window through which
the desktop wallpaper could be seen.
This problem manifested itself with glow backend and wayland compositor
(gnome).

This patch completely eliminates this problem. With it, the final image
produced by glow backend corresponds to the image produced by the wgpu
backend.

Fixes #1018

### DIFF
--- a/glow/src/window/compositor.rs
+++ b/glow/src/window/compositor.rs
@@ -26,7 +26,7 @@ impl iced_graphics::window::GLCompositor for Compositor {
 
         // Enable alpha blending
         gl.enable(glow::BLEND);
-        gl.blend_func(glow::SRC_ALPHA, glow::ONE_MINUS_SRC_ALPHA);
+        gl.blend_func(glow::DST_ALPHA, glow::ONE_MINUS_SRC_ALPHA);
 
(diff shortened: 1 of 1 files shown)


akhilman:
This is not correct solution. I'll continue to dig in.

PR #1021, opened 2021-09-02
Fix window holes behind transparent objects

### LINKED ISSUE #1018: Glow canvas handles transparency differently from wgpu

akhilman:
Without this change, the transparent objects composed over the opaque
background creates the transparency of the entire window through which
the desktop wallpaper could be seen.
This problem manifested itself with glow backend and wayland compositor
(gnome).

This patch completely eliminates this problem. With it, the final image
produced by glow backend corresponds to the image produced by the wgpu
backend.

Fixes #1018

### DIFF
--- a/glow/src/window/compositor.rs
+++ b/glow/src/window/compositor.rs
@@ -26,7 +26,12 @@ impl iced_graphics::window::GLCompositor for Compositor {
 
         // Enable alpha blending
         gl.enable(glow::BLEND);
-        gl.blend_func(glow::SRC_ALPHA, glow::ONE_MINUS_SRC_ALPHA);
+        gl.blend_func_separate(
+            glow::SRC_ALPHA,
+            glow::ONE_MINUS_SRC_ALPHA,
+            glow::ONE,
+            glow::ONE_MINUS_SRC_ALPHA,
+        );
 
         // Disable multisampling by default
         gl.disable(glow::MULTISAMPLE);


akhilman:
The `glBlendFunc` sets same function for all channels including alpha. Suppose we have `glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA)`, background colour `(1.0, 1.0, 1.0, 1.0)` as `(R. G, B, A)` and foreground colour `(0,5, 0,5, 0,5, 0,5)`. For all channels we will have `0.5 * 0.5 + 1.0 * (1.0 - 0.5)` which equals to `0.75`. So overall composition is transparent and we have hole in the window.

With `glBlendFuncSeparate(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA, GL_ONE, GL_ONE_MINUS_SRC_ALPHA)` function for alpha channel will be `0.5 * 1.0 + 1.0 * (1.0 - 0.5)` which equals to `1.0`, so overall composition is opaque.

hecrj (approved):
Good catch!

I noticed transparent colors not completely matching when comparing backends and couldn't exactly pin-point why. Thank you! :tada:

oknozor:
I just tested this on latest master : 
image
I think this also closes #1021

PR #1044, opened 2021-09-14
Add Align::Fill variant

### LINKED ISSUE #746: How can I flex a collection of elements to be the same width/fill remaining space?
How could I flex these buttons to fill the remaining space (designated by the red square)?

## Code

```rust
Container::new(Column::new()

    .push(

        Container::new(Column::new()
        .align_items(Align::Center)
        .width(Length::Shrink)

        .push(
            Image::new(image::Handle::from_memory(include_bytes!("../assets/gmpublisher.png").to_vec()))
            .width(Length::Units(128))
        )

        .push(
            Text::new(format!("gmpublisher v{} by Billy", env!("CARGO_PKG_VERSION")))
            .horizontal_alignment(HorizontalAlignment::Center)
            .color(Color::WHITE)
        )

        .push(
            Landing::button(star_btn, "Star on GitHub")
            .on_press(Message::GitHubStar)
        )

        .push(
            Landing::button(new_addon_btn, "New Addon")
            .on_press(Message::NewAddon)
        )

        .push(
            Landing::button(open_addon_btn, "Open Addon")
            .on_press(Message::OpenAddon)
        )

        .push(
            Landing::button(my_workshop_btn, "My Workshop")
            .on_press(Message::MyWorkshop)
        )

        .push(
            Landing::button(settings_btn, "Settings")
            .on_press(Message::Settings)
        )

        .push(
            Landing::button(help_btn, "Help")
            .on_press(Message::Help)
        )

        .push(
            Landing::button(close_btn, "Close")
            .on_press(Message::Close)
        )

    )
    .width(Length::Fill)
    .height(Length::Fill)
    .center_x()
    .center_y()

    )

    .push(
        Column::new()
        .align_items(Align::Center)
        .width(Length::Fill)

        .push(
            Text::new(
                match &app_vars.gmod {
                    Some(gmod) => gmod.to_string_lossy().to_string(),
                    None => String::from("Garry's Mod installation directory not found! Add it in Settings.")
                }
            )
            .horizontal_alignment(HorizontalAlignment::Center)
            .color(Color::WHITE)
        )

    )

)

.width(Length::Fill)
.height(Length::Fill)
.center_x()
.center_y()
.padding(20).into()
```

## Result

image

tarkah:
Adds a new `Align::Fill` variant that works similar to `align-items: stretch;`.

Currently, it's impossible to have all child elements of a row or column "fill" the remaining space based on the largest sized child. This makes using widgets like `Rule` impossible when nested inside a row or column that has no fixed size.

Here's an example of how it's possible to use `Rule` in the `pokedex` example, where it's height now stretches to fill the dynamic height of the row it's in. Without `Align::Fill`, the `Rule` expands to fill the entire viewport.

Before:

Screenshot from 2021-09-14 10:00:27

After:

Screenshot from 2021-09-14 10:01:56

Closes #746


native/src/layout/flex.rs
```diff
@@ -124,19 +145,34 @@ where
...
-            let (max_main, max_cross) =
-                axis.pack(max_main, axis.cross(limits.max()));
+            let (max_main, max_cross) = axis.pack(
+                max_main,
+                if align_items == Align::Fill {
+                    cross
+                } else {
+                    axis.cross(limits.max())
+                },
+            );
```
tarkah, line 164: With `Align::Fill`, I think there's an inconsistency in how the `child.layout` is being calculated for children with and without fill factor.

In the first iteration, I get the max `cross` of all non-fill on cross axis children. But the nodes on `fill_factor == 0` that are stored aren't bounded by that max `cross` that's calculated. However, I'm bounding the `fill_factor != 0` nodes by this max `cross`.

I think I need to add a new iteration to calculate `cross` up front, that I can then bound ALL nodes by.

tarkah: Fixed in 7631c70

native/src/layout/flex.rs
```diff
@@ -90,16 +116,37 @@ where
...
-            let child_limits =
-                Limits::new(Size::ZERO, Size::new(max_width, max_height));
+            let (max_width, max_height) = axis.pack(
+                available,
+                if align_items == Align::Fill {
+                    cross
+                } else {
+                    max_cross
+                },
+            );
```
hecrj, line 135: I wonder if we could group the branching here with:

```rust
let (min_height, max_height) = if align_items == Align::Fill {
    // ...
} else {
   // ...
};
```

It may help with readability!

native/src/layout/flex.rs
```diff
@@ -124,19 +171,34 @@ where
...
-            let (max_main, max_cross) =
-                axis.pack(max_main, axis.cross(limits.max()));
+            let (max_main, max_cross) = axis.pack(
+                max_main,
+                if align_items == Align::Fill {
+                    cross
+                } else {
+                    max_cross
+                },
+            );
```
hecrj, line 190: Same applies here!

native/src/layout/flex.rs
```diff
@@ -124,19 +171,34 @@ where
...
 
             let child_limits = Limits::new(
                 Size::new(min_main, min_cross),
                 Size::new(max_main, max_cross),
```
hecrj, line 194: While we are at it, we should probably rename these variables to be `*_width` and `*_height` like in the loop above.

hecrj (changes requested):
Looks great! :partying_face: 

It's really nice that we can extend the layout engine so easily with just an additional variant! :tada: 

I left a couple of suggestions. Let me know what you think!

native/src/layout/flex.rs
tarkah: That cleaned it up a bunch, thanks!

tarkah:
Thanks, @hecrj! Agreed, a surprisingly easy change for a big layout improvement.

One last question before we merge... I'm having a harder time reasoning about the potential impact of this change on `Container`, since it accepts `Align` in `align_x` and `align_y`, but it only has "one" child. Do we need to handle anything special in `layout` of `Container`?

hecrj:
@tarkah I think the changes in `Node::align` should take care of it. We could also create a new enum exclusively for cross alignment.

hecrj:
I have rebased and refactored the alignment types by moving them into a new `alignment` module.

Also, I have changed the `Container` API to leverage `alignment::Horizontal` and `alignment::Vertical` directly, which have no `Fill` variant.

hecrj (approved):

PR #1049, opened 2021-09-17
CI: Run a Windows application without starting the shell too

Luni-4:
When an application is executed on Windows, alongside with that, it also starts the shell. This PR adds to the CI a way not to run a shell through the use of a specific Rust attribute.

Thanks in advance for your review! :)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,6 +51,9 @@ jobs:
         mkdir .cargo
         echo '[target.x86_64-pc-windows-msvc]' >> .cargo/config
         echo 'rustflags = ["-Ctarget-feature=+crt-static"]' >> .cargo/config
+    - name: Run the application without starting the shell
+      run: |
+        sed -i '1 i\#![windows_subsystem = \"windows\"]' examples/todos/src/main.rs
     - name: Build todos binary
       run: cargo build --verbose --release --package todos
     - name: Archive todos binary


.github/workflows/build.yml
```diff
@@ -51,6 +51,9 @@ jobs:
         mkdir .cargo
         echo '[target.x86_64-pc-windows-msvc]' >> .cargo/config
         echo 'rustflags = ["-Ctarget-feature=+crt-static"]' >> .cargo/config
+    - name: Run the application without starting the shell
+      run: |
+        sed -i '1 i\#![windows_subsystem = \"windows\"]' examples/todos/src/main.rs
```
hecrj, line 56: I think we can just add this line to the `main` file directly.

hecrj (changes requested):
Thanks for giving this a shot! 🙇

.github/workflows/build.yml
Luni-4: I think that if we add that attribute to the `main` file directly we will prevent developers to start the shell. Perhaps we could do that only for release builds

hecrj: Good point! Let's merge then.

hecrj (approved):

.github/workflows/build.yml
Luni-4: Thank you for merging my PR! :)

Issue #553, opened 2020-10-06
[Proposal] Persistent widget tree


twitchyliquid64:
This issue attempts to condense discussions (on the zulip chat) over the last week into a single set of requirements, and a proposal for implementing a widget tree.

For context, it seems that a persistent (keeps state across layout/update/draw cycles) widget tree is necessary or ideal to implement a number of core features: animations, incremental draw/layout, retained-mode, and accessibility integration.

## Requirements

 * Motion (every frame) animations: Widgets need to be able to indicate that they need to be drawn every frame (ie: something fading in). Ideally, `draw()` would only be called on the widgets which need it: This will help us scale to large numbers of widgets.
 * Periodic (scheduled) animations: Widgets need to be able to schedule periodic updates at roughly some moment in the future (ie: blinking cursor). See #31 for more details on both of these.
 * Retained-mode primitives: For each widget, keep track of the display list, and whether the display list needs to be re-computed. This information can be used to cull the number of invocations to `draw()` on the widget, and in the future keep objects in GPU memory live.
 * Accessibility: We need a stable concept of widget identity (to communicate/key with an external, platform-specific API), children / parent widgets, and the ability to associate additional data with a widget for accessibility reasons (ie: A custom widget could markup logically-related widgets).
 * Tab/keyboard selection: The widget tree needs to keep track of sequential widgets so tab can advance through widgets.

## Proposal

I'll finish up a proper proposal tomorrow as its almost midnight. But this is what I have in mind so far:

The 'root' of a widget tree is a `UI` struct:

```rust
pub struct UI<'a> {
  root: Node<'a>,
  timers: Vec<Timer>, // sorted by soonest
}
```

And `Node` is a DOM-style doubly-linked tree of widgets (parent / next child / prev child / next sibling / prev sibling):

```rust
type Node<'a> = rctree::Node<Widget<'a>>;
```
(See [rctree](https://docs.rs/rctree/0.3.3/rctree/index.html) for more information on this tree scheme - such a well-connected structure seems ideal given the kinds of mutations and iterations we need to support.)

A `Widget` contains all the persistent state represented by a widget in the tree:

```rust
struct Widget<'a> {
  widget: Element<'a>,
  display_list: Option<Vec<Primitive>>,

  needs_draw: bool,
  children_need_draw: bool,

  needs_animation_tick: bool,
  child_needs_animation_tick: bool,

  needs_layout: bool, // also set if any children need layout
  layout: Option<Rectangle>,
}
```

[...]

PR #1057, opened 2021-09-26
Widget states storage

alex13sh:
I have been using iced for my projects  for a year now. My concern is that the Application::view function is mutable. I don't know, maybe you have your own solution, but I decided to offer my own. I know that you don't like to go to the code without a preliminary discussion, but my English is bad.

This is my first PR not only in this project, but also my first contribution to open source in general. So do not judge strictly.
This PR allows you to get rid of the changeable Application function::view, by storing states in a special WidgetStateStorage struct. This storage is formed during the deletion of the current Element tree and before creating a new Element tree, during the Application::view call. StateStorage is temporarily stored in the Cache and is used inside the UserInterface::builder.
Since widgets themselves store their state, now there is no longer a need to store widget states inside the Application, and therefore there is no need inside the Application::view function to pass a mutable reference to the state.
But what if you need to change the state of some widget inside the Application::update function? Then it is possible to use a mutable reference to StateStorage, which is passed to the argument of the update function.
Each State has its own ID, which is manually specified or automatically generated. To get a mutable reference to the state of a specific widget inside the Application::update function, you must assign an ID to this widget in advance.
A tree of elements of great depth is sometimes divided into several "components", and in this case, an ID conflict may occur. To avoid this, a multi-level ID in the form of PathBuf is used.

Two functions have been added for the widget function. "into_states" - the widget is deleted and returns its state, "apply_status" - the widget finds and assigns its state.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -28,6 +28,7 @@ mod point;
 mod rectangle;
 mod size;
 mod vector;
+// mod hash_states;
 
 pub use alignment::Alignment;
 pub use background::Background;
@@ -39,3 +40,4 @@ pub use point::Point;
 pub use rectangle::Rectangle;
 pub use size::Size;
 pub use vector::Vector;
+// pub use hash_states::HashStates;
--- a/examples/todos/src/main.rs
+++ b/examples/todos/src/main.rs
@@ -1,10 +1,9 @@
 use iced::alignment::{self, Alignment};
-use iced::button::{self, Button};
-use iced::scrollable::{self, Scrollable};
 use iced::text_input::{self, TextInput};
 use iced::{
-    Application, Checkbox, Column, Command, Container, Element, Font, Length,
-    Row, Settings, Text,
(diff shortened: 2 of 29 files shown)


hecrj:
We talked briefly about this in the last Discord meeting.

In short, there are many challenges here that need discussion:

- The strategy used to identify widgets and keep continuity (specially when the `view` tree changes)
  - Do we want to burden our users with generating unique identifiers?
  - Do we need a diffing algorithm?
- The changes to the `Widget` API
  - What is the simplest way to extend it to support the new use cases?
  - How do we store the states of each `Widget`? Do we really need to box everything and downcast?
 - Impact on other features
   - How do the changes relate to other future planned features (e.g. persistent widget tree #553).

I think we need to answer these questions first before we jump into writing code. @derezzedex is already looking into it! Feel free to join [the Discord server](https://discord.gg/3xZJ65GAhd) and chat about it.

PR #1065, opened 2021-09-29
Fix `examples/events`.

baehyunsol:
`iced::alignment` and `iced::Alignment` are deprecated. It uses `iced::Align`, `iced::HorizontalAlign` and `iced::VerticalAlign` now.

Also, `Application::update` takes 3 parameters but the older one only has 2.


hecrj:
It's the other way around. `Alignment` was introduced recently in #1044!

PR #1067, opened 2021-09-29
Honor Exif orientation in `iced_wgpu::Image`

akavel:
See: https://github.com/iced-rs/iced/discussions/1064

### DIFF
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -35,6 +35,7 @@ log = "0.4"
 guillotiere = "0.6"
 futures = "0.3"
 kamadak-exif = "0.5"
+bitflags = "1.3"
 
 [dependencies.bytemuck]
 version = "1.4"
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -35,7 +35,7 @@ log = "0.4"
 guillotiere = "0.6"
 futures = "0.3"
 kamadak-exif = "0.5"
-bitflags = "1.3"
+bitflags = "1.2"
 
 [dependencies.bytemuck]
 version = "1.4"
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -34,7 +34,7 @@ raw-window-handle = "0.3"
 log = "0.4"
 guillotiere = "0.6"
 futures = "0.3"
-kamadak-exif = "0.5"  # TODO(akavel): cfg(only when jpeg enabled) ?
+kamadak-exif = "0.5"
 
 [dependencies.bytemuck]
 version = "1.4"
--- a/wgpu/Cargo.toml
+++ b/wgpu/Cargo.toml
@@ -34,6 +34,7 @@ raw-window-handle = "0.3"
 log = "0.4"
 guillotiere = "0.6"
 futures = "0.3"
+kamadak-exif = "0.5"  # TODO(akavel): cfg(only when jpeg enabled) ?
 
 [dependencies.bytemuck]
 version = "1.4"
--- a/wgpu/src/image/raster.rs
+++ b/wgpu/src/image/raster.rs
@@ -175,16 +175,20 @@ impl Operation {
             .unwrap_or_else(Self::empty))
     }
 
-    fn perform<I>(self, mut image: I) -> I
+    fn perform<P>(
+        self,
+        image: image_rs::ImageBuffer<P, Vec<P::Subpixel>>,
+    ) -> image_rs::ImageBuffer<P, Vec<P::Subpixel>>
     where
-        I: image_rs::GenericImage,
+        P: image_rs::Pixel + 'static,
     {
         use image_rs::imageops;
 
-        if self.contains(Self::FLIP_DIAGONALLY) {
-            imageops::flip_horizontal_in_place(&mut image);
-            imageops::flip_vertical_in_place(&mut image);
-        }
+        let mut image = if self.contains(Self::FLIP_DIAGONALLY) {
+            flip_diagonally(image)
+        } else {
+            image
+        };
 
         if self.contains(Self::ROTATE_180) {
             imageops::rotate180_in_place(&mut image);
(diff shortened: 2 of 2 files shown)


wgpu/Cargo.toml
```diff
@@ -34,6 +34,7 @@ raw-window-handle = "0.3"
 log = "0.4"
 guillotiere = "0.6"
 futures = "0.3"
+kamadak-exif = "0.5"  # TODO(akavel): cfg(only when jpeg enabled) ?
```
akavel, line 37: Actually, since writing this TODO I realized that the library [seems to work for at least TIFF, JPEG, HEIF, PNG, and WebP](https://docs.rs/kamadak-exif/0.5.4/exif/index.html), so I planned to remove the comment before submitting, but missed it. Leaving the decision as to whether to clean it to you now then.

hecrj (approved):
Looks great! 🎉 

I have added an `Orientation` enum in 03fee3106f0d32f0883bf3ee293213059902f304 to improve readability a bit. Let me know what you think!

wgpu/Cargo.toml
hecrj: Yes. Since it's a small dependency, we can just add it unconditionally for simplicity.

wgpu/src/image/raster.rs
```diff
@@ -132,3 +145,67 @@ impl Cache {
...
+        match self {
+            Self::FlippedHorizontally => flip_horizontal_in_place(&mut img),
+            Self::Rotated180 => rotate180_in_place(&mut img),
+            Self::FlippedVertically => flip_vertical_in_place(&mut img),
+            Self::Rotated90AndFlippedHorizontally => {
+                img = rotate90(&img);
+                flip_horizontal_in_place(&mut img);
+            }
+            Self::Rotated90 => img = rotate90(&img),
+            Self::Rotated90AndFlippedVertically => {
```
akavel, line 201: [nit] So, in theory, I think one could argue that if *we* need to rotate the image 90°, it technically means *the source image* was transformed in reverse "from the correct orientation" - i.e. in case of l.200, the orientation of the JPEG is "Rotated270" from "what was really seen".

The "double transform" ones would then also become reversed in the order of transforms - l.201 being arguably `FlippedVerticallyAndRotated270`.

Alternatively, if we wanted to make it right, we could also try renaming the "Orientation" enum to something else to better capture that it represents "How To Fix" orientation, not just orientation that the specific JPEG "has". Unfortunately, I don't currently have a great idea for what name could be cool for that.

With that said, especially with this being private, that's *totally* a nitpick, so just saying for your awareness and decision at your discretion.

wgpu/src/image/raster.rs
```diff
@@ -132,3 +145,67 @@ impl Cache {
...
+    // Meaning of the returned value is described e.g. at:
+    // https://magnushoff.com/articles/jpeg-orientation/
+    fn from_exif<R>(reader: &mut R) -> Result<Self, exif::Error>
+    where
+        R: std::io::BufRead + std::io::Seek,
+    {
+        let exif = exif::Reader::new().read_from_container(reader)?;
+
+        Ok(exif
+            .get_field(::exif::Tag::Orientation, ::exif::In::PRIMARY)
```
akavel, line 171: Should we drop the `::` prefix also here and on l.188? I thought this one here would get confused with `exif` var defined on l.168, but I just tried and apparently it works (I'm still absolutely a Rust newbie)

akavel (commented):
🎉 Thanks! I'm still a Rust newbie, so I'm observing your tweaks with absolute fascination and desire to learn from them! 😍❤

I added two extra comments based on what came to my mind after reading your changes. Obviously, just thoughts left to your decision!

wgpu/src/image/raster.rs
hecrj: Good point. I have refactored the `Orientation` enum into an `Operation` bitflag in fb97fea438ff803c990e36119977c08b0e5773e1. I believe this also sets everything up for future optimizations (i.e. applying a transform matrix to the image once).

Hopefully I didn't break anything!

hecrj: It looks like flipping along the diagonal isn't what I think it is. Fixing!

hecrj: Ok, I think 65c2c17226ec8472066ea48d733a458cb40fe12c should be it!

akavel:
I wanted to test-run it on the sample images from https://github.com/recurser/exif-orientation-examples, but in my test app I'm getting a dependencies conflict:


```
C:\prog\deorient>cargo run --release
    Updating crates.io index
error: failed to select a version for `bitflags`.
    ... required by package `iced_wgpu v0.4.0 (C:\prog\iced\wgpu)`
    ... which is depended on by `iced v0.3.0 (C:\prog\iced)`
    ... which is depended on by `deorient v0.1.0 (C:\prog\deorient)`
versions that meet the requirements `^1.3` are: 1.3.2, 1.3.1

all possible versions conflict with previously selected packages.

  previously selected package `bitflags v1.2.1`
    ... which is depended on by `andrew v0.3.1`
    ... which is depended on by `smithay-client-toolkit v0.12.3`
    ... which is depended on by `winit v0.25.0 (https://github.com/iced-rs/winit?rev=44a9a6fc442fcfa3fa0dfc2d5a2f86fdf4aba10c#44a9a6fc)`
    ... which is depended on by `iced_winit v0.3.0 (C:\prog\iced\winit)`
    ... which is depended on by `iced v0.3.0 (C:\prog\iced)`
    ... which is depended on by `deorient v0.1.0 (C:\prog\deorient)`

failed to select a version for `bitflags` which could resolve this conflict

hecrj:
@akavel `cargo update` should fix that I believe. However, we should lower the minimum version requirement for consistency, since `iced_core` depends on `1.2`.

akavel:
Ok, so I added a commit to downgrade it!

akavel:
(Edited the last commit's message, the previous one was incorrect due to my misunderstanding)

hecrj (approved):
Awesome! Let's 🚢!

PR #1096, opened 2021-10-23
Experimental WebGL wgpu backend support

pacmancoder:
Made changes to allow the use of WebGL wgpu backend. The basic functionality works, but text rendering is still buggy - Somehow, if the initial value of the text control was changed, some parts of the text start to disappear/show artifacts. 

**UPD:** Strange issue with text rendering is only reproducing in Firefox (v93). It is not directly related to the iced, I created an issue in `wgpu_glyph` here: https://github.com/hecrj/wgpu_glyph/issues/78

The current implementation provides neither `Application` nor `Sandbox` implementations for this backend, and does not implement proper `Clipboard` in `iced_winit`, however, this should be at least some starting point.

Changes list:
- Added new `integration_wgpu_webgl` example, based on `integration_wgpu` and `wgpu`'s `hello-triangle` example (can't use `integration_wgpu` example directly because it fails with its spirv shaders on WebGL)
- Lifted `Sync` requirement for `Runtime` and `Tracker` in `iced_futures`
- Aded `webgl` feature to `iced_wgpu` which enables `wgpu/webgl`
- Created stub implementation for `Clipboard` in `iced_winit`

<details><summary>Images</summary>

iced running inside WebGL canvas:
image

Text issues after changing text value:
image

Same example running on desktop backend:
image

</details>


kaimast:
Did you test input with this branch yet? I seem to have a bunch of issues but they are probably caused by winit. See this bug report here: https://github.com/rust-windowing/winit/issues/2036 

Problem is that iced still uses a custom winit branch, so it's hard to figure out what's going on is fixed upstream or not.

native/src/subscription.rs
```diff
@@ -1,7 +1,12 @@
 //! Listen to external events in your application.
 use crate::event::{self, Event};
 use crate::Hasher;
-use iced_futures::futures::stream::BoxStream;
+
+#[cfg(not(target_arch = "wasm32"))]
+type BoxStream<'a, T> = iced_futures::futures::stream::BoxStream<'a, T>;
+
+#[cfg(target_arch = "wasm32")]
+type BoxStream<'a, T> = iced_futures::futures::stream::LocalBoxStream<'a, T>;
```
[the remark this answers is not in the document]
pacmancoder: fixed

futures/src/subscription/tracker.rs
```diff
@@ -3,6 +3,50 @@ use crate::{BoxFuture, Subscription};
...
+    {
+    }
+
+    impl<Message: TrackerMessage, T> TrackerMessageReceiver<Message> for T where
+        T: Sink<Message, Error = mpsc::SendError> + Unpin + Clone + 'static
+    {
+    }
+}
+
+pub use trait_aliases::{TrackerMessage, TrackerMessageReceiver};
```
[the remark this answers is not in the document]
pacmancoder: Sadly yes, this is necessary - we use `winit` here with WebGL backend and it (winit) does not implement `Send` on its `EventLoopProxy` on wsam32, so yea, because of that we should lift here requirement for `Send` for our types too

pacmancoder:
> 
> 
> Did you test input with this branch yet? I seem to have a bunch of issues but they are probably caused by winit. See this bug report here: [rust-windowing/winit#2036](https://github.com/rust-windowing/winit/issues/2036)
> 
> Problem is that iced still uses a custom winit branch, so it's hard to figure out what's going on is fixed upstream or not.

@kaimast Ah... I see... Sliders in the example were working fine then I just tried to add `TextInput` to the view and yep, after that, it broke (GUI started to ignore any mouse events to any GUI element). Looks like `winit` indeed could be broken for wasm32?

cc @hecrj

kaimast:
Could you rebase this on current master? iced has moved to wgpu 0.12 and it would be good to see if some of the issues are resolved.

pacmancoder:
@kaimast sure, I'll try to find som spare time on this week and do the rebase

examples/integration_wgpu/src/main.rs
```diff
@@ -14,11 +14,39 @@ use winit::{
...
+            .expect("Canvas with id `iced_canvas` is missing")
+    };
+    #[cfg(not(target_arch = "wasm32"))]
     env_logger::init();
 
     // Initialize winit
     let event_loop = EventLoop::new();
+
+    #[cfg(target_arch = "wasm32")]
+    let window = winit::window::WindowBuilder::new()
```
aentity, line 44: i think you can to save some `cfg` by doing:

```rust
let window = if cfg!(target_arch = "wasm32") {
    let canvas_element = {
        console_log::init_with_level(log::Level::Debug)
            .expect("could not initialize logger");
        std::panic::set_hook(Box::new(console_error_panic_hook::hook));

        web_sys::window()
            .and_then(|win| win.document())
            .and_then(|doc| doc.get_element_by_id("iced_canvas"))
            .and_then(|element| element.dyn_into::<HtmlCanvasElement>().ok())
            .expect("Canvas with id `iced_canvas` is missing")
    };
    let window = winit::window::WindowBuilder::new()
        .with_canvas(Some(canvas_element))
        .build(&event_loop)
        .expect("Failed to build winit window")
} else {
    env_logger::init();
    winit::window::Window::new(&event_loop).unwrap()
};
```

unless the `Window::new()` method returns separate types

aentity:
this is very interesting. does this patch allow to run in pure wgpu, without to use html elements for the widget?

kaimast:
> this is very interesting. does this patch allow to run in pure wgpu, without to use html elements for the widget?

yes

aentity:
Wow this is great news! To experiment with targetting webgpu, will this approach here block in future to do this? Do we need to add a webgpu future when/if we try such a thing?

kaimast:
I created a tiny pull request that should fix the input issues. I think it should be close to being merge-able now.

pacmancoder:
Thanks, @kaimast, merged

winit/src/application.rs
```diff
@@ -179,41 +183,81 @@ where
...
+
+                let poll = instance.as_mut().poll(&mut context);
+
+                *control_flow = match poll {
+                    task::Poll::Pending => ControlFlow::Wait,
+                    task::Poll::Ready(_) => ControlFlow::Exit,
+                };
+            }
+        });
+    }
```
kaimast, line 260: I'm trying to understand why these two different code paths are needed. Can you explain?

I will add a comment or, ideally, try to merge them somewhat.

pacmancoder: Actually, yes, this can be de-duplicated, I just had no enough time to do it, so I did it quick and dirty \*sorry\* and then forgot to fix this code chunk. @kaimast If you interested in this PR, I sent your invite to be a collaborator on the fork, you welcome to make direct changes to the PR.

kaimast:
I just removed the custom clipboard code, thank for updating `window_clipboard`, @hecrj !

winit/src/application.rs
```diff
@@ -179,7 +178,7 @@ where
 
     let mut context = task::Context::from_waker(task::noop_waker_ref());
 
-    event_loop.run_return(move |event, _, control_flow| {
+    event_loop.run(move |event, _, control_flow| {
```
kaimast, line 181: @hecrj do you think this would break anything? It simplifies the code a lot and the `winit` docs says about `run_return` "You are strongly encouraged to use run, unless the use of this is absolutely necessary."

hecrj:
I made some changes here! Will explain everything first thing tomorrow!

hecrj (approved):
Okay! First of all, thanks for all the efforts here. This is very cool!

As I mentioned yesterday, I made a bunch of changes, some of them quite important.

First, I have removed the `iced_web` subcrate completely from the repository in 825c7749ff364cf1f7ae5cab0c25f27ed85c7d82! `iced_web` has always been quite experimental and their dependencies have been outdated for a while now. Therefore, I believe it's better to move it to a different repository in the `iced-rs` organization. As a consequence, from now on when targetting Wasm, `iced` will leverage `wgpu` and / or `glow` to render into a `canvas` using the `iced_native` crate (see 26d95fdc4b7b6a66431d45f49edd0cc3ef19823f).

In my opinion, this is a better default than `iced_web` because it makes targetting Wasm quite easier for users. As there fewer differences between the native stack and the web stack, all of the native widgets should work on Wasm. If you have a native `iced` application, chances are you can compile it to Wasm! No additional feature flags necessary! Just slap a `--target wasm32-unknown-unknown` to your `cargo build` command.

That said, the plan is to still offer `iced_web` as a separate crate for users that need to leverage the DOM.

The rest of the changes are relatively minor:
- I changed the `Compositor` in `iced_wgpu` to properly use WebGL limits in 776724162aa4351310de9f88c7bb8d61b0abca04.
- I created a new `time` module that offers a cross-platform `Instant` implementation in 87b3e03d187237f665b376018ea5af0cc5f05814 and 83c649b574d90667d23c8430baaebcd0ef933055.
- I introduced a `MaybeSend` trait in `iced_futures` in 5dab5a327ef643ee38ac3e42ab35212fff445631, which allows us to get rid of all the `trait_aliases` modules.
- I have split the `iced_futures` executor implementations into different `backend` modules in 167be45a7db7c1f60a79116766bdf38300429c6a. Each `backend` should expose the same API!
- I implemented `time::every` for the `wasm_bindgen` backend in e730d97f61bc2edc77d2f061b6a763c4d0a948df.

Finally, I have added a couple of `index.html` files to the `todos` and `tour` examples. You can try running them on your browser with [`trunk`](https://trunkrs.dev/).

There are still some issues present related to rendering text. My MacBook Pro M1 displays completely mangled text in all of the examples both in Chrome and Firefox:

image

image

This seems to happen only to dynamic text, which seems to indicate there may be an issue in the cache upload logic in `wgpu_glyph`. Most likely related to https://github.com/hecrj/wgpu_glyph/issues/78.

That said, I am pretty happy with the changes here and I think we can merge this! 🎉  We can iron out the issues in future PRs.

m4b:
this is... awesome :joy_cat:

PR #1106, opened 2021-11-01
Remove unnecessary String allocation

RamType0:
Remove unnecessary `String` allocation by passing `&format!` or `&x.to_string` as `impl Into<String>`


hecrj (approved):
Nice!

I think we should get `clippy` involved in our CI eventually to avoid pitfalls like these ones when APIs change.

Issue #1050, opened 2021-09-22
Customizable Text/TextInput Representation


NathanSWard:
Currently `Text`/`TextInput` are defined as follows:

```rust
struct Text {
    content: String,
    ..
}

struct TextInput<..> {
    placeholder: String,
    ..
}
```

I have come across use cases where it would be nice to use a `Cow<str>` or just a `&'static str` to avoid unnecessary memory allocations. 
I believe it would be beneficial to make these type generic of the string type. 

e.g.

```rust
struct Text<S: AsRef<str>> {
    content: S,
    ..
}
```

Any thoughts?
I'd be willing to make a PR for this if people are willing to accept this change :smile:

NathanSWard:
Note: `AsRef<str>` wouldn't quite work for `TextInput` but there are other possible solutions :)

RamType0:
Sorry, I want this and I implemented it.

PR #1107, opened 2021-11-01
Text content by cow

RamType0:
#1050
Replace `Text.content` from `String` to `Cow<'a,str>` to reduce `String` allocation.

### DIFF
--- a/examples/todos/src/main.rs
+++ b/examples/todos/src/main.rs
@@ -469,7 +469,7 @@ const ICONS: Font = Font::External {
     bytes: include_bytes!("../../todos/fonts/icons.ttf"),
 };
 
-fn icon(unicode: char) -> Text {
+fn icon(unicode: char) -> Text<'static> {
     text(unicode.to_string())
         .font(ICONS)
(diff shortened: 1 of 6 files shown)


tema3210:
> #1050 Replace `Text.content` from `String` to `Cow<'a,str>` to reduce `String` allocation.

I wounder if it's feasible to do alter `TextInput` message to be `(String,TInputEvent)` where TInputEvent  contains various things that can happen with text input, like copy-paste events, `char` insertion at a position, etc.

Imagined API:
```
enum TInputEv {
   Paste( fn(&mut String) ), //this writes paste content to the string; when it's emmited message's `String` still contains old value
   Insertion{ at: Range<usize>, content: String} // range here indicates which indices of original string are going to be changed, content is payload; old value is still unchanged
  KeyPress(char) //technically a subset of previous, but this is so common that it deserves it's own variant, I think
}

hecrj (approved):
I have rebased the changes on top of `master`.

I believe we can start by chagning `Text::new` to take a `Cow` and eventually expose a new helper.

PR #1109, opened 2021-11-03
Fix color issue with Debugger::explain implementation for Renderer

mettz:
This PR should fix an issue with the actual implementation of `layout::Debugger` where the argument `color` of the `explain` method is ignored


hecrj (approved):
Wow, that's a silly bug!

Thank you! :bow:

PR #1112, opened 2021-11-07
Application::run returns on native platforms

AndiHofi:
Added an `on_exit() -> Option<Box<dyn FnOnce>>` function to the iced::Application trait that provides that callback.

When the returned closure does not actually exit the application then `iced::Application::run()` can return as well, allowing the program to continue when the main window is closed.

For iced_web it could make sense to invoke the on_exit() closure when unloading the page, did not try to implement that though.

The main use-case for this is to allow to change the exit code of the application.

### DIFF
--- a/glutin/src/application.rs
+++ b/glutin/src/application.rs
@@ -11,7 +11,6 @@ use iced_winit::futures;
 use iced_winit::futures::channel::mpsc;
 use iced_winit::{Cache, Clipboard, Debug, Proxy, Settings};
 
-use crate::glutin::platform::run_return::EventLoopExtRunReturn;
 use glutin::window::Window;
 use std::mem::ManuallyDrop;
 
@@ -29,6 +28,7 @@ where
     use futures::task;
     use futures::Future;
     use glutin::event_loop::EventLoop;
+    use glutin::platform::run_return::EventLoopExtRunReturn;
     use glutin::ContextBuilder;
 
     let mut debug = Debug::new();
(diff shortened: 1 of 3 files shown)


hecrj (changes requested):
Maybe I'm missing something, but if we are leveraging `run_return` and allowing `Application::run` to return control, then the `on_exit` method seems redundant.

In other words, users can just write:

```rust
let _ = Application::run()?;

on_exit();
```

AndiHofi:
yeah - I somehow overlooked that. I did not want to change the control flow of `Application::run()` it currently only returns when initializing the application fails.

I realized that the run() method may return normally when running through the code for the pull-request.

Letting it return in all cases would be simpler than any callback - unless returning from `Application::run` does not work with all backends.

AndiHofi:
Removed the on_exit() callback entirely and added an example that `Application::run` returns. Also that it may be called in a loop.

Added another commit that demonstrates the same when calling `iced_winit::run` or `iced_glutin::run`

hecrj (changes requested):
Can we keep the changes to the the bare minimum (i.e. just the `run_return`).

We can think about additional examples later. If we start adding examples for every change, they will be very hard to navigate 😅

AndiHofi:
sure, I get rid of the examples...

and rename the pull request. There is no more on_exit callback

hecrj (approved):
Looks good!

I made some minor changes, but it should be ready to go! :tada:

PR #1114, opened 2021-11-11
Don't rebroadcast redundant slider messages

tarkah:
Updates `Slider` to only broadcast `on_change` when the `Slider` value actually changes from it's previous value. I noticed a ton of redundant messages being broadcast for the same slider value (probably due to high DPI mouse). This causes the slider to become very laggy when driving a computationally heavy change.

`Slider` is now buttery smooth!

### DIFF
--- a/native/src/widget/slider.rs
+++ b/native/src/widget/slider.rs
@@ -144,7 +144,7 @@ where
 }
 
 /// The local state of a [`Slider`].
-#[derive(Debug, Clone, Copy, PartialEq, Default)]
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Looks great!

I have made a small change to use the `value` in `Slider` instead of introducing a new `previous` field in `State`. This way, we properly handle the scenario where the `value` may be changing from a different part of the GUI.

Let me know what you think!

tarkah:
Yup, that makes a lot more sense! Looks great.

PR #1117, opened 2021-11-14
Disable default features in twox-hash dependency.

TannerRogalsky:
We don't use twox-hash's random seeding features. Removing them makes
this crate leaner and also more portable (specifically for web).


hecrj (approved):
Good catch. Thanks! :bow:

PR #1118, opened 2021-11-15
Native web fixes.

TannerRogalsky:
`iced_futures`
* Since there are several "platform-specific" type aliases (and a new function), I moved them into their own `mod`. Just makes it easier to read and minimizes the number of `#[cfg(...)]` annotations.
* Added a new function `boxed_stream` that takes a stream of the type supported by the target architecture and boxes it accordingly.

`iced_native`
* Use the `iced_futures` types and functions to allow the `iced_native` to compile on the `wasm32-unknown-unknown` target.

### DIFF
--- a/futures/src/lib.rs
+++ b/futures/src/lib.rs
@@ -37,33 +37,58 @@ pub mod time;
 
 pub use command::Command;
 pub use executor::Executor;
+pub use platform::*;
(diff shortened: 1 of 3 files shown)


hecrj (approved):
Awesome! The module approach is much cleaner.

Thank you :bow:

PR #1122, opened 2021-11-18
Correct 0.3 docs to reference 0.3 source

quintenpalmer:
Previously, the docs for version 0.3 would point to the source code on
branch 0.2. This corrects the link to point to the 0.3 source.

I had initially seen this behavior from the docs.rs page:

https://docs.rs/iced/0.3.0/iced/index.html

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,7 +30,7 @@
 //! [windowing shell]: https://github.com/hecrj/iced/tree/master/winit
 //! [`dodrio`]: https://github.com/fitzgen/dodrio
 //! [web runtime]: https://github.com/hecrj/iced/tree/master/web
-//! [examples]: https://github.com/hecrj/iced/tree/0.2/examples
+//! [examples]: https://github.com/hecrj/iced/tree/0.3/examples
 //! [repository]: https://github.com/hecrj/iced
 //!
 //! # Overview


proudmuslim-dev:
If I recall correctly, the 0.3 docs mention 0.2 in a lot of places- I'll see if I can get links to specific examples of this

proudmuslim-dev:
Here are the ones I found at a glance, in descending order of how long they are:

- https://github.com/iced-rs/iced/blob/0.3/src/sandbox.rs#L39-L51
- https://github.com/iced-rs/iced/blob/0.3/futures/src/subscription.rs#L128-L130
- https://github.com/iced-rs/iced/blob/master/native/src/widget.rs#L96-L101
- https://github.com/iced-rs/iced/blob/0.3/wgpu/src/widget/pane_grid.rs#L9
- https://github.com/iced-rs/iced/blob/0.3/web/src/lib.rs#L52

Note: I haven't checked if these are fixed on the `main` branch

proudmuslim-dev:
It also seems like your PR is requesting a merge into the `master` branch instead of `0.3`

quintenpalmer:
Great call on the last comment, thank you! I've updated the target branch

quintenpalmer:
Good point on all of the other `0.2`s as well, I can update those and hunt for more if that would be useful and this branch is likely to be merged.

hecrj:
Thanks!

This was fixed in `master` recently and we will soon release `0.4.0`!

PR #1123, opened 2021-11-19
Clip overlay dropdown in picklist

tarkah:
Currently text inside the picklist dropdown overlay isn't clipped and bleeds over (when no scrollbar is present).

`List`, which is rendered as the overlay widget of `Picklist`, isn't clipped. Since it's placed in a `Scrollable`, it's clipped when the scrollbar is present (due to the implementation of `Scrollable`), but not when the scrollbar is hidden.

This fix draws the `List` primitives in it's own layer so that text is properly clipped.

### DIFF
--- a/native/src/overlay/menu.rs
+++ b/native/src/overlay/menu.rs
@@ -422,48 +422,50 @@ where
 
         let visible_options = &self.options[start..end.min(self.options.len())];
 
-        for (i, option) in visible_options.iter().enumerate() {
-            let i = start + i;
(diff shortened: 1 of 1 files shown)


clarkmoody (approved):
Looks simple enough to me 👍

hecrj (commented):
I have opened #1125 with a more generic fix. Could you try it out?

tarkah:
@hecrj #1125 did the trick! Much better :+1:

PR #1125, opened 2021-11-21
Fix overflow of graphics primitives in overlays

hecrj:
... by using `overlay_bounds` instead of `viewport` as clip bounds for overlay layer in `UserInterface::draw`.

Closes #1123.


tarkah (approved):
Fixes the issue addressed in #1123. Thanks!

PR #1127, opened 2021-11-22
Example fixed: counter

MarkLagodych:
Changed `Alignment` -> `Align`. Now it compiles!


hecrj:
`master` has breaking changes! `Alignment` is the proper type now :slightly_smiling_face:

PR #1140, opened 2021-12-06
Include alignment in the container layout hash

Imberflur:
I noticed the layout wasn't updating when I changed the alignment.


hecrj (approved):
Good catch! :+1:

PR #1147, opened 2021-12-08
Added icon in docs

daladim:
I just realized iced had an icon, and it is worth showing it in the generated docs!


hecrj (approved):
Awesome! Thank you 🙇

PR #1148, opened 2021-12-08
Added iced_winit::application::Application::run_with_window_configurator

AndiHofi:
Defined a new `WindowConfigurator` trait in iced_winit to configure the `WindowBuilder.`

The `SettingsWindowConfigurator` struct just applies the Window settings.

The new `iced_winit::application::run_with_window_configurator` uses a WindowConfigurator instead of the Settings struct. Except for the name, this makes quite a nice API for run as it avoids the generic Settings struct.

### DIFF
--- a/winit/src/application.rs
+++ b/winit/src/application.rs
@@ -17,7 +17,13 @@ use iced_graphics::window;
 use iced_native::program::Program;
 use iced_native::{Cache, UserInterface};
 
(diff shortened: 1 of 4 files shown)


hecrj:
I'm not completely sure which use cases we are solving with this code.

Generally, we prefer to tackle specific use cases directly in `iced`, instead of opening and exposing the internals directly.

I'll close this for now, but please feel free to start a discussion!

PR #1149, opened 2021-12-08
Fix `Widget::height` for `Component`

nicksenger:
This updates `Widget:height` for `Component` to use `height` rather than `width` of the cached element.


hecrj (approved):
Hah, too much copy-paste! 😅 

Thank you!

PR #1150, opened 2021-12-09
Make TextInput emit a Submit message when the key NumpadEnter is pressed

thenlevy:
Currently `TextInput`s only emit their `on_submit` message when the "main" Enter key is pressed. I believe they should also do so when the Enter key from the numpad is pressed.


hecrj (approved):
Fair enough! Thanks 🥂

PR #1151, opened 2021-12-09
Replace Zulip with Discord in README

x37v:


README.md
```diff
@@ -214,7 +214,7 @@ The development of iced is sponsored by the [Cryptowatch] team at [Kraken.com]
 [The Elm Architecture]: https://guide.elm-lang.org/architecture/
 [the current issues]: https://github.com/hecrj/iced/issues
 [contributing guidelines]: https://github.com/hecrj/iced/blob/master/CONTRIBUTING.md
-[Zulip server]: https://iced.zulipchat.com/
+Iced [Discord server]: https://discord.gg/3xZJ65GAhd
```
hecrj, line 217: The `Iced` word here is out of place. It should be:

```suggestion
[Discord server]: https://discord.gg/3xZJ65GAhd
```

hecrj (changes requested):
Yes, good catch!

Just a small thing we need to fix.

hecrj (approved):
Thank you! :tada:

x37v:
> Yes, good catch!
> 
> Just a small thing we need to fix.

yeah.. I added `Iced` specifically do differentiate between it and the rust sever below it but, if it makes sense to just call it "Discord Server" .. that is fine by me

PR #1154, opened 2021-12-10
Implement `overlay` for `Component`

nicksenger:
I saw the note about this method in `Component` so figure perhaps something like this has already been considered (or there are other plans for `overlay`), but here is something I came up with that seems to work in the limited tests I've performed.

The idea is to use a second self-referential struct to cache the cached element's overlay content, then return an overlay element that accesses the cache for each operation similar to how it's being done for the original element.

There's a couple potential problems with doing this:
- In each widget method for the original content we must call `into_heads` to access the element. I don't know what the performance implications of this are.
- Since we need to drop the cached overlay to access the original element, the returned overlay element may try to access the cached overlay content after it has already been dropped. I've tried to do the right thing in those cases, but I'm not sure what impact this could have on the UI in practice.

### DIFF
--- a/lazy/src/component.rs
+++ b/lazy/src/component.rs
@@ -24,7 +24,6 @@ where
         state: RefCell::new(Some(
             StateBuilder {
                 component: Box::new(component),
-                overlay_events: None,
                 cache_builder: |state| {
                     Some(
                         CacheBuilder {
@@ -56,7 +55,6 @@ struct Instance<'a, Message, Renderer, Event> {
 #[self_referencing]
 struct State<'a, Message: 'a, Renderer: 'a, Event: 'a> {
     component: Box<dyn Component<Message, Renderer, Event = Event> + 'a>,
-    overlay_events: Option<Vec<Event>>,
 
     #[borrows(mut component)]
     #[covariant]
@@ -105,17 +103,6 @@ impl<'a, Message, Renderer, Event> Instance<'a, Message, Renderer, Event> {
                 result
             })
     }
-
-    fn with_overlay_events_mut<T>(
-        &self,
-        f: impl FnOnce(&mut Option<Vec<Event>>) -> T,
-    ) -> T {
-        self.state
-            .borrow_mut()
-            .as_mut()
-            .unwrap()
-            .with_overlay_events_mut(f)
-    }
 }
 
 impl<'a, Message, Renderer, Event> Widget<Message, Renderer>
@@ -148,9 +135,7 @@ where
         clipboard: &mut dyn Clipboard,
         shell: &mut Shell<'_, Message>,
     ) -> event::Status {
-        let mut local_messages = self
-            .with_overlay_events_mut(|overlay_events| overlay_events.take())
-            .unwrap_or_default();
+        let mut local_messages = Vec::new();
         let mut local_shell = Shell::new(&mut local_messages);
 
         let event_status = self.with_element_mut(|element| {
@@ -183,7 +168,6 @@ where
             *self.state.borrow_mut() = Some(
                 StateBuilder {
                     component,
-                    overlay_events: None,
                     cache_builder: |state| {
                         Some(
                             CacheBuilder {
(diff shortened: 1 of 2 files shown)


hecrj:
> I saw the note about this method in Component so figure perhaps something like this has already been considered 

Yes! I originally tried to implement another self-referential `Overlay` struct:

```rust
#[self_referencing]
struct Overlay<'a, 'b, Message, Renderer, Event> {
    state: &'b mut Option<State<'a, Message, Renderer, Event>>,

    #[borrows(mut state)]
    #[covariant]
    overlay: Option<overlay::Element<'this, Event, Renderer>>,

    message: PhantomData<Message>,
}
```

So we can implement `Widget::overlay` as follows:

```rust
fn overlay(
    &mut self,
    layout: Layout<'_>,
) -> Option<overlay::Element<'_, Message, Renderer>> {
    let overlay = OverlayBuilder {
        state: &mut self.state,
        overlay_builder: |state| {
            state
                .as_mut()
                .unwrap()
                .with_element_mut(|element| element.overlay(layout))
        },
        message: PhantomData,
    }
    .build();

    if overlay.borrow_overlay().is_some() {
        Some(overlay::Element::new(layout.position(), Box::new(overlay)))
    } else {
        None
    }
}
```

But unfortunately, I don't think `ouroboros` can handle two lifetimes in a head field.

I did not think about relying on `RefCell` for internal mutations... That's pretty nifty!  Let me see if I can simplify the code a bit...

hecrj (approved):
I have removed some of the duplication in 48b2264bc6026e4b4b00791fef5ce34d464abc17 by creating a couple of helper methods.

Hopefully I didn't brek anything! Let me know what you think.

nicksenger:
> I have removed some of the duplication in [48b2264](https://github.com/iced-rs/iced/commit/48b2264bc6026e4b4b00791fef5ce34d464abc17) by creating a couple of helper methods.
> 
> Hopefully I didn't break anything! Let me know what you think.

I really like the `with_*` helper methods, they make this much easier to follow!

There is an edge case though where if `overlay.on_event()` causes a state mutation which results in the original element's `overlay` method now returning `None`, the application will panic because `hash_layout` or other methods may be invoked on `Overlay`, but there's no longer any cached overlay content for it to access.

I made an attempt to fix this by deferring the processing of the overlay's events by `component.update(..)` until `on_event` is invoked on the original element in 711dc9c26d424e41798d59fee9d13478c704ebae. This avoids rebuilding the cache in `Overlay::on_event`, so there should no longer be a risk of the overlay vanishing while it's still being operated on. It does change the order of operations a bit for components relative to other widgets in some cases though, so let me know whether or not you think this is a viable workaround.

hecrj:
> It does change the order of operations a bit for components relative to other widgets in some cases though, so let me know whether or not you think this is a viable workaround.

I'm not sure it will work properly. There is no guarantee that `Widget::on_event` will be called after `Overlay::on_event`. Therefore, this approach could produce weird behavior (overlays not closing until you move the mouse, etc.).

I think the best approach for now is to wrap the `overlay::Element` in an `Option` and simply drop overlay events if it turns out to be closed during the processing of an event batch.

nicksenger:
> I'm not sure it will work properly. There is no guarantee that `Widget::on_event` will be called after `Overlay::on_event`. Therefore, this approach could produce weird behavior (overlays not closing until you move the mouse, etc.).

That's a good point. It implicitly depends on the order in which methods are being called from `UserInterface::update` currently, which isn't really part of its contract.

> I think the best approach for now is to wrap the `overlay::Element` in an `Option` and simply drop overlay events if it turns out to be closed during the processing of an event batch. This is what you first implemented and I incorrectly removed, right? 😅

Yeah this is what I tried to do originally, but I was a bit concerned that returning an empty layout node from `Overlay::layout` in this case could cause problems. Thinking about it more though, since we won't be drawing anything from the overlay anyways, it shouldn't do any harm right? I've restored this approach in 44c0d75953984b0b31ac723bc040c016f572244d

hecrj:
> Thinking about it more though, since we won't be drawing anything from the overlay anyways, it shouldn't do any harm right?

It shouldn't be a problem. The layout nodes of each widget are opaque to `iced` and, as such, cannot be a part of any external contract.

hecrj (approved):
Thank you for this!

It's great to be able to make it work even when the current overlay implementation hinders composability!

Let's ship! 🎉

PR #1155, opened 2021-12-10
Decouple `image::Handle` from `iced_native` and lower trait bound on `text::Renderer::Font`

AlisCode:
As discussed on discord, the goal of this PR is to make `iced_native`'s widgets usable in more contexts (including potentially a Backend for a well-known Rust game engine, who knows :eyes: ) by : 

* Making the image renderer use an associated type `Handle` instead of relying on `iced_native`'s `image::Handle` which is impractical for custom backends 
* Lowering the bound of a Font handle from `Copy` to `Clone`   

This is a breaking change for `iced_native` because any custom renderer will need to add the associated type `Handle` to the `text::Renderer` implementation.

### DIFF
--- a/graphics/src/widget/image.rs
+++ b/graphics/src/widget/image.rs
@@ -13,6 +13,8 @@ impl<B> image::Renderer for Renderer<B>
 where
     B: Backend + backend::Image,
 {
+    type Handle = image::Handle;
+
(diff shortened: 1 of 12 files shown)


hecrj (approved):
Looks great! Thanks! 🎉

Issue #1156, opened 2021-12-11
404 on image in iced_web tour


n8henrie:
Following instructions at <https://github.com/iced-rs/iced/tree/master/web>, the image is blank on the page `An image that tries to keep its aspect ratio.`.

Console shows a 404 for `localhost:8000/images/ferris.png`; looks like the issue is that the tutorial has us run from `examples/` (the provided html includes `import init from "./tour/tour.js";`), so the path to `ferris.png` needs the `tour/` prefix.

PR #1157, opened 2021-12-11
Fix path to ferris.png

n8henrie:
Fixes https://github.com/iced-rs/iced/issues/1156


hecrj (approved):
Good catch. Thank you! 🙇

PR #1160, opened 2021-12-16
feat: add support to ARM devices (and older hardware)

### LINKED ISSUE #428: Iced on Raspberry PI 4 (almost working)
Hi, 

I'm trying to use iced on my Raspberry 4. So far I managed to make it compile. It even runs and the window is shown, but no frame is rendered. Eventually it exits with the following error:
```
thread 'main' panicked at 'Next frame: TimeOut', /home/pi/work/iced/wgpu/src/window/compositor.rs:110:21
```

Before I patched the `iced` dependencies it failed to compile:
```
error[E0308]: mismatched types
   --> /home/pi/.cargo/registry/src/github.com-1ecc6299db9ec823/font-kit-0.6.0/src/loaders/freetype.rs:804:20
    |
804 |                 x: vector.x() as i64,
    |                    ^^^^^^^^^^^^^^^^^ expected `i32`, found `i64`

error[E0308]: mismatched types
   --> /home/pi/.cargo/registry/src/github.com-1ecc6299db9ec823/font-kit-0.6.0/src/loaders/freetype.rs:805:20
    |
805 |                 y: -vector.y() as i64,
    |                    ^^^^^^^^^^^^^^^^^^ expected `i32`, found `i64`

error[E0308]: mismatched types
   --> /home/pi/.cargo/registry/src/github.com-1ecc6299db9ec823/font-kit-0.6.0/src/loaders/freetype.rs:808:21
    |
808 |                 xx: matrix.x() as i64,
    |                     ^^^^^^^^^^^^^^^^^ expected `i32`, found `i64`

error[E0308]: mismatched types
   --> /home/pi/.cargo/registry/src/github.com-1ecc6299db9ec823/font-kit-0.6.0/src/loaders/freetype.rs:809:21
    |
809 |                 xy: matrix.y() as i64,
    |                     ^^^^^^^^^^^^^^^^^ expected `i32`, found `i64`

error[E0308]: mismatched types
   --> /home/pi/.cargo/registry/src/github.com-1ecc6299db9ec823/font-kit-0.6.0/src/loaders/freetype.rs:810:21
    |
810 |                 yx: matrix.z() as i64,
    |                     ^^^^^^^^^^^^^^^^^ expected `i32`, found `i64`

error[E0308]: mismatched types
   --> /home/pi/.cargo/registry/src/github.com-1ecc6299db9ec823/font-kit-0.6.0/src/loaders/freetype.rs:811:21
    |
811 |                 yy: matrix.w() as i64,
    |                     ^^^^^^^^^^^^^^^^^ expected `i32`, found `i64`

error: aborting due to 6 previous errors

For more information about this error, try `rustc --explain E0308`.
error: could not compile `font-kit`.
```

I updated `iced_graphics`'  `font-kit` dependency to `0.9` and the error disappeared. I was able to build and execute examples, but rendering fails for some reason. Here's the typical output: https://gist.github.com/0x7CFE/3f29f6d7fccd67bd8804ad47c97daaa5 Also, please notice a ton of font errors.

Currently Raspberry Pi 4 has only [experimental but working Vulkan support](https://www.raspberrypi.org/blog/vulkan-update-now-with-added-source-code/). Yet I thought that for 2D stuff it should probably be enough.

I will be happy to hear any suggestions.

P.S.: Since iced is render agnostc, is it possible to build it on top of OpenGL? Maybe that would be easier than trying to debug alpha version of the Vulkan driver.

### LINKED ISSUE #931: OpenGL backend does not work on old OpenGL hardware (e.g. OpenGL ES 2)
The current OpenGL (glow) backend does not work on OpenGL ES 2 hardware (such as the PinePhone [Mali 400 GPU] in my case, with the latest GLSL version supported being 1.20).

However, it's not too far from being able to run there. By removing the use of instanced rendering and switching to older GLSL syntax, I managed to get it running on my PinePhone.

I have a fork at https://bics.ga/reivilibre/iced-mobile/src/commit/2dc42b4a22ffc31a83297bfed7641e3737a0a940 which I am using .

I'm not sure if I can contribute any changes back immediately (since there may be a performance cost for hardware which does support instanced rendering etc, but I felt compelled to leave a note here in case anyone may find it useful.

(as an aside: I have left a comment about my experience with Iced on the PinePhone on the mobile issue here: https://github.com/hecrj/iced/issues/302#issuecomment-869061787)

I wonder what the best approach might be here; would we prefer to have both a modern implementation and an old one and use the modern one if we can?

derezzedex:
This PR aims to add support to ARM devices and older hardware.
Since this depends on https://github.com/hecrj/glow_glyph/pull/5, the main idea of those changes are basically the same, so check that first.

### Changes
To maximize compatibility, this PR does a couple of things:
- Add targets for `armv7` and `aarch64`.
- Separate the `quad` pipeline into `core` and `compatibility`.
- Create and select different shaders based on context version.
- Add a `Setting` field to allow developers to try OpenGL ES first (this should be better for embedded devices).

There are also some "quality of life" changes:
- Add some logging to the changed areas (and `env_logger` to `game_of_life`).
- Export `iced_winit::conversion` in `iced_glutin`.
- Export `glow` and `winit` on their `iced_` version.
- ~~Fixes a problem with the `integration_opengl` example.~~ (#1181)
- Improve documentation related to current built-in renderers.

Closes #428.
Closes #931.


docker/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+### Raspberry Pi 2/3/4 (32 bits)
+```
+$ docker build -t iced-rs/armv7 -f Dockerfile.armv7-unknown-linux-gnueabihf .
+```
+
+### Raspberry Pi 3/4 (64 bits)
+```
+$ docker build -t iced-rs/aarch64 -f Dockerfile.aarch64-unknown-linux-gnu .
+```
```
hecrj, line 17: I think this `docker` folder is generic enough to become a standalone repository for all of the Rust ecosystem!

Instead of including it here, I believe we could create a GitHub Action to easily cross-compile any Rust project for Raspberry Pis. I'll gladly help you out with this!

Once we do that, we could then change the `build` workflow to additionally build binaries of the `todos` example for the Raspberry Pis and make sure we keep everything compatible moving forward.

glow/src/backend.rs
```diff
@@ -30,8 +31,60 @@ impl Backend {
...
+            version.is_embedded,
+        ) {
+            // OpenGL 3.0+
+            (3, 0 | 1 | 2, false) => (
+                format!("#version 1{}0", version.minor + 3),
+                format!(
+                    "#version 1{}0\n#define HIGHER_THAN_300 1",
+                    version.minor + 3
+                ),
+            ),
```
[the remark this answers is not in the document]
hecrj: Maybe it could be a part of the `program` module?

hecrj (changes requested):
Looks great! Really nice job here 🥳

In the future, once offscreen rendering lands, it'd be great to test the output of both pipelines (potentially in CI) to ensure minimal differences!

In any case, I left just a couple of suggestions! Let me know what you think.

glow/src/backend.rs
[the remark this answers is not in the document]
derezzedex: I've added a `Version` struct similar to what you described and further separated a new `Shader` struct from the `program::create`. Unfortunately, I can't simplify the `include_str!` since it doesn't accept `&'static str` (only string literals).

Check 424e1d3 and let me know what you think!

docker/README.md
derezzedex: Right now the `Dockerfile`s are tailored to `iced`, but if we wanted we could upload the images to 'Docker Hub'.
This could be useful to developers cross-compiling their `iced` apps, meaning they could build using our `Dockerfile`:
```Dockerfile
FROM iced-rs/iced:arm64
```

hecrj: I like the idea! I have created [an organization in Docker Hub](https://hub.docker.com/orgs/icedrs/repositories) and [a new `iced-rs/docker` repository](https://github.com/iced-rs/docker).

Let me know your username in Docker Hub and I will add you.

hecrj: I have added [a workflow in `iced-rs/docker` to build the images in GitHub CI and push them to Docker Hub](https://github.com/iced-rs/docker/blob/4a5994498743955fd730048ec0fe3d672b061afa/.github/workflows/publish.yml).

Also, I have removed the `docker` folder in 2b3528e9929e1b4bb563348cd5f694b83d0caec4 and added a `todos_raspberry` job to the `build` workflow in 522368e8af6947ea0676f62bf1ae5f68acb2058d. [Everything seems to work nicely!](https://github.com/iced-rs/iced/runs/4880885146?check_suite_focus=true)

It'd be great if you could try the binaries from this GitHub CI run in the Pis: https://github.com/iced-rs/iced/actions/runs/1722948462

glow/src/backend.rs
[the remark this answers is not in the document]
hecrj: Looks awesome! :ok_hand:

hecrj (approved):
Great work, Richard!

If you agree with my latest changes, I think this ship is ready to sail! :ship:

PR #1164, opened 2021-12-20
Update wgpu 0.12

PolyMeilex:
~~Blocked by wgpu_glyph: https://github.com/hecrj/wgpu_glyph/pull/82 or https://github.com/hecrj/wgpu_glyph/pull/83~~
Blocked by iced winit fork: We have to update winit to new version of raw-window-handle


genusistimelord:
You should add the winit fix with this otherwise the older wgpu will make it fail.

PolyMeilex:
> You should add the winit fix with this otherwise the older wgpu will make it fail.

I already mentioned this as a blocker in the description of the PR, I don't know what iced fork of winit adds or how it should be rebased, so I'm just mentioning it as a blocker instead of doing it myself. I don't use it, so I never had a reason to look into it.

hecrj (approved):
Thank you! 🙇 

I have rebased on top of #1182 and everything seems to be working.

PR #1168, opened 2021-12-23
Change GitHub namespace to new org to prevent link-rot

LordRatte:


LordRatte:
Feel free to close this if it isn't helpful

I see there are a lot of other `https://github.com/hecrj/iced` URLs that could be changed to `https://github.com/iced-rs/iced`
and I'd be happy to change those before merge if it will be helpful.

hecrj:
Sounds good! Go for it!

LordRatte:
So, just for transparency, this is the command I ran:

`find . -type f -not -path './.git/*' -print0 | xargs -0 sed -i 's;github.com/hecrj;github.com/iced-rs;g'`

I did a spot-check on links (especially issues etc.) and everything seems fine. I wasn't sure if it's appropriate to change historical records like `CHANGELOG.md`, so if that or anything else needed to stay the same I can revert them.

hecrj (approved):
Awesome! Thank you 🙇 

I've run the command again with the name of the repository as well. This way, we avoid rewriting some of the URLs that are pointing to my personal repos (like [`coffee`](https://github.com/hecrj/coffee)).

Issue #1170, opened 2021-12-26
Load icon from a file?


daladim:
Thanks for iced.

I'm using iced to make a native (Windows and Linux if that matters) app, and I'd like to give its window an icon.
I managed to add an icon using [Icon](https://docs.rs/iced/latest/iced/window/icon/struct.Icon.html) and passing it to my app's [`settings`](https://docs.rs/iced/latest/iced/window/struct.Settings.html#structfield.icon).

Is there a way (I haven't found in the doc, but I may not have been looking at the right places) to load an icon from a file, instead of giving a pre-computed array of pixels?

Especially, I'd like to load an SVG icon. Since you already provide [SVG widgets](https://docs.rs/iced/latest/iced/widget/svg/struct.Svg.html). I haven't read its source code yet, but maybe this could be somehow reused so that we could provide an `Icon::from_file()`, or at least `Icon::from_svg()`.
What do you think? Would that make sense?
Is there an easy way to rasterize an SVG into a pixel array already, and would it be better suited to build the array first, then give it to `Icon::from_rgba`?

I could help and make the MR myself, once you tell me what solution sounds the most sensible to you.
Thanks

daladim:
How does [`svg::Handle::data()`](https://docs.rs/iced/latest/iced/widget/svg/struct.Handle.html#method.data) relates to an RGBA array?

13r0ck:
This is how I did it in my project

1. [first create the icon data](https://github.com/13r0ck/bathtub/blob/main/src/main.rs#L39)
2. [pass that into the iced window](https://github.com/13r0ck/bathtub/blob/main/src/main.rs#L47)

PR #1174, opened 2021-12-28
Added convenience functions for window::icon::Icon

daladim:
See #1170 

Since the `image` dependency was already an (optional) dependency of `iced`, why not provide these convenience functions that save a few lines of code in applications that need an icon

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ default = ["wgpu"]
 # Enables the `iced_wgpu` renderer
 wgpu = ["iced_wgpu"]
 # Enables the `Image` widget
-image = ["iced_wgpu/image"]
+image = ["iced_wgpu/image", "image_rs"]
 # Enables the `Svg` widget
 svg = ["iced_wgpu/svg"]
 # Enables the `Canvas` widget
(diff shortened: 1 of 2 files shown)


13r0ck:
Thank you for the PR! 

Could you make it so that your convenience function is only available if the iced image feature is enabled? 

The image crate is quite large, and iced tries to give users the option for simplicity, but tries to be as light as it can be by default. This is the reason that image isn't already included by default.

daladim:
Well, I had already gated this behind an `image_rs` feature I had just added, for the very reasons you described.
I've changed this PR so that this optional feature is now tied to the already-existing `image` feature.

I'm afraid I cannot remove the `image_rs` feature and keep only the `image` feature, since `image` is both the name of a crate and a feature. To do so, we would need:
* either to rename the `image` feature so that it does not clash with the crate name
* wait for https://github.com/rust-lang/cargo/issues/5565 to be ever stabilized
* some other solution I would not be aware of?

daladim:
Also, I'm not sure how docs.rs generates its doc. Will these functions be shown in https://docs.rs/iced/latest/iced/?

There is already a 
```
[package.metadata.docs.rs]
rustdoc-args = ["--cfg", "docsrs"]
features = ["image", "svg", "canvas"]
```
...but I'm not sure how this will render the functions that are gated behind the `image_rs` feature instead of the `image` feature

daladim:
I did not manage to build the doc locally though the `docs.rs` project to test the `image_rs`-gated functions are visible, but since this feature is implied by the `image` one, it should be fine.

I think this MR is ready to be merged. What's your opinion on this?
Thanks

daladim:
I'm finally having a green CI on this PR.

Is this PR OK for you? Or do you need anything else from me before merging it?

Thanks

hecrj (approved):
Looks great! Thanks :bow:

PR #1175, opened 2021-12-29
Allow Sandbox applications to exit.

EkardNT:
I noticed that when using the `Sandbox` its not possible to graceful tell iced to quit. Although I understand that `Sandbox` is a simplified alternative to `Application`, it seemed to me that `Sandbox` users may still want the ability to exit.

To make it possible for `Sandbox` programs to quit, this PR adds a `fn should_exit(&self) -> bool` method to the `Sandbox`  trait. The `Sandbox` trait defines the default implementation of `should_exit` to return false, just like `Application` does, so `Sandbox` users who do not care about the ability to exit should notice no difference.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,6 +71,7 @@ members = [
     "examples/custom_widget",
     "examples/download_progress",
     "examples/events",
+    "examples/exit",
     "examples/game_of_life",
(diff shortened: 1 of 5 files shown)


hecrj (approved):
Looks great! Thank you 🙇

Issue #1087, opened 2021-10-16
Memory leak in OpenGL integration example


ELginas:
When moving cursor in OpenGL integration example, RAM usage rises per each user event (mouse move, key press, etc.) and it never releases that memory.
Tested on debug/release modes on master branch.
System: Windows 10 (build 19043.1288)
GPU: Intel(R) HD Graphics 530 (integrated)

hecrj:
Fixed by #1181.

PR #1181, opened 2022-01-02
Fix memory leak in opengl integration example

feylinara:
Fixes issue #1087, where any user input in the opengl integration example would leak memory. This was due to creating a new VAO and program for every call to the event handler.


hecrj (approved):
Good catch! Thank you 🥳

PR #1189, opened 2022-01-10
Conditionally allow mainline winit for external integration purposes

tasgon:
I'd like to integrate iced UI rendering in other projects, however the usage of the forked winit can break some dependency handling. I've made it so that users can toggle between it and the mainline version through a feature (defaulting to the forked version).

### DIFF
--- a/winit/Cargo.toml
+++ b/winit/Cargo.toml
@@ -20,8 +20,6 @@ thiserror = "1.0"
 
 [dependencies.winit]
 version = "0.26"
-git = "https://github.com/iced-rs/winit"
-rev = "02a12380960cec2f351c09a33d6a7cc2789d96a6"
 
 [dependencies.iced_native]
 version = "0.4"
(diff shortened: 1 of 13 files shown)


tasgon:
Just tested your branch, and it seems to work for me. Out of curiosity, are you planning on publishing this version? If it wouldn't be too much of a burden, would it be possible to have versions of this crate in sync with new versions of wgpu and winit on crates.io? If you'd like, I can assist with the updating process as necessary.

hecrj:
Yes, I have plans to publish `0.4` soon. But before that, I'd like to have the new website ready and the "Architecture" chapters of [the book](https://book.iced.rs/) finished.

At that point, it'll be just a matter of coming up with a proper announcement with a nice change log. [A lot has happened](https://github.com/iced-rs/iced/milestone/7?closed=1) since [the last release](https://github.com/iced-rs/iced/pull/806)!

I'd appreciate any help on these fronts!

PR #1196, opened 2022-01-15
Add application feature, which allows to build freestanding wgpu app

aentity:
This PR adds new "application" feature, which is default. When user builds with:

```
cargo build --no-default-features --features "wgpu"
```

it allows to user to build "freestanding" app, with all the nice building blocks of iced, but user responsible for window creation, and rendering.

This also fixes building for iOS, for example, because user can turn off application, and put together pieces they need, similar to integration example.

We can also reduce other dependencies in parts of the workspace I think -- make them optional as example--, to decrease compile time of iced and codegen in this case, but I leave it here for now, for @hecrj to consider. We can do such things in future if desired.

Thank you for your time!

I think this is a better answer than #1190

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,9 +13,7 @@ categories = ["gui"]
 resolver = "2"
 
 [features]
-default = ["wgpu", "application"]
-# Enables the building the application modules
-application = ["iced_winit/application"]
+default = ["wgpu"]
 # Enables the `Image` widget
 image = ["iced_wgpu/image", "image_rs"]
 # Enables the `Svg` widget
@@ -98,7 +96,7 @@ iced_core = { version = "0.5", path = "core" }
 iced_futures = { version = "0.4", path = "futures" }
 iced_native = { version = "0.5", path = "native" }
 iced_graphics = { version = "0.3", path = "graphics" }
-iced_winit = { version = "0.4", path = "winit" }
+iced_winit = { version = "0.4", path = "winit", features = ["application"] }
 iced_glutin = { version = "0.3", path = "glutin", optional = true }
 iced_glow = { version = "0.3", path = "glow", optional = true }
 thiserror = "1.0"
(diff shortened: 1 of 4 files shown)


aentity:
@hecrj what do you think of this pr, can we be reviewed?

aentity:
@hecrj hello i would love see this merged into crates release. perhaps 0.4.3? this is very useful to allow crates release supported to have ios/no application build. i will do whatever is needed to see merged, is there some blocking issue i can help resolve? Thank you

I can also rebase onto 0.4 branch if it is easier for you, just tell me what to do :)

hecrj (approved):
Thanks! :tada: 

I think we can just add the `application` feature to `iced_winit` for now.

aentity:
@hecrj I do not see how this can work, here you set feature unconditional: https://github.com/iced-rs/iced/pull/1196/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R99

so user cannot turn off by saying `default-features = false` when using iced-winit, because iced turns it on unconditional.

i think my original PR is correct...

if you try to build it will fail same as usual:

```
cargo check --target aarch64-apple-ios --no-default-features
 
error[E0432]: unresolved import `winit::platform::run_return`
   --> winit/src/application.rs:744:30
    |
744 |         use winit::platform::run_return::EventLoopExtRunReturn;
    |                              ^^^^^^^^^^ could not find `run_return` in `platform`

error[E0599]: no method named `run_return` found for struct `EventLoop<T>` in the current scope
   --> winit/src/application.rs:746:28
    |
746 |         let _ = event_loop.run_return(event_handler);
    |                            ^^^^^^^^^^ method not found in `EventLoop<T>`

Some errors have detailed explanations: E0432, E0599.
For more information about an error, try `rustc --explain E0432`.
```

nikolajevs86:
Hi @aentity I am trying to build app for ios  with `default-features = false` and wgpu feature but still is facing no run_return issue. Maybe you can share if you was able to do it and which iced version do you use?

PR #1205, opened 2022-01-20
Fix component overlay position

cossonleo:
The position change from component position to widget with overlay position.

### DIFF
--- a/lazy/src/component.rs
+++ b/lazy/src/component.rs
@@ -263,14 +263,16 @@ where
                     .build(),
                 );
 
-                cache.as_ref().unwrap().borrow_overlay().is_some()
+                cache
+                    .as_ref()
+                    .unwrap()
+                    .borrow_overlay()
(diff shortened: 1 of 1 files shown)


lazy/src/component.rs
```diff
@@ -263,14 +263,16 @@ where
...
+                    .map(|overlay| overlay.position())
             });
 
-        has_overlay.then(|| {
-            overlay::Element::new(
-                layout.position(),
-                Box::new(Overlay { instance: self }),
-            )
+        has_overlay.map(|pos| {
+            overlay::Element::new(pos, Box::new(Overlay { instance: self }))
```
hecrj, line 275: Let's avoid unnecessary abbreviations!

```suggestion
        has_overlay.map(|position| {
            overlay::Element::new(position, Box::new(Overlay { instance: self }))
```

hecrj (changes requested):

PR #1217, opened 2022-01-22
Fix links

SynergyBest:
I changed old links to new ones, like wgpu-rs to wgpu, or hecrj to iced-rs, etc.
I also changed file links to relative ones which makes them more portable (where possible.)
Enjoy!


SynergyBest:
I also optimized images/reduced their size with [ImgBot](https://imgbot.net/)

hecrj (approved):
Awesome! I didn't notice so many links were broken 😅 

Thank you 🙇

SynergyBest:
Of course! Any eta on merge?

SynergyBest:
Awesome! There are actually still a few links that need to be fixed (I think it was caused by the force-push) so I’ll get that done when I get home.

PR #1219, opened 2022-01-23
ContentFit support for images

### LINKED ISSUE #1014: How do I stretch an image
I want to create an image without keeping the aspect ratio

is that possible with iced

### LINKED ISSUE #1179: [Feature Request] Control image fit / scaling in image widget
As of right now, the `Image` widget displays an image downscaled to fit within the widgets bounds.

However, as mentioned in #885, some users may want their image upscaled within the widget bounds as well.  Beyond this, however, some users may want an image to completely cover (as opposed to just fit within) the image bounds, cropping out the edges.  This is especially useful for images meant to act as backgrounds, headers, or other places where the shape of the bounding box of the image is important.  The user in issue #1014 represents another possible usecase where the image should be stretched to fit without preserving the aspect ratio.

I'd like to propose a feature similar to [`object-fit` in css][1], where the user can select their preferred behavior by passing an enum to the Image widget.  I think that a feature like this would make designing UIs with different behaviors easier for users who might otherwise need to manually do the calculations to scale the image to fit their screen.

Thank you for your time and consideration!

[1]: https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

Alch-Emi:
Closes #1179
Closes #1014

Adds support for various image fit modes, documented and explained in the new [`fit`][1] module and explained in #1179.

Additionally, the `tour` example has been updated to show off the new functionality.  To this end, an additional image was added to the `tour` example which satisfies the need for a background (to clearly denote the bounds of the image) and a size smaller than the screen (so that the special behavior of the `ScaleDown` mode can be seen).

Breaking changes:
- Images now take up space slightly differently, although defaults are a very close approximation of original behavior
- The public `fit` field was added to the `Image` struct
- Images on web now follow their `height` property properly, if set

Non-breaking Changes:
- Addition of `ContentFit` to `iced_core` (and appropriate re-exports)
- Update of the `tour` example
- Addition of the `fit(ContentFit)` method to the `Image` widget
- Adjustments to the native rendering of images to support the new `ContentFit`s

Recommended future work:
- This PR does not provide any logic pertaining to the alignment of images within the space alotted to them, and chooses to always align images into the top-left corner, as this behavior is closest to the original behavior.  However, varied alignment behavior (or at least the ability to center the content within its frame) will doubtlessly be desired.

https://user-images.githubusercontent.com/38897201/150687061-e16d5f94-04f5-4709-bbdc-75667621bf37.mp4

[1]: https://github.com/iced-rs/iced/blob/91fd66181deef3288afad94ca435ba5401edcf72/core/src/fit.rs


examples/tour/src/main.rs
```diff
@@ -699,19 +733,36 @@ impl<'a> Step {
...
+    height: u16,
+    fit: Option<ContentFit>,
+) -> Container<'a, StepMessage> {
     Container::new(
         // This should go away once we unify resource loading on native
         // platforms
         if cfg!(target_arch = "wasm32") {
-            Image::new("tour/images/ferris.png")
+            if trans {
+                Image::new("tour/images/trans_ferris.png")
```
hecrj, line 747: I think we should keep the project decoupled from any symbolism related to gender and/or sex!

Maybe we could use the project logo instead?

core/src/fit.rs
```diff
@@ -0,0 +1,119 @@
+//! Control the fit of some content (like an image) within a space
+
+use crate::Size;
+
+/// How the image should scaled to fit the bounding box of the widget
```
hecrj, line 5: should scale*

core/src/fit.rs
```diff
@@ -0,0 +1,119 @@
...
+    ///
+    /// This is best for when you've sized the image yourself
+    None,
+
+    /// Scale the image down if it's too big for the space, but never scale it up
+    ///
+    /// This works much like [`Contain`](Self::Contain), except that if the
+    /// image would have been scaled up, it keeps its original resolution to
+    /// avoid the bluring that accompanies upscaling images
+    ScaleDown,
```
hecrj, line 62: Awesome docs! Thank you :bow: 

Nit: We are missing some `.` in many paragraphs.

hecrj (changes requested):
This is great! Thank you :partying_face:

I just have a couple of small suggestions!

examples/tour/src/main.rs
Alch-Emi: if i may, Ferris is [cannonically][1] [trans & nb][2], so any ferris is a trans ferris.  I'd be curious why you're concerned about explicitly including a trans flag tho?  that said, you're the maintainer, and if you really want the trans flag gone, ill take it out if you say the word.

[1]: https://twitter.com/whoisaldeka/status/1165147320964415489
[2]: https://www.rust-lang.org/learn/get-started#ferris

Alch-Emi:
Alright, that should be everything!  Any other thoughts?

Alch-Emi:
Should I include a path for the `iced_pure` image widget in this PR too now that that exists, or should that be a separate PR?

hecrj (approved):
Looks great! :partying_face: Thank you :bow: 

I made some changes here and there, since I realized we could add `ContentFit` support for the `Svg` widget as well!

Let me know what you think and we can merge this! :ship: 

> Should I include a patch for the iced_pure image widget in this PR too now that that exists, or should that be a separate PR?

Don't worry! The changes there haven't landed yet. I can deal with the conflicts :)

Alch-Emi:
Looks great!  My two concerns with the large & transparent Ferris image are that, because she doesn't have a background, it's hard to tell where the border is, and that it can't showcase the `OnlyScaleDown` option, since it can't scale up.  The border wouldn't normally be a huge deal, and to be fair, it still isn't, but an important part of understanding the image modes is understanding how they affect the border.  At the very least, I think we need to use a smaller Ferris, and preferably one with a non-white background.

hecrj:
I think `Contain`, `Cover`, and `Fill` are easy to grasp.

I have removed the other two from the `tour`, since we don't really need to be that exhaustive. Maybe we can come up with a dedicated `image` example to showcase all of the `ContentFit` variants properly.

PR #1225, opened 2022-01-27
Add line dash API

tarkah:
This PR adds `LineDash` configuration to `Stroke` and uses `RepeatedPattern` from `lyon_algorithms` to closely mimic the HTML Canvas APIs for adding stroked line dash styling.

I've updated the Solar System example to use a dashed line pattern for the orbit, which I think showcases this well.

A couple comments / questions:

- `Stroke` is no longer `Copy`, I see no way around this
- `LineDash` is only considered if `segments` is not empty. Should we instead use an `Option`?
- I've added no rail-guards to validate `segments`, as things may not render properly if negative or non-finite values are used. It is up to the caller to define this properly.
- If an odd number of segments are provided, they are concatenated again to the end. This is in line w/ the html canvas spec.

Relevant canvas API spec:
https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-setlinedash

Screenshot:
image

### DIFF
--- a/examples/clock/src/main.rs
+++ b/examples/clock/src/main.rs
@@ -100,14 +100,14 @@ impl canvas::Program<Message> for Clock {
 
             let wide_stroke = Stroke {
                 width: thin_stroke.width * 3.0,
-                ..thin_stroke
+                ..thin_stroke.clone()
             };
 
             frame.translate(Vector::new(center.x, center.y));
 
             frame.with_save(|frame| {
                 frame.rotate(hand_rotation(self.now.hour(), 12));
-                frame.stroke(&short_hand, wide_stroke);
+                frame.stroke(&short_hand, wide_stroke.clone());
             });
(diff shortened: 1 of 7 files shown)


hecrj (approved):
Looks great!

I have made a couple of changes:

- Replaced the `Vec` for `segments` with a simple `slice` by adding a lifetime to `LineDash` and `Stroke` in f56c8a7361ceb215bce68e88bd6ce402e2694693. This way, users can avoid reallocating on every draw and we get to keep the `Copy` trait!
- Removed the direct `lyon_algorithms` dependency in bace264bfe5e3cb5046867bd411e54969a637c79, as we can access it through `lyon` already.

Let me know what you think and we can merge this 🥳

tarkah:
Changes look great, good call on the short lifetime and slice usage!

PR #1227, opened 2022-01-31
`color!` macro

### LINKED ISSUE #375: Add hex constructors for Color
It would be nice to add a feature to create colors from a single hexadecimal number like this:
```rust
let color1 = Color::from_hex_rgb(0x3280c8);
let color2 = Color::from_hex_rgba(0x3280c86f);
```
The code could be something like this:
```rust
pub fn from_hex_rgb(hex: u32) -> Color {
    let r = (hex & 0xff0000) >> 16;
    let g = (hex & 0xff00) >> 8;
    let b = (hex & 0xff);

    Color {
        r: f32::from(r) / 255.0,
        g: f32::from(g) / 255.0,
        b: f32::from(b) / 255.0,
        a: 1.0f32,
    }
}

pub fn from_hex_rgba(hex: u32) -> Color {
    let r = (hex & 0xff000000) >> 24;
    let g = (hex & 0xff0000) >> 16;
    let b = (hex & 0xff00) >> 8;
    let a = (hex & 0xff);

    Color {
        r: f32::from(r) / 255.0,
        g: f32::from(g) / 255.0,
        b: f32::from(b) / 255.0,
        a: f32::from(a) / 255.0,
    }
}
```

LordRatte:
I spotted #375 and thought maybe a macro could do the trick.

Here is an idea of how I thought it could work but I'm very new to this so there could be any number of problems.

I got the idea from [macroquad](https://docs.rs/macroquad/latest/macroquad/macro.color_u8.html)

Closes #375.

### DIFF
--- a/core/src/color.rs
+++ b/core/src/color.rs
@@ -134,6 +134,39 @@ impl From<[f32; 4]> for Color {
     }
 }
 
+/// # Examples
+///
+/// ```
+/// # use iced_core::{Color, color};
+/// assert_eq!(color!(0, 0, 0), Color::from_rgb(0., 0., 0.));
+/// assert_eq!(color!(0, 0, 0, 0.), Color::from_rgba(0., 0., 0., 0.));
+/// assert_eq!(color!(0xffffff), Color::from_rgb(1., 1., 1.));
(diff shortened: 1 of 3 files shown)


hecrj (approved):
Looks cool! :tada: 

I think this should fit right in, since we introduced some macro helpers in #1393.

PR #1230, opened 2022-02-03
Pass ignored events to program

kaimast:
This pull request aims to make the issues addressed in #408 a little easier to solve. Currently, you have to re-implement `program::State` which seems like a bad idea just to capture events.

I think in the future iced needs to receive proper focus handling, but for now this change would be very helpful for me at least.

### DIFF
--- a/native/src/program/state.rs
+++ b/native/src/program/state.rs
@@ -1,8 +1,9 @@
 use crate::application;
+use crate::event::{self, Event};
 use crate::mouse;
 use crate::renderer;
 use crate::user_interface::{self, UserInterface};
-use crate::{Clipboard, Command, Debug, Event, Point, Program, Size};
+use crate::{Clipboard, Command, Debug, Point, Program, Size};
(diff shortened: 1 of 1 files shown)


native/src/program/state.rs
```diff
@@ -153,7 +152,17 @@ where
             self.cache = Some(user_interface.into_cache());
 
             Some(commands)
+        };
+
+        // Allow for the program to handle any uncaptured event
```
kaimast, line 157: I needed to move this down here to make the borrow checker happy

hecrj (approved):
Thanks! :bow: 

I think we can simply return the ignored events as a result of `program::State::update` instead of changing the `Program` trait as well.

I just rebased and did exactly that in `c9b8dbdb7a15a7a797eec78d1a42783e0ea0f8d6`.

kaimast:
Either is fine. Thanks so much for merging!

PR #1231, opened 2022-02-05
Allow the creation of a clipboard without a window.

TannerRogalsky:
I've been testing the recent wasm32 "native" shell changes and they work marvelously! It would be useful to me if we could create the "unavailable" clipboard without needing a `Window`.

The motivation is that I prefer to (on web) use iced without winit's event loop because of how implements the "does not return" behaviour via an uncaught exception. As such I also forgo their Window object.

Everything else is working exceptionally well! Thank you!

### DIFF
--- a/winit/src/clipboard.rs
+++ b/winit/src/clipboard.rs
@@ -26,6 +26,14 @@ impl Clipboard {
         Clipboard { state }
     }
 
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Seems reasonable! Thank you!

PR #1252, opened 2022-02-18
Background image

daladim:
Following #1235, here is my attempt at implementing background images. \
That was relatively easy, since the rendering of images existed already. \
This PR also includes an example of background image in the `tour` example, which has been correctly tested on Windows and Linux.

This is a basic support, background images are always stretched, and I think there should be more options in the future. I have left comments in the `BackgroundImageAlignment` enum to give such ideas.

I am not sure how this would affect performance, because I haven't understood yet how/if the `image::Handle` caches data after the first time it reads and decodes an image from the disk.


examples/tour/src/main.rs
```diff
@@ -591,6 +591,26 @@ impl<'a> Step {
...
+                "Images can also be used as background for other items:",
+            ))
+            .push(
+                Container::new(
+                    Container::new(
+                        Column::new()
+                            // We're cheating here. Since Background::Image does not (yet?) have advanced resizing options, let's include content that has roughly the same aspect ratio as the background image
+                            .push(Text::new("A first line"))
+                            .push(Text::new("A second line"))
+                            .push(Text::new("A third line")),
```
hecrj, line 605: Given the current implementation, I believe quads and triangle meshes will still be renderered under the image background. You can try pushing a `Button` here to see the issue.

As I mentioned in #1235, there is a bunch of limitations in our rendering pipelines that make background images hard.

daladim: Hmmm, that's a nice catch.

image

I'm not sure I'm knowledgeable enough to help here (graphical rendering is not my cup of tea). But maybe I can have a look if you're mentoring me or tell me "generally speaking" what should be done (and where).

Otherwise, do you think the first commits (that move `image::Handle` to `iced_core`) are worth being merged into master, even if we drop the (broken) commit that adds background images? Would this slight architectural change would benefit to the rest of us?

hecrj:
Given the limitation I mentioned earlier, I think it's better to wait until we figure out the best approach to tackle all of the issues I mentioned in #1235.

There are improvements coming to the renderer that should help us with layering and composition.

PR #1262, opened 2022-02-22
Fix: overlay layout for `Responsive`

nicksenger:
The layout node passed to a child of `Responsive` may be incorrect in some cases. This PR fixes the overlay.

### DIFF
--- a/lazy/src/responsive.rs
+++ b/lazy/src/responsive.rs
@@ -178,7 +178,7 @@ where
             let content_layout = state.layout(layout);
 
             match content {
-                Content::Pending(_) => false,
+                Content::Pending(_) => None,
                 Content::Ready(cache) => {
                     *cache = Some(
                         CacheBuilder {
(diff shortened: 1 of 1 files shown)


native/src/widget/pane_grid/content.rs
```diff
@@ -258,7 +258,7 @@ where
                 None => self.body.overlay(children.next()?, renderer),
             }
         } else {
-            self.body.overlay(layout, renderer)
+            self.body.overlay(layout.children().last()?, renderer)
```
hecrj, line 261: This seems incorrect, as we are returning the `body` overlay directly in `layout` when no title bar is present:

https://github.com/iced-rs/iced/blob/a536f5ee0e3b17d4962526038099197ed6ccdad1/native/src/widget/pane_grid/content.rs#L153-L155

lazy/src/responsive.rs
```diff
@@ -190,14 +190,19 @@ where
                         .build(),
                     );
 
-                    cache.as_ref().unwrap().borrow_overlay().is_some()
+                    cache
+                        .as_ref()
+                        .unwrap()
+                        .borrow_overlay()
+                        .as_ref()
+                        .map(|overlay| overlay.position())
```
hecrj, line 198: Ah, good catch!

hecrj (changes requested):

hecrj (approved):
Thanks!

PR #1264, opened 2022-02-23
Add clip to canvas

tarkah:
This adds the ability to draw on the canvas within a clip primitive. Similar to `with_save`, it "restores" itself after returning (we never actually modify `&mut self / Frame`). This is very useful when drawing on the canvas with multiple regions that have distinct geometry that can bleed over into the other regions.

Not 100% sure what the API should look like, but it made the most sense to supply a translation and size to define the clipped `Rectangle` relative to the `Frame`.

### DIFF
--- a/graphics/src/widget/canvas/frame.rs
+++ b/graphics/src/widget/canvas/frame.rs
@@ -244,6 +244,30 @@ impl Frame {
         self.transforms.current = self.transforms.previous.pop().unwrap();
     }
 
+    /// TODO...
+    #[inline]
+    pub fn with_clip(
+        &mut self,
+        translation: Vector,
+        size: Size,
+        f: impl FnOnce(&mut Frame),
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Cool! Thank you 🎉 

I have made some changes:

1. Replaced both `Vector` and `Size` with a `Rectangle` that represents the clipping region.
2. Changed the bounds of the new `Frame` to the `region` bounds.
3. Leveraged `Primitive::Translate` instead of `Frame::translate`, as it should be a bit more performant (`Frame::translate` transforms all the drawn paths, while `Primitive::Translate` defers the translation to the vertex shader on the GPU).

Hope I didn't break anything. Let me know what you think!

tarkah (commented):
This works perfect and the API is more clear. Thanks!

Issue #156, opened 2020-01-14
Support for text styling / rich text


Toqozz:
There is currently no way to style individual parts of text within a text block -- you can only change the style of the entire block.

I've seen a couple different ways of doing this on the user end:
- Markup in strings, like in Pango/GTK+ (https://developer.gnome.org/pygtk/stable/pango-markup-language.html)
- Setting state when inserting text into the field, like in Java/Swing (http://www.java2s.com/Code/Java/Swing-JFC/StyledText.htm)

From a user perspective, I personally prefer markup strings in the text.

`glyph_brush` already supports text styling using `glyph_brush_layout`, so implementation should just be a matter of parsing the text in whatever format to a `Vec<SectionText>`.

hecrj:
I think we should tackle this together with #34, once we figure out font fallback and text shaping.

String markup will force us to implement some kind of parser while setting state is quite a side-effect heavy approach.

I personally like the approach of [`elm-ui`](https://package.elm-lang.org/packages/mdgriffith/elm-ui/latest/Element#paragraph). It's explicit, self-contained, and composable.

Toqozz:
I like the elm-ui example, but I worry that it won't be as ergonomic in rust.  We'll probably just end up having people essentially write `SectionText` directly:

```rust
let glyphs = Layout::default().calculate_glyphs(
    &fonts,
    &SectionGeometry {
        screen_position: (150.0, 50.0),
        ..SectionGeometry::default()
    },
    &[
        SectionText {
            text: "hello ",
            scale: Scale::uniform(20.0),
            ..SectionText::default()
        },
        SectionText {
            text: "glyph_brush_layout",
            scale: Scale::uniform(25.0),
            font_id: FontId(1),
            color: [0.0, 1.0, 0.0, 1.0],
        },
    ],
);
```

hecrj:
We currently use the builder pattern to configure widgets. I'd use the same strategy here.

However, we are not quite close to implementing this yet. The API could change considerably before we get to tackle this.

PR #1276, opened 2022-03-04
Add basic support for text highlighting

Pixelstormer:
Adds support for changing the background colour of various parts of text, to enable things like highlighting search results, marking outstanding regions of text, etc.

You could achieve a similar effect before this PR by splitting your text into multiple `Text` structs and nesting the parts you want to highlight in a `Container` with a background colour set, but this is not a viable solution as doing so interferes with text layouting, as demonstrated in the below videos.

Text highlighting using the aforementioned containers-based method (Notice how the text fidgets around as the highlighting changes):
```rust
/* Snip */
let splitter = &self.highlight_input_value;
                            
let mut row = Row::new();
let mut split = s.split(splitter);
                            
if let Some(first) = split.next() {
    row = row.push(Text::new(first));
}
                            
for s in split {
    row = row
        .push(
            Container::new(Text::new(splitter))
                .style(ContainerStyle::red()),
        )
        .push(Text::new(s));
}
/* Snip */
```

https://user-images.githubusercontent.com/28542452/156685350-6d1837b7-fea1-4eaa-b445-6bd1c55b9818.mp4

Text highlighting using the `highlight` method introduced by this PR:
```rust
/* Snip */
let mut text = Text::new(s);
for (index, _) in s.match_indices(&self.highlight_input_value) {
    text = text.highlight(
        index,
        index + self.highlight_input_value.len(),
        Color::from_rgb8(0xff, 0xc0, 0xcb),
    );
}
/* Snip */
```

https://user-images.githubusercontent.com/28542452/156685426-c456722b-18e9-433f-9bb2-368e67aebad4.mp4

### DIFF
--- a/native/src/widget/text.rs
+++ b/native/src/widget/text.rs
@@ -160,6 +160,17 @@ where
         _cursor_position: Point,
         _viewport: &Rectangle,
     ) {
+        draw_highlights(
+            renderer,
+            layout,
+            &self.content,
+            self.font.clone(),
+            self.size,
+            &self.highlights,
+            self.horizontal_alignment,
+            self.vertical_alignment,
+        );
+
         draw(
             renderer,
             style,
@@ -168,7 +179,6 @@ where
             self.font.clone(),
             self.size,
             self.color,
-            &self.highlights,
             self.horizontal_alignment,
             self.vertical_alignment,
         );
@@ -193,7 +203,6 @@ pub fn draw<Renderer>(
     font: Renderer::Font,
     size: Option<u16>,
     color: Option<Color>,
-    highlights: &[Highlight],
     horizontal_alignment: alignment::Horizontal,
     vertical_alignment: alignment::Vertical,
 ) where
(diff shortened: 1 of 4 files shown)


native/src/widget/text.rs
```diff
@@ -186,9 +208,33 @@ pub fn draw<Renderer>(
...
+        let width_before_start =
+            renderer.measure_width(&content[..start], size, font.clone());
+
+        let width =
+            renderer.measure_width(&content[start..end], size, font.clone());
+
+        let quad = renderer::Quad {
+            bounds: Rectangle {
+                x: x + width_before_start,
+                y,
```
hecrj, line 223: It seems highlighting will not work with a `Text` of multiple lines?

hecrj:
I think this is related to #156.

We are not only interested in highlighthing support, but also changing color, font, size, and other text properties without affecting layout.

native/src/widget/text.rs
Pixelstormer: Yeah, I should clarify that this a very basic and naive implementation thus far; there's a lot of trivial cases that break it.

native/src/widget/text.rs
```diff
@@ -197,6 +235,158 @@ pub fn draw<Renderer>(
...
+        };
+
+        // Measure height of lines up until the line that contains the start of the highlight
+        let (_, mut height_offset) =
+            renderer.measure(above_lines, size, font.clone(), Size::INFINITY);
+
+        // If the highlight crosses over multiple lines, draw a seperate rect on each line
+        // BUG: This ignores single `\r` but Iced's text layouting does not (See above)
+        // BUG #2: Text wrapping caused by the text not being given wide enough bounds is not handled at all
+        //         (And furthermore it currently _can't_ be handled because there's no way to get information about it)
```
Pixelstormer, line 337: If there is a way to handle this, please correct me, but if not, this would be a blocker on ensuring correct behaviour.

Pixelstormer:
It's worth considering what the API for this should look like. If Iced ever gets full rich text support, this will likely be integrated with that, but until then this will need its own API.

Currently the API accepts grapheme indexes; graphemes are the most intuitive way to deal with partitioning rendered text, and indeed, there is no intuitive answer as to what a highlight should look like if it were to be specified by byte or char index instead, and ended up cutting a grapheme in half.

The issue with this is the std library provides no way to observe strings via their graphemes (Iced itself uses [this crate](https://crates.io/crates/unicode-segmentation)), instead dealing with `byte`s and `char`s . This is a problem as I think it would be reasonable to say that for most use cases of text highlighting, the region of text to highlight would be calculated programmatically, and thus would likely be represented (on the user's end) as bytes or chars, rather than graphemes (Unless the user specifically uses a grapheme processing crate).

The API could be changed to accept `byte`s or char`s, but that would either require extra input sanitisation to handle inputs that don't conform to a grapheme boundary (Do we silently ignore, return an error or panic?), or somehow interpreting what highlights would look like if they start or end in the middle of a grapheme.

On the other hand, keeping the API as-is would pass these burdens on to the user, requiring them to convert their data into grapheme indexes. We could expose helper methods to aid with this, but doing so could be scope creep.

I am certainly not acquainted enough with Iced's architecture and design philosophies to make any kind of informed decision on my own (Not only about API but the other topics mentioned in my above comments), so I'm willing to delegate to the opinion of someone more familiar with Iced.

hecrj:
I think there is still a lot of discussion that needs to happen before we can start thinking about code here.

I have some text rendering improvements coming soon in the pipeline. I am aware I am the main bottleneck here! Once those improvements land, we can review this approach and evaluate if it's suitable or whether we can reuse any parts or not.

Issue #1022, opened 2021-09-03
System themes


mmstick:
What's the stance on supporting system themes? It's often important to have a consistent UX across the desktop, with all applications sharing a common system theme by default. Will iced support allowing users and distributors to design system themes that can be applied universally to all iced applications?

yusdacra:
I'm not sure how feasible it would be, but perhaps writing iced backends for native toolkits of a platform (qt, gtk, etc.) would work? I know druid supports native platform toolkits so it should be possible.

mmstick:
Let's say that I have an OS, and I want to build that OS around a toolkit that isn't GTK or Qt. I'm already using GTK for building applications in this OS, but I want to see if there are better solutions that I could use should I want to transition away from GTK. So what I want isn't to use a toolkit that has a reliance on GTK or Qt, but a toolkit that I can use as an alternative to them. But that would require that such theoretical toolkit supports defining system themes.

yusdacra:
Ah so you mean as in, say, something like a "theming backend", which could perhaps support using GTK themes, or Qt themes etc.?

mmstick:
Having something like `/etc/iced/settings.{ron|toml}` that supports defining a theme by name, and having themes placed in `/usr/share/themes/{NameOfTheme}/iced/`, using whatever format works best for iced. This way, if I develop an application for this OS, it would use the system theme by default, but users could change the system theme to one they created. Applications could extend the theme or opt into their own custom theme at the application level.

mmstick:
I guess to be clear, support for GTK and Qt themes is not something I'd expect. Instead, a theme format specific to Iced. I don't generally think it wise to try to guess color schemes and layouts from a GTK theme.

hecrj:
I think it sounds like a good idea! I am not happy with our current default theme. This would be part of the efforts to improve it.

We could make `iced` look into user settings to obtain the default theme, instead of using the currently hardcoded one. For this, we would need to change the widgets slightly and pass the default theme around when rendering, since it will be chosen at runtime now.

13r0ck:
Where are the hard coded values? I am willing to try and make a draft for this functionality.

emann:
[...]

PR #1277, opened 2022-03-05
Application-wide Theming Set/Changed at Runtime

emann:
A precursor for #1022. Allows users to create entirely custom themes that can be set/updated during runtime.

## Implementation Details

- `renderer::Style` was already passed around to `Widget::draw()` calls so I just added fields to the struct to store a 
`Box<dyn a_widget::StyleSheet + 'static>` for each widget. 
- Added `fn styling(&self) -> renderer::Style` to `Sandbox`, `Application`, and a few other places where needed which allows the user to dynamically select which style to use. A default implementation is provided that returns what were previously hardcoded values for the current default style of an iced application.
- `Widget::draw()` implementations for widgets that allow custom styling have been updated to fall back on the `renderer::Style`'s style sheet if no custom styling was specified
  - the `style_sheet: Box<dyn StyleSheet + 'a>` field for said structs has been changed to 
  `custom_style_sheet: Option<Box<dyn StyleSheet + 'a>>`
- Added some methods to some `StyleSheet` traits that take some state data (e.g. `is_mouse_over`,) and return the style that should be used. The default implementation of these methods operate identically to the implementations that were previously in the `draw()` methods (delegating to one of the other style sheet's methods which return a `my_widget::Style` e.g. `hovered()`, `active()`, etc.) but both clean up the code in the `draw()` methods and allow for greater flexibility when creating custom themes.

## Example
The [`application_theme` example project](https://github.com/emann/iced/tree/feature/styling-passed-around-at-runtime/examples/application_theme)
application_theme_demo


emann:
@hecrj marking as a draft as there are still some ToDo items (see below) but wanted to see if you had any input/requests in the meantime.

### ToDo

- Rename `renderer::Style` to `renderer::Theme` and update params/docs/etc. accordingly
- Make sure all examples still work (or at least the ones that work on master)
- Maybe remove the `styling` example as it doesn't really show anything special anymore???

hecrj:
I'm not sure this is the right approach.

Instead of listing the `StyleSheet` of every widget, I believe we should come up with a `Theme` type that contains a specific color palette that all widgets can use (even custom ones!).

For instance, something like:

```rust
pub struct Theme {
    pub base: Color,
    pub action: Color,
    pub text: Color,
    pub positive: Color,
    pub negative: Color,
    // and more...
}

Then, all of the widgets can use the colors of this `Theme` to style themselves by default, using some colors directly or maybe mixing some of them!

Before we write any code, the first step would be deciding the different fields of this `Theme` struct. Then, we can start by migrating the widgets to the new styling strategy one by one.

emann:
@hecrj I love the idea of a color palette but we also need to be able to specify widget-specific styling e.g. button border radius, slider handle shape, etc. How should we go about doing so? Adding fields to the `Theme` struct for each widget's specific styling seems to make sense, but then what if someone wants to change those depending on `active` `hovered` etc.? I've definitely seen buttons change shape when hovered for example.

artursapek:
> I'm not sure this is the right approach.
> 
> Instead of listing the `StyleSheet` of every widget, I believe we should come up with a `Theme` type that contains a specific color palette that all widgets can use (even custom ones!).
> 
> For instance, something like:
> 
> ```rust
> pub struct Theme {
>     pub base: Color,
>     pub action: Color,
>     pub text: Color,
>     pub positive: Color,
>     pub negative: Color,
>     // and more...
> }
> ```
> 
> Then, all of the widgets can use the colors of this `Theme` to style themselves by default, using some colors directly or maybe mixing some of them!
> 
> Before we write any code, the first step would be deciding the different fields of this `Theme` struct. Then, we can start by migrating the widgets to the new styling strategy one by one.

this looks familiar :-D

hecrj:
Closing since the RFC efforts have landed with #1362. Thank you for starting all of it! 😁

Issue #1077, opened 2021-10-08
unresolved import `iced::Alignment`


anzhi0708:
Was trying to run the example code "Counter"
```
[rust-analyzer unresolved-import] [E] unresolved import
   ───────────────────────────────────────────────────────
   [rustc E0432] [E] unresolved import `iced::Alignment`
   no `Alignment` in the root
```
```
error[E0432]: unresolved import `iced::Alignment`
 --> src/main.rs:2:13
  |
2 |     button, Alignment, Button, Column, Element, Sandbox, Settings, Text,
  |             ^^^^^^^^^ no `Alignment` in the root

For more information about this error, try `rustc --explain E0432`.
```

anzhi0708:
@tarkah Thanks, problem solved.

PR #1279, opened 2022-03-07
Fix `counter` example

Daggy1234:
Incorrect `Alignment` import over `Align`


hecrj:
[The `README` in the `examples` folder](https://github.com/iced-rs/iced/blob/master/examples/README.md) mentions:

> __Iced moves fast and the `master` branch can contain breaking changes!__ If you want to learn about a specific release, check out [the release list].

[the release list]: https://github.com/iced-rs/iced/releases

`Alignment` is correct. See #1065, #1077, and #1079.

Issue #1232, opened 2022-02-06
Running example on WSL: Error: GraphicsAdapterNotFound


Ploppz:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

When running the `clock` example in WSL with WSLg installed (graphics support for WSL), the example exits with `Error: GraphicsAdapterNotFound`.
Note that the `xclock`command  runs fine in the same terminal.
I found that in the `wgpu` repo, `cd wgpu ; cargo run --example cube` works fine also. So it seems to be an iced issue.

### What is the expected behavior?

It should work without error

### Version

master

### Operative System

Windows

hecrj:
> When running the clock example in WSL with WSLg installed (graphics support for WSL), the example exits with Error: GraphicsAdapterNotFound.

What about the other examples?

Could you try setting the `antialiasing` flag to `false` here?

https://github.com/iced-rs/iced/blob/c15701581e52ae838f8e7153ae7e639dd86b1ef6/examples/clock/src/main.rs#L9

Ploppz:
Tried a few other examples like geometry, pick_list - it's the same.

With `antialiasing: false` it's the same.

hecrj:
Thanks for trying it out.

Could you maybe try setting the environment variable `WGPU_BACKEND` to `vulkan`, `dx12`, `dx11`, and `gl`? Let me know if it helps!

Ploppz:
All the options gave the same error except vulkan:
```
thread 'main' panicked at 'Error in Surface::get_preferred_format: no suitable format found', /home/ploppz/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.12.0/src/backend/direct.rs:214:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Same story with antialiasing true and false.

genusistimelord:
is this WSLv1 or WSLv2 and do you have CPU virtualization enabled?

Ploppz:
v2. And yes I believe CPU virtualization is enabled as I had to enable it in order to get WSL to work at all? It was a bios option iirc

hecrj:
Could you provide a backtrace?

I really think the issue here is either environment dependent or `wgpu`, since we are not doing anything special during initialization.

[...]

PR #1280, opened 2022-03-09
Power and Limits Auto configure and setting.

genusistimelord:
Adjusted Limits so it uses what the adapter allows also changed the Power settings so it checks for High powered adapters first and then low powered.

This will attempt to use high powered GPU first <PCIe> if it fails to find any it will then use Low Powered GPU's 2nd Processor embedded and software based. 

It will pull the required limits for None WASM to what the default adapter supports rather than setting it all to the lowest of low.
Should fix some issues with #1232 .  We might also be able to add a 3rd check for fallback software devices which might allow WSL to work better when neither GPU are pass-thru acceptable. But will need to test that too.

### DIFF
--- a/wgpu/src/settings.rs
+++ b/wgpu/src/settings.rs
@@ -83,6 +83,7 @@ fn backend_from_env() -> Option<wgpu::Backends> {
             "gl" => wgpu::Backends::GL,
             "webgpu" => wgpu::Backends::BROWSER_WEBGPU,
             "primary" => wgpu::Backends::PRIMARY,
+            "secondary" => wgpu::Backends::SECONDARY,
             other => panic!("Unknown backend: {}", other),
         }
     })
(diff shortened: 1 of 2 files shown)


genusistimelord:
honestly the power part should not matter but I am wondering if something is broken on the wgpu side as to why changing it to low power mode for someone worked in a VM while it should have worked regardless. 

Generally based on https://github.com/gfx-rs/wgpu/blob/df50028e49e1448b56ce71f7904921a95c15594c/wgpu-core/src/instance.rs#L735

which is suppose to parse thru each one till it finds an acceptable adapter the only difference is low and high just puts discrete or integrated first in the list.  So I am wondering if it doesn't fail at https://github.com/iced-rs/iced/blob/ac3e477bb34fb804cdacc1b02023a6315b156e43/wgpu/src/window/compositor.rs#L35

But fails at 
https://github.com/iced-rs/iced/blob/ac3e477bb34fb804cdacc1b02023a6315b156e43/wgpu/src/window/compositor.rs#L66  due to it not having access to the GPU directly. this needs to be tested so hoping @skydig and @Ploppz can test this.

PR #1286, opened 2022-03-15
Slider fix

ThisIsRex:
Fixed rail position. Fixed incorrect handle position with large radius/width/border width

### DIFF
--- a/native/src/widget/slider.rs
+++ b/native/src/widget/slider.rs
@@ -258,7 +258,7 @@ pub fn draw<T>(
         renderer::Quad {
             bounds: Rectangle {
                 x: bounds.x,
-                y: rail_y,
+                y: rail_y - 1.0,
                 width: bounds.width,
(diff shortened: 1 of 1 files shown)


hecrj:
Thanks for looking into this!

Could you share an [SSCCE](http://sscce.org/) of the issue these changes are meant to fix?

ThisIsRex:
> Thanks for looking into this!
> 
> Could you share an [SSCCE](http://sscce.org/) of the issue these changes are meant to fix?

Sure, here it is [main.rs.txt](https://github.com/iced-rs/iced/files/8318416/main.rs.txt)
I used a git repository as a dependecy

before:
ezgif-4-1e40769b21

after:
ezgif-4-38e225cc87

hecrj (approved):
Thank you!

Not sure what I was thinking when I wrote the handle position logic :sweat_smile:

PR #1294, opened 2022-03-26
Forward overlay() calls in iced_pure::element::Map

marienz:
If Map does not override overlay(), calling map() on a pure Element
breaks any pick_list inside it (its overlay does not appear).

Fix it by implementing overlay() the same way iced_native::element::Map
does.

### DIFF
--- a/native/src/overlay/element.rs
+++ b/native/src/overlay/element.rs
@@ -41,7 +41,7 @@ where
     where
         Message: 'a,
         Renderer: 'a,
-        B: 'static,
+        B: 'a,
     {
         Element {
             position: self.position,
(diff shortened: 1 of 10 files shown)


marienz:
I copied the lifetime changes from iced_native because the code didn't compile without them. I didn't look into whether they could be avoided (which would take me a while, I'm a Rust beginner).

hecrj (approved):
Good catch! I always forget to implement `overlay` :sweat_smile: 

> I copied the lifetime changes from iced_native because the code didn't compile without them. I didn't look into whether they could be avoided (which would take me a while, I'm a Rust beginner).

Thanks for reminding me of this! I have properly loosened the lifetimes of `Message` and `Renderer` in d9d6dff550a917d2b73c7ebf96cab2c1472e3243.

I believe we can merge this! :ship:

PR #1298, opened 2022-03-29
Implement `Default` for `pure::State`

0x192:
Allow to use a `pure::State` in a struct deriving Default

This PR set default to `pure::State::new()`


hecrj (approved):
Great! Thank you 🙇

PR #1301, opened 2022-04-06
Fix nested pure components

nicksenger:
Currently, using a pure component within another pure component will cause the application to panic in certain cases. As far as I can tell, this is because the creation of a component's element is deferred until `diff` is called on the corresponding widget tree, which may not happen before `on_event` if that component is nested within another component which undergoes some internal state change.

Repro case here: https://gist.github.com/nicksenger/543e1574b4a5519d9775e94359ae66eb

Diffing the widget tree after processing local events rebuilds the element for any sub-components, which resolves the issue.


nicksenger:
Going to go ahead and open this up. I've tested it pretty extensively at this point and it seems to be working as intended.

nicksenger:
It seems the fix for this is much more straightforward than I previously thought. We simply need to diff the widget tree after rebuilding the cached element in on_event.

hecrj (approved):
I see! That makes way more sense :tada: 

Thank you!

PR #1308, opened 2022-04-18
fix: tree node passed in `component::OverlayInstance::on_event`

nicksenger:
I was seeing some strange behavior when using overlays inside of components. Currently when rebuilding the overlay in `on_event` of `OverlayInstance`, the widget tree node for the component itself is being used rather than the tree node for the cached element. In `Instance::overlay`, the correct tree node is already being used, so no change is required there.

### DIFF
--- a/lazy/src/pure/component.rs
+++ b/lazy/src/pure/component.rs
@@ -470,7 +470,7 @@ where
                             .as_ref()
                             .unwrap()
                             .as_widget()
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Great catch!

I wish there was a way to encode these constraints in the type system for guaranteed safety. But since the widget and state trees are meant to be completely decoupled, we have to deal with this...

PR #1314, opened 2022-04-21
feat: fetch system information

derezzedex:
The goal of this PR is to improve ways to debug Iced's built-in backends creation/boot.

For that, it currently:
- Adds a new `Error::ContextCreationFailed` in `iced_graphics`.
- Adds some logging around context creation in both renderers (`iced_glow` and `iced_wgpu`).
- Adds a new `System` variant to `Action`. 
- Adds a new `System::QueryInformation` action to fetch system and hardware information (gated by the `sysinfo` feature flag).
- Exposes a generic `system::Information` struct containing acquired hardware information.

If there are any other points that could be worth logging or any new fields that we could add to `system::Information`, please let me know!


winit/src/application.rs
```diff
@@ -536,8 +543,10 @@ pub fn run_command<Message: 'static + std::fmt::Debug + Send, E: Executor>(
     clipboard: &mut Clipboard,
     proxy: &mut winit::event_loop::EventLoopProxy<Message>,
     window: &winit::window::Window,
+    graphics_info: &window::GraphicsInformation,
```
derezzedex, line 546: I couldn't think of other ways to do this, I'd be happy to receive feedback on this.

winit/src/system.rs
```diff
@@ -0,0 +1,53 @@
...
+    };
+
+    Some(information)
+}
+
+#[cfg(not(feature = "sysinfo"))]
+pub(crate) fn get_information(
+    _graphics_info: &window::GraphicsInformation,
+) -> Option<Information> {
+    None
```
derezzedex, line 52: I wasn't sure about removing the `system::Information` struct or/and removing the `System::QueryInformation` enum variant if `sysinfo` wasn't enabled, so I decided to turn this into an `Option`.

Any improvements on this would be nice!

hecrj (changes requested):
This is awesome. Great work! 👍 

I left some suggestions here and there. Let me know what you think!

winit/src/application.rs
derezzedex: Done in https://github.com/iced-rs/iced/pull/1314/commits/6e167675d6a51a8a78d93439719ebffe35dcfdef.

winit/src/system.rs
derezzedex: Simply conditionally compiling the `system::information` helper wouldn't work
because the `run_command` helper matches on every variant of `Action` and so
would throw the error for anyone not using `sysinfo` feature.

So, I've gone ahead and completely remove the `sysinfo` flag in https://github.com/iced-rs/iced/pull/1314/commits/e6b09e9194011c2a588cab0d9f1af1b9a35da787.
By not storing and delaying the fetching of information until requested,
we remove the need of feature flagging.

Reverting this should be as simples as yanking the last commit.

graphics/src/window/compositor.rs
```diff
@@ -38,6 +38,9 @@ pub trait Compositor: Sized {
         height: u32,
     );
 
+    /// Returns [`GraphicsInformation`] used by this [`Compositor`].
+    fn get_information(&self) -> GraphicsInformation;
```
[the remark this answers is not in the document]
derezzedex: Renamed in https://github.com/iced-rs/iced/pull/1314/commits/005e516b5e1e8bb22f2da8524ffe4529f3b60ba1.

Since retrieving this information is done through a dependency (that uses system calls), I decided to stick with `fetch_`, although this is done synchronously.

src/error.rs
```diff
@@ -11,9 +11,9 @@ pub enum Error {
     #[error("the application window could not be created")]
     WindowCreationFailed(Box<dyn std::error::Error + Send + Sync>),
 
-    /// A suitable graphics adapter or device could not be found.
-    #[error("a suitable graphics adapter or device could not be found")]
-    GraphicsAdapterNotFound,
+    /// The application context could not be created.
+    #[error("the application context could not be created")]
+    ContextCreationFailed(iced_graphics::Error),
```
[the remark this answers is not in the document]
derezzedex: Done in https://github.com/iced-rs/iced/pull/1314/commits/18ecec2bbd9478151b8c65df0644278a174d0847.

wgpu/src/window/compositor.rs
```diff
@@ -28,6 +29,17 @@ impl Compositor {
...
 
+        log::info!("{:#?}", settings);
+
+        #[cfg(not(target_arch = "wasm32"))]
+        {
+            let available_adapters: Vec<_> = instance
+                .enumerate_adapters(settings.internal_backend)
+                .map(|adapter| adapter.get_info())
+                .collect();
+            log::info!("Available adapters: {:#?}", available_adapters);
```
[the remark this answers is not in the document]
derezzedex: Done in https://github.com/iced-rs/iced/pull/1314/commits/7a13928130803629b15999923da3a9747e110e7a.

winit/src/system.rs
```diff
@@ -0,0 +1,53 @@
+//! Access the native system.
+use crate::command::{self, Command};
+pub use iced_native::system::*;
+
+use iced_graphics::window;
+
+/// Query for available system information.
+///
+/// Returns `None` if not using the `sysinfo` feature flag.
+pub fn information<Message>(
```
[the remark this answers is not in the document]
derezzedex: Done in https://github.com/iced-rs/iced/pull/1314/commits/8643fbef90fc16371e27ae0142eb7e4c6b432d29.

winit/src/system.rs
```diff
@@ -0,0 +1,53 @@
...
+        Box::new(f),
+    )))
+}
+
+#[cfg(feature = "sysinfo")]
+pub(crate) fn get_information(
+    graphics_info: &window::GraphicsInformation,
+) -> Option<Information> {
+    use sysinfo::{ProcessExt, ProcessorExt, System, SystemExt};
+    let mut system = System::new_all();
```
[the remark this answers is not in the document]
derezzedex: Because of https://github.com/iced-rs/iced/commit/5eefa5d4ead9ebfac7dab1db9aebf9797d2dad38, I've decided to maintain this here (this should isolate the crate to the Action itself).
And it seems that refreshing and creating are almost the same.

winit/src/system.rs
hecrj: Gave it a shot in 93bfe2c75ec97ef78f993926c703f040dde4a5f3.

derezzedex: Interesting, would it be possible to also gate the `iced_native` side? That would make it impossible for someone to misuse this.

I've also commited a simple fix to another issue in https://github.com/iced-rs/iced/pull/1314/commits/a4477723d44c35effc8f4bb0ff0cfce93fdf4695 (which I was planning on squashing, but didn't for now).

hecrj (approved):
Looks good!

I noticed that the action blocks the main thread for a second or two. It's not a big deal since you shouldn't be calling this often, but it's still a bit jarring to freeze the GUI to users. So I have moved the slow call to a different thread in c2bbce9775b601413500c889c7203353ed5c00d3.

Let me know what you think and we can merge this!

PR #1315, opened 2022-04-22
Implemented Tooltip as Pure

casperstorm:
This PR introduces a pure `Tooltip`.

* (Re)implemented `Tooltip` as `pure`
* Added Example to `/examples/pure/tooltip`
* Corrected variable name in `native/src/widget/tooltip.rs`

tooltip

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,6 +93,7 @@ members = [
     "examples/pure/pick_list",
     "examples/pure/todos",
     "examples/pure/tour",
+    "examples/pure/tooltip",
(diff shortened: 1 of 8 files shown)


hecrj (approved):
Looks great!

I made some small changes to expose the `Position` enum from `iced_native` instead of duplicating it in 011b7d11126404eaeff1562e49d81693d4d852b2 and refactored the draw logic so we can reuse it in e33f43af31888b8b0795c2287aed9f26b4e1d699.

I think we can merge this! 🥳

PR #1318, opened 2022-04-25
Fix `PickList` text wrapping

tarkah:
PickList text is wrapped if the pick list is set to a fixed width and the selected text is wider than the fixed width. This resolves the issue by giving the rendered text bounds an infinite width so it doesn't wrap. A layer / clip needs to be provided to prevent text overflow.

I can't imagine wrapping is intended since the height is fixed to `text_size` and 2 lines of wrapped text inside `text_size` doesn't look good at all. This seems like a more sane default to handle text overflow.

### DIFF
--- a/native/src/widget/pick_list.rs
+++ b/native/src/widget/pick_list.rs
@@ -402,16 +402,18 @@ pub fn draw<T, Renderer>(
     if let Some(label) =
         label.as_ref().map(String::as_str).or_else(|| placeholder)
     {
+        let text_size = f32::from(text_size.unwrap_or(renderer.default_size()));
+
         renderer.fill_text(Text {
             content: label,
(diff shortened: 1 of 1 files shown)


hecrj:
This was intended to some extent. Basically, I wanted to avoid an unconditional `with_layer` call, which breaks text caching currently.

Since a `PickList` is a much smaller widget than a `Scrollable`, there is a high chance users will use a bunch of them at once. Therefore, I considered the use of `with_layer` here a very potential performance hog.

We may be able to fix the issue for now by just changing the `vertical_alignment` to `Top` so the wrapped text is completely out of bounds. But of course, it will still cause the wrapped text to disappear which is not the expected behavior.

Once the text caching issues are fixed, we can revisit this!

hecrj (approved):
Awesome! Thanks :bow:

PR #1319, opened 2022-04-25
Allow use of `pure::Component` within `pure::Responsive`

nicksenger:
Currently, using an `iced_lazy::pure::Component` within the closure provided to an `iced_lazy::pure::Responsive` will cause the application to panic. This is because `layout` on the component is invoked before the widget tree is diffed, so the component's cache has not yet been populated. It's straightforward to reproduce by wrapping the pure/component example in a responsive.

Simply moving the call to diff the widget tree up a couple lines so that it happens before producing the layout appears to resolve the issue.

### DIFF
--- a/examples/pure/component/src/main.rs
+++ b/examples/pure/component/src/main.rs
@@ -1,7 +1,6 @@
 use iced::pure::container;
 use iced::pure::{Element, Sandbox};
 use iced::{Length, Settings};
-use iced_lazy::pure::responsive;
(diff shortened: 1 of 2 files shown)


hecrj (approved):
Huh, good catch! Thanks!

It's a bit worrying that the structure of a widget can cause panics in a different one so easily. Food for thought 🤔

In any case, I think we can merge!

PR #1326, opened 2022-05-02
add pure version of color_palette example

maxwell8888:


hecrj (approved):
Looks great! Thank you :bow:

PR #1332, opened 2022-05-04
Fix vertical rule helper

tarkah:
`vertical_rule` helper is returning `Rule::horizontal`


hecrj:
Hah! Classic copy-paste! :sweat_smile: 

Thank you!

hecrj (approved):

PR #1337, opened 2022-05-06
Take `Borrow<dyn Widget>` instead of `Element` in `Tree` API

hecrj:
It should make things easier when implementing custom widgets that do not necessarily own `Element` directly.

This should be a non-breaking change, since `Element` should be accepted just as before.

### DIFF
--- a/pure/src/element.rs
+++ b/pure/src/element.rs
@@ -321,3 +321,11 @@ where
             .map(move |overlay| overlay.map(mapper))
     }
 }
+
+impl<'a, Message, Renderer> AsRef<dyn Widget<Message, Renderer> + 'a>
+    for Element<'a, Message, Renderer>
+{
+    fn as_ref(&self) -> &(dyn Widget<Message, Renderer> + 'a) {
+        self.widget.as_ref()
+    }
+}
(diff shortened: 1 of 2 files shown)


genusistimelord:
this will require all the widgets to get an AsRef<dyn Widget<Message, Renderer> + 'a> implemented for them for this to work for the widget to get used within the Tree directly.   OR we will get this
```
the trait bound `iced_pure::widget::TextInput<'_, Message, Renderer>: std::convert::AsRef<dyn iced_pure::Widget<_, _>>` is not satisfied
required because of the requirements on the impl of `std::convert::AsRef<dyn iced_pure::Widget<_, _>>` for `&iced_pure::widget::TextInput<'_, Message, Renderer>`
```

hecrj:
@genusistimelord I assumed this implementation would be enough: https://doc.rust-lang.org/std/convert/trait.AsRef.html#impl-AsRef%3CU%3E

If that doesn't work, then maybe casting the `Widget` to `as &dyn Widget<Message, Renderer>` directly does?

hecrj:
@genusistimelord 59aeb20600661e55e59223a309146eea68b4c32f shoud allow you to pass a `&dyn Widget` to the `Tree` methods.

genusistimelord:
this works  

Example for people who find this later.
```rust
 fn children(&self) -> Vec<Tree> {
        vec![Tree::new(&self.content as &dyn Widget<_, _>)]
    }

    fn diff(&self, tree: &mut Tree) {
        tree.diff_children(&[&self.content as &dyn Widget<_, _>]);
    }
    ```

genusistimelord (approved):
This looks good and works.

PR #1340, opened 2022-05-09
Add missing versions to issue template

mtkennerly:


.github/ISSUE_TEMPLATE/BUG-REPORT.yml
```diff
@@ -60,6 +60,9 @@ body:
       description: What version of iced are you using?
       options:
         - master
+        - 0.4.2
+        - 0.4.1
+        - 0.4.0
         - 0.3.0
```
hecrj, line 66: We should just drop support for `0.3`, given it's extremely outdated.

I think we can simply list `master` and `0.4`, and ask users here to make sure they are using the latest patch version.

hecrj (changes requested):

.github/ISSUE_TEMPLATE/BUG-REPORT.yml
mtkennerly: Sounds good to me. Just updated the PR.

hecrj (approved):
Thank you! :bow:

PR #1341, opened 2022-05-11
fix: prevent building component overlay with stale layout

nicksenger:
In the implementation of `on_event` for a component's overlay (both pure and impure), after processing local messages which may mutate the state and in turn alter the output of `Component::view`, the cached element and overlay must be rebuilt. Currently, it is rebuilt with a layout tree which is potentially stale, because the shape of the layout tree is dependent upon the cached element which may have changed.

I've added a preliminary fix here, which is to rebuild the cached element only after processing any local messages, then attempt to rebuild the overlay at the beginning of `on_event` if it is absent.

I'll open this as a draft because it feels a bit odd to intentionally leave the cached overlay element as `None` after `on_event`. I haven't yet thought of another way to resolve this though.

### DIFF
--- a/lazy/src/component.rs
+++ b/lazy/src/component.rs
@@ -376,38 +376,6 @@ where
         let mut local_messages = Vec::new();
         let mut local_shell = Shell::new(&mut local_messages);
 
-        if self
-            .instance
-            .state
-            .borrow()
-            .as_ref()
-            .and_then(|state| state.borrow_cache().as_ref())
-            .and_then(|cache| cache.borrow_overlay().as_ref())
-            .is_none()
-        {
-            let component =
-                self.instance.state.take().unwrap().into_heads().component;
-
-            self.instance.state = RefCell::new(Some(
-                StateBuilder {
-                    component,
-                    message: PhantomData,
-                    cache_builder: |state| {
-                        Some(
-                            CacheBuilder {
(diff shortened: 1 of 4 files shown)


hecrj (commented):
Good catch!

I think the root cause of the issue here is in the `UserInterface::update` implementation. Specifically, we are not relayouting the `base` layer when an overlay invalidates the layout through a `Shell`, causing the layout to be stale for further events.

I have tried to fix this in 296e157dcf2efe1ae52ae6cc35060b2f0cf212bf. I needed to rewrite the code to use a `for` loop instead of iterators as well as use `ManuallyDrop` to please the borrow checker, but it seems to be happy in the end.

Could you test if these changes fix the issue?

nicksenger:
That explanation makes sense, and the solution looks good!

I had to make one small change in order for it to fix the issue though:

Since the component overlay will always invalidate the layout when there are local messages produced in `on_event`, and invalidation of the layout now results in the overlay being rebuilt in `UserInterface::update`, the overlay no longer needs to be rebuilt after processing local messages in the component's `on_event` (which is what triggered the panic, since it was still using the same layout to produce the overlay from the newly rebuilt cached element).

hecrj (approved):
Makes sense. Less code to worry about!

Let's merge :ship:

PR #1346, opened 2022-05-20
Add profile release

Luni-4:
This PR adds a `profile.release` to reduce binaries size, but at the same time enabling some optimizations. 
Using this profile, I was able to reduce the `bezier_tool` binary from 15.6 MB to 7.1MB, I haven't checked the other binaries though.

I hope this is appreciated, thanks in advance for your review!


Luni-4:
@hecrj  Is there any update on this one?

Luni-4:
Yeah, I understand what you're saying, since `release` mode might be used in the development phase too, increasing the compilation time could not be an optimal solution. 

Perhaps we can explore the new [custom profiles](https://doc.rust-lang.org/cargo/reference/profiles.html#custom-profiles) feature in order to reduce the binary size of an example. 

Perhaps we can have something like this: 
```
cargo build --profile release-less-binary-size
```

In this way we don't affect the default `release` profile at all. What do you think of this solution @hecrj?

We can also add the `release` configurations of this PR and the new custom profiles in documentation or iced book.

hecrj:
> In this way we don't affect the default release profile at all. What do you think of this solution @hecrj?

That sounds great. If we can just have these settings in a `release-opt` profile then that would be awesome:

```rust
[profile.release-opt]
inherits = "release"
codegen-units = 1
debug = false
lto = true
incremental = false
opt-level = 3
overflow-checks = false
strip = "debuginfo"
```

hecrj:
> Do you also want me to create a ci-profile to remove those lines https://github.com/iced-rs/iced/blob/master/.github/workflows/build.yml#L93-L96 ?

Actually, we can just use the new custom profile! I just did that in 7022090a97c19c8175ae103a6d7bea8a1055c301.

hecrj (approved):
Looks great now! Thank you! :partying_face:

Issue #1177, opened 2021-12-31
Intercepting Paste Events


Alch-Emi:
I'm trying to write an app which includes custom behavior for pasting, and especially multi-line pasting.  It's fairly easy to listen for a Ctr+V event, check if a relevant input is selected, grab the clipboard, and then run the custom behavior.

However, by the time the message reaches my code, the UI will have already filled the widget with the text clipboard.  I can work around this by just guessing the length of the pasted text (i think it's just the text length - the # of newlines) and truncating the text in the UI, but this feels a little messy.

It would be nice to be able to intercept a paste event, maybe by setting a property on the `TextInput` component that produces a `Message` instead of pasting text, or just disabling pasting into the `TextInput`.

brightly-salty:
Just a possible solution, not a permanent one: you could copy the code of this file and remove the linked lines to create a widget which doesn't handle pasting, or just insert your custom logic into that area
https://github.com/iced-rs/iced/blob/5466d6a11d0a4bee01e954b7c727b3fd67cea02a/native/src/widget/text_input.rs#L680-L710

Alch-Emi:
ooh that's handy, thank you!

hecrj:
We could extend the `TextInput` widget with additional methods to override paste behavior.

Could you elaborate a bit on your use case? Do you want to simply perform a transformation on the pasted text?

Alch-Emi:
Sorry I didn't get back to this sooner!  The project I'm working on is displaying several lines of text in a context where it makes sense to use individual, single-line text inputs.  However, I'd like the user to be able to do a multi-line paste, so I'm trying to fudge it a little bit by intercepting the paste and breaking up the text between lines.

That said, I could definitely see this being used in other usecases where transformations might need to be applied as well.

PR #1350, opened 2022-05-28
Added a paste Event to TextInput

wyatt-herkamp:
If no paste event is defined it will just act like a normal on_change event

Closes https://github.com/iced-rs/iced/issues/1177

### DIFF
--- a/native/src/widget/text_input.rs
+++ b/native/src/widget/text_input.rs
@@ -237,7 +237,7 @@ pub fn update<'a, Message, Renderer>(
     font: &Renderer::Font,
     is_secure: bool,
     on_change: &dyn Fn(String) -> Message,
-    on_paste: &Option<Box<dyn Fn(String) -> Message + 'a>>,
+    on_paste: Option<&dyn Fn(String) -> Message>,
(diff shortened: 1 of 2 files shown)


examples/todos/src/main.rs
```diff
@@ -88,6 +89,9 @@ impl Application for Todos {
                     Message::InputChanged(value) => {
                         state.input_value = value;
                     }
+                    Message::InputPasted(value) => {
+                        state.input_value = value;
+                    }
```
hecrj, line 94: I don't think we should change the example. The behavior will be exactly the same as before, so we are just showcasing how to overcomplicate things.

hecrj (changes requested):

examples/todos/src/main.rs
wyatt-herkamp: Ok. Will remove when I get home.

wyatt-herkamp: Alright the test has been removed

hecrj (approved):
Looks great! Thank you :bow:

PR #1353, opened 2022-05-30
Don't process captured overlay statuses

tarkah:
Captured events processed by overlays are still getting processed by the widget. This skips processing the event w/ the widget if it was `Captured` by the overlay.


hecrj (approved):
Good catch! Thanks!

PR #1356, opened 2022-06-02
fix: corrected images in `README.md` to be similar height.

casperstorm:
This is just a tiny PR which ensures the two gifs in the README is equal heights. It looks a lot better, when Github is switched to Dark. I also took the liberty to correct a few markdown lints.

Screenshot 2022-06-02 at 10 11 52 AM


hecrj (approved):
Awesome! Thank you 🥳

PR #1361, opened 2022-06-06
Prevent pane grid title bar content and controls from overlapping

mtkennerly:
This fixes the content/controls becoming unreadable when they overlap in narrow panes.

Should this be configurable? For example:

```rust
enum Overlap {
    Allow, // existing behavior
    HideContent, // proposed behavior
    HideControls,
}
```

### Before
https://user-images.githubusercontent.com/4934192/172265902-715c0a3f-bab7-41d9-8ad5-1fcbd6dc214a.mp4

### After
https://user-images.githubusercontent.com/4934192/172265908-66f7ec7e-2121-4c82-9f4a-fa12f663cb7a.mp4

### DIFF
--- a/native/src/widget/pane_grid/title_bar.rs
+++ b/native/src/widget/pane_grid/title_bar.rs
@@ -105,19 +105,17 @@ where
 
         let mut children = padded.children();
         let title_layout = children.next().unwrap();
-
-        self.content.draw(
-            renderer,
-            &inherited_style,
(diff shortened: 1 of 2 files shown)


hecrj (approved):
Looks great! It's definitely better than displaying overlapping elements.

Maybe we could make it configurable in a different PR? Your idea seems great.

Thank you 🙇

PR #1370, opened 2022-06-22
Added more clarification as to how the position and alignment of the …

jhannyjimenez:
…Canvas Text structure interact


hecrj (approved):
Good catch! Thank you! :bow:

PR #1371, opened 2022-06-23
Allow overriding pure text_input value during draw

tarkah:
`iced_native::widget::TextInput` defines a `draw` method to override it's `Value` when drawn. This adds the same functionality to the pure version.

Difference here is caller must provide the `Tree` with associated state in it for the text input and it's up to them to make sure this is safe.


hecrj (approved):
Thanks! :bow:

Certainly necessary to implement some custom widgets with composition.

PR #1373, opened 2022-06-30
Update import path in pure canvas example

fralonra:
In current version, `iced::pure::canvas` would be unresolved.
It should be `iced::pure::widget::canvas`.


hecrj (approved):
Good catch. Thank you! :bow:

fralonra:
Why it failed on Ubuntu?

PR #1379, opened 2022-07-03
Address Clippy lints

PolyMeilex:
A bunch of stylistic changes based on Clippy lints


hecrj (approved):
Awesome! Thank you :partying_face: Lots of needless borrows I see!

I have gone ahead and fixed all of the reasonable lints in all of the crates and disabled the unreasonable ones.

Additionally, I have introduced a new `lint` workflow on GitHub CI that should error if we ever make `clippy` angry. Let's see how this pans out!

PolyMeilex:
Thanks, I was not sure which choices were intentional and which ones should be fixed, that's why my PR was fixing only some of them, so it's great that you did all the remaining ones that I was not sure about. :ok_hand:

PR #1383, opened 2022-07-10
`row_with_children` and `column_with_children` pure function helper methods

cryptoquick:
Adds row_with_children and column_with_children pure function helper methods.

The inconsistency regarding lifetimes is due to clippy providing guidance for unused lifetimes.

Regarding https://github.com/iced-rs/iced/discussions/1382


hecrj:
The main reason these do not exist yet is because of some WIP exploration I have around useful macros, which will need to wait until we deprecate the original widget API.

API changes normally need some kind of discussion, planning, and iteration!

hecrj:
#1393 changes `row` and `column` to use `with_children`.

Issue #715, opened 2021-01-26
Stop scrollview from consuming the interface


rnd-ash:
My application has a main UI page which is broken down into 2 parts:

1. main interface which can change
2. Bottom status bar which is permanently on screen

With the UI with explain enabled, you can see this:
![image](https://user-images.githubusercontent.com/14551965/105879691-64a61200-5ffa-11eb-8786-a1fdcaafe474.png)

Under log view in this image is a scrollview. Below that is a space followed by the status bar. It is drawn like so:

```rust
let mut c: Element<_> = Column::new()
                .push(view_contents) // Draw the custom view on screen
                .push(Space::with_height(Length::Fill)) // Pad the gap
                .push(Rule::horizontal(1)) // Draw a line
                .push(s_bar) // Draw the status bar
                .into();
if themes::is_debug() {
    c = c.explain(iced::Color::BLACK);
}
container(c)
    .height(Length::Fill)
    .width(Length::Fill)
    .into()

```

however, when I keep adding elements to the scrollview, this eventually happens:
![image](https://user-images.githubusercontent.com/14551965/105879941-ac2c9e00-5ffa-11eb-88f1-b644fabd6120.png)

Here, the scrollview has completely consumed the interface.

How would it be possible to stop this behavior?. I don't want to use the max_height function on the scrollview as i still want the entire UI to be adaptive according to the window size rather than creating a hard limit.

NathanSWard:
+1 for this

hecrj:
This is caused by a limitation / bug of the current layout engine and widget implementation.

In order to fix this, you have to explicitly state `height(Length::Fill)` for all of the parent containers of the `Scrollable`.

In any case, we should really tackle this limitation soon. I will work on it right away!

hecrj:
Fixed by #2192.

lobneroO:
I think this is still happening with version 0.12. I have a scrollable which can become quite lengthy and two buttons beneath it. If the scrollable takes up more space than the window offers, it pushes the button out of the window.

[My proejct](https://github.com/lobneroO/pro-tool/commit/fc0f1fd650e8468e7eb7c491ff3b2e8554415d99)
You will have to choose the attached input file, then click "Create Personal Running Order".

Input file:
[data_summerbreeze2022.csv](https://github.com/user-attachments/files/16102841/data_summerbreeze2022.csv)


This is the default for me when it opens:
[...]

PR #1390, opened 2022-07-18
Fix `max_width` and `max_height` for `Container`

### LINKED ISSUE #1388: `max_height` seems to be ignored since 0.4.0

### What happened?

`max_height` does not seem to have an effect on container height, starting in 0.4.0 and still the case in master.

Gist: https://gist.github.com/mtkennerly/f248f51b3f1c25622f63d5b6d94eb92d

From left to right, the columns are configured as:

* no `.height()`, no `.max_height()`
* `.height(Length::Fill)`, no `.max_height()`
* no `.height()`, `.max_height(200)`
* `.height(Length::Fill)`, `.max_height(200)`

The ones with `max_height` look the same as the ones without it:

> iced-max-height-0 4 0

This was the previous behavior on 0.3.0. When `max_height` is set, the container height is limited:

> iced-max-height-0 3 0

### Version

0.4

### Operative System

Windows

_No response_

hecrj:
Fixes #1388.

### DIFF
--- a/native/src/widget/container.rs
+++ b/native/src/widget/container.rs
@@ -127,12 +127,20 @@ pub fn layout<Renderer>(
     limits: &layout::Limits,
     width: Length,
     height: Length,
+    max_width: u32,
+    max_height: u32,
     padding: Padding,
     horizontal_alignment: alignment::Horizontal,
     vertical_alignment: alignment::Vertical,
     layout_content: impl FnOnce(&Renderer, &layout::Limits) -> layout::Node,
 ) -> layout::Node {
-    let limits = limits.loose().width(width).height(height).pad(padding);
+    let limits = limits
+        .loose()
+        .max_width(max_width)
+        .max_height(max_height)
+        .width(width)
+        .height(height)
+        .pad(padding);
 
     let mut content = layout_content(renderer, &limits.loose());
     let size = limits.resolve(content.size());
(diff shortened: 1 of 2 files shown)


native/src/widget/scrollable.rs
```diff
@@ -162,9 +162,10 @@ pub fn layout<Renderer>(
     limits: &layout::Limits,
     width: Length,
     height: Length,
+    max_height: u32,
     layout_content: impl FnOnce(&Renderer, &layout::Limits) -> layout::Node,
 ) -> layout::Node {
-    let limits = limits.width(width).height(height);
+    let limits = limits.max_height(max_height).width(width).height(height);
```
mtkennerly, line 168: The output on 0.3.0 and on this branch look the same now :tada:

For `Scrollable`, I was surprised that `max_height()` gets superseded by `height(Length::Fill)` (far right column below), since that's not the case for `Container`, but maybe that's intentional and/or related to #715? It respects the `max_height()` if you set `height(Length::Shrink)` instead or just don't set a height.

Gist for screenshot: https://gist.github.com/mtkennerly/c67a18ffcef3b926926f012f7b6d1512

> image

mtkennerly (approved):

native/src/widget/scrollable.rs
hecrj: Yeah. The layout engine overall needs a review and a proper test suite to ensure consistency.

I believe `max_` and `min_` attributes should eventually be moved to a single widget (maybe `Container`?).

PR #1396, opened 2022-08-02
Allow specification of `border_radius` for `pick_list::Menu`

nicksenger:
Currently a `border_radius` may be applied to the pick list button, but not its menu. This PR updates the menu to support border radius.


hecrj (approved):
Sweet. Thanks!

PR #1398, opened 2022-08-04
Add `min_width` and `min_height` to `Container`

nicksenger:
This PR adds `min_width` and `min_height` to the `Container` widget. I'm not sure if there's a reason these are absent, but I have a use-case for min-width and it seems to be working as expected when provided in this way.

### DIFF
--- a/native/src/widget/container.rs
+++ b/native/src/widget/container.rs
@@ -127,7 +127,9 @@ pub fn layout<Renderer>(
     limits: &layout::Limits,
     width: Length,
     height: Length,
+    min_width: u32,
     max_width: u32,
+    min_height: u32,
     max_height: u32,
     padding: Padding,
     horizontal_alignment: alignment::Horizontal,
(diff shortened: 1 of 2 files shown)


nicksenger:
Nevermind, found a way to do what I'm trying to do without this.

bungoboingo:
This would be very useful to my current layouting predicament if there are no issues

bungoboingo:
Actually this does not function as intended when the container is inside a row or column, min width/height are completely ignored. Probably needs some more adjustments, will look in to it

nicksenger:
Haha, haven't thought about this in a while. iirc there is(was?) some directionality associated with `layout::flex::resolve` causing child elements to collapse first from one side or the other, which is probably what I was trying to work around.

Idk your use-case, but if it's isolated enough I'd be inclined to try a custom `Row`-like widget with a layout method implemented specifically for the situation. It'd be cool if a general solution like this could be made to work though.

PR #1400, opened 2022-08-05
`Lazy` widget

nicksenger:
In some cases computations are performed in each view regardless of whether their inputs have actually changed. This PR adds a widget `Cached`, which takes some dependency and a function producing a view. The view function will only be invoked to generate a new element when the dependency produces a different hash.

The example displays a sorted list of items based on data stored in a set, and is optimized to only produce the sorted list when either the sort order or number of items in the list changes. The view function is not re-invoked in view after the user types into the text input, because the content of the input is not a dependency of the sorted list.


hecrj (approved):
This is great! Similar in spirit to the [`Html.lazy`](https://guide.elm-lang.org/optimization/lazy.html) in Elm.

I renamed the widget to `Lazy`, since it's actually the main idea behind the `iced_lazy` crate!

PR #1402, opened 2022-08-07
Impl operate for Component widget

tarkah:
`operate` was missing for `Component` widget implementation. Now operations like `focus_next` will work for text inputs inside a `Component`.


hecrj (approved):
Good catch! Thank you :bow:

PR #1403, opened 2022-08-07
Minor style changes to counter example

proudmuslim-dev:


hecrj:
I find expression with side-effects to be a bit weird. I feel it's clearer to have these as a statement.

PR #1404, opened 2022-08-07
Take advantage of derive default feature for `Theme` enum

proudmuslim-dev:


tarkah:
This can be extended to all the other `Default` impls in this module for the different Style enums

proudmuslim-dev:
I can go through and do that, if you'd like 

@tarkah

hecrj:
I'm afraid this will bump the minimum supported Rust version considerably. Maybe we should wait a bit first!

PR #1405, opened 2022-08-09
Fix Tooltip inside Scrollable

tarkah:
`Tooltip` inside `Scrollable` is currently broken since the `viewport` passed to the `Tooltip` is modified by `Scrollable` to be it's visible bounds instead of the application viewport bounds. This causes the `Tooltip` to position itself within these bounds.

I've added a builder to `Tooltip` to disable the positioning adjustment that ensures its always inside the `viewport` with `snap_within_viewport(bool)`. I've then also removed the layer clip when drawing the tooltip so it's not clipped by the scrollable bounds (tooltip is effectively an overlay and should never be clipped).

Commit b3f5f74 shows the bug and 046e7e0 shows it being fixed. These commits can get dropped.

I realize this is a hacky solution, since `snap_within_viewport` shouldn't be something that's exposed or even considered and isn't really honest about why it's there. However, since we don't get the _actual_ window viewport in the widget, we have no way of know when or when not to adjust to the viewport.

Should we do something like the following, so there is a `window` bounds that can be used that we know is unchanged during `draw`. `Tooltip` can then use this for it's positioning adjustment to the window bounds.

```rs
fn draw<'a>(
    &'a self,
    ...,
    viewport: &Rectangle,
    window: &'a Rectangle,
)
```

### DIFF
--- a/native/src/widget/tooltip.rs
+++ b/native/src/widget/tooltip.rs
@@ -349,7 +349,7 @@ pub fn draw<Renderer>(
                 viewport.y + viewport.height - tooltip_bounds.height;
         }
 
-        renderer.with_layer(*viewport, |renderer| {
+        renderer.with_layer(Rectangle::with_size(Size::INFINITY), |renderer| {
             container::draw_background(renderer, &style, tooltip_bounds);
 
             draw_text(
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks! This should be good enough until we generalize our overlay support.

PR #1411, opened 2022-08-15
Don't double translate text clip inside canvas

tarkah:
The clip bounds is already inside a `Translate`, so using `region` offsets the clip bounds by translate again. Clip within the translate should start at origin w/ region size.


hecrj (approved):
Thanks! Good catch.

PR #1414, opened 2022-08-17
Don't constrict tooltip text layout to viewport size

tarkah:
Oops, I forgot to remove this constraint. The text layout also shouldn't be constrained to the viewport size (when viewport is set by scrollable).


hecrj (approved):
:+1:

PR #1424, opened 2022-08-26
When pane grid title bar is cramped, still show content until hover and allow dragging where content would've been

mtkennerly:


hecrj (approved):
Awesome :tada:

PR #1362, opened 2022-06-08
Theming

hecrj:
This PR completely overhauls styling by implementing first-class support for themes, as described in the recently merged RFC: https://github.com/iced-rs/rfcs/pull/6.

The default built-in themes and their corresponding color palettes are mostly placeholders and will most likely change as we fine-tune and iterate on the default looks of `iced`.

PR #1425, opened 2022-08-27
Make `overlay::menu` accessible

### LINKED ISSUE #1406: No way to access `menu::Stylesheet`

### What happened?

On the current master branch, there is no way to access the `menu::StyleSheet` from `iced` itself, so you can't implement theming for a picklist without pulling in `iced_style` manually.

There should be a `menu` mod or something in `iced`. 

### Version

master

### Operative System

Windows

_No response_

0x192:
Implementing `pick_list::StyleSheet` for a custom `Theme` requires to also implement `menu::StyleSheet` which is currently not accessible.

This pull request creates an alias of `overlay::menu` in the root so you can import it with `use iced::overlay::menu`.

Fixes #1406

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -174,6 +174,7 @@ pub mod clipboard;
 pub mod executor;
 pub mod keyboard;
 pub mod mouse;
+pub mod overlay;
 pub mod settings;
 pub mod time;
 pub mod widget;
--- a/src/overlay.rs
+++ b/src/overlay.rs
@@ -1,0 +1,10 @@
+//! Build and show dropdown menus.
+///
+/// This is an alias of an `iced_native` element with a default `Renderer`.
+pub mod menu {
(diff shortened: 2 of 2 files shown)


0x192:
Note sure about the `improvement` label. It's more of a `bugfix` of than an `improvement`.

#1362 introduced the theme support allowing to implement a custom style for all widgets except the `pick_list` because you need to implement `menu::stylesheet` which is not accessible.

hecrj (approved):
Thanks! :tada:

0x192:
Thanks to you!

Why did you need to add this? I'm not sure to understand

```rust
pub type Element<'a, Message, Renderer = crate::Renderer> =
    iced_native::overlay::Element<'a, Message, Renderer>;
``` 

(no problem if you don't have time to answer)

hecrj:
Just figured we should expose `overlay::Element` as well, since we are exposing `overlay::Menu` an it is part of its public API.

PR #1393, opened 2022-07-27
Replace stateful widgets with the new `iced_pure` API

hecrj:
This PR replaces the current stateful widgets with the new widget API introduced by `iced_pure` in #1284.

The changes here remove a lot of code and prepare the codebase for the soon-to-land widget operations described in https://github.com/iced-rs/rfcs/pull/7.

Some of the most relevant changes are:

- `iced_pure` code has been removed and the crate is deprecated.
- The `pure` feature has been removed.
- Widget types are no longer re-exported in the root `iced` crate. Instead, they have to be properly imported with the full path (eg. `use iced::widget::Text`).
- The widget function helpers in `iced_pure` have been moved to `iced::widget`.
- A couple of new macros have been added in `iced::widget`: `column` and `row`. They are analogous to `vec!`. Therefore, the old approach
    ```rust
    use iced::widget::Column;

    Column::new().push("I am a column").push("with 2 children");
    ```
    can now be written as follows:

    ```rust
    use iced::widget::column;

    column!["I am a column", "with 2 children!"]
    ```
- The `row` and `column` function helpers are also available, and they are equivalent to `Row::with_children` and `Column::with_children`. These are useful when you have an iterator or a `Vec` of elements!
- `Text::new` now can take any `ToString` instead of `Into<String>`.
- All of the internal `State` types are no longer exported in `iced`.

Thanks to all of these changes, all of the examples are now much simpler and less verbose! For instance, here is the new `counter` example:

```rust
use iced::widget::{button, column, text};
use iced::{Alignment, Element, Sandbox, Settings};

pub fn main() -> iced::Result {
    Counter::run(Settings::default())
}

struct Counter {
    value: i32,
}

#[derive(Debug, Clone, Copy)]
enum Message {
    IncrementPressed,
    DecrementPressed,
}

impl Sandbox for Counter {
    type Message = Message;

    fn new() -> Self {
        Self { value: 0 }
    }

    fn title(&self) -> String {
        String::from("Counter - Iced")
    }

    fn update(&mut self, message: Message) {
        match message {
            Message::IncrementPressed => {
                self.value += 1;
            }
            Message::DecrementPressed => {
                self.value -= 1;
            }
        }
    }

    fn view(&self) -> Element<Message> {
        column![
            button("Increment").on_press(Message::IncrementPressed),
            text(self.value).size(50),
            button("Decrement").on_press(Message::DecrementPressed)
        ]
        .padding(20)
        .align_items(Alignment::Center)
        .into()
    }
}
```

[...]

PR #1429, opened 2022-09-01
Remove old pure feature from docs.rs build metadata

pheki:
It looks like the `pure` feature was removed in #1393


hecrj (approved):
Thank you! Good catch 😄

PR #1432, opened 2022-09-07
Add custom palette to built in theme

wash2:
This PR adds an option for a custom palette in the built in theme as described in the theme RFC with a small change to include the generated Extended struct in the Custom variant

### DIFF
--- a/examples/styling/src/main.rs
+++ b/examples/styling/src/main.rs
@@ -159,6 +159,6 @@ impl Sandbox for Styling {
     }
 
     fn theme(&self) -> Theme {
-        self.theme
+        self.theme.clone()
     }
 }
--- a/examples/styling/src/main.rs
+++ b/examples/styling/src/main.rs
@@ -53,7 +53,10 @@ impl Sandbox for Styling {
         };
         let extended = Extended::generate(palette);
         Styling {
-            custom_theme: Theme::Custom { palette, extended },
+            custom_theme: Theme::Custom {
+                palette: Box::new(palette),
+                extended: Box::new(extended)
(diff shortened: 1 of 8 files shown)


examples/styling/src/main.rs
```diff
@@ -1,14 +1,35 @@
...
+        primary: Color::from_rgb(0.5, 0.5, 0.0),
+        success: Color::from_rgb(0.0, 1.0, 0.0),
+        danger: Color::from_rgb(1.0, 0.0, 0.0),
+    };
+    let extended = Extended::generate(palette);
+    CUSTOM_THEME.set(Theme::Custom { palette, extended }).unwrap();
     Styling::run(Settings::default())
 }
 
+static CUSTOM_THEME: OnceCell<Theme> = OnceCell::new();
```
IohannRabeson, line 24: Instead of using this static variable CUSTOM_THEME, I would instead store the custom theme in the enum ThemeType::Custom:
```rust
enum ThemeType {
    Light,
    Dark,
    Custom(Theme),
}
```
The update would be something like:
```rust
Message::ThemeChanged(theme) => self.theme = match theme {
    ThemeType::Light => Theme::Light,
    ThemeType::Dark => Theme::Dark,
    ThemeType::Custom(custom_theme) => custom_theme,
},
```

hecrj (approved):

I have implemented the opaque `Custom` type we talked about in the last monthly meeting. This way, we ensure that the `Extended` palette is always consistent with the provided `Palette` input.

The new `Custom` struct also allows us to only `Box` once :)

PR #1437, opened 2022-09-11
fix: sync size of window and viewport

wash2:
I noticed that `iced::window::resize` did change the size of the window, but didn't sync with the viewport size, like how a user resize would. This syncs the size of the window and viewport.

### DIFF
--- a/winit/src/application/state.rs
+++ b/winit/src/application/state.rs
@@ -3,7 +3,6 @@ use crate::conversion;
 use crate::{Application, Color, Debug, Point, Size, Viewport};
 
 use std::marker::PhantomData;
-use winit::dpi::PhysicalSize;
 use winit::event::{Touch, WindowEvent};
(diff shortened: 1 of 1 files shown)


hecrj:
Interesting... I figured this branch in the `update` logic would deal with it:

https://github.com/iced-rs/iced/blob/465541125612ef6e70623098b956d5c757b0b1cf/winit/src/application/state.rs#L124-L133

But maybe resizing through the `Window` API does not trigger a `Resized` event in `winit`?

wash2:
> But maybe resizing through the `Window` API does not trigger a `Resized` event in `winit`?

Ya that seemed to be the case. I can look into making an issue / PR at winit for it as well.

hecrj:
@wash2 That would be great!

hecrj (approved):

hecrj:
Thank you :bow:

PR #1440, opened 2022-09-15
Document that `window::Action::Move` is unsupported on Wayland

ids1024:
https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_outer_position notes that this isn't supported on Wayland.

Wayland by design doesn't allow applications to position windows arbitrarily. GTK4 in comparison removed `gtk_window_move()` (which naturally didn't work on Wayland).

### DIFF
--- a/native/src/window/action.rs
+++ b/native/src/window/action.rs
@@ -13,6 +13,8 @@ pub enum Action<T> {
         height: u32,
     },
     /// Move the window.
+    ///
+    /// Unsupported on Wayland.
     Move {
         /// The new logical x location of the window
         x: i32,


13r0ck:
This is a limitation of [smithay-client-toolkit](https://github.com/smithay/client-toolkit) not exposing a way for winit to specify a window as xdg_popup, once that is in place the new [multi-window](https://github.com/iced-rs/iced/pull/1439) should be able to expose the type of window for wayland (either xdg_popup or xdg_toplevel)

ids1024:
Then there will need to be a different API to provide a relative position of the popup using [`xdg_positioner`](https://wayland.app/protocols/xdg-shell#xdg_positioner) (and similar relative positioning on X/Windows/macOS/etc.).

ids1024:
> This is a limitation of [smithay-client-toolkit](https://github.com/smithay/client-toolkit) not exposing a way for winit to specify a window as xdg_popup

It looks like the git version of `smithay-client-toolkit` adds some helpers for popups, that don't exist in the release that's used in winit (though the api of `smithay-client-toolkit` and `wayland-rs` have changed a lot in breaking ways since the last release). But it should also possible to do things like that without helpers from `smithay-client-toolkit`, it just takes more manual code.

13r0ck:
> But it should also possible to do things like that without helpers from smithay-client-toolkit, it just takes more manual code.

PR's welcome :zany_face:

PR #1443, opened 2022-09-19
Allow to replace an element instead of append to body for web rendering

traxys:
In case you want to have the application at the start of the page and some other content after it just appending to the body won't work

With this change you can add a  `<div  id="iced_root"></div>` and it will be replaced by the canvas

### DIFF
--- a/winit/src/application.rs
+++ b/winit/src/application.rs
@@ -159,9 +159,14 @@ where
         let document = window.document().unwrap();
         let body = document.body().unwrap();
 
-        let _ = body
-            .append_child(&canvas)
-            .expect("Append canvas to HTML body");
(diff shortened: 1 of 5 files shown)


13r0ck:
How cool would it be to have an Elm lang package that allows for Iced embeds using this!
Elm could handle all the SEO, pages, general web stuff, and Iced could be for the parts that don't need SEO and/or rendering of more complicated processes.

hecrj (approved):
Thanks! This was long overdue :bow: 

I have introduced a `target` field to our `PlatformSpecific` settings that allows users to provide the identifier of the DOM node they want to target. This way, we avoid hardcoding `iced_root` and, at the same time, we allow users to easily embed multiple `iced` applications in a single website.

PR #1448, opened 2022-09-29
Add linear gradient support to canvas widget

bungoboingo:
# **GRADIENTS**

### This PR adds support for linear gradients (radial & conical TBD) as a `Fill` style for canvas widgets.

Please note this does **_NOT_** add support for gradients for quads (yet), only 2d meshes on a canvas.

Big thanks to @tarkah for letting me yoink his original linear-gradient wgpu PR to get my feet wet in Iced. 🤗 

### Known issues:
~1) When running examples with both a solid & gradient shader with the `glow` backend, only one shader type will be drawn until the window is resized.~
~2) Rainbow example is broken, need to rework it 😢~ <- this was mutually agreed upon to be removed
~3) Declaring a color stop offset out of order (e.g. "0.3", "0.6", "0.2", then "1.0") causes the gradient to straight up not work~

## Usage

When creating a primitive you can now change the `FillStyle` to be either `Solid` or a `Gradient`. This changes the original field from `color` to `style`. E.g.:

### Before:

```rust
Fill {
    color: Color::BLACK,
    .. Default::default()
}
```

### After:

```rust
Fill {
    style: fill::Style::Solid(Color::BLACK),
    .. Default::default()
}

//or!

let gradient = Gradient::linear([Point::ORIGIN, Point::new(bounds.width, bounds.height)])
    .add_stop(0.0, Color::from_rgb(1.0, 0.0, 0.0)) //color at the beginning of the gradient (red)
    .add_stop(0.5, Color::from_rgb(0.0, 1.0, 0.0)) //color half-way through the gradient (green)
    .add_stop(0.1, Color::from_rgb(0.0, 0.0, 1.0)) //color at the end of the gradient (blue)
    .build()
    .unwrap(); //invalid color stop locations will panic

Fill {
    style: fill::Style::Gradient(&gradient),
    .. Default::default()
}
```

Any direction of linear gradient can be created, with `start` and `end` points. Color "stops" are added as a percentage along the total length of the gradient, akin to [making a linear-gradient with CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient). So a color stop with an offset of "0.5" would mean the color is fully rendered at 50% of the total length of the gradient. A color stop with an offset of "0.0" would mean the color is fully rendered at 0% of the total length, etc.

This builder was created by @tarkah (I yoinked it from his PR) and not sure that git will attribute it correctly, so big thanks to him!!

## Implementation Details (Feedback appreciated!)

I am still fairly new to graphics programming, WGPU in particular, so any feedback would be appreciated! 🤗 

Iced currently has no support for multiple shader types. In this PR I've attempted to modularize adding new pipelines (wgpu)/programs(opengl) so we can add more shader types easier in the future.

My first change is to remove color data from the current Mesh2D attribute data, leaving only position data as a `Vertex2D`. For gradients this attribute is unneeded and was wasting quite a bit of bytes in the attribute buffers, so it has been purged. For solids as well this wasn't really needed as a fill generally does not switch colors per-vertex, but was probably more efficient than adding a uniform write potentially every draw. I've switched all color information to being uniform based. Performance implications of this TBD.

Other areas that were changed:
- Modularized the WGPU/Glow rendering backends to be a little more scalable and tried to remove as much code duplication as possible (within reason!)
- Added different abstractions for buffer types in WGPU backend (static & dynamic, more can be added as needed e.g. a static uniform buffer)
- Added a new dependency to WGPU backend called `encase` ([crates.io](https://crates.io/crates/encase)). This ensures compile-time safety for alignment & padding requirements for all uniforms in accordance with the WebGPU specification. This crate is recommended by the WGPU team and used in other large projects using WGPU for easier data management (Bevy being the biggest one). It introduces a derived `ShaderType` trait which generates metadata to perform checks on. It also comes with its own CPU buffer implementation which will pad data on write, making it very easy to avoid alignment issues.

## Known areas to be improved

I will be doing some heavy profiling to see bottlenecks, but there are some obvious areas of improvement that I know will increase performance without doing so.

1) Add implementation for OpenGL using VBOs for OpenGL 3.1+
2) Add *another* implementation using SSBOs for OpenGL 4.3+ 😢 
    - Due to these minimum version requirement for these features currently the gradient implementation on OpenGL has a hard-coded limit of 16 color stops. Any additional stops added after this limit will be ignored.
    - In contrast, the technical color stop limit on WGPU is 524,288 stops. 😿
anyways. But still good to do now if there is a uniform that passes the 256 byte threshold in the future.
3) We do not need to be writing to the GPU buffer every draw call for WGPU & Glow backends both; the only issue right now is that for a Mesh we are storing attribute data in a buffer that can stay the same exact size but have its content change (e.g. in an example of a single rectangle with a fill that is update to be a new rectangle with a different fill; the attribute data would be the same size, but the uniform data must be changed). Thinking about best way to guard against this.
    - This would involve putting the burden of calculating a diff on the CPU vs just rewriting every draw() on the GPU. Performance could be better in either situation depending on hardware. For now we are just leaving it as writing to the GPU every draw.
~4) Uniforms can be rearranged in WGPU to pack them more tightly; working on this now in addition to the known issue(s) above.~
5) Creating an [UBER shader](https://radeon-pro.github.io/RadeonProRenderDocs/en/uber/about.html#:~:text=The%20Uber%20Shader%20is%20a,ProRender%20updated%20with%20new%20materials.) instead of having separate shaders might be a performance increase. Will have to see how much of a bottleneck swaping shader programs is during profiling.
6) Remove branching from both fragment shaders for better performance since the `else` case will get executed every time on the GPU unless it performs some compilation-time unrolling, which, to my knowledge after some small amount of research, seems pretty hit or miss.

## Areas to improve usability

1) Right now it's slightly awkward to make a gradient for a rectangle and have to match the start/end points to the bounds to get it to fill the whole rectangle. I'd like to implement something like `deg(0)` that you can do with CSS gradients to just have a gradient fill the primitive at the right angle. This would be fairly trivial and just involve a wee bit of math. Keywords like `left` `top` etc would also be useful and make this process less awkward.

**Edit:** I have added a way to use relative positioning, but it still has the requirement of a bounds to know what to position relative *to*. A tad more cumbersome than the CSS implementation where you know the bounds of the object beforehand so don't need to specify it, but otherwise might be more comfortable to use in certain use-cases. For quad support I believe this bounds check can be builtin.

```rust
    Position::Relative {
        top_left: Point::ORIGIN,
        size: bounds.size(),
        start: Location::TopLeft,
        end: Location::BottomRight
    }
)
    .add_stop(0.0, Color::from_rgb(1.0, 0.0, 0.0)) //color at the beginning of the gradient (red)
    .add_stop(0.5, Color::from_rgb(0.0, 1.0, 0.0)) //color half-way through the gradient (green)
    .add_stop(0.1, Color::from_rgb(0.0, 0.0, 1.0)) //color at the end of the gradient (blue)
    .build()
    .unwrap(); //invalid color stop locations will panic
```

2) ??? Seeking feedback! What do you think?


graphics/src/widget/canvas/gradient/linear.rs
```diff
@@ -0,0 +1,73 @@
...
+            return None;
+        }
+
+        Some(Gradient::Linear(Linear {
+            start: self.start,
+            end: self.end,
+            color_stops: self
+                .stops
+                .into_iter()
+                .map(|f| ColorStop {
```
clarkmoody, line 66: Do the `ColorStop`s need to be sorted?

examples/arc/src/main.rs
```diff
@@ -2,7 +2,7 @@ use std::{f32::consts::PI, time::Instant};
 
 use iced::executor;
 use iced::widget::canvas::{
-    self, Cache, Canvas, Cursor, Geometry, Path, Stroke,
+    self, Cache, Canvas, Cursor, Geometry, Path, Stroke, StrokeStyle,
```
hecrj, line 5: Probably worth exposing a `stroke` module and simply define `Style` in there. E.g.

```rust
use widget::canvas::stroke::{self, Stroke};
```

Then write `stroke::Style`.

examples/geometry/src/main.rs
```diff
@@ -1,6 +1,9 @@
 //! This example showcases a simple native custom widget that renders using
 //! arbitrary low-level geometry.
+//!
+//TODO need to update this now that vertex data doesn't contain color data
 mod rainbow {
```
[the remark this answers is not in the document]
hecrj: We don't necessarily need to support meshes with color interpolation. I believe it's a pretty weird use case.

Is any of the watchers of the repository using this feature?

examples/modern_art/src/main.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+#[derive(Debug, Clone, Copy)]
+enum Message {}
+
+struct ModernArt {
+    cache: Cache,
+}
+
+impl Application for ModernArt {
```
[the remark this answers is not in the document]
hecrj: It's great! Examples should be fun!

glow/src/shader/common/gradient.frag
```diff
@@ -0,0 +1,48 @@
+// GLSL does not support dynamically sized arrays without SSBOs
+#define MAX_STOPS 64
```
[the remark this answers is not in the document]
hecrj: Seems more than plenty! I don't think you should ever use more than 3 stops in practice...

glow/src/triangle/gradient.rs
```diff
@@ -0,0 +1,189 @@
+use crate::program::Version;
+use crate::triangle::{simple_triangle_program, update_transform};
+use glow::{Context, HasContext, NativeProgram};
+use iced_graphics::gradient::Gradient;
+use iced_graphics::widget::canvas::gradient::Linear;
+use iced_graphics::Transformation;
+
+#[derive(Debug)]
+pub(super) struct GradientProgram {
```
hecrj, line 9: Nit: Do these need to be `super`? I'm not sure `GradientProgram` should be part of the public API of `iced_glow`, and if it's not already... Then it's fine to just leave it public for the whole crate.

glow/src/triangle/gradient.rs
```diff
@@ -0,0 +1,189 @@
...
+                                        .color_stops_locations[index]
+                                        .color,
+                                ),
+                                stop.color.r,
+                                stop.color.g,
+                                stop.color.b,
+                                stop.color.a,
+                            );
+                        }
+                    }
```
hecrj, line 103: I wonder if there is a way to avoid many memory transfers and instead perform a single one for all the stops?

graphics/src/gradient.rs
```diff
@@ -0,0 +1,103 @@
...
+/// or conically (TBD).
+pub enum Gradient {
+    /// A linear gradient interpolates colors along a direction from its [`start`] to its [`end`]
+    /// point.
+    Linear(Linear),
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+/// A point along the gradient vector where the specified [`color`] is unmixed.
+pub struct ColorStop {
```
hecrj, line 17: Is just `Stop` still understandable in the `gradient` context? `gradient::Stop`...

graphics/src/layer.rs
```diff
@@ -323,3 +334,21 @@ unsafe impl bytemuck::Zeroable for Quad {}
...
+        self.0
+            .iter()
+            .map(|Mesh { buffers, .. }| {
+                (buffers.vertices.len(), buffers.indices.len())
+            })
+            .fold((0, 0), |(total_v, total_i), (v, i)| {
+                (total_v + v, total_i + i)
+            })
+    }
+}
```
hecrj, line 354: Instead of a newtype without opacity, we could expose this method as a function somewhere else and make it take `&[Mesh<'a>]`.

graphics/src/widget/canvas/fill.rs
```diff
@@ -19,24 +21,42 @@ pub struct Fill {
...
-            color,
+            style: FillStyle::Solid(color),
             ..Fill::default()
         }
     }
 }
 
+/// The color or gradient of a [`Fill`].
+#[derive(Debug, Clone)]
+pub enum FillStyle<'a> {
```
hecrj, line 44: We can drop the prefix, since this is the `fill` module already!

graphics/src/shader.rs
```diff
@@ -0,0 +1,23 @@
+//! Supported shaders;
+
+use crate::Color;
+use crate::gradient::Gradient;
+
+#[derive(Debug, Clone)]
+/// Supported shaders for primitives.
+pub enum Shader {
```
hecrj, line 8: I'd say this makes more sense as a `mesh::Style` or similar!

graphics/src/widget/canvas/frame.rs
```diff
@@ -331,52 +331,24 @@ impl Frame {
...
-                    vertices: self.buffers.vertices,
-                    indices: self.buffers.indices,
-                },
-                size: self.size,
-            });
+        for (buffer, shader) in self.buffers {
+            if !buffer.indices.is_empty() {
+                self.primitives.push(Primitive::Mesh2D {
+                    buffers: triangle::Mesh2D {
+                        vertices: vertices_from(buffer.vertices),
```
hecrj, line 338: I wonder if we could reuse `lyon::math::Point` internally to avoid copies here.

wgpu/src/buffers.rs
```diff
@@ -0,0 +1,3 @@
+//! Utilities for buffer operations.
+pub mod buffer;
+pub mod dynamic_buffers;
```
hecrj, line 3: Maybe just a `buffer` module with a `dynamic` one inside?

wgpu/src/buffers/buffer.rs
```diff
@@ -0,0 +1,91 @@
+//! Utilities for static buffer operations.
+
+/// A generic buffer struct useful for items which have no alignment requirements
+/// (e.g. Vertex, Index buffers) and are set once and never changed until destroyed.
+///
+/// This buffer is mapped to the GPU on creation, so must be initialized with the correct capacity.
+#[derive(Debug)]
+pub(crate) struct StaticBuffer {
```
hecrj, line 8: I wonder if we should add a generic type `T` with `PhantomData` to ensure type safety.

wgpu/src/buffers/buffer.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /// Returns whether or not the buffer needs to be recreated. This can happen whenever the mesh 
+    /// data is re-submitted.
+    pub fn needs_recreate(&self, new_size: usize) -> bool {
+        self.size != new_size as u64
+    }
+
+    /// Writes the current vertex data to the gpu buffer with a memcpy & stores its offset.
+    pub fn write(&mut self, offset: u64, content: &[u8]) {
```
hecrj, line 48: If we have a generic type `T` we can ask for a slice of `&[T]` here and require the `bytemuck` traits to convert the slice to bytes in this method.

hecrj (changes requested):
Looking great! Excellent 🎉 

I left some feedback here and there after a quick glance at the code. Let me know what you think!

graphics/src/widget/canvas/frame.rs
```diff
@@ -107,25 +109,24 @@ impl Frame {
                 &options,
                 &mut buffers,
             )
-        };
+        }.expect("Tessellate path.");
 
-        result.expect("Tessellate path");
+        self.buffers.push((buf, style.into()))
     }
```
tarkah, line 115: The Gradient start / stop isn't currently affected by the current transform. So if we are not on the identity transform, the gradient direction vector will be inaccurate compared to this path. Instead of having a `From<(Stroke|Fill)Style> for Shader`, can there instead be a `build` type method that accepts `Transform`? We can then expose a `Transform::transform_point` helper and apply it against the gradients start / stop before storing it as a shader.

examples/geometry/src/main.rs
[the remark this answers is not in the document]
tarkah: Agreed. I think gradients pretty much solve whatever use case there is for manually assigning vertex colors, which is to effectively create some sort of gradient.

tarkah (changes requested):
Awesome work taking this over @bungoboingo! I'm very excited to see this working on the opengl side now, too!

glow/src/triangle/gradient.rs
```diff
@@ -0,0 +1,189 @@
+use crate::program::Version;
+use crate::triangle::{simple_triangle_program, update_transform};
+use glow::{Context, HasContext, NativeProgram};
+use iced_graphics::gradient::Gradient;
+use iced_graphics::widget::canvas::gradient::Linear;
```
tarkah, line 5: I'm getting an import error here?

```suggestion
use iced_graphics::gradient::Linear;
```

graphics/src/gradient.rs
```diff
@@ -0,0 +1,103 @@
+//! For creating a Gradient.
+use iced_native::Color;
+use crate::gradient::linear::Linear;
```
tarkah, line 3: Nit: Make this `pub use` so we can `use gradient::Linear` instead of `gradient::linear::Linear`

examples/modern_art/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+use crate::canvas::{Cursor, FillStyle, Geometry};
+use iced::widget::canvas::{Cache, Fill, Frame};
+use iced::widget::{canvas, Canvas};
+use iced::Settings;
+use iced::{
+    executor, Application, Color, Command, Element, Length, Point, Rectangle,
+    Renderer, Size, Theme,
+};
+use iced_graphics::gradient::Gradient;
+
```
tarkah, line 11: Shouldn't `Gradient` be available under the `iced::widget::canvas` since it's only use is in the canvas API? This extends to the `gradient` module as having access to `gradient::Builder` in application code will be needed.

graphics/src/widget/canvas/frame.rs
tarkah: or something like `Gradient::from(style).transform(self.transforms.current)`

graphics/src/widget/canvas.rs
```diff
@@ -37,6 +37,8 @@ use iced_native::{
     Clipboard, Element, Length, Point, Rectangle, Shell, Size, Vector, Widget,
 };
 
+pub use crate::gradient::Gradient;
+
```
tarkah, line 41: Nit: Can be group with the `use::crate` import above

tarkah: Yeah just location wise. I know we don't have any consistency, but I try to strive for `StdExternalCrate` as seen here:

https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=import+gr#group_imports

tarkah (commented):
Can't comment much on the glow side since I don't know opengl, but the changes made to the wgpu triangle pipeline & new buffer types are really nice!! It is much more readable and there's a very clear path for adding new pipelines now. Great job!

A few nits but otherwise this looks good to me!

graphics/src/widget/canvas.rs
```diff
@@ -28,7 +28,7 @@ pub use stroke::{LineCap, LineDash, LineJoin, Stroke};
 pub use text::Text;
 
 use crate::{Backend, Primitive, Renderer};
-pub use crate::gradient::Gradient;
+pub use crate::gradient::{Gradient, Position, Location};
```
tarkah, line 31: I'm not sure we should re-export `Position` and `Location` directly from `canvas` since they won't have the context of `gradient`.

```rs
use iced::widget::canvas::gradient;

gradient::Position {
    ..
}
```

vs 

```rs
use iced::widget::canvas::Position;

// Too generic, Position can be for so many things in a 2D canvas API
Position {
    ..
}
```

graphics/src/layer.rs
```diff
@@ -234,92 +246,14 @@ impl<'a> Layer<'a> {
...
-    /// The border color of the [`Quad`], in __linear RGB__.
-    pub border_color: [f32; 4],
-
-    /// The border radius of the [`Quad`].
-    pub border_radius: f32,
-
-    /// The border width of the [`Quad`].
-    pub border_width: f32,
+/// Returns the number of total vertices & total indices of all [`Mesh`]es.
+pub fn attribute_count_of<'a>(meshes: &'a [Mesh<'a>]) -> (usize, usize) {
```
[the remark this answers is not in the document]
tarkah: > Missing some aspects of Kotlin especially their extension function syntax e.g. something like `Vec<Mesh<'a>>.attribute_count(&self) -> (usize, usize)` which would be nice and clean but Rust doesn't like for me to play around with types I didn't make, the bastards

I think you can accomplish this by creating a Trait and impl it on `Vec<Mesh<'a>>` then you can `meshes.attribute_count()` but then you have to import the trait so it's not that great still.

glow/src/triangle.rs
```diff
@@ -207,50 +137,74 @@ impl Pipeline {
...
+                match mesh.style {
+                    mesh::Style::Solid(color) => {
+                        self.programs.solid.use_program(gl, &color, &transform);
+                    }
+                    mesh::Style::Gradient(gradient) => {
+                        self.programs
+                            .gradient
+                            .use_program(gl, &gradient, &transform);
+                    }
+                }
```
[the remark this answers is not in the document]
tarkah: Wouldn't this break layering as determined by how the canvas API was used? I'd expect my stroke / fill calls to be rendered in the order I call them.

graphics/src/widget/canvas/frame.rs
```diff
@@ -15,7 +18,7 @@ use lyon::tessellation;
 #[allow(missing_debug_implementations)]
 pub struct Frame {
     size: Size,
-    buffers: lyon::tessellation::VertexBuffers<triangle::Vertex2D, u32>,
+    buffers: Vec<(tessellation::VertexBuffers<Vertex2D, u32>, mesh::Style)>,
```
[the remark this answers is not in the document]
tarkah: Similar to my comment above, we wouldn't be able to retain proper render ordering if we make this change

examples/solar_system/src/main.rs
```diff
@@ -198,15 +200,19 @@ impl<Message> canvas::Program<Message> for State {
...
+                        start: Point::new(-Self::EARTH_RADIUS, 0.0),
+                        end: Point::new(Self::EARTH_RADIUS, 0.0)
+                    }
+                )
+                .add_stop(0.2, Color::from_rgb(0.15, 0.50, 1.0))
+                .add_stop(0.8, Color::from_rgb(0.0, 0.20, 0.47))
+                .build()
+                .unwrap();
+
+                frame.fill(&earth, &earth_fill);
```
clarkmoody, line 215: Really nice addition here 👌

examples/solar_system/src/main.rs
```diff
@@ -215,14 +221,6 @@ impl<Message> canvas::Program<Message> for State {
                     let moon = Path::circle(Point::ORIGIN, Self::MOON_RADIUS);
                     frame.fill(&moon, Color::WHITE);
                 });
-
-                frame.fill(
-                    &shadow,
-                    Color {
-                        a: 0.7,
-                        ..Color::BLACK
-                    },
```
clarkmoody, line 224: wen moon shadow?

clarkmoody (approved):
Examples working for me on Apple M1. Gradient API looks good. I'm not qualified to comment on the shaders.

hecrj (approved):
Awesome work!!! 🥳

I took a final look and made some changes here and there. Specifically, I have

- Implemented a `BufferStack` in `canvas::Frame` so we can reuse the last buffer if the `mesh::Style` matches in 20a0577034b40a6bbabee9bbbfc085f3fd5016c0. We could eventually use a `HashMap` or `BTreeMap` instead, but this simple approach seems to work quite well since it's likely for users to draw similar styled geometry together. The `game_of_life` went from having 100+ buffers, to simply having 4!
- Implemented color conversion to linear RGB before uploading colors to the GPU in 62465842099908f9e50b8edabfec709b37b1ade3 and 9a02d60ba51a91049466caa46db73d741520e051. Generally, shader input/output should be in linear RGB. Otherwise, color blending (and any other color operations) will simply not work as expected.
- Introduced a flag to reuse the latest pipeline if possible in `iced_wgpu` in 93e309f491a8941bafb919e75d660e65071475f4. I noticed `set_pipeline` has a noticeable overhead, even if the pipeline is the same as the old one (not completely sure of this, it may have been just a fluke!).
- Run `cargo fmt` and fixed a bunch of `clippy` lints in b95745340441835bd25b5cadc2342254631f8c05 and 7e22e2d45293c5916812be03dc7367134b69b3ad, respectively.
- Refactored a bunch of imports and exports, renamed some modules, and moved some types around to remove cyclical dependencies in the other commits. I tried to keep the commits self-contained so you can take a look at them!

With these changes, the `game_of_life` example seems to run at pretty much the same speed as in `master` (maybe even a bit faster!).

Really great stuff here! Thanks again 🎉 Take a look at the latest changes, make sure I didn't break anything, and then I believe we can merge this!

graphics/src/widget/canvas/frame.rs
```diff
@@ -18,13 +18,41 @@ use lyon::tessellation;
...
+        match self.stack.last_mut() {
+            Some((_, current_style)) if current_style == &mesh_style => {}
+            _ => {
+                self.stack
+                    .push((tessellation::VertexBuffers::new(), mesh_style));
+            }
+        };
+
+        tessellation::BuffersBuilder::new(
+            &mut self.stack.last_mut().unwrap().0,
```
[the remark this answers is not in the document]
tarkah: `get` is still called sequentially as the canvas API is used. Iterating over the internal vec should be the order things were created in. Unless I'm missing something too?

hecrj:
No worries! Great work here :tada:

Issue #580, opened 2020-10-26
Bundle images


fenhl:
Is there a way to bundle images into the binary at compile time, rather than loading them from the file system at runtime (which means the app can't be a standalone `.exe`)?

fenhl:
I managed to find a workaround using [`iced::image::Handle::from_memory`](https://docs.rs/iced_native/0.2.2/iced_native/widget/image/struct.Handle.html#method.from_memory) and [`std::include_bytes`](https://doc.rust-lang.org/std/macro.include_bytes.html), but that involves cloning the images.

twitchyliquid64:
Seems related to #549 & #570. I agree we need a better way to load images into memory, and pass that through to the backend efficiently.

As for the specific problem of bundling images, I think include_bytes or writing into an ELF section in the binary is the way to go - I can't think of anything iced-specific here.

twitchyliquid64:
Btw, you can 'bundle' images into the binary using the `lazy_static` macro/crate combined with the `include_bytes` builtin macro. The 'bundling' looks like this:

```rust
    pub static ref SOIC8: image::Handle =
        image::Handle::from_memory(include_bytes!("../../res/package/SOIC-8.png").to_vec());
```

From there, you can clone it + instantiate a new image:

```rust
Image::new(SOIC8.clone())
```

Note that this approach does an allocation + bytes copy as part of the render path, which isn't ideal. In practice this is fast enough for icon-sized images **in release mode**.

PR #1453, opened 2022-10-03
Use `Cow<'static, [u8]>` in image/svg, add constructors taking &[u8]`

ids1024:
This should resolve https://github.com/iced-rs/iced/issues/580 by providing a way to use an image included with `include_bytes!` without needing to copy it to a `Vec` to create an image handle.

It would be nice if these methods could also be `const`, but that isn't possible due to the hashing being done.

This is technically a breaking change since `Handle::data()` is public. But if that is used, it's most likely in used somewhere that only relies on the type derefing to `&[u8]`.

### DIFF
--- a/native/src/image.rs
+++ b/native/src/image.rs
@@ -1,6 +1,7 @@
 //! Load and draw raster graphics.
 use crate::{Hasher, Rectangle};
 
+use std::borrow::Cow;
 use std::hash::{Hash, Hasher as _};
 use std::path::PathBuf;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


4gboframram:
About the constness issue, you could fix it yourself by using lazy_static, although it is naïve workaround and causes runtime overhead on the first dereference, but I agree that `Handle` should have this option.

ids1024:
Yep. Or perhaps `once_cell::sync::Lazy`, which seems to be favored over `lazy_static` now (and is in the standard library now, but not yet stabilized). That is a general solution to things like this, but it would be more elegant to have a `const` constructor.

hecrj (approved):
Awesome!

We can leverage `impl Into<Cow<'static, [u8]>>` to avoid writing new methods and still keep backwards compatibility!

PR #1464, opened 2022-10-07
Fix image/svg widget size allocation to use `full_size`, not `raw_size`

ids1024:
Previously only had an impact on the final size of dimensions set to `Shrink`. I'm guessing this was a mistake but anyway...

Before this change, setting `Fill` for height and width of an image, with `ContentFit::Contain`, caused it to take all available space in both dimensions. Where the ratio of the images differs from that of the space it has available, it preserves the ratio and centers along the dimension with extra space. I want to to allocate only the space it needs, without centering, and I don't see a way to do that currently.

The centering may be useful in some cases but should involve another property or widget.

### DIFF
--- a/native/src/widget/image.rs
+++ b/native/src/widget/image.rs
@@ -100,11 +100,11 @@ where
     let final_size = Size {
         width: match width {
             Length::Shrink => f32::min(raw_size.width, full_size.width),
-            _ => raw_size.width,
+            _ => full_size.width,
         },
         height: match height {
             Length::Shrink => f32::min(raw_size.height, full_size.height),
-            _ => raw_size.height,
+            _ => full_size.height,
         },
     };
 
--- a/native/src/widget/svg.rs
+++ b/native/src/widget/svg.rs
@@ -99,11 +99,11 @@ where
         let final_size = Size {
             width: match self.width {
                 Length::Shrink => f32::min(raw_size.width, full_size.width),
-                _ => raw_size.width,
+                _ => full_size.width,
             },
             height: match self.height {
                 Length::Shrink => f32::min(raw_size.height, full_size.height),
-                _ => raw_size.height,
+                _ => full_size.height,
             },
         };
 


ids1024:
Hm, looking at https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit, I guess "letterboxing" is the behavior of `contain` there?

I haven't written much css there, I suppose I should see how what I'm trying to do would be done there...

ids1024:
Ah, I guess what I'd use in CSS is `width: 100%; height: auto;`. So I guess if it's trying to follow a CSS like model, we'd need a `Length::Auto`? Maybe not the way Iced wants to handle it though.

13r0ck:
> So I guess if it's trying to follow a CSS like model, we'd need a Length::Auto? 

At least for this question with a fixed dimensiom image, Length::Shrink should do the same as your requested Length::auto

Length::Shrink meaning, as small as your smallest bound on this axis. Which is generally the same as what auto does in CSS

ids1024:
> Which is generally the same as what auto does in CSS

This is not true when `width` and `height` are set differently, when scaling up, at least.

You can generate a 1x1 pixel image with `convert -size 1x1 canvas:black 1x1.png`. With this, `` sets the width to fill the available space, and sets the height to the same to preserve aspect ratio (`auto` is the default behavior). In Iced, with `Shrink`, it doesn't allow a height greater than 1px here.

So CSS and GTK both have ways for the height allocated to depend on the width (and vice versa). Iced's image widget is supporting that only with `Shrink`.

ids1024:
It seems even `width: 100%; height: min-content` expands the height this way? I might expect `min-content` to be more link `Shrink`.

`object-fit: contain; width: 100%; height: 100%` does seem to match what Iced is doing with `Fill` and `Contain` though. (So this PR definitely isn't right if the behavior should match CSS, but the existing behavior isn't either.)

hecrj:
Since it's unclear what we actually want to do here, let's avoid breaking anything and have a proper discussion first about what the new behavior should be (RFC?).

Issue #759, opened 2021-03-02
Support for drawing custom headerbars


kaimast:
On most Unix desktop environments applications can draw an application-specific CSD that replaces the native window title bar.
Here are screenshots of Telegram doing this: https://github.com/electron/electron/issues/11907#issuecomment-615404410

It would be cool to do this in iced as well. For example draw a semi-transparent title bar over a 3d scene. 
This should be possible to implement once `winit` lands support for it (see [here](https://github.com/rust-windowing/winit/issues/221))

tonogram:
It would also be a nice touch if we can make the title bar transparent on MacOS, [like Tauri lets you do](https://tauri.app/blog/2022/11/08/tauri-1-2/#enhanced-titlebar-configuration-on-macos). Not sure what the best interface for this would be, seeing as it's a platform-specific thing, but it looks pleasant and would certainly be good to have available.

juxuanu:
Given this is not closed, does it mean one cannot achieve headerbars akin to GTK's/Adwaita's in Iced? Just curious.

kaimast:
> Given this is not closed, does it mean one cannot achieve headerbars akin to GTK's/Adwaita's in Iced? Just curious.

You can disable window decorations in winit and then draw whatever you want. There are just no widgets to do this easily.

I was hoping iced could at least have a widget that lets me drag the associated window, so I can build my own title bar with it.

tsuza:
> > Given this is not closed, does it mean one cannot achieve headerbars akin to GTK's/Adwaita's in Iced? Just curious.
> 
> You can disable window decorations in winit and then draw whatever you want. There are just no widgets to do this easily.
> 
> I was hoping iced could at least have a widget that lets me drag the associated window, so I can build my own title bar with it.

Use a `MouseArea` and then use `window::drag` when you're inside it.

kennethnym:
for anyone who wants to achieve transparent title bar with content underneath, i pieced together a small snippet of doing so, which sets the style whenever a new window is opened:

```rust
impl App {
    pub fn subscription(&self) -> Subscription<Message> {
        iced::event::listen().map(Message::IcedEventReceived)
    }

    pub fn update(&mut self, message: Message) -> Task<Message> {
        match message {
            Message::IcedEventReceived(iced::event::Event::Window(iced::window::Event::Opened { .. })) => {
                iced::window::get_latest().and_then(|id| {
                    iced::window::run_with_handle(id, |handle| {
                        on_window_opened(handle);
[...]

PR #1471, opened 2022-10-11
feat: Additional actions for window controls

mmstick:
Adds new window actions to enable the possibility to create client-side decorations in Iced applications. libcosmic has a functioning headerbar widget which utilizes these actions and their functions. We could add a headerbar widget example for demonstrating the functionality.

- `Action::Drag`
- `Action::Maximize(bool)`
- `Action::Minimize(bool)`
- `Action::ResizeDrag`
- `Action::ToggleMaximize`

`ResizeDrag` is currently a no-op because it requires a patch to winit. I am working on this patch at the moment.

Partially implements #759

Related RFC: https://github.com/iced-rs/rfcs/pull/13

### DIFF
--- a/native/src/window/action.rs
+++ b/native/src/window/action.rs
@@ -20,6 +20,8 @@ pub enum Action<T> {
     },
     /// Sets the window to maximized or back
     Maximize(bool),
+    /// Set the window to minimized or back
+    Minimize(bool),
     /// Move the window.
     ///
     /// Unsupported on Wayland.
@@ -50,6 +52,7 @@ impl<T> Action<T> {
             Self::Drag => Action::Drag,
             Self::Resize { width, height } => Action::Resize { width, height },
             Self::Maximize(bool) => Action::Maximize(bool),
+            Self::Minimize(bool) => Action::Minimize(bool),
             Self::Move { x, y } => Action::Move { x, y },
             Self::SetMode(mode) => Action::SetMode(mode),
             Self::ToggleMaximize => Action::ToggleMaximize,
@@ -68,6 +71,7 @@ impl<T> fmt::Debug for Action<T> {
                 width, height
             ),
             Self::Maximize(value) => write!(f, "Action::Maximize({})", value),
+            Self::Minimize(value) => write!(f, "Action::Minimize({}", value),
             Self::Move { x, y } => {
                 write!(f, "Action::Move {{ x: {}, y: {} }}", x, y)
             }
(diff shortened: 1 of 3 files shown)


winit/src/application.rs
```diff
@@ -615,12 +615,25 @@ pub fn run_command<A, E>(
...
                         height,
                     });
                 }
+                window::Action::ResizeDrag => {
+                    // TODO: Requires patch to winit
+                    // let _res = window.drag_resize_window();
+                }
+                window::Action::Maximize => {
+                    window.set_maximized(!window.is_maximized());
+                }
```
tarkah, line 633: Nit: I think this should this action should be called `ToggleMaximized`. Maybe we also have a `Action::Maximize(bool)` as well?

winit/src/application.rs
```diff
@@ -615,12 +615,25 @@ pub fn run_command<A, E>(
...
+                window::Action::Minimize => {
+                    window.set_minimized(true);
+                }
```
tarkah, line 636: Does it make sense to take a bool here? `Action::Minimize(bool)`

native/src/window/action.rs
```diff
@@ -5,13 +5,12 @@ use std::fmt;
...
 pub enum Action<T> {
-    /// Resize the window.
-    Resize {
-        /// The new logical width of the window
-        width: u32,
-        /// The new logical height of the window
-        height: u32,
-    },
+    /// Starts a window drag while mouse button is held.
+    Drag,
```
tarkah, line 9: I don't suppose there's any way we can guarantee this is called after a left button press is received? Maybe add a disclaimer similar to the one on the winit docs:

```
There’s no guarantee that this will work unless the left mouse
button was pressed immediately before this function is called.
```

tarkah (commented):
These actions are long overdue, very nice!

winit/src/application.rs
mmstick: From what I've read, only X11 supports setting false, but yeah it could be supported.

winit/src/application.rs
mmstick: Sure

winit/src/application.rs
ids1024: Looking at the implementation in winit and at https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_minimized, `set_minimized(false)` works on Windows, macOS, and X, but not Wayland.

hecrj (approved):
Great stuff! Thanks for the RFC as well :)

PR #1479, opened 2022-10-17
Simplify code, avoid unstable function

nilsmartel:
Makes code more obvious by reducing usage of an uncommon macro and an unstable feature.

While this is a tiny change, I believe it's nicer to use "more standard" (e.g. more common) code. I always like to believe obvious code should be preferred, especially when coding as a community.

Here we sacrifice close to nothing in terms of character count, while avoiding the use of an uncommonly used macro (avoiding this pattern is a personal opinion)
and avoiding an _unstable_ standard-library feature: https://doc.rust-lang.org/std/primitive.bool.html#method.then_some

### DIFF
--- a/native/src/program/state.rs
+++ b/native/src/program/state.rs
@@ -119,8 +119,9 @@ where
             .queued_events
             .iter()
             .zip(event_statuses)
-            .filter_map(|(event, status)| {
-                matches!(status, event::Status::Ignored).then_some(event)
+            .filter_map(|(event, status)| match status {
+                event::Status::Ignored => Some(event),
+                _ => None,
             })
             .cloned()
             .collect();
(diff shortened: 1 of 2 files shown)


Niedzwiedzw:
`matches!` is unstable?

geom3trik:
No, they are referring to the `then_some` method for bool, but actually it's only the const version which is unstable. Also I don't think the `matches!` macro is that uncommon personally.

13r0ck (approved):

nilsmartel:
I personally think there are very few and narrow cases, where matches! Should be preferred over match, especially since we would only ever safe  3-5 characters every now and  then while invoking some macro to avoid a feature that is as fundamental as the if statement.

More importantly I think matches! should never be used where == can be applied

fu5ha:
Clippy would disagree :P https://rust-lang.github.io/rust-clippy/master/#match_like_matches_macro though of course cilppy is not necessary the arbiter of all that is right in the world, it does show general community sentiment

nilsmartel:
That is a good hint, though this specific lint would not apply for the proposed code changes.

hecrj (commented):
I think these changes are fairly subjective and do not warrant a discussion.

PR #1480, opened 2022-10-17
More pane grid fixes

tarkah:
- dca99f3 - Fixes mouse interactions 
- fb03652 - Renders picked pane last. Similar to #1463, if panes have an internal canvas clip / `renderer.with_layer`, it's possible for their content to be drawn on top of the picked / dragging pane if the pane is iterated after the picked pane. This ensures picked pane is always rendered last and will therefore always come on top.

### DIFF
--- a/native/src/widget/pane_grid.rs
+++ b/native/src/widget/pane_grid.rs
@@ -757,7 +757,12 @@ pub fn draw<Renderer, T>(
         cursor_position
     };
 
-    for ((id, pane), layout) in elements.zip(layout.children()) {
+    // Render picked pane last
+    let mut elements = elements.zip(layout.children()).collect::<Vec<_>>();
+    elements
+        .sort_by_key(|((id, _), _)| picked_pane.map(|(id, _)| id) == Some(*id));
+
+    for ((id, pane), layout) in elements {
         match picked_pane {
             Some((dragging, origin)) if id == dragging => {
                 let bounds = layout.bounds();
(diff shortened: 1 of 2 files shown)


native/src/widget/pane_grid.rs
```diff
@@ -756,7 +757,12 @@ pub fn draw<Renderer, T>(
         cursor_position
     };
 
-    for ((id, pane), layout) in elements.zip(layout.children()) {
+    // Render picked pane last
+    let mut elements = elements.zip(layout.children()).collect::<Vec<_>>();
+    elements
+        .sort_by_key(|((id, _), _)| picked_pane.map(|(id, _)| id) == Some(*id));
```
hecrj, line 763: Allocating and sorting seems a bit overkill. Maybe we could loop for the non-picked panes first, and then execute a single iteration for the picked one?

hecrj (changes requested):

native/src/widget/pane_grid.rs
tarkah: Agreed, let me fix

tarkah: Resolved in 470723c. No reason to iterate twice, I just store the needed info in an Option and render it after the iterations

hecrj: Nice :tada:

hecrj (approved):
Thanks! :bow:

tarkah:
> Thanks! bow

Wait! Don't merge, I didn't test and I think I found a bug :P

Issue #674, opened 2020-12-27
`glow` backend does not support `Image`


rednaz1337:
Is it possible to run the `tour` example with the glow backend? when I add the `glow` feature to the example's `Cargo.toml` I get the following error:

```
error[E0277]: the trait bound `iced_glow::backend::Backend: iced_graphics::backend::Image` is not satisfied
   --> examples/tour/src/main.rs:675:5
    |
675 |     Container::new(
    |     ^^^^^^^^^^^^^^ the trait `iced_graphics::backend::Image` is not implemented for `iced_glow::backend::Backend`
    |
    = note: required because of the requirements on the impl of `iced_native::widget::image::Renderer` for `iced_graphics::renderer::Renderer<iced_glow::backend::Backend>`
    = note: required because of the requirements on the impl of `From<Image>` for `iced_native::element::Element<'_, _, iced_graphics::renderer::Renderer<iced_glow::backend::Backend>>`
    = note: required because of the requirements on the impl of `Into<iced_native::element::Element<'_, _, iced_graphics::renderer::Renderer<iced_glow::backend::Backend>>>` for `Image`
    = note: required by `iced_native::widget::container::Container::<'a, Message, Renderer>::new`
```

It seems like the `Image` widget is not supported on the `glow` backend.

cryptoquick:
Yeah, it looks like there's no `iced_glow/image` implementation, like there is for `iced_wgpu/image`.

cryptoquick:
Related to #354

genbyte:
Probably not the right place to ask, but is anyone working on `Image` for `iced_glow`? I'd very much like it to be a thing and would be happy to give `Image` a glow if someone isn't already working on it.

hecrj:
The current `Image` implementation has a bunch of underlying issues! https://github.com/iced-rs/iced/labels/image

Instead of porting the existing code to the `glow` backend, we should focus on coming up with a proper implementation that addresses all of these issues.

PR #1485, opened 2022-10-25
Glow image rendering support; move image/svg code to iced_graphics

ids1024:
Fixes https://github.com/iced-rs/iced/issues/674 and closes #1489.

This works, but duplicates code from `iced_wgpu` that should ideally be shared, and the cache never evicts.

The next step here is to work on an API design to move some of the image loading and caching logic from the backend to `iced_graphics` (or elsewhere?).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,9 +15,9 @@ resolver = "2"
 [features]
 default = ["wgpu"]
 # Enables the `Image` widget
-image = ["iced_wgpu/image", "image_rs"]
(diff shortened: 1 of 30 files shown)


hecrj (approved):
This has been a long-awaited feature! Thanks :partying_face: 

We are not leveraging a texture atlas in `iced_glow`, but this gets us one step closer!

I have made some changes here and there; check the latest commits! In any case, I think this is ready to go :ship:

PR #1489, opened 2022-10-26
Change uses of BGRA for images to RGBA and use image-rs 0.24

ids1024:
Is there a reason this was using BGRA? RGBA is more common, though BGR was historically popular at one point. `image` 0.24 has removed support for BGR images.

`image::Handle::from_pixels()` is renamed to `image::Handle::from_rgba()` to be more descriptive, and force any users to use RGBA instead of silently interpreting their pixels incorrectly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,7 +104,7 @@ iced_glow = { version = "0.3", path = "glow", optional = true }
 thiserror = "1.0"
 
 [dependencies.image_rs]
-version = "0.23"
+version = "0.24"
 package = "image"
 optional = true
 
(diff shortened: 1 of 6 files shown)


ids1024:
Updated to fix svg rendering (by removing the RGBA -> BGRA conversion, so this is an improvement).

Apparently wgpu [only guarantees](https://docs.rs/wgpu/latest/wgpu/struct.SurfaceConfiguration.html) `Bgra8Unorm` and `Bgra8UnormSrgb` are supported formats for surfaces, but I think for a texture RGBA should always be available?

PR #1490, opened 2022-10-26
Implement dynamic linking

DSeeLP:
Used by enabling the `dynamic` feature.
This reduces the compile time from about 30 seconds to 5-8 seconds for me in a personal project.

See #849


hecrj (changes requested):
Is there any chance this could be implemented completely on top of the current `iced` crate?

In other words, do we really need to introduce two new subcrates and support this feature ourselves?

PR #1491, opened 2022-10-26
wgpu: Do not offset `_position_in_atlas` and `_size_in_atlas`

ids1024:
Based on my testing, removing the `+ 0.5` and `- 1.0` here makes off-screen rendering of an image primitive output exactly the same data as the input image. Changing to `Nearest` interpolation has the same impact. Previously this resulted in a somewhat blurred output.

Unless there's some subtle issue I'm missing when there are more images in the atlas, this seems to be correct.

### DIFF
--- a/wgpu/src/image.rs
+++ b/wgpu/src/image.rs
@@ -550,12 +550,12 @@ fn add_instance(
         _position: position,
         _size: size,
         _position_in_atlas: [
-            (x as f32 + 0.5) / atlas::SIZE as f32,
-            (y as f32 + 0.5) / atlas::SIZE as f32,
+            x as f32 / atlas::SIZE as f32,
+            y as f32 / atlas::SIZE as f32,
         ],
         _size_in_atlas: [
-            (width as f32 - 1.0) / atlas::SIZE as f32,
-            (height as f32 - 1.0) / atlas::SIZE as f32,
+            width as f32 / atlas::SIZE as f32,
+            height as f32 / atlas::SIZE as f32,
         ],
         _layer: layer as u32,
     };


hecrj:
I think the reason for this logic was that, when using texture filtering, there may be bleeding when images are next to each other in the atlas.

You can try this out by reducing the `SIZE` constant in the `atlas` module and running the `svg` example:

image

ids1024:
Yeah, I was wondering if it was done this way for a reason like that.

The current way it's implemented doesn't look bad for a lot of images, but results in blurry output when used with text (this became rather apparent using an image primitive for text in `cosmic-text`). So in one way or another it needs to be fixed.

> Careful alignment may be needed to avoid bleeding between sub textures when used with [mipmapping](https://en.wikipedia.org/wiki/Mipmapping) and [texture compression](https://en.wikipedia.org/wiki/Texture_compression).

- https://en.wikipedia.org/wiki/Texture_atlas

Hm, `mip_map_level` is set to `1`, and I don't think it should be using texture compression...

My understanding is that texture atlases aren't necessarily the most efficient solution with modern graphics cards and APIs. (Relative to using texture arrays or multiple samplers). Still need to work on bench-marking that properly (https://github.com/iced-rs/iced/pull/1492).

hecrj:
> The current way it's implemented doesn't look bad for a lot of images, but results in blurry output when used with text

I wouldn't reuse the logic of an image atlas with a text atlas. They are completely different use cases.

hecrj:
> Also I think the advantage of atlasing is usually related to reducing draw calls? Currently there's one draw call per image. And the overhead is more a concern when drawing hundreds and thousands of objects a frame.

Not sure what makes you think that? We use instancing for drawing images. We can draw a 1000 images with a single draw call.

hecrj:
> (Relative to using texture arrays or multiple samplers)

The current atlas *is* a texture array as well :)

ids1024:
> I wouldn't reuse the logic of an image atlas for a text atlas. They are completely different use cases.

Probably, though I don't know what would be correct there. But anyway, images should still be implemented in a way that doesn't inherently add blurring like this, including images that may contain text.

> Not sure what makes you think that? We use instancing for drawing images. 

Ah, I see, the third argument to `draw_indexed()` is the number of instances. I'm more familiar with OpenGL so I was expecting a function with a name more like `glDrawElementsInstanced` and the use of something like `gl_InstanceID` in the shader. Maybe I should have looked up how instancing is handled in wgpu :laughing:.

Still worth benchmarking, I think. And figuring that "typical" uses cases to optimize for. 1000s of images on screen at once is probably quite unusual in a GUI app, while maybe 100 isn't too unsual with a lot of icons, with a file manager than shows previews of files, etc.

ids1024:
Performance with GPUs isn't always particularly logical. I've seen comments about instanced rendering preforming worse than a theoretically less "optimal" approach. And worse, it varies by GPU and driver.

hecrj:
Since this approach introduces artifacts, I believe we can close this.

Maybe we could try playing with the `position` and the `size` of the `Instance` instead? Or maybe we need to pad the images in the texture atlas?

PR #1492, opened 2022-10-26
WIP benchmarking and testing for renderers

ids1024:
This shows `iced_glow` outperforming `iced_wgpu`. Probably not accurate, some something may be wrong in the rendering and timing here? It should also test with more primitivies.

Tests pass when combined with https://github.com/iced-rs/iced/pull/1485 and https://github.com/iced-rs/iced/pull/1491.


ids1024:
I wanted benchmarking of some kind to answer questions about how the glow and wgpu backends compare in performance, whether or not the texture atlas performs better than separate textures or a texture array, etc. Though the numbers don't entirely make sense currently so more work here is probably needed.

But similar logic helps to automatically test that rendering produces the expected output. Might be good to get that working on CI if possible (should be possible to use both glow and wgpu with software rendering?).

ids1024:
Good point about Bevy. It makes sense that this is something they have worked on too.

hecrj:
This is cool! Thank you.

I think we can come back to it when we start focusing on optimization.

PR #1494, opened 2022-10-27
Constrain padding to resolved inner / outer widget sizes

tarkah:
During layout, padding isn't constrained to the difference between the child widget and parent widgets max limits. This causes the parent widget to have the wrong resolved size and for the child widget to get offset by `padding` instead of the constrained padding.

This fixes the issue by constraining the padding to the min between itself and `limits.max() - child.size()`. So far I've fixed Container, Button & Text Input as I've seen instances of this for all of them, but haven't combed the other widgets yet for padding.

Below is an example of the fix implemented on `Todos`, where both text input and buttons are bleeding over when their size should be 0. I've tested these fixes on some larger projects I maintain and couldn't find any regression in layout.

Feel free to rename things, not sure `constrain` is the best, I suck at naming :P

| Before | After |
| - | - |
| Screenshot from 2022-10-27 12:12:02 | Screenshot from 2022-10-27 12:11:35 |

### DIFF
--- a/core/src/padding.rs
+++ b/core/src/padding.rs
@@ -1,3 +1,5 @@
+use crate::Size;
+
 /// An amount of space to pad for each side of a box
 ///
 /// You can leverage the `From` trait to build [`Padding`] conveniently:
@@ -70,6 +72,18 @@ impl Padding {
     /// Returns the total amount of horizontal [`Padding`].
     pub fn horizontal(self) -> u16 {
         self.left + self.right
+    }
+
+    /// Constrains the padding to fit between the inner & outer [`Size`]
+    pub fn constrain(self, inner: Size, outer: Size) -> Self {
+        let available = (outer - inner).max(Size::ZERO);
+
+        Padding {
+            top: self.top.min((available.height / 2.0) as u16),
+            right: self.right.min((available.width / 2.0) as u16),
+            bottom: self.bottom.min((available.height / 2.0) as u16),
+            left: self.left.min((available.width / 2.0) as u16),
+        }
     }
 }
 
--- a/core/src/padding.rs
+++ b/core/src/padding.rs
@@ -87,7 +87,7 @@ impl Padding {
     }
 }
 
-impl std::convert::From<u16> for Padding {
+impl From<u16> for Padding {
     fn from(p: u16) -> Self {
         Padding {
             top: p,
@@ -98,7 +98,7 @@ impl std::convert::From<u16> for Padding {
     }
 }
 
-impl std::convert::From<[u16; 2]> for Padding {
+impl From<[u16; 2]> for Padding {
     fn from(p: [u16; 2]) -> Self {
         Padding {
             top: p[0],
@@ -109,7 +109,7 @@ impl std::convert::From<[u16; 2]> for Padding {
     }
 }
 
-impl std::convert::From<[u16; 4]> for Padding {
+impl From<[u16; 4]> for Padding {
     fn from(p: [u16; 4]) -> Self {
         Padding {
             top: p[0],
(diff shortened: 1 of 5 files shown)


core/src/padding.rs
```diff
@@ -71,6 +73,18 @@ impl Padding {
     pub fn horizontal(self) -> u16 {
         self.left + self.right
     }
+
+    /// Constrains the padding to fit between the inner & outer [`Size`]
+    pub fn constrain(self, inner: Size, outer: Size) -> Self {
```
Mingun, line 78: The algorithm of constraining in min-max range usually named `clamp`, so
```suggestion
    pub fn clamp(self, inner: Size, outer: Size) -> Self {
```

core/src/size.rs
```diff
@@ -34,6 +34,22 @@ impl Size {
             height: self.height + padding.vertical() as f32,
         }
     }
+
+    /// Returns the minimum of each component of this size and another
+    pub fn min(self, other: Self) -> Self {
```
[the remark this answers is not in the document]
Mingun: It could be worth to use [euclid](https://lib.rs/crates/euclid) for geometric primitives. It also allow to create different types for different coordinate systems in order to not mix them.

core/src/padding.rs
tarkah: I thought of `clamp`, but since this isn't passing it a min / max `Size` I was unsure if it made sense

tarkah: > constrain a value..

Welp, guess constrain works too :P

tarkah: `clamp` just seems like it should only take `clamp(Self, Self)` is my only concern. There's no concept for `Padding.min(Size)` / `Padding.max(Size)`, so `Padding.clamp(Size, Size)` feels awkward

core/src/size.rs
```diff
@@ -68,3 +84,14 @@ impl From<Size> for Vector<f32> {
         Vector::new(size.width, size.height)
     }
 }
+
+impl std::ops::Sub for Size {
```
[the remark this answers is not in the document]
hecrj: Not sure these make sense for `Size`. Which property would we use for comparison? Area? Where would that be useful?

core/src/padding.rs
hecrj: I renamed it to `fit` in 7476663069572adec25161b46c26570f864f736f.

core/src/padding.rs
```diff
@@ -71,9 +73,21 @@ impl Padding {
...
+
+    /// Fits the [`Padding`] between the provided `inner` and `outer` [`Size`].
+    pub fn fit(self, inner: Size, outer: Size) -> Self {
+        let available = (outer - inner).max(Size::ZERO);
+
+        Padding {
+            top: self.top.min((available.height as u16) / 2),
+            right: self.right.min((available.width as u16) / 2),
+            bottom: self.bottom.min((available.height as u16) / 2),
+            left: self.left.min((available.width as u16) / 2),
```
hecrj, line 85: I have changed these lines so we perform the cast to `u16` first and then divide by 2. Integer division by 2 is way faster, and I believe both approaches produce equivalent results in the end.

hecrj (approved):
Looks good! Thanks! :bow: 

I made some small changes here and there. Let me know if I broke anything, and we can merge!

tarkah:
> Not sure these make sense for Size. Which property would we use for comparison? Area? Where would that be useful?

I struggled with this as well, but it seems other mathcrates include these methods on size so 🤷‍♂️

Changes look great, `fit` is a fit naming choice!

Issue #457, opened 2020-07-22
Feasible to enable OpenGL as a fallback instead of an override?


mtkennerly:
If I understand correctly, the `glow` feature makes the OpenGL support totally override the default WebGPU support. Would it be feasible to have some way of trying both in a single build? As in, try WebGPU first, but if Vulkan/etc isn't available on the user's system, then try OpenGL as a fallback. Right now, I'm producing two separate builds (a main version with WebGPU and a secondary version with OpenGL in case people need it), but I was thinking it might be nice to have an all-inclusive build so users don't have to worry about which one to use.

manokara:
I think that would be great, but only if the `glow` feature is enabled, just like how `wgpu` is also optional. I guess Hector would want to finish the glow backend first before adding it a fallback though, so that the fallback is seamless.

hecrj:
Is there any particular reason you are interested in having `iced_wgpu` as the default?

If you do not need image or SVG support, why not ship `iced_glow` by default, as it should run on older hardware?

mtkennerly:
I have a few reasons, roughly in order of importance:

* On Mac, OpenGL is deprecated in favor of Metal.

* I prefer the default font for WebGPU more than the default font for OpenGL. Is there a way to make them the same?

  WebGPU:
  > ![image](https://user-images.githubusercontent.com/4934192/88420148-6de8c480-cdb4-11ea-96d8-427b305d3412.png)

  OpenGL:
  > ![image](https://user-images.githubusercontent.com/4934192/88420155-6fb28800-cdb4-11ea-973b-6cb31662b827.png)

* Philosophically, I like the idea of using/supporting newer technologies by default to keep the ecosystem moving forward.

hecrj:
> On Mac, OpenGL is deprecated in favor of Metal.

I know, "OpenGL is dead". This is the main reason `iced_wgpu` exists. However, this doesn't change the fact that OpenGL is the most supported graphics API. If I am not mistaken, Catalina still supports OpenGL 4.1.

> I prefer the default font for WebGPU more than the default font for OpenGL. Is there a way to make them the same?

You need to enable the `glow_default_system_font` feature separately.

> Philosophically, I like the idea of using/supporting newer technologies by default to keep the ecosystem moving forward.

Fair enough! However, keep in mind these newer technologies may not be able to satisfy all your needs right away.

--------

Just to be clear, I do believe we should eventually offer rendering fallbacks, specifically:

```
[...]

PR #1495, opened 2022-10-27
WIP Update to glutin 0.30

ids1024:
Glutin 0.30 drops the dependency on `winit`, allowing it to be used with winit or other things that can provide a raw window handle. (Perhaps this could change the relationship between `iced_winit` and `iced_glutin`?)

This needs some fixes to error handling, and a few other things.

### DIFF
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -18,9 +18,7 @@ system = ["iced_winit/system"]
 version = "0.4"
 
 [dependencies.glutin]
-version = "0.29"
-git = "https://github.com/iced-rs/glutin"
-rev = "da8d291486b4c9bec12487a46c119c4b1d386abf"
+version = "0.30"
 
 [dependencies.iced_native]
 version = "0.5"
@@ -34,3 +32,6 @@ path = "../winit"
 version = "0.3"
 path = "../graphics"
 features = ["opengl"]
+
+[dependencies.raw-window-handle]
+version = "0.5.0"
--- a/glutin/src/application.rs
+++ b/glutin/src/application.rs
@@ -12,10 +12,15 @@ use iced_winit::futures;
 use iced_winit::futures::channel::mpsc;
 use iced_winit::renderer;
 use iced_winit::user_interface;
+use iced_winit::winit;
 use iced_winit::{Clipboard, Command, Debug, Proxy, Settings};
 
-use glutin::window::Window;
+use glutin::prelude::*;
+use raw_window_handle::{HasRawDisplayHandle, HasRawWindowHandle};
+use winit::window::Window;
+
 use std::mem::ManuallyDrop;
+use std::num::NonZeroU32;
 
 /// Runs an [`Application`] with an executor, compositor, and the provided
 /// settings.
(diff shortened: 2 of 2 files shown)


ids1024:
So currently `iced_winit` requires a `Compositor`, which means a wgpu compositor? I wonder what API would be best to make `iced_winit` rendering backend agnostic, so it could be used with wgpu, glutin 0.30, or a software renderer.

ids1024:
I'm thinking the way this works could be improved by doing something like this:
* Unify `Compositor` and `GlCompositor` as a single trait, with an `Api` associated type
  - The types methods take and return could depend on this
* Make `iced_winit` generic over this API (using some trait it implements for APIs it supports)
* Remove `iced_glutin` and move context creation and a couple other small details using glutin 0.30 to `iced_winit` (behind a feature flag)

Advantages
* Window creation, input event handling, etc. is handled in one place, instead of being duplicated between `iced_winit` and `iced_glutin`
* It should be fairly easy to also add support for a software renderer with a new `Api`, and a bit of support code in `iced_winit` (in addition to implementing the software renderer backend)

I'm not sure how dynamic selection of a renderer (https://github.com/iced-rs/iced/issues/457) would fit in here, but otherwise this seems like a good way to improve things (assuming actually implementing it works out well).

ids1024:
> e.g. winit and glutin, possibly druid-shell

And `iced_sctk` for Wayland layer-shell surfaces. And supporting things like `xdg_popup` (handling those sorts of things with winit remains an issue).

> I also think we should unify Compositor, but I don't see the need of adding the Api associated type

Hm, I guess the traits' definitions aren't so different. I was thinking `.present()` has different type signatures, but that's actually for methods in the backends that aren't part of the trait. But anyway there needs to be logic somewhere to setup the OpenGL context (using Glutin) and do a couple other things an OpenGL renderer is in use; so that needs to be controlled by a type generic or something else.

Using Glutin here is non-trivial and is needs to call things from winit to deal with `XlibErrorHookRegistrar`, and make sure the context and window are created in the right order for the platform (https://github.com/rust-windowing/glutin/issues/1511)

> Additionally, I did this exact work (https://github.com/derezzedex/iced/commit/ef0097a89f59537bd5ea67091cd1290d9559d171), of switching to glutin 0.30 on my branch of multi_window, maybe we could unify this.

Yep. I guess I need to look at what changes multi-window support entails. Though if the glutin update is somewhat independent of the multi-window changes and could be merged earlier it would be good to not have one thing blocked on the other.

ids1024:
@wash2 should have some ideas about what changes would be needed to accommodate things like iced-sctk as a "shell" without having to patch other parts of Iced.

wash2:
iced_sctk probably can't exist as a shell without adding some platform specific actions and events as well as the application and sandbox traits. It could probably share the same application & sandbox trait as multi-window, though at the moment there is a separate view function for each windows, layer surfaces, and popups.

While working on iced-sctk, I've pretty much combined iced-glutin & iced-winit & an sctk 0.30 replacement for winit to roughly make iced-sctk. I think all of my changes to iced itself are behind feature flags and work alongside iced_winit with its single + multi-window as they are. iced_sctk currently couldn't be switched to at runtime; however if we are ok with it sharing a trait with multi-window & including all of the platform specific additions that are behind a feature right now, i think it probably could use the same or a similar approach to multi_window.

kchibisov:
You might want to try https://crates.io/crates/glutin-winit for winit integration. If you find lucking some features from it please open an issue in glutin tracker.

ids1024:
`glutin-winit` would probably help here, but since Iced also has a `wgpu` renderer and may want to add a software renderer, it may be better to unify the uses of `winit` for glow/wgpu and just use glutin directly. Though I'm not sure exactly what that would look like.

ids1024:
> This should be fine, the codebase already has platform specific events and we won't be able to offer cross-platform "native" support without these.

If the "shell" was somehow represented with a trait, it could have an associated type for shell-specific events. Then an out-of-tree shell implementation could add events without forking Iced, shared crates aren't polluted with events that are specific to iced-sctk/etc., and it could be checked at compile time if those events are available.

This is similar to the idea I had recently of a renderer-specific primitive, to be able to have a primitive render directly with OpenGL/wgpu, or use a texture imported from a dmabuf.

ids1024:
Superseded by https://github.com/iced-rs/iced/pull/1688.

PR #1496, opened 2022-10-28
Fix drop down not closing when inside scrollable

tarkah:
Fix drop down not closing when inside scrollable and user clicks outside the scrollable. This is because the scrollable sets -1.0 on cursor.y for any events where cursor is outside it's bounds. I'm not sure why picklist had this logic to stay open on -1.0 / -1.0, any click outside the overlay should close it.


hecrj (approved):
Makes sense! I believe this logic is a leftover from way before event capturing was a thing in the library.

PR #1497, opened 2022-10-28
Use `once_cell` instead of `lazy_static`

ids1024:
This seems to be generally considered the preferred, idiomatic solution now. This is in the standard library behind a feature flag (apparently now called `std::sync::LazyLock`).


hecrj (approved):
Awesome! Always great to see some macro magic being dispelled!

PR #1500, opened 2022-11-01
[WIP] Horizontal scrollable implemetation

bungoboingo:
Still need to do some edge-case testing & api refinement, but thought I'd post it to get feedback on my initial approach.

I originally tried including a `direction` enum in the existing `Scrollable` struct directly but that was an issue for two reasons:
1) The API design is bad/unclear due to having stuff like `fn height()` available as an option for a scrollable whose direction is `horizontal` when it was ignored internally. This is what CSS does and I find it incredibly annoying.
2) Properties like `scrollbar_width` don't make sense in a horizontal scrollable so I messed around with renaming it `length` or other names before deciding it was more clear just to separate it out.

I also tried just duping most of the code over between the two `Horizontal` & `Vertical` structs & swapping the w/h x/y bits as necessary, but the amount of duplicated code was pretty immense. I have a significant amount of match cases in `scrollable` mod right now that could be reduced by just duping the code between `Horizontal` / `Vertical` modules. On the fence about this.

I ALSO tried doing some funky stuff with just modifying the bounds before use in a horizontal scrollable but that was superrrr confusing to read.

**Note:** No scrolling keybinds are currently supported other than just the good ol' grab & drag. Using mousewheel seemed counter intuitive when scrolling x-y, also would cause conflicts with nested scrollables.

For now the new (breaking) implementation is this:

```rust
scrollable::vertical(content) //or
scrollable::horizontal(content)

//can also call the variants directly e.g.
scrollable::Vertical::new(content)
scrollable::Horizontal::new(content)
```

I've modified the `scrollable` example to showcase the new options.

Again, just want to reiterate -> **I have not tested this very thoroughly at ALL**, just want to get some feedback on the API! Thanks!

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -99,7 +99,7 @@ A bunch of simpler examples exist:
 - [`pick_list`](pick_list), a dropdown list of selectable options.
 - [`pokedex`](pokedex), an application that displays a random Pokédex entry (sprite included!) by using the [PokéAPI].
 - [`progress_bar`](progress_bar), a simple progress bar that can be filled by using a slider.
-- [`scrollable`](scrollable), a showcase of the various scrollbar width options.
+- [`scrollable`](scrollable), a showcase of various scrollable content configurations.
 - [`sierpinski_triangle`](sierpinski_triangle), a [sierpiński triangle](https://en.wikipedia.org/wiki/Sierpi%C5%84ski_triangle) Emulator, use `Canvas` and `Slider`.
 - [`solar_system`](solar_system), an animated solar system drawn using the `Canvas` widget and showcasing how to compose different transforms.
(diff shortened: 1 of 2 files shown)


ids1024:
In GTK a `ScrolledWindow` handles both horizontal and vertical scrolling. So with this implementation, if you need both (which is typical if you want to allow a window to be shrunk small enough to need both) you'd compose one within the other?

That isn't an inherently bad (theoretically it's quite elegant), but I wonder if the single widget handling both might handle at least certain things better.

ids1024:
My thought is that having both generally doesn't provide a good experience, but if you want your application to not be broken when the window is too small, you may want it show a horizontal scrollbar as a last resort at least. I think that behavior is pretty common.

PR #1504, opened 2022-11-03
Add pane maximize / restore for `PaneGrid`

tarkah:
Adds the ability to maximize / restore a pane in the pane grid. When maximized, dragging is disabled (we can remove this, since dragging is enabled when only 1 pane exists currently).

### API

Two new methods are exposed on `State` to allow the user control of maximizing and restoring a pane:

```rs
/// Maximize the given [`Pane`]. Only this pane will be rendered by the
/// [`PaneGrid`] until [`Self::restore()`] is called.
pub fn maximize(&mut self, pane: &Pane);

/// Restore the currently maximized [`Pane`] to it's normal size. All panes
/// will be rendered by the [`PaneGrid`]
pub fn restore(&mut self);
```

Once a pane is maximized, a new `bool` is passed on the `view` function to specify if the pane is maximized or not. Maybe we create an enum here for a stronger type since the closure is not descriptive:

```rs
PaneGrid::new(&self.panes, |id, pane, is_maximized| {
    ..
})
```

### Implementation 

`State` stores the maximized pane in a new field: `maximized: Option<Pane>` and is set by the above API.

Instead of storing `Vec<Content>` in the `PaneGrid`, I've created a new helper `Contents` that is built considering the maximized pane. This can be iterated over for building the widget tree & layout.

```rs
pub enum Contents<T> {
    All(Vec<(Pane, T)>),
    Maximized(Pane, T),
}
```

A new `state::Scoped` has been added which either wraps the internal state when no field is maximized, or specifies the maximized pane. The widget methods now take this for determining the layout Node of the `PaneGrid`.

```rs
/// The scoped internal state of the [`PaneGrid`]
pub enum Scoped<'a> {
    /// The state when all panes are visible
    All(&'a Internal),
    /// The state when a pane is maximized
    Maximized(Node),
}

impl<'a> Scoped<'a> {
    /// The layout [`Node`] of the [`Scope`] state
    pub fn layout(&self) -> &Node {
        match self {
            Scoped::Normal(Internal { layout, .. }) => layout,
            Scoped::Maximized(layout) => layout,
        }
    }
}
```

Iterating on `Contents` and calling `Scoped::layout` allowed for very minimal adjustments of the existing widget implementation code.

### Demo

https://user-images.githubusercontent.com/10239377/199626412-7507c106-185f-4d64-8e93-b55943abae49.mp4

### DIFF
--- a/native/src/widget/pane_grid.rs
+++ b/native/src/widget/pane_grid.rs
@@ -221,6 +221,12 @@ where
         self.style = style.into();
         self
     }
+
+    fn drag_enabled(&self) -> bool {
+        (!self.elements.is_maximized())
+            .then(|| self.on_drag.is_some())
+            .unwrap_or_default()
+    }
 }
 
 impl<'a, Message, Renderer> Widget<Message, Renderer>
@@ -296,6 +302,11 @@ where
     ) -> event::Status {
         let action = tree.state.downcast_mut::<state::Action>();
 
+        let on_drag = self
+            .drag_enabled()
+            .then_some(&self.on_drag)
+            .unwrap_or(&None);
+
         let event_status = update(
             action,
             self.elements.node(self.state),
@@ -306,7 +317,7 @@ where
             self.spacing,
             self.elements.iter(),
             &self.on_click,
-            &self.on_drag,
+            on_drag,
             &self.on_resize,
         );
 
@@ -361,7 +372,7 @@ where
                         cursor_position,
                         viewport,
                         renderer,
-                        self.on_drag.is_some(),
+                        self.drag_enabled(),
                     )
                 })
                 .max()
@@ -960,5 +971,10 @@ impl<T> Elements<T> {
             Elements::Normal(_) => state.layout(),
             Elements::Maximized(_, _, node) => node,
         }
+    }
+
+    /// TODO
+    pub fn is_maximized(&self) -> bool {
+        matches!(self, Self::Maximized(..))
     }
 }
--- a/native/src/widget/pane_grid.rs
+++ b/native/src/widget/pane_grid.rs
@@ -100,8 +100,8 @@ where
     Renderer: crate::Renderer,
     Renderer::Theme: StyleSheet + container::StyleSheet,
 {
-    state: &'a state::Internal,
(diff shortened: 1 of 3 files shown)


casperstorm:
This is a great addition to the pane grid.
Code change is easy to follow and the example worked as expected.

native/src/widget/pane_grid.rs
```diff
@@ -897,3 +935,41 @@ fn hovered_split<'a>(
...
+    fn iter_mut(&mut self) -> Box<dyn Iterator<Item = (Pane, &mut T)> + '_> {
+        match self {
+            Contents::All(contents) => Box::new(
+                contents.iter_mut().map(|(pane, content)| (*pane, content)),
+            ),
+            Contents::Maximized(pane, content) => {
+                Box::new(std::iter::once((*pane, content)))
+            }
+        }
+    }
```
hecrj, line 970: Great! We avoid a lot of code changes with this API :) Iterators are pretty nifty.

native/src/widget/pane_grid.rs
```diff
@@ -119,22 +119,37 @@ where
...
-                .iter()
-                .map(|(pane, pane_state)| (*pane, view(*pane, pane_state)))
-                .collect()
+        let (contents, state) = if let Some((pane, pane_state)) =
+            state.maximized.and_then(|pane| {
+                state.panes.get(&pane).map(|pane_state| (pane, pane_state))
+            }) {
+            (
+                Contents::Maximized(pane, view(pane, pane_state, true)),
+                state::Scoped::Maximized(Node::Pane(pane)),
```
hecrj, line 133: These seem inherently coupled, but the current data model allows us to potentially have desync. Could we simply merge `Scoped` with `Contents`?

hecrj (changes requested):
Looking good! Just one suggestion!

native/src/widget/pane_grid.rs
tarkah: :+1: Good call, fixed in 7de9d24

hecrj (approved):
Awesome! Let's sail :ship:

PR #1505, opened 2022-11-03
Combine `glow_default_system_font` and `default_system_font` features

ids1024:
Apparently "weak dependency features" have been stable since Rust 1.60.0, allowing a feature to enable a feature of a dependency only if that dependency is enabled elsewhere. So having a separate feature for this with glow is unnecessary now.

Also remove the `resolver` setting which is redundant on edition 2021.


hecrj (approved):
Feels good when your tools keep getting better consistently. Thanks!

PR #1506, opened 2022-11-03
non uniform border radius for quads

rksm:
This allows quads to be rendered with different border radius values. It only modifies quads, not yet any widgets based on that. I tried to open a discussion about the implications of changing the API of widgets such as button over [here](https://github.com/iced-rs/iced/discussions/1486) but didn't receive any feedback.

non-uniform-quad-example

A project showing the use of that can be found over here: https://github.com/rksm/iced-non-uniform-border-radius.


rksm:
This feature is quite useful to for example render more sophisticated buttons. It's similar to the [CSS border-radius property](https://www.w3schools.com/CSSREF/css3_pr_border-radius.php).

image

Happy to address potential concerns.

hecrj (approved):
This is awesome! :partying_face: Thank you.

I have added your `custom_quad`  as an official example as well.

rksm:
Thanks for adding this! Stoked to see how much development is happening around iced!

PR #1507, opened 2022-11-05
Don't convert svg to BGRA before passing to shader

ids1024:
Now that the shader is using RGBA, this incorrectly swaps the components.


hecrj (approved):
Whoops! Good catch.

PR #1518, opened 2022-11-08
Fixed issues with old GL versions ( <= 2.1 )

bungoboingo:
Fixes an issue with older GL versions <= 2.1 with new shader changes.


ids1024:
What is the best way to test things with an old OpenGL version? I know when I've tried in the past it's been problematic since the driver just gave a context for a newer (compatible) version.

If possible it would be good to have some automated CI testing of this. It's something that breaks easily if changes aren't being tested with the older version.

hecrj (approved):
Looks great! Thanks for the quick fix :partying_face:

PR #1521, opened 2022-11-10
Update iced version to 0.5

mq1:


hecrj (approved):
Thanks! Classic oversight!

PR #1524, opened 2022-11-10
Additional cfg for wasm target for gradients

bungoboingo:
Storage buffers aren't supported by wgpu WASM + webgl which gradients currently use. For now I've just added a `#cfg` around relevant bits of code so WASM targets that don't use gradients will compile as expected.

Not quite sure if there is a preferred way to separate out this code, but I figured writing a separate implementation for WASM which uses a static uniform array like the GL target would be a relatively simple process so it didn't make sense to reorganize the code to reduce the number of target-specific compilation exclusions.

### DIFF
--- a/glow/src/triangle.rs
+++ b/glow/src/triangle.rs
@@ -141,6 +141,7 @@ impl Pipeline {
                     triangle::Style::Solid(color) => {
                         self.programs.solid.use_program(gl, color, &transform);
                     }
+                    #[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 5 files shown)


tarkah:
Since our canvas API is already limited due to the opengl implementation, should we just use fixed # stops / uniform data on wgpu? I think a single implementation that is compatible with all targets is preferable to a bunch of code split. I really don't think it's an issue since hitting the max 16 color stops is such a far out edge case.

hecrj (approved):
Looks good! Thanks!

PR #1525, opened 2022-11-11
add more window actions

Night-Hunter-NF:
users should be able to with window::Actions:
- toggle decorations live 
- toggle always on top i.e. for pinning a window
- change icon live to show changes in app state etc  

There may be a better way to handle the type for the Icon enum but you cant import winit into iced_native due to cyclic package dependency's. feel free to say a better way then i have done.

Also what do people think the Action::Fullscreen should do? currently it full screens on primary monitor but IMO i think it should full screen on current monitor. 

And should it be an option to full screen exclusive. currently you can only full screen windowed borderless

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -321,7 +321,8 @@ Many thanks to...
 ### Added
 - First release! :tada:
 
-[Unreleased]: https://github.com/iced-rs/iced/compare/0.5.0...HEAD
+[Unreleased]: https://github.com/iced-rs/iced/compare/0.6.0...HEAD
+[0.6.0]: https://github.com/iced-rs/iced/compare/0.5.0...0.6.0
 [0.5.0]: https://github.com/iced-rs/iced/compare/0.4.2...0.5.0
 [0.4.2]: https://github.com/iced-rs/iced/compare/0.4.1...0.4.2
 [0.4.1]: https://github.com/iced-rs/iced/compare/0.4.0...0.4.1


native/src/window/action.rs
```diff
@@ -35,16 +35,42 @@ pub enum Action<T> {
...
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Web / Wayland / macOS:** Unsupported.
+    ///
+    /// - **Windows:** Sets `ICON_SMALL`. The base size for a window icon is 16x16, but it's
+    ///   recommended to account for screen scaling and pick a multiple of that, i.e. 32x32.
+    ///
+    /// - **X11:** Has no universal guidelines for icon sizes, so you're at the whims of the WM. That
+    ///   said, it's usually in the same ballpark as on Windows.
+    SetWindowIcon(I),
```
tarkah, line 63: This generic is polluting everything. We should have an `Icon` type that can go here and be converted to `winit::Icon`.

There is already an `Icon` in `iced`, maybe it can live in `iced_native` instead and hold the raw pixels `Vec<u8>` / width / height so it can be converted to the winit Icon outside this crate (iced_native doesn't / shouldn't depend on winit)

tarkah:
```rs
// iced_native::window::icon

struct Icon {
    data: Vec<u8>,
    width: u32,
    height: u32,
}

// iced_winit::window

pub struct Icon(winit::window::Icon);

impl From<iced_native::window::Icon> for Icon {
    ..
}

impl From<Icon> for winit::window::Icon {
    ..
}

// iced_winit::settings

pub struct Window {
    icon: Option<crate::window::Icon>,
    ..
}

// iced::window
pub use iced_native::icon;
pub use iced_native::icon::Icon;
```

Night-Hunter-NF: ive done that still need to add some more functions but the main biggest issue is error checking.
```rs
impl From<iced_native::window::Icon> for Icon {
    fn from(value: iced_native::window::Icon) -> Self {
        let (rgba, width, height) = value.get_icon();
        Icon(winit::window::Icon::from_rgba(rgba, width, height).unwrap())
    }
}
```
that has no checks an winit icon has quite a few. so i could either write those checks onto iced_native::icon or figure out another way of doing it

Night-Hunter-NF: i will do some more work on it tm

Night-Hunter-NF:
@tarkah I'm trying to add 
https://github.com/iced-rs/iced/blob/b4aedceac6f3937eabec90fc84d5eda61a000575/src/window/icon.rs#L28-L33
https://github.com/iced-rs/iced/blob/b4aedceac6f3937eabec90fc84d5eda61a000575/src/window/icon.rs#L40-L56
to the my iced native icon https://github.com/Night-Hunter-NF/iced/blob/master/native/src/window/icon.rs
but it doesn't have image as a dep. 
just wondering what you thoughts are on this

hecrj (changes requested):
I feel we are tackling multiple use cases here and some of the actions seem quite straightforward while some others may need further discussion.

Since there is not an RFC for these changes, I feel we should at least split this PR into smaller ones; one for each new action.

Night-Hunter-NF:
ok yup will work on splitting them up into different PRs

Night-Hunter-NF:
split into prs above

PR #1526, opened 2022-11-11
Add widget operation to find currently focused widget

mtkennerly:
This also adds conversions from widget-specific IDs to the generic ID type, so that you can use the ID returned by this operation to match against specific `TextInput`s, for example.

More context: https://discord.com/channels/628993209984614400/1040413102606393434


hecrj (approved):
Sweet! Thanks! :tada:

Issue #1527, opened 2022-11-11
If you expand and collapse a PickList, then expand a different one, the first one briefly re-expands


mtkennerly:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

* If you expand PickList 1, do not collapse it, and then expand PickList 2, it transitions smoothly.
* If you expand PickList 1, collapse it, and then expand PickList 2, the issue occurs. PickList 1 will re-expand briefly before PickList 2 expands.

> https://user-images.githubusercontent.com/4934192/201330897-85157cbd-cda1-4e2c-b30c-3e642454e150.mp4

To reproduce this, run the `pick_list` example with these changes (relative to commit 23299a555f8b7e908a6a14915307792a7cf97b9a):

```diff
diff --git a/examples/pick_list/src/main.rs b/examples/pick_list/src/main.rs
index 9df1f5c7..f0fa42dd 100644
--- a/examples/pick_list/src/main.rs
+++ b/examples/pick_list/src/main.rs
@@ -35,17 +35,25 @@ impl Sandbox for Example {
     }
 
     fn view(&self) -> Element<Message> {
-        let pick_list = pick_list(
+        let pick_list1 = pick_list(
             &Language::ALL[..],
             self.selected_language,
             Message::LanguageSelected,
         )
         .placeholder("Choose a language...");
 
+        let pick_list2 = pick_list(
+            &Language::ALL[..],
+            self.selected_language,
+            Message::LanguageSelected,
+        )
+        .placeholder("Choose another language...");
+
         let content = column![
             vertical_space(Length::Units(600)),
             "Which is your favorite language?",
-            pick_list,
+            pick_list1,
+            pick_list2,
             vertical_space(Length::Units(600)),
         ]
         .width(Length::Fill)
```

### What is the expected behavior?

After collapsing PickList 1, it should not re-expand before PickList 2 expands.

### Version

master

### Operative System

Windows

mtkennerly:
This is similar/related to #1484, but that one is about mismatched options (PickList 2's options appear in PickList 1's overlay just before PickList 1 collapses), whereas this ticket is about PickList 1's overlay reappearing.

PR #1528, opened 2022-11-11
#1484, #1527: Fix overlay transition

### LINKED ISSUE #1484: Switching between two pick lists is not seamless

### What happened?

Switching between two pick lists causes the text in the dropdown of the pick list to change a half second or so before the dropdown is moved to the new pick list. Due to another bug (#1312 ?), OBS crashes when I try to record a video of the bug.

Here is a (poorly written) example:
```rust
use std::borrow::Cow;

use iced::{Sandbox, Element};
use iced::widget::{pick_list, row};

fn main() -> iced::Result {
    App::run(iced::Settings::default())?;
    Ok(())
}

struct App {
    list_one: Vec<&'static str>,
    list_one_selected: Option<&'static str>,
    list_two: Vec<&'static str>,
    list_two_selected: Option<&'static str>
}
#[derive(Debug, Copy, Clone)]
enum Message {
    ListOne(&'static str),
    ListTwo(&'static str)
}

impl Sandbox for App {
    type Message = Message;
    fn new() -> Self {
        Self { list_one: vec!("one", "two", "three"), list_one_selected: Option::None, list_two: vec!("alpha", "bravo", "charlie"), list_two_selected: Option::None}
    }
    fn title(&self) -> String {
        String::from("Bugcheck")
    }
    fn update(&mut self, _message: Message) {
        return
    }
    fn view(&self) -> Element<Self::Message> {
        let list_one =  pick_list(Cow::from(self.list_one.clone()), self.list_one_selected.clone(), Message::ListOne);
        let list_two = pick_list(Cow::from(self.list_two.clone()), self.list_two_selected.clone(), Message::ListTwo);

        row![list_one, list_two].into()
    }
}
```

Switching between pick lists should be seamless.

### Version

master

### Operative System

Linux

_No response_

mtkennerly:
This fixes #1484 and #1527, at least on my system (Windows 11).

⚠️ I don't really understand what I did here! ⚠️

I just figured it seemed like the old overlay was being held onto, and so I tried to find where it might need to be flushed. This works, but I haven't studied the other `user_interface` code to see how this aligns with the rest of the design, so I don't know if it's a good/correct solution.

I tested the `pick_list` and `tour` examples with this change. Here's a modified `pick_list` to show the new behavior:

> https://user-images.githubusercontent.com/4934192/201336666-23823cb0-add6-49c7-8eb6-9a682d30e0fb.mp4

### DIFF
--- a/native/src/user_interface.rs
+++ b/native/src/user_interface.rs
@@ -255,7 +255,7 @@ where
                 cursor_position
             };
 
-            self.overlay = None;
+            self.overlay = Some(layout);
 
             (base_cursor, event_statuses)
         } else {
@@ -284,6 +284,10 @@ where
                     clipboard,
                     &mut shell,
                 );
+
+                if matches!(event_status, event::Status::Captured) {
+                    self.overlay = None;
+                }
 
                 shell.revalidate_layout(|| {
                     self.base = renderer.layout(
--- a/native/src/user_interface.rs
+++ b/native/src/user_interface.rs
@@ -255,7 +255,7 @@ where
                 cursor_position
             };
 
-            self.overlay = Some(layout);
+            self.overlay = None;
 
             (base_cursor, event_statuses)
         } else {


hecrj (approved):
Thanks for taking a look!

This bug has been alive for a while. Basically, the issue is that the active overlay may change after an `Event` in the base layer is processed. Reusing the layout of the old overlay is incorrect in that case!

Your inital fix consisted in always invalidating the cache of the overlay layout. I changed it in bbd55ff3a9338775cf87f236265369d9eddae4bd so we only invalidate when an `Event` in the base layer is `Captured`, which I believe is reasonable.

Hopefully this still works!

mtkennerly:
> Hopefully this still works!

Looks good on my end :D Thanks!

PR #1529, opened 2022-11-11
Add text input operations

tarkah:
We're missing operations for manipulating the text input cursor / selection of text.

TODOs example is updated to `select_all` text when an entry is edited, which is how the impure example used to work.

### DIFF
--- a/native/src/widget/text_input.rs
+++ b/native/src/widget/text_input.rs
@@ -233,6 +233,7 @@ where
         let state = tree.state.downcast_mut::<State>();
 
         operation.focusable(state, self.id.as_ref().map(|id| &id.0));
+        operation.text_input(state, self.id.as_ref().map(|id| &id.0));
     }
 
     fn on_event(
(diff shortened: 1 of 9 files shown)


native/src/widget/text_input.rs
```diff
@@ -337,6 +338,32 @@ pub fn focus<Message: 'static>(id: Id) -> Command<Message> {
...
+}
+
+/// Produces a [`Command`] that moves the cursor of the [`TextInput`] with the given [`Id`] to the
+/// front.
+pub fn move_cursor_to_front<Message: 'static>(id: Id) -> Command<Message> {
+    Command::widget(operation::text_input::move_cursor_to_front(id.0))
+}
+
+/// Produces a [`Command`] that moves the cursor of the [`TextInput`] with the given [`Id`] to the
+/// provided position.
```
Mingun, line 354: What's happens if `position` would be outside range of text? Can be this be prevented by using custom types? That should be documented

tarkah: If it's greater than the number of characters in the text input, it just puts the cursor at the end, which is expected.

hecrj (approved):
Looks preem!

PR #1533, opened 2022-11-13
Fixed Operations within PaneGrid

### LINKED ISSUE #1532: Operations Broken in PaneGrid

### What happened?

When the target of an [Operation](https://github.com/iced-rs/iced/pull/1399) lies anywhere within a `PaneGrid`, the operation is silently ignored.

You can reproduce the issue by running this [modified example](https://github.com/iced-rs/iced/commit/4c4105fdfbbb0f98fb160fdc76d68da4395d4797).
I've added a 'snapTo 0.0' whenever you click a Pane, so you can:
1. You can resize the window such that the scrollbar appears
2. Scroll that bar down a bit.
3. Then, clicking anywhere in any pane **should** cause all scrollbars to snap to the top (but doesn't without the fix))

_Note: I removed the `responsive` part of the example for this reproduction, because it also breaks Operations. I'll open a separate issue for that._

I think the fix is pretty straightforward (`PaneGrid` is simply missing its `operate` method), so I'll put out a PR for this shortly.

Operations should work within PaneGrids

### Version

master

### Operative System

Windows

_No response_

rs017991:
Fixes #1532

### DIFF
--- a/native/src/widget/pane_grid.rs
+++ b/native/src/widget/pane_grid.rs
@@ -38,6 +38,7 @@ use crate::mouse;
 use crate::overlay;
 use crate::renderer;
 use crate::touch;
+use crate::widget;
 use crate::widget::container;
 use crate::widget::tree::{self, Tree};
 use crate::{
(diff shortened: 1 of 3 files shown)


rs017991:
Oops, I missed a file when committing. Coming right up...

hecrj:
No worries! Just fixed implemented the missing stuff myself :)

rs017991:
Ah, you beat me to it. Sorry about that

hecrj (approved):
No worries! I had the feeling I missed some stuff here. Thanks for cathing it!

PR #1536, opened 2022-11-13
Fix mapped `Operation` always returning `Outcome::None`

### LINKED ISSUE #1535: Operations Broken when mapped

### What happened?

When you call `.map` on an [Operation](https://github.com/iced-rs/iced/pull/1399), the operation is silently ignored.

You can reproduce the issue by running [these modified examples](https://github.com/iced-rs/iced/compare/master...rs017991:iced:op_bug_reprod_map).
* In the case of the `scrollable` example: the 'Scroll to top/bottom' buttons no longer function
* In the case of the `todo` example: The various focusing behaviours no longer function

I've tracked this down to missing function implementations in [MapRef](https://github.com/iced-rs/iced/blob/23299a555f8b7e908a6a14915307792a7cf97b9a/native/src/widget/action.rs#L58).
I have a partial fix, and will open a PR shortly.

Operations should work when mapped to different message types

### Version

master

### Operative System

Windows

_No response_

rs017991:
Fixes #1535.

This fixes the `scrollable` issue and the basic `focus` issue from the reproduction in the linked issue.

However, `focus_previous` and `focus_next` remain broken.
I suspect that this is due to missing `finish` implementations in both [Map](https://github.com/iced-rs/iced/blob/23299a555f8b7e908a6a14915307792a7cf97b9a/native/src/widget/action.rs#L43) and [MapRef](https://github.com/iced-rs/iced/blob/23299a555f8b7e908a6a14915307792a7cf97b9a/native/src/widget/action.rs#L58).
After staring at it for a while, I was unable to divine what that method is even supposed to do in this situation, so I'll leave that to someone more knowledgeable.

### DIFF
--- a/native/src/widget/action.rs
+++ b/native/src/widget/action.rs
@@ -3,6 +3,8 @@ use crate::widget::Id;
 
 use iced_futures::MaybeSend;
 
+use std::rc::Rc;
+
 /// An operation to be performed on the widget tree.
 #[allow(missing_debug_implementations)]
 pub struct Action<T>(Box<dyn Operation<T>>);
@@ -24,7 +26,7 @@ impl<T> Action<T> {
     {
         Action(Box::new(Map {
             operation: self.0,
-            f: Box::new(f),
+            f: Rc::new(f),
         }))
     }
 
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Good catch! I have implemented `finish` for `action::widget::Map`. `MapRef` doesn't necessarily need to implement it, since calling `finish` on an `Operation` that you do not own breaks its contract.

PR #1538, opened 2022-11-16
Group all solid triangles independently of color

hecrj:
This PR enhances our `triangle` pipelines to batch all solid geometry independently of color.

As a result, all solid geometry drawn sequentially in a `Canvas` will be batched together into a single draw call independently of color. The previous behavior produced as many draw calls as different colors used.

We achieve this by introducing a new `ColoredVertex2D` type and separating `Mesh` into `SolidMesh` and `GradientMesh` variants in `Primitive`.

No user-facing APIs have changed.


artursapek:
did we mention it's fast

hecrj:
> Only nit I have re: nesting solid & gradient mods in the triangle.rs. I found it easier to read when they were separated out into their own files. Realize this is subjective so don't feel particularly strongly about it.

I know it's pretty subjective, but I felt like the boundaries of the modules were a bit blurry (e.g. only some fields were public, methods needed to be called in a particular order and with correct indices, etc.). I usually prefer no boundaries at all than blurry ones.

We could still move the modules to their own files but, since everything is pretty coupled, I feel the pipelines are easier to edit this way.

PR #1542, opened 2022-11-16
feat(winit): client-side window drag resize support

mmstick:
The patch was accepted into winit early this year. Iced has also already pulled a winit release that contains it. So it's just a matter of utilizing it here.

~~Requires https://github.com/iced-rs/winit/pull/8~~
Implements https://github.com/iced-rs/rfcs/pull/14


n1ght-hunter:
is there a reason this isnt merged yet?

PR #1546, opened 2022-11-18
Fix overlay widget operations

tarkah:
Widget operations are broken in a couple ways for Overlay.

- `Overlay::operate` must be mutable so that state can be operated on
- A new user interface is built before processing operations. `Overlay::layout` is only cached during `UserInterface::draw`. This ensures a new overlay layout is cached before processing the operation.

I've also added a new example `modal` which showcases building with the `Overlay` API and has working widget operations occurring while the overlay is active.

https://user-images.githubusercontent.com/10239377/202817143-768add3d-8546-4483-a37e-10bc688ff576.mp4


examples/modal/src/main.rs
```diff
@@ -0,0 +1,481 @@
...
+            self.base.as_widget().operate(
+                &mut state.children[0],
+                layout,
+                operation,
+            );
+        }
+    }
+
+    struct Overlay<'a, 'b, Message, Renderer> {
+        content: &'b RefCell<Element<'a, Message, Renderer>>,
```
tarkah, line 344: @hecrj This `RefCell` is ultimately unnecessary, but required since overlay is restricted to `&self`. If we change `Widget::overlay` to take `&mut self` this wouldn't be necessary. I've got a patch making this change and everything compiles (including changes in `lazy` crate w/ ouroboros use).

If a `Widget` needs mutability during `on_event`, an overlay holding reference to a widget should also be allowed mutability during `Overlay::on_event`. 

Is there a reason for this restriction?

hecrj: Makes sense. I think we can go ahead with the proposed change!

tarkah: Great! I'll submit a separate PR and rebase this to target that.

tarkah: Resolved by #1563. This targets that branch and should merge after (I can't change the base branch for some reason)

hecrj (approved):
Looks good! Great example to showcase custom overlays as well :)

Thank you :partying_face:

GyulyVGC:
I find the Modal example really interesting and I would like to do a similar thing in my application.
I'm asking to be sure: as of today it doesn't exist a modal widget, is it right?
If I'd like to implement a modal do I have to follow your example and duplicate the code you wrote from [this point](https://github.com/iced-rs/iced/blob/a6d0d5773f0561a841a84b538523cbd97e91eccd/examples/modal/src/main.rs#L180-478) on? @tarkah

tarkah:
> I find the Modal example really interesting and I would like to do a similar thing in my application.
I'm asking to be sure: as of today it doesn't exist a modal widget, is it right?
If I'd like to implement a modal do I have to follow your example and duplicate the code you wrote from [this point](https://github.com/iced-rs/iced/blob/a6d0d5773f0561a841a84b538523cbd97e91eccd/examples/modal/src/main.rs#L180-478) on? @tarkah

Correct, it's not included as a built-in in iced, so definitely feel free to use the example as inspiration! 

Also, feel free to checkout my latest PR here https://github.com/iced-rs/iced/pull/1655. It makes overlays more powerful by allowing for multiple to be returned from the view tree. It also adds a new `toast` example which leverages overlays and also new widget animations. Be warned it's not merged yet.

PR #1550, opened 2022-11-19
[Feature] Multidirectional scrolling

bungoboingo:
This PR adds support for scrolling horizontally in Iced's `Scrollable' widget.

**This PR does not create any breaking changes with the current Scrollable API, except for scrollable::snap_to() operation.** All other api additions for horizontal scrolling are opt-in!

https://user-images.githubusercontent.com/4241774/202867014-b91f0f59-e662-4a35-94d9-b89e0f629e97.mov

**Tasks:**
- [x] Update `Scrollable` to include horizontal scrolling capabilities
- [x] Update scrollable demo to reflect new changes
- [x] Introduce styling support for horizontal scrollbar independently of vertical scrollbar

**Known issues:**
- There is currently an issue where if you use the `snap_to()` operation on a Scrollable and don't trigger any action which requests a redraw, the operation will not be reflected until the next redraw. I'm pretty sure this is just inherent to operations and not something that occurred due to my changes, but I'm not 100% certain.

**Implementation Details:**
- I've moved around several items in the internal state of a `Scrollable` widget in order to prevent instances where the offset of the Scrollable could be adjusted independently of the actual status of the scrollbar, leading to potential out of sync states. I moved `Scrollbar` into `scrollable::State` as `Option<Scrollbar>` for both the x & y scrollbars, and moved some methods that once existed in `scrollable::State` into `Scrollbar` & `Scroller`. This does increase the memsize of the internal Scrollable state very slightly. Let me know what you think of these changes! Definitely open to feedback!! 
- Currently styling support for horizontal scrollbars will default to using whatever style is in place for vertical scrollbars, with optional customization.
- I've updated the `snap_to()` operation to take a `Vector<f32>` vs `f32` to reflect ability to scroll on both axis.

**Behavior Details:**
- Horizontal scrollbars will always be rendered so that if a vertical scrollbar is present, it will adjust its width to always be shunted to the left so that no overlap will occur. You can see this in action in the video I posted above. This made the most sense to me and seems to be how literally every UI widget framework does it.
- Some UI frameworks (e.g. browsers rendering overflow-y & overflow-x) render a weird little "dead space" box in between horizontal & vertical scrollbars when they are present, which I omitted here. I just am rendering nothing there. Let me know if you think something should be put there, or if this option should be configurable!

Very much open to feedback! Let me know what you think, thanks!


jaunkst:
Looking forward to this feature landing. I've been working off this branch for my project and touch events are working as expected.

examples/scrollable/Cargo.toml
```diff
@@ -7,3 +7,4 @@ publish = false
 
 [dependencies]
 iced = { path = "../..", features = ["debug"] }
+lazy_static = "1.4"
```
hecrj, line 10: We are using `once_cell` instead of `lazy_static` now. See #1497.

examples/scrollable/src/main.rs
```diff
@@ -1,44 +1,60 @@
...
 }
 
 struct ScrollableDemo {
-    theme: Theme,
-    variants: Vec<Variant>,
+    scrollable_direction: Direction,
+    scrollbar_width: u16,
+    scrollbar_margin: u16,
+    scroller_width: u16,
+    current_scroll_offset: Vector<f32>,
```
hecrj, line 23: I'm not sure using a `Vector` for storing the offsets is appropriate. I see offsets as specific points in the scrollable, rather than a length with a direction. In other words, they represent a position rather than a "movement".

Maybe we should use `Point` instead.

examples/scrollable/src/main.rs
```diff
@@ -1,44 +1,60 @@
...
-    ScrollToTop(usize),
-    ScrollToBottom(usize),
-    Scrolled(usize, f32),
+    SwitchDirection(Direction),
+    ScrollbarWidthChanged(u16),
+    ScrollbarMarginChanged(u16),
+    ScrollerWidthChanged(u16),
+    ScrollToBeginning(scrollable::Direction),
+    ScrollToEnd(scrollable::Direction),
+    Scrolled(Vector<f32>),
```
hecrj, line 41: Same here. We can use `Point` for everything I think.

examples/scrollable/src/main.rs
```diff
@@ -50,209 +66,333 @@ impl Application for ScrollableDemo {
...
+            scroller: Scroller {
+                color: Color::from_rgb8(250, 85, 134),
+                border_radius: 0.0,
+                border_width: 0.0,
+                border_color: Default::default(),
             },
-        ]
+        }
     }
+}
```
hecrj, line 384: Oh! That's a cool pattern.

examples/scrollable/src/main.rs
```diff
@@ -50,209 +66,333 @@ impl Application for ScrollableDemo {
...
+                .scroller_width(self.scroller_width)
+                .horizontal_scroll(
+                    scrollable::Horizontal::new()
+                        .scrollbar_height(self.scrollbar_width)
+                        .scrollbar_margin(self.scrollbar_margin)
+                        .scroller_height(self.scroller_width),
+                )
+                .style(theme::Scrollable::Custom(Box::new(
+                    ScrollbarCustomStyle,
+                )))
```
hecrj, line 306: May be a good idea to add a `custom` helper to `theme::Scrollable` that does the boxing internally so the resulting `view` code is cleaner.

native/src/widget/scrollable.rs
```diff
@@ -37,8 +35,9 @@ where
     scrollbar_width: u16,
     scrollbar_margin: u16,
     scroller_width: u16,
+    scroll_horizontal: Option<Horizontal>,
```
hecrj, line 38: It seems `Horizontal` includes the same fields of the vertical scrollbar as well. I propose renaming `Horizontal` to `Scrollbar` and use `width` instead of `height`:

```rust
/// Properties of a scrollbar within a [`Scrollable`].
#[derive(Debug)]
pub struct Scrollbar {
    width: u16,
    margin: u16,
    scroller_width: u16,
}
```

And then here:

```rust
vertical_scrollbar: Scrollbar,
horizontal_scrollbar: Option<Scrollbar>,
```

We can rename the current private `Scrollbar` type to something else.

native/src/widget/scrollable.rs
```diff
@@ -153,13 +191,60 @@ where
...
+        tree::State::new(State::new())
+    }
+
+    fn children(&self) -> Vec<Tree> {
+        vec![Tree::new(&self.content)]
+    }
+
+    fn diff(&self, tree: &mut Tree) {
+        tree.diff_children(std::slice::from_ref(&self.content))
+    }
```
hecrj, line 246: Let's keep the ordering of these, for consistency!

Maybe we should just reorder the declarations in the `Widget` trait.

native/src/widget/scrollable.rs
```diff
@@ -342,7 +394,10 @@ impl From<Id> for widget::Id {
 
 /// Produces a [`Command`] that snaps the [`Scrollable`] with the given [`Id`]
 /// to the provided `percentage`.
-pub fn snap_to<Message: 'static>(id: Id, percentage: f32) -> Command<Message> {
+pub fn snap_to<Message: 'static>(
+    id: Id,
+    percentage: Vector<f32>,
```
hecrj, line 399: As I said, I think `Point` makes more sense.

native/src/widget/scrollable.rs
```diff
@@ -390,36 +461,39 @@ pub fn update<Message>(
...
-    let scrollbar = scrollbar(
-        state,
+    state.create_scrollbars_maybe(
+        horizontal,
         scrollbar_width,
         scrollbar_margin,
         scroller_width,
         bounds,
         content_bounds,
     );
```
hecrj, line 476: Let's avoid methods with hidden side-effects like this one!

Instead, we can create a new `Scrollbars` type and return it:

```rust
let scrollbars = scrollbars(...);
```

native/src/widget/scrollable.rs
```diff
@@ -435,18 +509,18 @@ pub fn update<Message>(
...
-                    mouse::ScrollDelta::Lines { y, .. } => {
-                        // TODO: Configurable speed (?)
-                        state.scroll(y * 60.0, bounds, content_bounds);
-                    }
-                    mouse::ScrollDelta::Pixels { y, .. } => {
-                        state.scroll(y, bounds, content_bounds);
+                let delta = match delta {
+                    mouse::ScrollDelta::Lines { x, y } => {
+                        // TODO: Configurable speed/friction (?)
+                        Vector::new(x * 60.0, y * 60.0)
```
hecrj, line 518: Yes! `Vector` makes sense here since this represents a `delta` (i.e. a movement to be applied).

native/src/widget/scrollable.rs
```diff
@@ -700,34 +891,8 @@ pub fn draw<Renderer>(
...
-                            bounds: scrollbar.scroller.bounds,
-                            border_radius: style.scroller.border_radius,
-                            border_width: style.scroller.border_width,
-                            border_color: style.scroller.border_color,
-                        },
-                        style.scroller.color,
-                    );
-                }
+                draw_scrollbar(renderer, state.scrollbar_y.as_ref());
+                draw_scrollbar(renderer, state.scrollbar_x.as_ref());
```
hecrj, line 895: I don't think we need this function helper, since we are calling it only twice with completely different directions. We can just inline each one of the branches here.

native/src/widget/scrollable.rs
```diff
@@ -970,10 +1311,56 @@ impl Scrollbar {
...
+                Direction::Horizontal => {
+                    ((content_bounds.width - bounds.width) * percentage)
+                        .max(0.0)
+                }
+                Direction::Vertical => {
+                    ((content_bounds.height - bounds.height) * percentage)
+                        .max(0.0)
+                }
+            },
+        }
```
hecrj, line 1363: I think we can just split this into `absolute_x` and `absolute_y` and just call the method accordingly, since we never have a dynamic `Direction`.

native/src/widget/scrollable.rs
```diff
@@ -736,227 +901,403 @@ pub fn draw<Renderer>(
...
-        ));
-    }
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// The direction of the [`Scrollable`].
+pub enum Direction {
+    /// X or horizontal
+    Horizontal,
+    /// Y or vertical
+    Vertical,
 }
```
hecrj, line 955: I don't think this type is necessary. We always statically know the `Direction` of the scrollbar we are dealing with in the code.

hecrj (changes requested):
Looking great! Nice work :partying_face: This was long overdue, thanks!

I left some feedback here and there that I think should help simplfiy some of the code.

native/src/widget/scrollable.rs
hecrj: I think length is more confusing. In my mind, the length of a scrollbar is its longitude, while the width of a scrollbar is the cross-axis length.

Furthermore, we are already using width similarly in other instances like border width, stroke width, etc.

lfrancke:
Thanks a lot for this. I did not look at the code but I did test it in my application and it seems to work fine. I also tried touch control (scrolling horizontally by swiping) and that worked (at least on Windows).

lfrancke:
Happy to try, anything specific you have in mind?

Just running tests or something more elaborate?

lfrancke:
I tried it today on a Windows laptop (Surface) using touch-screen and it worked beautifully.

There is one thing that's a bit weird and I'll try to describe, maybe that's just a feeling though:

- When I touch the screen/content (not scrollbar) and move my finger to the right, everything moves left (including the scrollbar) -> This is just as I'd expect
- When I touch the scrollbar itself, the same thing happens. What I expect is different though. I want to "drag" the scrollbar thing to the left but it actually moves to the right.

This is different than it works in other UIs. There it works as expected: Drag the content and it moves the opposite direction, drag the scrollbar itself and it moves along with your finger.

I hope that makes sense?
I also tried to cross-compile to a Raspberry but the compilation failed (unrelated to this), I'll try again later.
This looks very good though!

hecrj (approved):
Awesome stuff @bungoboingo! Looks great :partying_face: 

I made some smaller changes here and there to bring it to the finish line. Let me know what you think, and I believe we can merge!

@lfrancke I believe I have fixed this behavior you mention in 2d007474dd002dfc936f75f7379f1bbae72c5ad4.

hecrj:
@bungoboingo Life changer! You are welcome :D

PR #1551, opened 2022-11-20
image: Allow any kind of data that implements `AsRef<[u8]>` for the i…

sdroege:
…mage data

It's not required anywhere for it to be a plain slice or a `Vec` and this makes it possible to use data allocated in a different way without copying.

Also store the data in an `Arc` to allow cheap cloning.

### DIFF
--- a/examples/pokedex/src/main.rs
+++ b/examples/pokedex/src/main.rs
@@ -193,7 +193,7 @@ impl Pokemon {
         {
             let bytes = reqwest::get(&url).await?.bytes().await?;
 
-            Ok(image::Handle::from_memory(bytes.as_ref().to_vec()))
+            Ok(image::Handle::from_memory(bytes))
         }
 
         #[cfg(target_arch = "wasm32")]
--- a/native/src/image.rs
+++ b/native/src/image.rs
@@ -9,7 +9,7 @@ use std::sync::Arc;
 #[derive(Debug, Clone)]
 pub struct Handle {
     id: u64,
-    data: Arc<Data>,
+    data: Data,
 }
 
 impl Handle {
@@ -33,7 +33,7 @@ impl Handle {
         Self::from_data(Data::Rgba {
             width,
             height,
-            pixels: ImageBytes::new(pixels),
+            pixels: Bytes::new(pixels),
         })
     }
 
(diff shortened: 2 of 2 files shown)


13r0ck:
Image handle already uses an arc
https://github.com/iced-rs/iced/blob/370fa14efb8d41bcf81e4a0ead6a9ab7ab750bee/native/src/image.rs#L11
https://github.com/iced-rs/iced/blob/master/examples/pokedex/src/main.rs#L113

sdroege:
Ah I see, thanks! Then this can also become a plain `Box<dyn _>` instead.

Is the `Clone` impl on `Data` then actually needed? If it's needed this needs a bit more code (as `Clone` is not object-safe) but also not much of a problem.

native/src/image.rs
```diff
@@ -84,14 +85,70 @@ impl Hash for Handle {
     }
 }
 
+/// Reference counted wrapper around raw image data.
+///
+/// Behaves like a `&[u8]`.
+pub struct ImageBytes(Box<dyn ImageBytesTrait>);
+
+trait ImageBytesTrait: AsRef<[u8]> + Send + Sync + 'static {
+    fn clone_boxed(&self) -> Box<dyn ImageBytesTrait>;
```
sdroege, line 94: The `Clone` trait bound everywhere and this new trait wouldn't be needed if it would be OK to not have `Data` implement `Clone`.

With `Handle` being `Clone`, I don't see much use for it but I might be missing something. What do you think?

examples/pokedex/src/main.rs
```diff
@@ -195,7 +195,7 @@ impl Pokemon {
         {
             let bytes = reqwest::get(&url).await?.bytes().await?;
 
-            Ok(image::Handle::from_memory(bytes.as_ref().to_vec()))
+            Ok(image::Handle::from_memory(bytes))
```
sdroege, line 198: And here an example how this can be used to avoid one useless copy

hecrj (approved):
Thanks for this!

I have moved the `Arc` from `Handle` to `Bytes` since the other fields in `Data` should be fairly cheap to `Clone` and we can get rid of the awkward trait this way.

native/src/image.rs
```diff
@@ -85,55 +85,38 @@ impl Hash for Handle {
...
-        Self(Box::new(data))
-    }
-}
-
-impl Clone for ImageBytes {
-    fn clone(&self) -> Self {
-        ImageBytes(self.0.clone_boxed())
+impl Bytes {
+    /// Creates new [`Bytes`] around `data`.
+    pub fn new(data: impl AsRef<[u8]> + Clone + Send + Sync + 'static) -> Self {
```
sdroege, line 96: You don't really need the `Clone` here anymore as the `Arc` already gives you that :)
Do you want me to submit a PR for removing the trait bound or will you take care of the change?

hecrj: Good catch. Just opened #1717.

sdroege: Thanks!

PR #1557, opened 2022-11-23
fix: scissor layer bounds for images when using glow

wash2:
This fixes an issue where images show up outside of the bounds of their parent, for example when in a scrollable.


wash2:
I think actually this fix needs a bit more testing and adjustment, so i've made it a draft.

hecrj (approved):
Good catch! Thanks :bow:

PR #1558, opened 2022-11-24
fix: Don't require text::Stylesheet::Style to be Copy

truelossless:
Hello,

In 18fb74f20092b2703a90afdb01f39754445998da the trait bounds for most widgets' `Stylesheet::Style` were relaxed from `Default + Copy` to `Default` or `Default + Clone`. However, `Text` was not changed, by mistake I assume. As a result, trying to implement a custom style for `Text` yields an error as `iced::Theme` is not `Copy`.

```rust
pub struct MenuTextStyle;

impl text::StyleSheet for MenuTextStyle {
    type Style = Theme;
    // the trait bound `iced::Theme: std::marker::Copy` is not satisfied
    // the trait `std::marker::Copy` is not implemented for `iced::Theme`

    fn appearance(&self, style: Self::Style) -> text::Appearance {
        let palette = style.palette();

        text::Appearance {
            color: Some(palette.background),
        }
    }
}
```

This PR fixes this issue and changes the signature of the `appearance` function to `fn appearance(&self, style: &Self::Style) -> text::Appearance`, to be consistent with the other widgets.

### DIFF
--- a/native/src/widget/text.rs
+++ b/native/src/widget/text.rs
@@ -163,7 +163,7 @@ where
             &self.content,
             self.size,
             self.font.clone(),
-            theme.appearance(self.style),
+            theme.appearance(&self.style),
             self.horizontal_alignment,
             self.vertical_alignment,
         );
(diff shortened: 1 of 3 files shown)


truelossless:
Oops, turned out this is not needed since `theme::Text` doesn't have a `Custom` enum variant as the other widgets do. While this still brings the API closer to the other widgets this is not needed. I'm closing this PR, sorry for the noise.

PR #1560, opened 2022-11-25
docs: Remove invalid links.

l1Dan:
Remove invalid links in examples/README.md file, because the `geometry` example  has been removed 30432cbade3d9b25c4df62656a7494db3f4ea82a .


hecrj (approved):
We will most likely re-introduce this example in #1538, but we can merge this anyways :) Thanks!

PR #1563, opened 2022-11-29
Allow &mut self in overlay

tarkah:
As discussed in #1546


hecrj (approved):
Looks good! Thanks :bow:

PR #1568, opened 2022-12-01
[DRAFT] Additional focus implementations for essential widgets

jaunkst:
This PR adds more focus controls and keyboard controls to offer a near complete input workflow without a mouse.

It was mentioned that we should possibly consider a global focus state. This is already apparent in my example. Keyboard navigation focus vs mouse focus. We do not know to release an existing focus triggered my keyboard navigation when the mouse triggers focus on another component.

This could be resolved here if focusable widget set their is_focused state to false on a mouse click event when they are not in the cursors bounds.

RFC:
https://github.com/jaunkst/iced-rfcs/blob/keyboard_navigation/text/1568-keyboard-navigation.md


jaunkst:
I am going to try implementing an global focus_id, and define implicit focus styles instead of reusing hover.

PR #1569, opened 2022-12-01
#1397: Fix fill + max

### LINKED ISSUE #1397: Problems with Container when combining `width(Length::Fill)` and `max_width`

### Master
On master (c44267b85f7aaa2997e3caf1323b837d95818c22), `max_width` is respected, but surrounding containers using `width(Length::Fill)` no longer fill the whole parent. In the video, the container labelled `end` does not go all the way to the right when the middle container has a max width.

Gist: https://gist.github.com/mtkennerly/a61f945a652d2d7cd54181fc16f37538

> https://user-images.githubusercontent.com/4934192/182501466-72872566-7382-46f4-8065-10e54a31facd.mp4

### 0.4.2
On 0.4.2, `max_width` is ignored in favor of `width` when both are set. The video shows `width(Length::Fill)`, but it also happens with `Length::Units`.

Gist: https://gist.github.com/mtkennerly/7d9518abde5e1060427ad8fa503fdaf2

> https://user-images.githubusercontent.com/4934192/182501842-8c02c59c-964c-4ca2-95b8-698cee794dad.mp4

* Surrounding containers with `width(Length::Fill)` should fill the whole parent.
* `max_width` should override `width`

### Version

master

### Operative System

Windows

_No response_

mtkennerly:
This adds a pass so that, for example, when a container is both `.width(Length::Fill)` and `.max_width(200)`, if it ends up being capped by its max, then its sibling containers can fill the unused space.

This fixes #1397.

### DIFF
--- a/native/src/layout/flex.rs
+++ b/native/src/layout/flex.rs
@@ -81,6 +81,9 @@ where
     let mut nodes: Vec<Node> = Vec::with_capacity(items.len());
     nodes.resize(items.len(), Node::default());
 
+    let mut capped: Vec<bool> = Vec::with_capacity(items.len());
+    capped.resize(items.len(), false);
+
     if align_items == Alignment::Fill {
         let mut fill_cross = axis.cross(limits.min());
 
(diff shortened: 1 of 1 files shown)


native/src/layout/flex.rs
```diff
@@ -147,6 +150,54 @@ where
...
+            };
+            if unused > 0.0 {
+                capped[i] = true;
+                fill_sum -= fill_factor;
+                available -= axis.main(size);
+                nodes[i] = layout;
+            }
+        }
+    }
+
```
hecrj, line 200: I really really woud like to avoid introducing additional cost to this function. It's a very critical path for our layout performance.

Instead of introducing additional logic, I'd like to explore other options first. Like maybe changing `Widget::width` and `Widget::height` to return further information, enhancing our `Limits` type, etc.

The most important shortcoming of the current layout logic here is that a `Fill` inside a `Shrink` will cause the `Shrink` to fill all the remaining space in the parent, independently of anything. A container with a `Fill` child should somehow be `Fill` as well.

I believe we should focus on rethinking our layout engine to deal with this basic shortcoming first, as doing so may shine a light on other inconsistencies (like this one).

mtkennerly: > Instead of introducing additional logic, I'd like to explore other options first. Like maybe changing `Widget::width` and `Widget::height` to return further information, enhancing our `Limits` type, etc.

Please correct me if I'm mistaken, but wouldn't any `Widget` changes still be dependent on receiving the available space from `resolve`? Like in this specific example, the widget width/height either needs to act like `Length::Units(max)` (if available space > max) or `Length::Fill` (if available space <= max), and the outcome would affect the available space for any remaining `Fill` widgets. I'm not quite sure how that would be possible without changing this function (or some equivalent of it in an updated layout engine).

mtkennerly: Do you think it would make sense to have `Length` variants like this? (just `FillMax` would solve #1397, but the others are for the sake of example)

```
pub enum Length {
    // Either a few very specific combinations:
    FillMin(u16),
    FillMax(u16),
    FillWithin {
        min: u16,
        max: u16,
    },

    // Or something more generic that may allow unwanted combinations:
    FillAll {
        portion: Option<u16>,
        min: Option<u16>,
        max: Option<u16>,
    }
}
```

That way, `resolve` would know if any of the widgets have a max set, and it could only trigger the extra code/cost if there are widgets that need it.

If it doesn't belong in `Length` directly, then maybe have `Widget::width` return a `Width` like this?

```rust
struct Width {
    length: Length,
    max: Option<u16>,
}
```

hecrj:
As I mentioned in #1722, I think we should explore the limits of a single pass layout engine a bit more before we decide to introduce more complexity.

On a related note, I am working on fixing some layout inconsistencies in the [`fix/layout-inconsistencies`](https://github.com/iced-rs/iced/tree/fix/layout-inconsistencies) branch that will hopefully make reasoning about layout a bit easier.

Let's close this for now, and revisit it in the future!

PR #1570, opened 2022-12-02
Fix custom style of PickList

PolyMeilex:


hecrj (approved):
Good catch! Thanks :)

PR #1582, opened 2022-12-08
full screen example

dansgithubuser:
Couldn't find anything on this and ended up asking in Discord. Contributing back an example.


hecrj (commented):
Thank you!

For discoverability, I believe we shouldn't have an example to showcase every small feature.

Maybe you could try adding fullscreen functionality to an existing example? `game_of_life` sounds like a good candidate. We could toggle it on `Alt + Enter`.

PR #1585, opened 2022-12-09
add action to make window focus

Night-Hunter-NF:
add action to make window focus

### DIFF
--- a/native/src/window/action.rs
+++ b/native/src/window/action.rs
@@ -68,7 +68,7 @@ pub enum Action<T> {
     /// ## Platform-specific
     ///
     /// - **Web / Wayland:** Unsupported.
-    FocusWindow,
+    GainFocus,
 }
 
 impl<T> Action<T> {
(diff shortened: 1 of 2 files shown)


native/src/window/action.rs
```diff
@@ -35,6 +35,17 @@ pub enum Action<T> {
...
+    /// Brings the window to the front and sets input focus. Has no effect if the window is
+    /// already in focus, minimized, or not visible.
+    ///
+    /// This method steals input focus from other applications. Do not use this method unless
+    /// you are certain that's what the user wants. Focus stealing can cause an extremely disruptive
+    /// user experience.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **iOS / Android / Web / Wayland:** Unsupported.
```
hecrj, line 47: No need to mention `iOS / Android` since we do not officially support mobile platforms!

hecrj (changes requested):

hecrj (approved):

PR #1586, opened 2022-12-09
add action to set window grab mode

Night-Hunter-NF:
add action would allows users to lock the cursor to be inside window i.e. for games


hecrj:
Not sure if this is something we want to expose as an action or as a `mouse::Interaction` in the `Widget` API.

PR #1589, opened 2022-12-09
add action to get window id

Night-Hunter-NF:
add action that returns the windows id. not accessible any other way currently.

### DIFF
--- a/native/src/window/action.rs
+++ b/native/src/window/action.rs
@@ -76,6 +76,8 @@ pub enum Action<T> {
     ///
     /// - **Web / Wayland:** Unsupported.
     ChangeAlwaysOnTop(bool),
+    /// Fetch an identifier unique to the window.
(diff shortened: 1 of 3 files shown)


native/src/window/action.rs
```diff
@@ -37,6 +37,8 @@ pub enum Action<T> {
     ToggleMaximize,
     /// Fetch the current [`Mode`] of the window.
     FetchMode(Box<dyn FnOnce(Mode) -> T + 'static>),
+    /// Fetch the window id from winit
```
hecrj, line 40: We can't mention `winit` in `iced_native`, since it's a downstream dependency!

I would just say: `Fetch the id of the window`, and maybe explain the meaning of the `u64` a bit.

hecrj (changes requested):

Eraden:
Hi, I could really use this. Is there a chance for merge? :pray:

hecrj (approved):

PR #1590, opened 2022-12-09
Add `ChangeIcon` to `window::Action`

Night-Hunter-NF:
add action to set the icon live to show updates in state etc.


native/src/window/icon.rs
```diff
@@ -0,0 +1,73 @@
...
+    rgba: Vec<u8>,
+    width: u32,
+    height: u32,
+}
+
+impl Icon {
+    /// return the icon data
+    pub fn get_icon(self) -> (Vec<u8>, u32, u32) {
+        (self.rgba, self.width, self.height)
+    }
```
hecrj, line 27: I would rewrite this as `into_raw(self) -> (Vec<u8>, Size<u32>)`.

hecrj (changes requested):

hecrj (changes requested):

hecrj (approved):
Awesome! I've tidied some stuff up (mainly, renaming the action and using free functions instead of an extension trait).

I think we can merge this! Thanks :bow:

PR #1594, opened 2022-12-12
`MouseArea` widget

mmstick:
Implements https://github.com/iced-rs/rfcs/pull/12

Run example with `cargo run --release -p headerbar`

Not a breaking change

### DIFF
--- a/native/src/widget/mouse_listener.rs
+++ b/native/src/widget/mouse_listener.rs
@@ -112,6 +112,14 @@ where
     Renderer: crate::Renderer,
     Message: Clone,
 {
+    fn tag(&self) -> tree::Tag {
+        tree::Tag::of::<State>()
+    }
+
+    fn state(&self) -> tree::State {
+        tree::State::new(State::default())
+    }
+
     fn children(&self) -> Vec<Tree> {
         vec![Tree::new(&self.content)]
     }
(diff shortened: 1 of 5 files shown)


ErnWong:
Hi! Just a heads up that with the current implementation it looks like `on_mouse_enter` and `on_mouse_exit` doesn't react to layout changes if the mouse doesn't move.

https://user-images.githubusercontent.com/2609018/211186248-a7cd5e60-e12b-4c54-88e9-b39a48bc13b1.mov

(Thank you @13r0ck for pointing out this scenario over on Discord for my [mouse_area](https://github.com/ErnWong/iced-prototype-mouse-area/tree/f78a41c3ce4480026ae49e7aa2728b88f055e85b) prototype, which also suffers from this problem).

native/src/widget/mouse_listener.rs
```diff
@@ -0,0 +1,403 @@
...
+    if let Some(message) = widget.on_release.clone() {
+        if let Event::Mouse(mouse::Event::ButtonReleased(mouse::Button::Left))
+        | Event::Touch(touch::Event::FingerLifted { .. }) = event
+        {
+            shell.publish(message);
+            return event::Status::Captured;
+        }
+    }
+
+    if let Some(message) = widget.on_right_press.clone() {
```
IohannRabeson, line 340: Hi! You could clone the message only if it published.

hecrj (approved):
Looks good! Thanks and sorry for the wait! :bow: 

I had to unfortunately remove the `on_mouse_enter` and `on_mouse_exit` events since, as others pointed out in the comments. in order to support it we would need some kind of continuity guarantees in the persistent widget tree that we currently lack. Certainly something we should revisit in the future!

In any case, I think we can merge this simplified version :ship:

PR #1596, opened 2022-12-12
Vertical orientation added to Slider.

casperstorm:
This PR introduces the ability to change orientation of the `slider`.
The default orientation remains `horizontal` but can be changed like so:

```rust
slider(0..=100, value, Message::SliderChanged).orientation(Vertical)
```

I've added an example, which can be run like so: `cargo run --package slider`

sliders

Breaking changes:

- `Height` is now a `Length` rather than `u16`. This is because when `Orientation` is `Vertical` we want `Height` to behave like `Width` does for `Horizontal` mode. And vice versa.


hecrj (commented):
Cool! Thank you.

I wonder if splitting the new logic completely into a new widget (e.g. `VerticalSlider`) would be a better approach, since the API guarantees around layout seem to be different between both orientations.

Duplication is not that much of a concern to me :)

casperstorm:
> I wonder if splitting the new logic completely into a new widget (e.g. `VerticalSlider`) would be a better approach, since the API guarantees around layout seem to be different between both orientations.

Sure!

hecrj (approved):
Awesome! Thanks!

PR #1597, opened 2022-12-13
[Feature] Gradients for Backgrounds

bungoboingo:
### Gradients are now usable as a `Background` variant
When creating a background with a gradient, the syntax is slightly different than when creating it on a `Canvas`.

Example of building a gradient for background:
```rust
container::Appearance {
    background: Gradient::linear(Radians(PI/3.0)) //or Degrees(45.0)!
        .add_stop(0.0, Color::from_rgb8(255, 0, 0))
        .add_stop(0.5, Color::from_rgb8(0, 255, 0))
        .add_stop(1.0, Color::from_rgb8(0, 0, 255))
        .build()
        .into(),
    ..Default::default()
}
```

Users can declare a `Gradient` with a specified `Angle` (rads or degrees), and add up to =8 `ColorStop`s along that angle. The bounds of the gradient will be calculated from the bounds of the quad that it is filling.

### Changes to `Canvas` gradient syntax
Users can no longer specify a `Position::Absolute` or a `Position::Relative`; now users can only specify an absolute `start` & `end` position when using a gradient as a `Fill` on a `Canvas`. This change was made to simplify the gradient API. Calculating bounds is possible with a mesh but would increase the cost of our tessellation phase which is not ideal.

New syntax for creating a `Gradient` fill on a `Canvas` (very similar to old one):

```rust
    Point::ORIGIN,
    Point::new(
        bounds.x + bounds.size().width,
        bounds.y + bounds.size().height,
    ),
)
.add_stop(0.0, Color::from_rgb8(255, 0, 0))
.add_stop(0.5, Color::from_rgb8(0, 255, 0))
.add_stop(1.0, Color::from_rgb8(0, 0, 255))
.build();
    
frame.fill_rectangle(Point::ORIGIN, bounds.size(), gradient);
```

### Some API notes
Due to this differing requirement for both `Canvas` and `Background` gradient types, I've had to redeclare `Gradients` separately in `core` & `graphics`. I messed about with keeping one builder for both, but it led to too many unexpected behaviors. Duping the `Gradient` declaration between the two now has no unintended consequences of, for example, declaring a `Gradient` with an angle as a fill for a mesh, which is undefined behavior. Let me know what you think!

### Changes to Gradient rendering pipeline

Gradients are now stored as a vertex type, `GradientVertex2D` which takes a position & flattened gradient data. This change allows us to stay fast & reduce draw calls in quad instancing/triangle drawing. This will also make gradients more accessible for different platforms (eg for WASM target), as it removes the requirements of storage buffer support.

Old gradients bench (rendering 5151 static gradients):
Screen Shot 2022-12-12 at 6 21 23 PM

New gradients bench (same 5151 static gradients)!:
Screen Shot 2022-12-12 at 6 22 48 PM

GPU is now much more happy!

This does have the tradeoff of forcing a lower maximum number of color stops, which is now 8 compared to.. like 186,000. If anyone has any issues with this please let me know in the comments below!

### Addition of several new `benches` crates
~~I've added four new static benchmarks to `examples/benches`. Not sure this is the best place for them, I could see them perhaps being added to the `profiling` crate in my profiling PR, or to a separate repo entirely. Let me know what you think!~~
I've moved these to https://github.com/bungoboingo/iced_benches for the time being until we can figure out how we want to integrate the benches into our CI.

### Support added
WASM should now be able to compile gradient shaders & use gradients as I've removed the need of storage buffers. Let me know if you see any issues!

### Testing
**Tested:**
M1 mac (WGPU + metal, tiny-skia)
Linux (WGPU + Vulkan, tiny-skia)
Windows 10 (WGPU + DirectX, tiny-skia)

:rainbow: :rainbow: :rainbow: 
Welcoming all feedback on the API & other changes!  :tada:


glow/src/quad/core.rs
```diff
@@ -110,135 +133,427 @@ impl Pipeline {
...
+                        (11, "i_position_and_size"),
+                        (12, "i_border_color"),
+                        (13, "i_border_radius"),
+                        (14, "i_border_width"),
+                    ],
+                )
+            };
+
+            let (vertex_array, instances) =
+                unsafe { Self::create_instance_buffer(gl, MAX_INSTANCES) };
```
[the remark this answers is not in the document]
hecrj: We can fine-tune this stuff once we focus on optimization, but maybe we can reduce the `MAX_INSTANCES` to `10_000`.

glow/src/shader/quad/core/gradient.frag
```diff
@@ -0,0 +1,69 @@
...
+precision mediump float;
+#endif
+#endif
+
+#ifdef HIGHER_THAN_300
+out vec4 fragColor;
+#define gl_FragColor fragColor
+#endif
+
+// #includes
```
hecrj, line 14: Since we are already hinting this here... It'd be nice to list the actual includes directly here instead of hardcoding it in Rust.

wgpu/src/quad.rs
```diff
@@ -167,119 +89,128 @@ impl Pipeline {
...
+            0,
+            &[Uniforms::new(transformation, scale)],
+        );
+
+        // resize buffers if necessary
+        let _ = self.solid.instances.resize(device, instances.solids.len());
+        let _ = self
+            .gradient
+            .instances
+            .resize(device, instances.gradients.len());
```
hecrj, line 112: So we are resizing the buffers now instead of looping and uploading multiple times.

I think that should be fine, and it fits the architecture of the incoming overhaul. Maybe we can reduce the `MAX_INSTANCES` to `10_000` here as well.

hecrj: We should probably rename `MAX_INSTANCES` to `INITIAL_INSTANCES` or similar.

wgpu/src/shader/triangle/solid.wgsl
```diff
@@ -25,6 +25,6 @@ fn vs_main(input: VertexInput) -> VertexOutput {
 }
 
 @fragment
-fn fs_main(input: VertexOutput) -> @location(0) vec4<f32> {
+fn fs_main(input: SolidVertexOutput) -> @location(0) vec4<f32> {
     return input.color;
 }
```
hecrj, line 30: Should we unify `triangle/gradient.wgsl` and `triangle/solid.wgsl` into `triangle.wgsl` like `quad.wgsl`?

hecrj (changes requested):
Looking awesome!

No specific sjhowstoppers for me :tada: Just a bunch of suggestions here and there. Let me know what you think!

greatest-ape:
I would love to see this get merged, as it also seems to fix https://github.com/iced-rs/iced/issues/1625!

graphics/src/gradient.rs
```diff
@@ -75,20 +70,27 @@ pub mod linear {
...
+                        if index < 8 {
+                            self.stops[index] =
+                                Some(ColorStop { offset, color });
+                        }
                     }
                     Err(index) => {
                         if index < 8 {
-                            self.stops[index] = ColorStop { offset, color };
+                            self.stops[index] =
+                                Some(ColorStop { offset, color });
```
clarkmoody: Branches are identical?

PR #1610, opened 2022-12-20
Vectorial text for `Canvas`

hecrj:
This PR implements vectorial text support for the `Canvas` widget by leveraging [`cosmic-text`](https://github.com/pop-os/cosmic-text) :tada:

This means that text now is no different than any other kind of geometry in a `Canvas` and, as a result, it can be transformed (translated, rotated, and scaled) and composed (layering!) freely.

Also, `cosmic-text` implements both text shaping and font fallback, so a `Canvas` should now be able to support different scripts as well.

https://user-images.githubusercontent.com/518289/208642436-759ad908-e4ac-4333-b55a-933bf3ad5036.mp4

The bitmap font support is wonky (and very inefficient), as you can see in the example above. This is just the start of a series of PRs that will improve our text handling and rendering. Further improvements are coming!

### DIFF
--- a/graphics/src/widget/canvas/frame.rs
+++ b/graphics/src/widget/canvas/frame.rs
@@ -351,7 +351,7 @@ impl Frame {
         for run in layout.iter() {
             for glyph in run.glyphs.iter() {
                 let start_x = translation_x + glyph.x + glyph.x_offset;
-                let start_y = translation_y as f32 + glyph.y_offset - text.size;
+                let start_y = translation_y + glyph.y_offset - text.size;
 
                 let offset = Vector::new(start_x, start_y);
 
--- a/graphics/src/widget/canvas/frame.rs
+++ b/graphics/src/widget/canvas/frame.rs
@@ -5,7 +5,7 @@ use crate::triangle;
 use crate::widget::canvas::{path, Fill, Geometry, Path, Stroke, Style, Text};
 use crate::Primitive;
 
-use iced_native::{Font, Point, Rectangle, Size, Vector};
+use iced_native::{Color, Font, Point, Rectangle, Size, Vector};
 
 use lyon::geom::euclid;
 use lyon::tessellation;
@@ -355,10 +355,10 @@ impl Frame {
 
                 let offset = Vector::new(start_x, start_y);
 
-                if let Some(commands) = cache
-                    .swash
-                    .borrow_mut()
-                    .get_outline_commands(glyph.cache_key)
+                let mut swash_cache = cache.swash.borrow_mut();
+
+                if let Some(commands) =
+                    swash_cache.get_outline_commands(glyph.cache_key)
                 {
                     let glyph = Path::new(|path| {
                         use cosmic_text::Command;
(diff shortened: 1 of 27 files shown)


graphics/src/widget/canvas/frame.rs
```diff
@@ -295,46 +297,133 @@ impl Frame {
...
+                    });
+
+                    self.fill(&glyph, text.color);
+                } else {
+                    // TODO: Raster image support for `Canvas`
+                    let [r, g, b, a] = text.color.into_rgba8();
+
+                    swash_cache.with_pixels(
+                        glyph.cache_key,
+                        cosmic_text::Color::rgba(r, g, b, a),
```
tarkah, line 408: I did some profiling and the bitmap pixels -> lyon tessellation is a huge performance killer. Even on release mode, I'm getting primitive generation times for size 80 text in the `8ms` range. This drops to `~400μs` when removing the emoji.

Maybe we add a disclaimer to the `fill_text` docs until there is better support for rastered images w/ transformations.

graphics/src/widget/canvas/frame.rs
```diff
@@ -295,46 +297,133 @@ impl Frame {
...
-            );
+        let metrics =
+            cosmic_text::Metrics::new(text.size as i32, text.size as i32);
+        let attrs = match text.font {
+            Font::Default => cosmic_text::Attrs::new(),
+            Font::External { name, .. } => cosmic_text::Attrs {
+                family: cosmic_text::Family::Name(name),
+                ..cosmic_text::Attrs::new()
+            },
+        };
```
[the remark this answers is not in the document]
hecrj: I mentioned some ideas in the Discord, but basically we will change the way `iced` deals with fonts.

`cosmic-text` should be able to add support to load fonts from bytes by leveraging [`Database::load_font_data`](https://docs.rs/fontdb/0.10.0/fontdb/struct.Database.html#method.load_font_data).

graphics/src/widget/canvas/frame.rs
tarkah: I ported the new example to `master`, but it's near impossible compare performance differences between the old / new `fill_text` since previously this was GPU rendered text and now it's tessellated by CPU. New changes are exposed in `primitive generation` numbers, but previously the impact would be hidden in `render`. Both maintain 60fps in release mode.

Obviously the canvas will most likely sit behind a `canvas::Cache` and in most workflows, the text is fairly static outside applying transformations.

hecrj: Yes, as I mentioned it's pretty inefficient!

The bitmap support is just a nice to have. For vectorial text, you'll want to deal with fonts that have outline information, so I don't think it's a big deal.

That said, we can probably skip the generic tessellation algorithm and hardcode the triangles ourselves.

graphics/src/widget/canvas/frame.rs
```diff
@@ -295,46 +297,133 @@ impl Frame {
...
-        let position = if self.transforms.current.is_identity {
-            text.position
-        } else {
-            let transformed = self.transforms.current.raw.transform_point(
-                lyon::math::Point::new(text.position.x, text.position.y),
-            );
+        let metrics =
+            cosmic_text::Metrics::new(text.size as i32, text.size as i32);
+        let attrs = match text.font {
+            Font::Default => cosmic_text::Attrs::new(),
```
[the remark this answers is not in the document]
hecrj: Eventually! Currently just focused on improving the text foundations. API should be accommodated later.

graphics/src/widget/canvas/frame.rs
hecrj: Also, important to note that once we remove `glyph-brush`, we will be able to compose primitives much more efficiently. We may be able to use the GPU glyph atlas approach for text that isn't transformed for better readability and performance.

Lots of doors will open once we get that bottleneck out of the way.

tarkah: Makes sense!

> That said, we can probably skip the generic tessellation algorithm and hardcode the triangles ourselves.

That would be nice. Even though tessellating rectangles should be very cheap, I'd imagine there's still some overhead with `lyon` that we can get around building the mesh ourselves.

hecrj: I believe it's as cheap as any other path, since we are using the generic path algorithm and not `fill_rectangle` (which only works with axis-aligned rectangles).

PR #1616, opened 2022-12-22
Use same name & order for toggler::new and helper

13r0ck:
The helper function for the toggler widget switched the order and name of the arguments passed when creating the toggler widget. This just standardizes the order whether the dev is using the helper or the associated function.


hecrj (approved):
Thanks!

I believe `Checkbox` could use the same treatment.

PR #1620, opened 2022-12-25
Pass command buffer generically into application to reduce allocation pressure

udoprog:
Think of this PR as an idea suggestion. I don't intend it to be merged in its current state.

This changes how commands are passed out of applications by doing it through a generic `impl Commands<Self::Message>` trait instead of having constructor and update functions return a `Command`.

```rust
impl Application for MyAwesomeApplication {
    type Message = Message;
    // snip

    fn new(flags: Self::Flags, commands: impl Commands<Self::Message>) -> Self {
        // snip
    }
    
    fn update(message: Self::Message, commands: impl Commands<Self::Message>) {
        // snip
    }
}
```

The `Commands trait is (as the time of writing) the following`:

```rust
/// Send commands to an iced application.
pub trait Commands<T> {
    /// Type of the reborrowed command buffer.
    ///
    /// See [Commands::as_mut].
    type AsMut<'a>: Commands<T>
    where
        Self: 'a;

    /// Helper to generically reborrow the command buffer.
    ///
    /// This is useful if you have a function that takes `mut commands: impl
    /// Commands<T>` and you want to use a method such as [Commands::map] which
    /// would otherwise consume the command buffer.
    ///
    /// This can still be done through an expression like `(&mut
    /// commands).map(/*  */)`, but having a method like this reduces the number
    /// of references involves in case the `impl Command<T>` is already a
    /// reference.
    fn as_mut(&mut self) -> Self::AsMut<'_>;

    /// Perform a single asynchronous action and map its output into the message
    /// type `M`.
    fn perform<F>(
        &mut self,
        future: F,
        map: impl Fn(F::Output) -> T + MaybeSend + Sync + 'static,
    ) where
        F: Future + 'static + MaybeSend;

    /// Insert a command into the command buffer.
    fn command(&mut self, command: Command<T>);

    /// Extend the current command buffer with an iterator.
    #[inline]
    fn extend<I>(&mut self, iter: I)
    where
        I: IntoIterator<Item = Command<T>>;

    /// Map the current command buffer so that it can be used with a different
    /// message type `U`.
    fn map<M, U>(self, map: M) -> Map<Self, M>
    where
        Self: Sized,
        M: MaybeSend + Sync + Clone + Fn(U) -> T;
}

/// Output of [Commands::map].
#[derive(Debug)]
pub struct Map<C, M> {
    /* snip */
}

impl<T: 'static, C, M: 'static, U: 'static> Commands<U> for Map<C, M>
where
    C: Commands<T>,
    M: MaybeSend + Sync + Clone + Fn(U) -> T;
```

The reason here is to reduce the number of allocations my application performs. I do a lot of command batching and each time you construct a batch you allocate a vector for it.

After something like this is available I want to take a stab at removing most allocations related to futures using an abstraction similar to [`reusable-box-futures`](https://docs.rs/reusable-box-future). The idea there would be that when we call `perform` we know of the layout of the future and can use a custom allocator to allocate space for it. Once the future has been handed to the executor and been polled to completion we can re-use that space. Since async Rust mostly produces uniformly sized futures this should be fairly doable.

Note that I've also added `Command::map` which is something I use for components in my application, but I couldn't find any such example in iced. But it works roughly the same as `Command::map`. You map the message of any sub-components ([Example use](https://github.com/udoprog/ontv/blob/commands/src/application.rs#L317)).

Also note `Command::command`, once something like this has been accepted I intend to [strip most of the `Internal` functionality](https://github.com/iced-rs/iced/blob/5acba65c1135d8ebbda94efc74ea0bebb29c438a/futures/src/command.rs#L6) since `None` and `Batch` will no longer be needed.

Thank you!


udoprog:
I ported my application to use this branch: <https://github.com/udoprog/ontv/tree/commands>.

The only outstanding concern I personally have is in case a really large command buffer is allocated. It currently will not be truncated which might be desirable with some kind of heuristic.

udoprog:
Thought I should motivate `by_ref` here:

The naive approach would be to simply do `&mut commands` whenever you want to reborrow a command buffer which actually works (since we provide a `&mut C where C: Commands<T>` implementation.

But this has two downsides:
* Nesting components can cause complex types to be generated, such as `&mut &mut &mut &mut playground::Foo` [in this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=442ff812cbffa1dddbb376ccf9d73936). In generic code in particular this can lead to problems such as very large type signatures or even be flat out rejected by rustc for the type being too large ([note the difference with `Trait::by_ref`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=47a54551e72ffa86702fdf91980a414c)).
* `commands.by_ref().map(/*  */)` has in my mind better ergonomics than `(&mut commands).map(/*  */)`, see also [`Iterator::by_ref`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.by_ref) so even if we remove the GAT it's in my mind worth keeping.

13r0ck:
This is going to require an RFC https://github.com/iced-rs/rfcs

udoprog:
Ok, here you go: https://github.com/iced-rs/rfcs/pull/20

PR #1628, opened 2022-12-30
Add `operate` method to `Component` trait

nicksenger:
This PR adds the option for components to mutate their internal state in response to widget operations. As a consequence of this the stored element is rebuilt for any operation. The component will still delegate operations to the rebuilt element.

### DIFF
--- a/lazy/src/component.rs
+++ b/lazy/src/component.rs
@@ -46,6 +46,14 @@ pub trait Component<Message, Renderer> {
     /// Produces the widgets of the [`Component`], which may trigger an [`Event`](Component::Event)
     /// on user interaction.
     fn view(&self, state: &Self::State) -> Element<'_, Self::Event, Renderer>;
+
+    /// Update the [`Component`] state based on the provided [`Operation`](widget::Operation)
+    fn operate(
+        &self,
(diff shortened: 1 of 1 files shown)


13r0ck:
Why do you think this is necessary?

nicksenger:
> Why do you think this is necessary?

`Component` allows the user to implement custom widgets as though they were `iced` applications with encapsulated state, but currently that state cannot be manipulated by way of an operation like it can in a custom implementation of the `Widget` trait.

Say, for example, some `Component` is created which should be focusable (its produced `view` depends on whether or not it is currently focused). This is currently very cumbersome to express in a way that's compatible with other focusable widgets, because the component's state cannot be manipulated through operations. As a workaround you can:

- create a wrapper `Widget` whose sole purpose is to operate on the focused state, then carefully share this state with the `Component`
- abandon using `Component` and just implement `Widget` directly to better leverage operations

The current behavior seems unnecessarily limiting to me given that `Component::State` is the same as the widget's internal state.

hecrj (approved):
Yeah, I believe this makes sense since a `Component` is meant to be used for implementing custom widgets.

PR #1633, opened 2023-01-03
use same name & order for checkbox::new and helper

13r0ck:
The helper function for the checkbox widget switched the order and name of the arguments passed when creating the checkbox widget. This just standardizes the order whether the dev is using the helper or the associated function.

Continuation of https://github.com/iced-rs/iced/pull/1616


hecrj (approved):
Thanks! :bow:

PR #1638, opened 2023-01-05
Fix shadowed variable

tarkah:
I created this bug with #1563 :see_no_evil: . In trying to please the compiler to satisfy borrow constraints, I accidentally shadowed the `layout` variable passed into the `overlay` method with  the content's layout. This causes the offset to be incorrect and therefore causes any overlay under a `Responsive` to be positioned incorrectly.

### DIFF
--- a/lazy/src/responsive.rs
+++ b/lazy/src/responsive.rs
@@ -280,12 +280,14 @@ where
                 );
 
                 let Content {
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Oh, damn... Pesky one!

Thanks!

PR #1640, opened 2023-01-05
Add focus operation and style for button

13r0ck:
This lets buttons also be tab navigated like the text_input.
It should be possible to tab cycle buttons with the `widget::focus_next` & `widget::focus_previous`, as well as focus by ID. Then when pressing `Enter` submits the same message as `on_press()`.

By default this just sets a button to the hovered styling when focused.

### DIFF
--- a/native/src/widget/button.rs
+++ b/native/src/widget/button.rs
@@ -2,16 +2,18 @@
 //!
 //! A [`Button`] has some local [`State`].
 use crate::event::{self, Event};
+use crate::keyboard;
 use crate::layout;
 use crate::mouse;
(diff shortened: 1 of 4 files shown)


hecrj (commented):
This makes sense, but I believe we need a more generic solution that can easily support other interactive widgets (checkboxes, radio buttons, togglers, etc.) and that can guarantee only a single widget will be focused at a time.

13r0ck:
This needs more discussion then, we are already running away with this then at system76 https://github.com/pop-os/iced/pull/22

PR #1643, opened 2023-01-07
Fix slider colors and add rail height

Night-Hunter-NF:
## Bug fixs
the rail color is vertical not horizontal
before:
image
after:
image

## features added
added rail height to style

## discussions
you can currently set the slider height but i think the height should be determined off the size of the handle or rail

### DIFF
--- a/style/src/slider.rs
+++ b/style/src/slider.rs
@@ -5,9 +5,18 @@ use iced_core::Color;
 #[derive(Debug, Clone, Copy)]
 pub struct Appearance {
     /// The colors of the rail of the slider.
-    pub rail_colors: (Color, Color),
+    pub rail: Rail,
     /// The appearance of the [`Handle`] of the slider.
     pub handle: Handle,
+}
+
+/// The appearance of a slider rail
+#[derive(Debug, Clone, Copy)]
+pub struct Rail {
+    /// The colors of the rail of the slider.
+    pub rail_colors: (Color, Color),
+    /// The appearance of the [`Handle`] of the slider.
+    pub rail_height: f32,
 }
 
 /// The appearance of the handle of a slider.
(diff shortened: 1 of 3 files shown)


hecrj:
The rail having 2 vertical colors is on purpose! It's meant to allow creating a rail with a subtle shadow.

Having different rail colors for left and right makes complete sense, however.

Night-Hunter-NF:
oh it was lol. didn't event occur to me haha

style/src/slider.rs
```diff
@@ -5,11 +5,20 @@ use iced_core::Color;
...
     pub handle: Handle,
 }
 
+/// The appearance of a slider rail
+#[derive(Debug, Clone, Copy)]
+pub struct Rail {
+    /// The colors of the rail of the slider.
+    pub rail_colors: (Color, Color),
+    /// The height of a slider rail and the width of a vertical slider.
+    pub rail_size: f32,
```
casperstorm, line 19: I say we rename these: `colors` and `size` since it feels redundant to have it prefixed with `rail_`

casperstorm, line 19: Now we are doing this we should add `border_radius`, and maybe `border_color` as well as `border_width`. Then we can get come nice rounded sliders.

Screenshot 2023-01-26 at 9 58 00 AM

hecrj (approved):
Looks great! Thanks :partying_face: 

I made some minor changes:

- Simplified the drawing logic in 45cfce3f6dcf2773bc8ccc7e356906cb778f2f27. We can draw the rail behind the handle. No need to subtract its width everywhere.
- Removed `border_width` support in c2cc9a835d8281c1d26c861ecac7991a924c3785. I believe it could lead to weird-looking results with thick borders.
- Changed the rail `border_radius` for the built-in theme to `2.0` in 6e6804c5c9d0dba921f2e25d2c888b136d22d35e. It looks nice!
- Renamed `Rail::size` to `width` in 9b39a17628e92b66ac3649e879478ed23635fa76. The way I see it, the width of a rail is similar to the width of a stroke.

We can wrap this up and merge! :ship:

PR #1644, opened 2023-01-11
Fix `overlay` implementation for `Lazy` widget

### LINKED ISSUE #1583: `Lazy` example broken

### What happened?

I ran `cargo r` in `examples/lazy`.

I then get:

```text
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', C:\<path>\iced\lazy\src\lazy.rs:221:22
```

I tried the current `main` @ 7ceb8b8d406cb72ea0e0c341ad755d69165d95f9 ,
as well as using iced 0.6 + iced_lazy 0.3 in a standalone project (copy pasting the lazy example).

The example to run.

### Version

master

### Operative System

Windows

_No response_

nicksenger:
Fixes #1583 

Currently the `Lazy` widget in `iced_lazy` is broken because we assume we can `get_mut` an `Rc` which is not unique - both the element & widget tree hold references to this data.

Since we now need a mutable reference to self in `overlay` to support operations, the compiler will prevent us from producing an overlay with a mutable reference to the same element that is being held by the `Lazy` & its widget tree node. As a workaround, I've temporarily removed the element from the references held on the `Lazy` & in the widget tree while the `Overlay` is in use, and restored it during the overlay's `drop`. Note that this will break if any widget methods from the parent widget are invoked prior to the overlay being dropped. Currently `UserInterface::update` is written such that this doesn't happen, but I'm not sure if it's necessarily guaranteed by Iced.

I've also added pick_lists to each item in the `Lazy` example to demonstrate that the overlays are working properly.


tarkah:
Looks good! Thanks for fixing this. `Option` ftw. I hate how overly complex things are due to `ouroboros` but not sure we can get around it w/ these use cases.

tarkah (approved):

hecrj (approved):

hecrj (approved):
Looks great! Thanks!

PR #1646, opened 2023-01-11
Provide `&Dependency` to `Lazy` widget view fn

nicksenger:
This is a small ergonomics improvement that I noticed would be beneficial when working with the `Lazy` widget. The change is just to call the provided view function with a reference to the specified `Dependency`.

This is useful when the data provided for `Dependency` originates from the widget tree state, and so we don't have a reference to it which lives for the required `'a` lifetime bound on `Dependency`. In this case we need to take ownership of the data to use it for `Dependency`. Unfortunately, the provided view function must also satisfy the `'a` bound, so the data needs to be cloned again if we want to move it into the view function for use there.

By providing view with a reference to the dependency stored in the lazy widget, we can defer this second clone until the cache is actually invalidated rather than doing it on every view. I think it also makes using `Lazy` a bit more intuitive.

I'd be happy to write an RFC if more explanation/examples are required to make this change

### DIFF
--- a/examples/lazy/src/main.rs
+++ b/examples/lazy/src/main.rs
@@ -167,7 +167,7 @@ impl Sandbox for App {
     }
 
     fn view(&self) -> Element<Message> {
-        let options = lazy(self.version, || {
+        let options = lazy(self.version, |_| {
             let mut items: Vec<_> = self.items.iter().cloned().collect();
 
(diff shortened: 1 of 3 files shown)


hecrj (approved):
I think it makes sense! Thanks :bow:

PR #1647, opened 2023-01-12
Widget-driven animations

### LINKED ISSUE #31: Animations

### LINKED ISSUE #560: [RFC] Blinking a `TextInput` cursor using periodic animation

hecrj:
Fixes #31 and closes #560.

This PR introduces a new method `request_redraw` to the `Shell` type which enables a `Widget` to request a redraw during `on_event`.

`Shell::request_redraw` takes a `window::RedrawRequest`, which is defined as follows:

```rust
/// A request to redraw a window.
#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
pub enum RedrawRequest {
    /// Redraw the next frame.
    NextFrame,

    /// Redraw at the given time.
    At(Instant),
}
```

Additionally, a new variant `RedrawRequested(Instant)` has been added to `window::Event`. This event is triggered before the user interface is redrawn.

As a result, widgets can now choose to draw whenever necessary; making animations possible! For instance, the `TextInput` widget has now a blinking cursor:

https://user-images.githubusercontent.com/518289/211976226-47b10743-0648-498b-98d3-8167191f41d3.mp4

Furthermore, these changes allow us to finally implement a subscription to subscribe to the frame rate of a window. The brand new `window::frames` subscription produces an `Instant` at the same frequency as the frame rate of the screen of the user. This can be very useful to smoothly animate layout or different parts of an application (as opposed to using `time::every` with a fixed frame rate). The `solar_system` example now uses the `window::frames` subscription!

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -15,4 +15,4 @@ version = "0.6"
 optional = true
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-wasm-timer = { version = "0.2" }
+instant = "0.1"
--- a/core/src/time.rs
+++ b/core/src/time.rs
@@ -1,9 +1,13 @@
 //! Keep track of time, both in native and web platforms!
 
 #[cfg(target_arch = "wasm32")]
-pub use wasm_timer::Instant;
+pub use instant::Instant;
+
+#[cfg(target_arch = "wasm32")]
+pub use instant::Duration;
 
 #[cfg(not(target_arch = "wasm32"))]
 pub use std::time::Instant;
 
+#[cfg(not(target_arch = "wasm32"))]
 pub use std::time::Duration;
--- a/examples/events/Cargo.toml
+++ b/examples/events/Cargo.toml
@@ -6,5 +6,5 @@ edition = "2021"
 publish = false
 
 [dependencies]
-iced = { path = "../.." }
+iced = { path = "../..", features = ["debug"] }
 iced_native = { path = "../../native" }
--- a/examples/events/src/main.rs
+++ b/examples/events/src/main.rs
@@ -1,11 +1,12 @@
 use iced::alignment;
 use iced::executor;
 use iced::widget::{button, checkbox, container, text, Column};
+use iced::window;
 use iced::{
     Alignment, Application, Command, Element, Length, Settings, Subscription,
     Theme,
 };
-use iced_native::{window, Event};
+use iced_native::Event;
 
 pub fn main() -> iced::Result {
     Events::run(Settings {
@@ -18,7 +19,6 @@ pub fn main() -> iced::Result {
 struct Events {
     last: Vec<iced_native::Event>,
     enabled: bool,
-    should_exit: bool,
 }
 
 #[derive(Debug, Clone)]
@@ -50,29 +50,27 @@ impl Application for Events {
                 if self.last.len() > 5 {
                     let _ = self.last.remove(0);
                 }
+
+                Command::none()
             }
             Message::EventOccurred(event) => {
                 if let Event::Window(window::Event::CloseRequested) = event {
-                    self.should_exit = true;
+                    window::close()
+                } else {
+                    Command::none()
                 }
             }
             Message::Toggled(enabled) => {
                 self.enabled = enabled;
-            }
-            Message::Exit => {
-                self.should_exit = true;
-            }
-        };
 
-        Command::none()
+                Command::none()
+            }
+            Message::Exit => window::close(),
+        }
     }
 
     fn subscription(&self) -> Subscription<Message> {
         iced_native::subscription::events().map(Message::EventOccurred)
-    }
-
-    fn should_exit(&self) -> bool {
-        self.should_exit
     }
 
     fn view(&self) -> Element<Message> {
--- a/examples/exit/src/main.rs
+++ b/examples/exit/src/main.rs
@@ -1,5 +1,7 @@
+use iced::executor;
 use iced::widget::{button, column, container};
-use iced::{Alignment, Element, Length, Sandbox, Settings};
+use iced::window;
+use iced::{Alignment, Application, Command, Element, Length, Settings, Theme};
 
 pub fn main() -> iced::Result {
     Exit::run(Settings::default())
@@ -8,7 +10,6 @@ pub fn main() -> iced::Result {
 #[derive(Default)]
 struct Exit {
     show_confirm: bool,
-    exit: bool,
 }
 
 #[derive(Debug, Clone, Copy)]
@@ -17,28 +18,27 @@ enum Message {
     Exit,
 }
 
-impl Sandbox for Exit {
+impl Application for Exit {
+    type Executor = executor::Default;
     type Message = Message;
+    type Theme = Theme;
+    type Flags = ();
 
-    fn new() -> Self {
-        Self::default()
+    fn new(_flags: ()) -> (Self, Command<Message>) {
+        (Self::default(), Command::none())
     }
 
     fn title(&self) -> String {
         String::from("Exit - Iced")
     }
 
-    fn should_exit(&self) -> bool {
-        self.exit
-    }
-
-    fn update(&mut self, message: Message) {
+    fn update(&mut self, message: Message) -> Command<Message> {
         match message {
-            Message::Confirm => {
-                self.exit = true;
-            }
+            Message::Confirm => window::close(),
             Message::Exit => {
                 self.show_confirm = true;
+
+                Command::none()
             }
         }
     }
--- a/examples/solar_system/src/main.rs
+++ b/examples/solar_system/src/main.rs
@@ -9,7 +9,6 @@
 use iced::application;
 use iced::executor;
 use iced::theme::{self, Theme};
-use iced::time;
 use iced::widget::canvas;
 use iced::widget::canvas::gradient::{self, Gradient};
 use iced::widget::canvas::stroke::{self, Stroke};
@@ -90,7 +89,7 @@ impl Application for SolarSystem {
     }
 
     fn subscription(&self) -> Subscription<Message> {
-        time::every(time::Duration::from_millis(10)).map(Message::Tick)
+        window::frames().map(Message::Tick)
     }
 }
 
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -15,8 +15,9 @@ trace = ["iced_winit/trace"]
 debug = ["iced_winit/debug"]
 system = ["iced_winit/system"]
 
-[dependencies.log]
-version = "0.4"
+[dependencies]
+log = "0.4"
+instant = "0.1"
 
 [dependencies.glutin]
 version = "0.29"
--- a/glutin/Cargo.toml
+++ b/glutin/Cargo.toml
@@ -17,7 +17,6 @@ system = ["iced_winit/system"]
 
 [dependencies]
 log = "0.4"
-instant = "0.1"
 
 [dependencies.glutin]
 version = "0.29"
--- a/glutin/src/application.rs
+++ b/glutin/src/application.rs
@@ -12,10 +12,11 @@ use iced_winit::futures;
 use iced_winit::futures::channel::mpsc;
 use iced_winit::renderer;
 use iced_winit::user_interface;
-use iced_winit::{Clipboard, Command, Debug, Proxy, Settings};
+use iced_winit::{Clipboard, Command, Debug, Event, Proxy, Settings};
 
 use glutin::window::Window;
 use std::mem::ManuallyDrop;
+use std::time::Instant;
 
 #[cfg(feature = "tracing")]
 use tracing::{info_span, instrument::Instrument};
@@ -131,7 +132,8 @@ where
         })?
     };
 
-    let (mut sender, receiver) = mpsc::unbounded();
+    let (mut event_sender, event_receiver) = mpsc::unbounded();
+    let (control_sender, mut control_receiver) = mpsc::unbounded();
 
     let mut instance = Box::pin({
         let run_instance = run_instance::<A, E, C>(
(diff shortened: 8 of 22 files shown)


ids1024:
I guess this currently depends on GL `swap_buffers` or wgpu `present` blocking until vsync? (With an appropriate swap interval or present mode.) I guess that's the right way to do it until https://github.com/rust-windowing/winit/issues/2412 is implemented.

On Wayland, Mesa's EGL/WSI implementation uses [`wl_surface::frame`](https://wayland.app/protocols/wayland#wl_surface:request:frame) here, which may block indefinitely for occluded windows (does this block other parts of Iced in an undesirable way?). So it should be noted that the event won't necessarily be emitted regularly. I'm not sure exactly what the behavior is like on other platforms.

13r0ck:
> Furthermore, these changes allow us to finally implement a subscription to subscribe to the frame rate of a window.

That is very exciting

hecrj:
> I guess this currently depends on GL `swap_buffers` or wgpu `present` blocking until vsync? (With an appropriate swap interval or present mode.) I guess that's the right way to do it until [rust-windowing/winit#2412](https://github.com/rust-windowing/winit/issues/2412) is implemented.

Yes. Without vsync, the `RedrawRequested` event will be produced as fast as possible when subscribed to `window::frames`. Since redraws are graphical, it seems appropriate to let the graphics driver control the framerate (instead of trying to figure out the screen frame rate and throttle ourselves).

> On Wayland, Mesa's EGL/WSI implementation uses [`wl_surface::frame`](https://wayland.app/protocols/wayland#wl_surface:request:frame) here, which may block indefinitely for occluded windows (does this block other parts of Iced in an undesirable way?). So it should be noted that the event won't necessarily be emitted regularly. I'm not sure exactly what the behavior is like on other platforms.

I can clarify in the docs that the subscription matches the maximum refresh rate of the window at any given moment, which is in turn controlled by graphics drivers and the OS.

PR #1648, opened 2023-01-12
Fix comment description on crate example

wiktor-k:
This tiny PR aligns the comment on the crate example to align to the comment that is in README.md.


hecrj (approved):
Good catch! Thanks :bow:

PR #1655, opened 2023-01-18
Group Overlay

### LINKED ISSUE #1591: Only first overlay element is being rendered

### What happened?

When trying to add multiple overlay elements, it renders the only first. I took a Floating Button example and got this bug. I also tried to implement my own overlay container but got the same result. For the second overlay even doesn't call an `overlay` function and the layout isn't being calculated.

Iced version is 0.5, but it can't be chosen from the version dropdown list.

Screenshot:
image

The code:
```rust
use iced::widget::Row;
use iced::{
    widget::{Button, Column, Container},
    Element, Length, Sandbox, Settings,
};

use iced_aw::floating_element::{Anchor, FloatingElement};
use iced_aw::Badge;

fn main() -> iced::Result {
    FloatingElementExample::run(Settings::default())
}

#[derive(Debug, Clone)]
enum Message {
    ButtonPressed,
}

struct FloatingElementExample {
    lines: Vec<String>,
}

impl Sandbox for FloatingElementExample {
    type Message = Message;

    fn new() -> Self {
        FloatingElementExample { lines: Vec::new() }
    }

    fn title(&self) -> String {
        String::from("FloatingButton example")
    }

        match message {
            Message::ButtonPressed => self.lines.push("This is a newly added line.".into()),
        }
    }

        let row = Row::new()
            .spacing(8)
            .push(
                FloatingElement::new(
                    Button::new("Underlay1").on_press(Message::ButtonPressed),
                    || Badge::new("Ovr1").into(),
                )
                .anchor(Anchor::NorthEast),
            )
            .push(
                FloatingElement::new(
                    Button::new("Underlay2").on_press(Message::ButtonPressed),
                    || Badge::new("Ovr2").into(),
                )
                .anchor(Anchor::NorthEast),
            );

        Container::new(Column::new().push(row))
            .width(Length::Fill)
            .height(Length::Fill)
            .padding(10)
            .center_x()
            .center_y()
            .into()
    }
}
```

I expect the all overlay elements to be rendered

### Version

0.4

### Operative System

Windows

_No response_

tarkah:
This adds a new `Group` overlay that allows for grouping multiple overlays returned by container widgets such as row, column & pane grid.

A new `Overlay::is_over` method needs to be added so we can properly validate whether the children of the Group are hovered, and not the group itself.

I've added a new example `toast` that shows what a toast notification system might look like in iced. This is really cool because it leverages widget animations to request redraw when toasts are set to expire and need to be removed.

As you can see below, both the `toast` overlay and the `picklist` overlay are able to show up at the same time!

https://user-images.githubusercontent.com/10239377/213056986-766fdce0-7de1-41c6-909a-94b3d339737c.mp4

Fixes #1591.

### DIFF
--- a/examples/toast/Cargo.toml
+++ b/examples/toast/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "toast"
+version = "0.1.0"
+authors = ["tarkah <admin@tarkah.dev>"]
+edition = "2021"
+publish = false
+
+[dependencies]
+iced = { path = "../..", features = [] }
+iced_native = { path = "../../native" }
--- a/examples/toast/src/main.rs
+++ b/examples/toast/src/main.rs
@@ -652,11 +652,7 @@ mod toast {
                 .unwrap_or_default()
         }
 
-        fn contains_cursor(
-            &self,
-            layout: Layout<'_>,
-            cursor_position: Point,
-        ) -> bool {
+        fn is_over(&self, layout: Layout<'_>, cursor_position: Point) -> bool {
             layout
                 .children()
                 .any(|layout| layout.bounds().contains(cursor_position))
--- a/examples/toast/src/main.rs
+++ b/examples/toast/src/main.rs
@@ -1,0 +1,674 @@
+use iced::widget::{
+    self, button, column, container, pick_list, row, slider, text, text_input,
+};
+use iced::{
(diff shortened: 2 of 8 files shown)


hecrj (approved):
Looks awesome! Thanks :partying_face:

PR #1660, opened 2023-01-19
Update example/game_of_life: use [default] on enum

Jedsek:
See title and changed file.

### DIFF
--- a/examples/game_of_life/src/preset.rs
+++ b/examples/game_of_life/src/preset.rs
@@ -1,7 +1,7 @@
 #[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
 pub enum Preset {
     Custom,
-    [default]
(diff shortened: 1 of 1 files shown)


examples/game_of_life/src/preset.rs
```diff
@@ -1,6 +1,7 @@
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
 pub enum Preset {
     Custom,
+    [default]
```
hecrj, line 4: Pretty sure this should be

```suggestion
    #[default]
```

hecrj (changes requested):
Thanks! Just a small thing!

hecrj (approved):

PR #1663, opened 2023-01-19
Update sandbox.rs

Zaplae:


Zaplae:
The `subscription` method should exist in `Sandbox`

Issue #1573, opened 2022-12-05
Scroller in scrollable is no longer visible when hovering its content


ben-wallis:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

In iced `0.4` the scroller within the scrollbar is only visible when mousing over the content - in `0.5` the scroller is permanently visible with seemingly no way to recreate the behavior of `0.4`. This is demonstrable by running the `scrollable` example for both versions.

### What is the expected behavior?

Either the default behavior of 0.4 is restored, or it is possible to optionally specify this behavior for a `scrollable`.

### Version

master

### Operative System

Windows

ben-wallis:
Thanks for the quick fix!

GyulyVGC:
I just want to add that now the scroller is not visible at all when the scrollable area is not hovered (I'm using v0.6.0).

Could this be solved? I feel it's useful to see the scroller even when not hovering the scrollable content.

PR #1669, opened 2023-01-23
Added scrollable style `focused` to be displayed when mouse is over the scrollable area

GyulyVGC:
Hi! It's my first PR here for iced!

I added a style `focused` thought to be displayed when the scrollable area is hovered (not the scrollbar itself).

In my opinion, this is a more flexible solution to #1573.
In this way, a user can decide to hide the scroller when the area is not hovered or simply to have a different style from the `active` case, which would now be used when the scrollable area is not hovered.
If `focused` is not implemented, `active` will be used anyway.

If you don't like the name "focused", suggestions are welcomed!

https://user-images.githubusercontent.com/100347457/214104009-cbd3a5d6-336a-4478-9fd4-879c2059bb70.mp4


hecrj (changes requested):
I think the name should be `hovered` instead of `focused`, like in a `button::StyleSheet`.

GyulyVGC:
> I think the name should be `hovered` instead of `focused`, like in a `button::StyleSheet`.

The point is that `hovered` is already used to refer to the case when just the Scrollbar is hovered (instead of the whole scrollable area).

GyulyVGC:
@hecrj may I ask you what can I do to make this PR ready to be merged?

hecrj:
@GyulyVGC We need a better name here. `focused` is misleading.

GyulyVGC:
You are free to suggest a name of your preference. 

Hovered would be awesome, but the problem is that it's already used.

hecrj:
I think we could just add a boolean `over_scrollbar` argument to the `hovered` method.

GyulyVGC:
> I think we could just add a boolean `over_scrollbar` argument to the `hovered` method.

Wouldn't this method also require to pass two distinct `style` arguments?

hecrj:
> Wouldn't this method also require to pass two distinct style arguments?

No. The `Style` is always the same for both states of the scrollable.

GyulyVGC:
> I think the name should be `hovered` instead of `focused`, like in a `button::StyleSheet`.

Following your suggestion, I renamed it to `hovered`. 
The style that was previously `hovered` has been renamed to `hovered_scrollbar`.

hecrj (approved):
Thanks!

I just added the `is_mouse_over_scrollbar` flag as I suggested in the end.

GyulyVGC:
Amazing! Thanks for the corrections!

PR #1671, opened 2023-01-24
Remove Application::title() and replace it with an action.

thunderstorm010:
Needs review (I probably don't know what I'm doing)

Removes the method `Application::title()`.
Introduces a new field in `Settings`: `title: String`
Introduces a new action: `action::Title(String)`

**This is a breaking change.**

### DIFF
--- a/glutin/src/application.rs
+++ b/glutin/src/application.rs
@@ -65,8 +65,8 @@ where
     };
 
     let context = {
-        let builder = settings.window.into_builder(
-            &application.title(),
+        let builder = settings.window.clone().into_builder(
+            &settings.window.title,
             event_loop.primary_monitor(),
             settings.id,
         );
--- a/glutin/src/application.rs
+++ b/glutin/src/application.rs
@@ -65,11 +65,9 @@ where
     };
 
     let context = {
-        let builder = settings.window.clone().into_builder(
-            &settings.window.title,
-            event_loop.primary_monitor(),
-            settings.id,
-        );
+        let builder = settings
+            .window
+            .into_builder(event_loop.primary_monitor(), settings.id);
 
         log::info!("Window builder: {:#?}", builder);
 
--- a/native/src/window/action.rs
+++ b/native/src/window/action.rs
@@ -70,7 +70,7 @@ pub enum Action<T> {
     /// - **Web / Wayland:** Unsupported.
     GainFocus,
     /// Change the title of the application.
-    Title(String)
+    ChangeTitle(String),
 }
 
 impl<T> Action<T> {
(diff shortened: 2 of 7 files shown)


glutin/src/application.rs
```diff
@@ -65,8 +65,8 @@ where
     };
 
     let context = {
-        let builder = settings.window.into_builder(
-            &application.title(),
+        let builder = settings.window.clone().into_builder(
+            &settings.window.title,
```
hecrj, line 69: Since the `title` is in the `settings.window` already, we can just remove the argument here and access it directly in `into_builder`.

native/src/window/action.rs
```diff
@@ -69,6 +69,8 @@ pub enum Action<T> {
     ///
     /// - **Web / Wayland:** Unsupported.
     GainFocus,
+    /// Change the title of the application.
+    Title(String)
```
hecrj, line 73: Let's call it `ChangeTitle`.

src/window/settings.rs
```diff
@@ -47,6 +50,7 @@ impl Default for Settings {
             transparent: false,
             always_on_top: false,
             icon: None,
+            title: String::from("Iced")
```
hecrj, line 53: Maybe something like "A cool application" would be _cooler_.

winit/src/application/state.rs
```diff
@@ -29,7 +28,7 @@ where
 {
     /// Creates a new [`State`] for the provided [`Application`] and window.
     pub fn new(application: &A, window: &Window) -> Self {
-        let title = application.title();
+
```
hecrj, line 31: Unnecessary blank line here :)

winit/src/application/state.rs
```diff
@@ -181,14 +179,6 @@ where
...
     /// [`Application::update`]: crate::Program::update
     pub fn synchronize(&mut self, application: &A, window: &Window) {
-        // Update window title
-        let new_title = application.title();
-
-        if self.title != new_title {
-            window.set_title(&new_title);
-
-            self.title = new_title;
-        }
```
hecrj, line 182: Same here. We can delete this blank line.

winit/src/settings.rs
```diff
@@ -93,6 +93,9 @@ pub struct Window {
 
     /// Platform specific settings.
     pub platform_specific: platform::PlatformSpecific,
+
+    /// Title of the window.
```
hecrj, line 97:
```suggestion
    /// The title of the window.
```

For consistency.

hecrj (changes requested):

I have some suggestions. Let me know!

thunderstorm010 (commented):
There's some formatting changes done by `cargo fmt`. Are those okay?

hecrj:
Closing as discussed on Discord:

> After some thought while changing the examples... I am not sure this is a good change.
> The title of an application is normally inherently related to a bunch of application state. Therefore, using a Command to update it seems inappropriate since you will have to potentially couple this state in many different parts of your update logic.
> Some examples become more complex, so I don't see a net gain here.

PR #1673, opened 2023-01-25
Attempt to load_system_fonts() when loading SVGs

### LINKED ISSUE #870: Invisible text in svgs after updating to 0.3.0
An svg in iced 0.2.0:

svg in iced 0 2 0

The same svg in iced 0.3.0:

svg in iced 0 3 0

Code:

[iced_svg_test.zip](https://github.com/hecrj/iced/files/6441193/iced_svg_test.zip)

ripytide:
The default `usvg::Options` skips text elements. This is because it doesn't attempt to `load_system_fonts()`.

This PR may fix #870

### DIFF
--- a/examples/checkbox/README.md
+++ b/examples/checkbox/README.md
@@ -6,7 +6,7 @@ The __[`main`]__ file contains all the code of the example.
 
 You can run it with `cargo run`:
 ```
-cargo run --package pick_list
+cargo run --package checkbox
 ```
 
 [`main`]: src/main.rs


hecrj (changes requested):
Is there any overhead from calling `load_system_fonts`?

> may fix https://github.com/iced-rs/iced/issues/870

Could we get some confirmation?

Also, a `cargo fmt` is needed!

ripytide:
> Is there any overhead from calling load_system_fonts?

According to the docs it looks in a few predefined system directories to look for available fonts. Not sure if that is classed as too much overhead.

> Could we get some confirmation?

It fixed text rendering on my machine. I didn't get time to check using the code given in #870

Closing since you updated the `resvg` version to `0.29` and the `load_system_fonts()` function no longer exists, or at least it has been moved according to https://github.com/RazrFalcon/resvg/blob/master/CHANGELOG.md#removed-1

PR #1675, opened 2023-01-26
Extended `pick_list::Handle` to be more flexible

casperstorm:
I have extended the  `pick_list::Handle` slightly:

```
{
    Arrow,
    Static({ .. }),
    Dynamic {
        closed: { .. },
        open: { .. },
    },
    None,
}
``` 

This was done because I felt the previous solution was too limited by not allowing the user to choose different content depending on the state of the `pick_list`. The main use case is to display a different icon, like ▲ and ▼, depending on the open/close state.

Below is a small gif show casing the different states.

pick_list_handle

### DIFF
--- a/native/src/widget/pick_list.rs
+++ b/native/src/widget/pick_list.rs
@@ -224,7 +224,7 @@ where
             self.selected.as_ref(),
             &self.handle,
             &self.style,
-            tree.state.downcast_ref::<State<T>>(),
+            || tree.state.downcast_ref::<State<T>>(),
         )
     }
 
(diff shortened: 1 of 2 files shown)


hecrj (approved):
Looks great! Thanks.

I have simplified some stuff here and there:

- Renamed `HandleContent` to `Icon` and made the `text` be a `code_point` instead.
- Changed the generic type to be the `Font` type directly.
- Inlined the `Handle::content` helper since it;s only used in `draw` once.

Let's merge! :ship:

PR #1676, opened 2023-01-27
Annotate `Command` and `Subscription` with `#[must_use]`

ids1024:
Calling a function returning one of these types without using it is almost certainly a mistake. Luckily Rust's `#[must_use]` can help warn about this.


ids1024:
I guess the clippy error is unrelated and added in Rust 1.67.

hecrj (approved):

hecrj:
Sweet!

PR #1678, opened 2023-01-27
Mapped operations is missing text_input()

frey:
This fixes a bug where widget operations could be dropped. In some layouts, an `operation` for widgets of type `text_input` are not mapped.


hecrj (approved):
Good catch. Thanks!

PR #1688, opened 2023-01-31
glutin: Rewrite using Glutin 0.30 to provide `Compositor`

ids1024:
This provides a `Compositor` adapting a `GLCompositor`, that thus can be used with `iced_winit` or with a different shell.

Seems to be working for me on Wayland, but it needs more testing. And there are a few remaining questions:

* How should it handle the case where `compatible_window` is `None`? Is it reasonable to return an error in that case? Should the trait not take an `Option` here?
* Is using the same config and context for all windows reasonable? I guess different windows may have different requirements in some cases on some platforms?
* It probably should do something fancier than selecting the first config returned by `find_configs`.
* Should the `create_surface` method of the trait return a `Result`?


ids1024:
The CI failure is in `integration_opengl`. https://github.com/iced-rs/iced/pull/1689 has some initial changes to make that not depend on Glutin 0.29.

ids1024:
If https://github.com/iced-rs/iced/pull/1689 is merged that will fix the build failure. Still leaving the questions above.

ids1024:
Perhaps `Compositor` could also benefit from a way to control vsync behavior. On Wayland, `present` may block indefinitely on vsync for an occluded window, since the compositor does not send frame callbacks for windows that are not visible.

Not sure if that's an issue with Iced currently, but it would impact multi-window and may be a problem for some integrations of Iced.

PR #1689, opened 2023-02-01
Use `glutin-winit` in `integration_opengl` example

ids1024:
Glutin 0.30 changes the API to not depend on winit, and instead take a raw window handle. `glutin-winit` is a helper to make it a bit easier to create a GL context on top of winit.

Things like `iced_winit::conversion::window_event` are problematic here since we end up using a different winit version from `iced_winit`, and that function is quite complicated to duplicate...

### DIFF
--- a/examples/integration_opengl/Cargo.toml
+++ b/examples/integration_opengl/Cargo.toml
@@ -6,7 +6,10 @@ edition = "2021"
 publish = false
 
 [dependencies]
-iced_glutin = { path = "../../glutin" }
 iced_glow = { path = "../../glow" }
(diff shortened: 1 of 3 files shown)


ids1024:
I guess for `window_event`, the expected way to do it in an integration like this might be to only convert the event types that are relevant to the application?

ids1024:
This seems to be working well now, with some event conversion code copied from `iced_winit`.

An OpenGL integration is expected to be used with things other than `winit`, and in any case not the Iced fork of `winit`, so this is perhaps better as a minimal example of what an integration like this requires.

PR #1691, opened 2023-02-02
Logarithmic slider

GyulyVGC:
This PR introduces a function `logarithmic_scale()` to `slider` and `vertical_slider` to set the distribution of values of the slider range to logarithmic, instead of linear.

This is particularly useful in the case of very wide ranges for which is wanted to have a good precision in selecting the range lower values.

Of course, since the logarithm is not defined for non-positive numbers, it is meaningful to use this option only for ranges of values greater than zero.

Below an example of implementation.

https://user-images.githubusercontent.com/100347457/216378631-b8ed5637-b945-4464-a557-2fea7ab4f209.mp4

### DIFF
--- a/core/src/vector.rs
+++ b/core/src/vector.rs
@@ -15,6 +15,11 @@ impl<T> Vector<T> {
     }
 }
 
+impl Vector {
+    /// The zero [`Vector`].
+    pub const ZERO: Self = Self::new(0.0, 0.0);
+}
+
 impl<T> std::ops::Add for Vector<T>
 where
     T: std::ops::Add<Output = T>,
--- a/examples/download_progress/src/main.rs
+++ b/examples/download_progress/src/main.rs
@@ -177,7 +177,7 @@ impl Download {
                     .into()
             }
             State::Downloading { .. } => {
-                text(format!("Downloading... {:.2}%", current_progress)).into()
+                text(format!("Downloading... {current_progress:.2}%")).into()
             }
             State::Errored => column![
                 "Something went wrong :(",
--- a/examples/events/src/main.rs
+++ b/examples/events/src/main.rs
@@ -77,7 +77,7 @@ impl Application for Events {
         let events = Column::with_children(
             self.last
                 .iter()
-                .map(|event| text(format!("{:?}", event)).size(40))
+                .map(|event| text(format!("{event:?}")).size(40))
                 .map(Element::from)
                 .collect(),
         );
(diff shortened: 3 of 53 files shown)


hecrj:
Since the behavior seems different from a normal slider, I have a feeling that this should be a new widget with a new design as well.

GyulyVGC:
> Since the behavior seems different from a normal slider, I have a feeling that this should be a new widget with a new design as well.

I don't feel this needing, if you think about it, it is still a slider just with a different values distribution.
Its behaviour is actually the same, but now the range is not linear but logarithmic.
Creating a new widget for it would mean to duplicate a lot of code, since I don't see how the design should be changed either.
If values in the range are expressed as power of ten (Kilo, Mega, Giga...or 10^2, 10^3...) as in my use case, its apparent behaviour would be exactly that of a standard slider. There are just context in which the linear scale is not appropriate.

Btw, I fixed the last clippy lints and now everything should pass.

Just let me know if for you it's good that in case the user sets a negative values range it panics or if in this case the `logarithmic_scale()` should have no effect (keeping it to linear even if this function is called).

GyulyVGC:
> Since the behavior seems different from a normal slider, I have a feeling that this should be a new widget with a new design as well.

I don't feel this needing, if you think about it, it is still a slider just with a different values distribution.
Its behaviour is actually the same, but now the range is not linear but logarithmic.
Creating a new widget for it would mean to duplicate a lot of code, since I don't see how the design should be changed either.
If values in the range are expressed as power of ten (Kilo, Mega, Giga...or 10^2, 10^3...) as in my use case, its apparent behaviour would be exactly that of a standard slider. There are just context in which the linear scale is not appropriate.

Btw, I fixed the last clippy lints and now everything should pass.

Just let me know if for you it's good that in case the user sets a negative values range it panics or if in this case the `logarithmic_scale()` should have no effect (keeping it to linear even if this function is called).

hojjatabdollahi:
We needed something like this to change audio volume. We just used a normal slider and added the log and power function when reading or setting the value. That enabled us to create different curves and find a natural response (slider to volume). So I don't think having this flag on slider would've helped us, as we needed to tweak the functions to get the best curve. Also, I don't think in most cases the user needs to know that the slider is logarithmic (they move the slider to the middle and expect the audio to feel half as loud), so a new widget for this with different UI may not be needed either. But this was just in our case, maybe there are other usecases that it makes sense to somehow tell the user that they are changing a logarithmic slider.

hecrj:
I agree with @hojjatabdollahi in that this functionality seems quite specific to a particular use case. I would expect the user of the library to build it on top of the current slider.

PR #1692, opened 2023-02-02
Use overlay for tooltip

tarkah:
This converts the tooltip to an `Overlay` instead of using `renderer.with_layer`.

However, I just realized as I'm posting this that it wasn't #1655 that would unblock this, but instead the addition of support for "nested" overlays.

I'll leave this open since the work won't change once unblocked.


hecrj (approved):
Awesome! Rebased on top of `master`. Let's merge! :tada:

PR #1694, opened 2023-02-06
Accept FnOnce instead of Fn in canvas cache draw

### LINKED ISSUE #1119: Why not use `FnOnce` rather than `Fn` in `Cache::draw`?
Hello, I'm sorry about my broken English. So let me ask the question directly. In file [cache.rs](https://github.com/iced-rs/iced/blob/master/graphics/src/widget/canvas/cache.rs), the method `draw` use a `impl Fn(&mut Frame)` as its parameter:

```rust
pub fn draw(&self, bounds: Size, draw_fn: impl Fn(&mut Frame)) -> Geometry
```

So, I can't move a iterator into the closure and have to collect the iterator before calling `draw` method. I wonder why not just use a `FnOnce`?

sushigiri:
Use FnOnce in `draw` function signature instead of `Fn`, permitting the use of iterators and other one-time functions. Resolves #1119

### DIFF
--- a/graphics/src/widget/canvas/cache.rs
+++ b/graphics/src/widget/canvas/cache.rs
@@ -49,7 +49,11 @@ impl Cache {
     /// Otherwise, the previously stored [`Geometry`] will be returned. The
     /// [`Cache`] is not cleared in this case. In other words, it will keep
     /// returning the stored [`Geometry`] if needed.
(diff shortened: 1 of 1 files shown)


sushigiri:
Looks like tests keep failing at installing dependencies

PR #1695, opened 2023-02-10
Provide widgets IDs to `Operation::container`

nicksenger:
Use of `operation::scoped` requires a widget which provides its ID to `Operation::container`. Currently none of the provided container widgets do this, so using scoped operations requires implementation of a custom widget. This PR adds the option to provide an ID to each container widget that doesn't have it already (button, container, row, column, pane grid) and provides this ID to `Operation::container` in `Widget::operate`.

### DIFF
--- a/native/src/widget/button.rs
+++ b/native/src/widget/button.rs
@@ -8,7 +8,7 @@ use crate::overlay;
 use crate::renderer;
 use crate::touch;
 use crate::widget::tree::{self, Tree};
-use crate::widget::{self, Operation};
+use crate::widget::Operation;
 use crate::{
     Background, Clipboard, Color, Element, Layout, Length, Padding, Point,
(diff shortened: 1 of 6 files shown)


hecrj (commented):
I wonder if we could just get away with adding the `id` to just the `Container` widget?

tarkah:
> I wonder if we could just get away with adding the `id` to just the `Container` widget?

I agree with this. Wrap whatever with a `Container` if you need to use `scoped`

nicksenger:
Sounds good, I cut this down to just the container 👍

hecrj (approved):
Great! I have reverted 63cfec6532b56f74143aab958cb4b5f6f3f365d2, since `Scrollable` already has an `Id` and it makes sense to provide it.

PR #1699, opened 2023-02-11
Implement `is_over` for iced_lazy widget `Overlay`s 

nicksenger:
The lazy widgets all share this pattern of storing some element which may or may not produce an overlay, but they are not delegating `is_over` to the stored element's overlay. This PR adds the `is_over` method to each of the lazy widgets' overlay implementations, forwarding the call to the contained overlay.

### DIFF
--- a/lazy/src/component.rs
+++ b/lazy/src/component.rs
@@ -563,4 +563,11 @@ where
 
         event_status
     }
+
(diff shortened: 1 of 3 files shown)


hecrj (approved):
Good catch! Thanks :bow:

PR #1700, opened 2023-02-11
Fix panic when overlay event processing removes overlay

nicksenger:
Currently if processing an event in the overlay both invalidates the layout and results in the overlay no longer being present (we hit line 251), the application will panic. This PR fixes the issue by using `cursor_position` for the base cursor position if the overlay no longer exists.

Example application that panics: https://gist.github.com/nicksenger/13f051ebdf34fbede31abe91981d70a6

### DIFF
--- a/native/src/user_interface.rs
+++ b/native/src/user_interface.rs
@@ -263,16 +263,16 @@ where
                 }
             }
 
-            let base_cursor = if manual_overlay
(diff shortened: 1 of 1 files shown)


hecrj (approved):
This is what happens when you use `unwrap`, kids. Be careful out there.

Thanks again!

PR #1702, opened 2023-02-13
Added `Icon` to `text_input`

casperstorm:
This PR adds the ability to add a `Icon` to a `text_input`. This opens up for quite a lot of new styling possibilities to `text_input` as a `Icon` is defined by:

```rust 
let handle = text_input::Icon { 
    font: ICON_FONT,
    code_point: '\u{e900}',
    size: Some(14),
    position: text_input::IconPosition::Left,
};

let text_input = text_input(..).icon(icon);
```
There are multiple use-cases with the handle. Eg; login and search inputs with icons to the left and inputs which has invalid or unexpected content could show a warning icon to the right.

Screenshot 2023-02-13 at 11 44 43 AM

text_input


examples/text_input/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
```
hecrj, line 1: We probably want to avoid having an example for every piece of functionality.

Maybe we could instead change the `tour`?

native/src/widget/text_input.rs
```diff
@@ -827,7 +880,40 @@ pub fn draw<Renderer>(
...
+                IconPosition::Right => Rectangle {
+                    x: bounds.x,
+                    width: bounds.width - (width + padding),
+                    ..bounds
+                },
+            }
+        } else {
+            bounds
+        }
+    };
```
hecrj, line 916: Could we compute this only once in `layout`? We could add the icon child node after the `text` child, that way event logic won't need to change at all.

hecrj (changes requested):

examples/text_input/Cargo.toml
casperstorm: Yeah let me extend `tour` instead!

casperstorm: Done in [1a0f0c2](https://github.com/iced-rs/iced/pull/1702/commits/1a0f0c2a5368561984b5a124bb70e7ec6d900715).

nicoburns:
This could always be added in a future PR, but it would be cool if these icons could handle click events. The use case I am thiking of is a "show password" icon on a password input. Or a magnifying glass search button on a search input.

Night-Hunter-NF:
is there a reason to not make the icon impl Into<Element>?
so you could have an image or svg or text. can also have buttons etc which all seem like useful things

hecrj (approved):
Good stuff! :partying_face:

I made some changes here and there. The most important ones:

- Fixed mismatched types related to #1711 in 0e2fc99eb864800d2d1522c015054d84cad078f4.
- Move the icon layout logic in the `layout` function and added a `spacing` field to `Icon` in 9852b4b36442ef036f0b308f798e892ddaa06c2d.
- Renamed `text_input::IconPosition` to `text_input::Side` in cf9d8e01048845ee503a62eb55e634a76a0e9163.

With this, I think we can merge this! We can tackle further use cases separately :ship:

PR #1703, opened 2023-02-13
glow: enable GL_ARB_explicit_attrib_location to fix crashes

greatest-ape:


ids1024:
Presumably it would be safer (though annoying) to always query the attribute location and not rely on explicit locations, since it would work on GL implementations that only support lower versions and don't have this extension.

Though hopefully this shouldn't break things anywhere it's currently working, so it should strictly improve support relative to the existing version, at least.

greatest-ape:
@ids1024 Fair point. There is a larger rewrite of the gradient code which seems to also fix the issue [here](https://github.com/iced-rs/iced/pull/1597). The benefit of the simple fix is that it is small and that it is not breaking and could therefore be part of a point release.

hecrj (approved):
Thanks! Let's apply this fix for now.

Waiting for a couple of things to line up before the gradient rewrite can land.

PR #1706, opened 2023-02-14
Pad after setting width in text input

tarkah:
This fixes an issue where using `Length::Units` and `Padding` with a text input causes it to be bigger than `Length::Units` by `Padding`.

We need to apply `pad` to `Limits` AFTER `width`, otherwise `width` will set `Limits` back to a fixed width if `Length::Units` is used, overwriting padding.


casperstorm:
Good catch 👍🏻

hecrj (approved):
Thanks!

PR #1707, opened 2023-02-16
Added the ability to change `Checkbox` icon

casperstorm:
This PR extends `Checkbox` so its now possible to customize the icon.
The `Icon` is defined by the following. 

```rust
pub struct Icon<Font> {
    pub font: Font,
    pub code_point: char,
    pub size: Option<u16>,
}
```
I renamed `checkmark_color` to `icon_color` inside `checkbox::Appearance` since it feels more fitting with a custom icon that might not be a checkmark.
 
To test run `cargo run --package checkbox`
 

Screenshot 2023-02-16 at 2 13 23 PM


hecrj (approved):
Looks awesome! Thanks! :tada:

PR #1717, opened 2023-02-18
Remove `Clone` bound for `Bytes::new` in `image`

hecrj:


sdroege:
FWIW this is a backwards-compatible change and could also be added to a 0.8.1 release. I don't need it myself but it might be useful for someone else :)

hecrj:
Yes, I released this change as a minor patch of `iced_native`.

PR #1719, opened 2023-02-18
Nested overlays

tarkah:
This implements a new overlay which allows for nesting. It's fairly inefficient as it calls `Overlay::overlay` within every method, but this reduces complexity as dealing w/ lifetimes here will be really rough.

The nice part is this doesn't touch the `UserInterface` code at all except to wrap the root overlay into a `Nested`. `Nested` is only public to the crate and only meant to be used as the root overlay used by the user interface. This means that the layout caching done in user interface works for the `Nested` layout as well.

I haven't implemented the nested `Overlay::overlay` on any of the `lazy` overlay implementations yet, but I can implement if we feel this is a good approach.

If approved, we can rebase #1692 as that should now work w/ this.

https://user-images.githubusercontent.com/10239377/219901259-7a187e90-f4d6-43d3-ae86-c25716c08f66.mp4

### DIFF
--- a/widget/src/pick_list.rs
+++ b/widget/src/pick_list.rs
@@ -157,11 +157,11 @@ where
         From<<Renderer::Theme as StyleSheet>::Style>,
 {
     fn tag(&self) -> tree::Tag {
-        tree::Tag::of::<State<T>>()
+        tree::Tag::of::<State>()
     }
 
     fn state(&self) -> tree::State {
-        tree::State::new(State::<T>::new())
+        tree::State::new(State::new())
     }
 
     fn width(&self) -> Length {
@@ -209,7 +209,7 @@ where
             self.on_selected.as_ref(),
             self.selected.as_ref(),
             &self.options,
-            || tree.state.downcast_mut::<State<T>>(),
+            || tree.state.downcast_mut::<State>(),
         )
     }
 
@@ -249,7 +249,7 @@ where
             self.selected.as_ref(),
             &self.handle,
             &self.style,
-            || tree.state.downcast_ref::<State<T>>(),
+            || tree.state.downcast_ref::<State>(),
         )
     }
 
@@ -259,7 +259,7 @@ where
         layout: Layout<'_>,
         renderer: &Renderer,
     ) -> Option<overlay::Element<'b, Message, Renderer>> {
-        let state = tree.state.downcast_mut::<State<T>>();
+        let state = tree.state.downcast_mut::<State>();
 
         overlay(
             layout,
@@ -269,6 +269,7 @@ where
             self.text_shaping,
             self.font.unwrap_or_else(|| renderer.default_font()),
             &self.options,
+            &self.on_selected,
             self.style.clone(),
         )
     }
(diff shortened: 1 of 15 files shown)


tarkah:
CI appears to be on drugs, `native/src/user_interface.rs:452:18` is nowhere near any call to `is_over`

tarkah:
I'm getting a layout related panic (unwrap) in the `modal` example when interacting with the pick_list, but I can't always recreate it.

native/src/overlay/nested.rs
```diff
@@ -0,0 +1,359 @@
...
+                        event,
+                        layout,
+                        cursor_position,
+                        renderer,
+                        clipboard,
+                        shell,
+                    )
+                } else {
+                    status
+                }
```
tarkah, line 234: Still need to encode cursor availability here when the overlay a layer higher `is_over`.

I was going to, but there was an issue in the `modal` example because of how `pick_list` handles mouse clicks on it's overlay menu. It doesn't capture the event when clicked in the overlay, and instead lets the base widget handle "closing" and publishing to shell the selection. 

This is an issue because then the nested pick_list (higher layer) reports `is_over`, but doesn't capture the event. So if we set cursor to `(-1, -1)`, the "modal" overlay will think it's clicked outside and close.

I think we need to refactor `pick_list` to "capture" the mouse click in the overlay menu.

tarkah: I've resolved this with a617f3a and 4d439e4.

You can compare the behavior prior to a617f3a on `modal` example. If you select the middle picklist option over the text input, the text input focuses after selecting the option.

After a617f3a, the undesired behavior mentioned above occurs (modal closes when clicking outside menu bounds / onto modal).

After 4d439e4, the picklist menu now properly captures the status of the event in the overlay when an option is selected. Everything now works as expected.

tarkah:
I think we can just naively store the overlay element of the ouroboros builders tail field for the lazy widgets in a `Nested` and get nesting there without touching the existing overlay implementation. I don't even think it'll be possible to return an overlay from within those overlay implementations anyways. Maybe hackable for lazy / responsive but definitely not component.

hecrj (approved):
Awesome! Thanks :partying_face: 

I have rebased on top of `master` and everything seems to be working.

widget/src/pick_list.rs
```diff
@@ -570,7 +560,11 @@ where
             &mut state.menu,
             options,
             &mut state.hovered_option,
-            &mut state.last_selection,
+            |option| {
+                state.is_open = false;
+
+                (on_selected)(option)
+            },
```
hecrj, line 567: I was quite surprised when this compiled. Rust seems to be capable to partially capture fields of a struct in a closure now.

tarkah: It makes a big difference!

tarkah:
> Awesome! Thanks partying_face This is a huge win.
> 
> I have rebased on top of `master` and everything seems to be working. Hopefully I didn't break anything!

Looks good to me and modal example is still working :+1:

MichalLebeda:
@tarkah As you said "Nested is only public to the crate and only meant to be used as the root overlay used by the user interface" I wonder why is it used in `Responsive`?

PR #1721, opened 2023-02-20
Sliders no longer bleed over rail

casperstorm:
This PR is a suggestion for `Slider` to not bleed over the rails.   

The default behaviour is that `Slider` comes to a stop when the handle is `handle_width / 2` over the rail. I would like to propose we only move the handle within the rail. The gif below illustrate the current behaviour as well as my suggestion:

Bleed (current)           |  No bleed (this PR)
:-------------------------:|:-------------------------:
slider_with_bleed  |  slider_no_bleed

I believe this is the most common approach. macOS and Windows references below:

macos

windows

### DIFF
--- a/native/src/widget/slider.rs
+++ b/native/src/widget/slider.rs
@@ -423,8 +423,8 @@ pub fn draw<T, R>(
     let handle_offset = if range_start >= range_end {
         0.0
     } else {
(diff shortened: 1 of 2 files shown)


hecrj (approved):
Makes sense! Drawing outside of the widget bounds is forbidden as well :)

Thanks!

PR #1725, opened 2023-02-22
Impl std::ops for Size, Vector #1724

ArtHome12:
I found that the mathematical operations on Size, Vector are only partially implemented and prepared changes to support +, -, *, /, %, as well as +=, -=, *=, /=, %=. The second operand can be either a scalar or Self.


ArtHome12:
This is an incomplete solution. It is better to replace the built-in types with similar ones from the geometric crate, for example, Euclid

PR #1726, opened 2023-02-22
Added `on_unfocus` method to `text_input`

GyulyVGC:
I added the possibility to set a `Message` when the Text Input passes from a focused to an unfocused state.

This is similar to `on submit` in a certain way, and it is helpful in scenarios in which we only want to produce a message when the user has terminated his actions, without the need to press the return key.

### DIFF
--- a/native/src/widget/text_input.rs
+++ b/native/src/widget/text_input.rs
@@ -68,6 +68,7 @@ where
     on_change: Box<dyn Fn(String) -> Message + 'a>,
     on_paste: Option<Box<dyn Fn(String) -> Message + 'a>>,
     on_submit: Option<Message>,
+    on_unfocus: Option<Message>,
     style: <Renderer::Theme as StyleSheet>::Style,
 }
 
@@ -99,6 +100,7 @@ where
             on_change: Box::new(on_change),
             on_paste: None,
             on_submit: None,
+            on_unfocus: None,
             style: Default::default(),
         }
     }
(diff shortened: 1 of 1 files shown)


hecrj (commented):
The main concern with this listener is that it won't trigger when the input loses focus because of changes in `view`.

What is the use case you had in mind here? I believe most of the use cases satisified by this functionality could be better achieved with a timer and waiting for the user to stop typing.

GyulyVGC:
It seems natural to me that the value should be processed only when a user finishes his operation, and one way to achieve it is looking at when the text input unfocuses.

There are various cases where users finish typing without expecting to hit the return key, and in these cases `on_submit` wouldn't be called.

An example that comes to my mind: you have a GUI that performs computations in real-time on the basis of the text input value and you would not like to update the GUI behaviour while the user is typing (e.g., a numerical text input => if you want to change the current value you may have to clear and rewrite everything). 
Of course this could be solved adding a "confirm" button or something like that, but this wouldn't permit to build a minimal and functional interface.

GyulyVGC:
Also, doing this task with a timer is not very flexible to me, since in that case we'd assume something about the writing speed of the user, and the assumption could be wrong.

hecrj:
I don't see how forcing to user to unfocus a field to process a value is more functional or intuitive than a confirm button or a timer.

I think we need to review the use cases further here. We may need [an RFC](https://github.com/iced-rs/rfcs) discussing these kinds of handlers.

PR #1730, opened 2023-02-23
Platform specific `window` options.

casperstorm:
This PR enables us to add some platform specific `window` options.
For macOS it means we can set three new variables:

```rust    
    /// Hides the window title.
    pub title_hidden: bool,
    /// Makes the titlebar transparent and allows the content to appear behind it.
    pub titlebar_transparent: bool,
    /// Makes the window content appear behind the titlebar.
    pub fullsize_content_view: bool,
```

An example below is shown where all three of the above mentioned variables are enabled:

```rust
Settings {
            platform_specific: iced::window::PlatformSpecific {
                title_hidden: true,
                titlebar_transparent: true,
                fullsize_content_view: true,
            },
            ..Default::default()
        },
    ..Default::default()
}
```

Screenshot 2023-02-23 at 2 21 16 PM


src/window/settings.rs
```diff
@@ -32,6 +42,9 @@ pub struct Settings {
 
     /// The icon of the window.
     pub icon: Option<Icon>,
+
+    /// Platform specific settings.
+    pub platform_specific: platform::PlatformSpecific,
```
tarkah, line 47: Nit: redundant import path. `PlatformSpecific` is already in scope in this module.

casperstorm: If i remove that line, what do you then want me to export here?

```rust
// src/window.rs
pub use settings::{PlatformSpecific, Settings};
```


```bash
error[E0432]: unresolved import `settings::PlatformSpecific`
 --> src/window.rs:9:20
  |
9 | pub use settings::{PlatformSpecific, Settings};
  |                    ^^^^^^^^^^^^^^^^ no `PlatformSpecific` in `window::settings`
  |
help: consider importing one of these items instead
  |
9 | pub use settings::{crate::Event::PlatformSpecific;
  |                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
9 | pub use settings::{crate::event::PlatformSpecific;
  |                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
9 | pub use settings::{crate::runtime::settings::PlatformSpecific;
  |                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
9 | pub use settings::{iced_native::Event::PlatformSpecific;
  |                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    and 4 other candidates
```

hecrj (approved):
Thank you!

I think we can expose the `iced_winit` struct directly, since the platform specific settings are already internal details. I did this in the last commit.

PR #1731, opened 2023-02-23
Remove logging large bytes arrays

tarkah:
I've noticed a huge impact on startup time when logging is enabled w/ `iced` due to large byte arrays getting printed to the terminal. This is due to the font bytes that are debugged, as well as the winit icon bytes that is debugged.

We log the `winit::WindowBuilder` directly, so I don't have access to customize how that is logged. This will require an upstream fix, if accepted. In the meantime, I think it's fair to log that struct as `debug` instead of `info`. We already log the iced window settings on `info` anyways.

I noticed there was already a fix for this on the glow side for default_font bytes. I've flipped this to `is_some` since it makes more sense the `default_font: true` means `default_font: is_some / Some`, meaning a default font was provided.


hecrj (approved):
Thanks! Good catch.

Hopefully one day something like this lands: https://github.com/rust-lang/rust/issues/37009

PR #1735, opened 2023-02-27
Remove `Fill` variant for `Alignment`

hecrj:
Implementing this generically in our `flex` logic has an exponential cost. Let's explore other options!

### DIFF
--- a/core/src/alignment.rs
+++ b/core/src/alignment.rs
@@ -11,9 +11,6 @@ pub enum Alignment {
 
     /// Align at the end of the axis.
     End,
-
-    /// Fill the entire axis.
-    Fill,
 }
 
 impl From<Horizontal> for Alignment {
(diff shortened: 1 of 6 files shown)


nicoburns:
This leads me to a good example of a layout that isn't achievable in a single pass: Even height columns where the height of all the columns should be determined by the content of the tallest column:

Screenshot 2023-02-28 at 20 31 08

hecrj:
It's certainly possible if we tackle this example directly:

1. Measure all the text.
2. Find maximum height.
3. Set each outer container to the maximum height.

We only need multi-pass if we want to solve a generic problem (i.e. support any kind of content, not just text).

fhilgers:
Sorry if this is a naive question, but how would I achieve the layout in the picture above with the current features?

PR #1738, opened 2023-02-28
Update to winit 0.28

nicoburns:
This updates Iced to work with a rebased version of Iced's winit fork to latest winit master (nominally 0.28)

This depends on:

- https://github.com/iced-rs/winit/pull/9/
- https://github.com/hecrj/window_clipboard/pull/20

This PR will need to be updated with the new dependency versions once those have been merged.

The only code change in this PR is due winit now having a `Window::set_window_level(WindowLevel)` method rather then `Window::set_always_on_top(bool)`. WindowLevel as defined as:

```rust
pub enum WindowLevel {
    AlwaysOnBottom,
    Normal,
    AlwaysOnTop,
}
```

This PR just maps the boolean from Iced's existing API to the correct value in the new enum. It could alternatively expose this new enum to Iced users if that's considered desirable.

### DIFF
--- a/core/src/window.rs
+++ b/core/src/window.rs
@@ -2,12 +2,14 @@
 pub mod icon;
 
 mod event;
+mod level;
 mod mode;
 mod redraw_request;
 mod user_attention;
 
 pub use event::Event;
 pub use icon::Icon;
+pub use level::Level;
 pub use mode::Mode;
 pub use redraw_request::RedrawRequest;
 pub use user_attention::UserAttention;
--- a/core/src/window/level.rs
+++ b/core/src/window/level.rs
@@ -1,0 +1,19 @@
+/// A window level groups windows with respect to their z-position.
+///
+/// The relative ordering between windows in different window levels is fixed.
+/// The z-order of a window within the same window level may change dynamically
+/// on user interaction.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum Level {
+    /// The default behavior.
+    #[default]
+    Normal,
+
+    /// The window will always be below normal windows.
+    ///
+    /// This is useful for a widget-based app.
+    AlwaysOnBottom,
+
+    /// The window will always be on top of normal windows.
+    AlwaysOnTop,
+}
--- a/runtime/src/window.rs
+++ b/runtime/src/window.rs
@@ -5,7 +5,7 @@ pub use action::Action;
 
 use crate::command::{self, Command};
 use crate::core::time::Instant;
-use crate::core::window::{Event, Icon, Mode, UserAttention};
+use crate::core::window::{Event, Icon, Level, Mode, UserAttention};
 use crate::futures::subscription::{self, Subscription};
 
 /// Subscribes to the frames of the window of the running application.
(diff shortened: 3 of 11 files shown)


BBaoVanC:
Doesn't seem to compile:

```
error[E0432]: unresolved import `winit::platform::unix`
   --> winit/src/settings.rs:170:36
    |
170 |             use ::winit::platform::unix::WindowBuilderExtUnix;
    |                                    ^^^^ could not find `unix` in `platform`

error[E0599]: no method named `with_name` found for struct `WindowBuilder` in the current scope
   --> winit/src/settings.rs:173:49
    |
173 |                 window_builder = window_builder.with_name(id.clone(), id);
    |                                                 ^^^^^^^^^
    |
    = help: items from traits can only be used if the trait is in scope
help: the following traits are implemented but not in scope; perhaps add a `use` for one of them:
    |
2   | use winit::platform::wayland::WindowBuilderExtWayland;
    |
2   | use winit::platform::x11::WindowBuilderExtX11;
    |
help: there is a method with a similar name
    |
173 |                 window_builder = window_builder.with_theme(id.clone(), id);
    |                                                 ~~~~~~~~~~
```

Is there something I need to change? I really need to use this because the old version of winit has some conflicting dependencies with another crate I have in my project.

nicoburns:
@hecrj Could I get CI approved on this? This builds on macOS, but clearly there are issues on other platforms.

@BBaoVanC The thing to do would be to look at the winit changelog between 0.27 and 0.28 and work out what needs to be updated here to make this compile (there's no trick here - the error shown needs to be fixed).

hecrj (commented):
We should wait until the `advanced-text` changes land. `iced_glutin` does not exist in that branch.

BBaoVanC:
For documentation, I made a branch on my fork to use as a dependency, and after merging both `advanced-text` and this PR, I think it's getting a lot closer:

```
error[E0277]: the trait bound `iced_graphics::Renderer<iced_wgpu::Backend, <A as application::Application>::Theme>: iced_winit::Renderer` is not satisfied
   --> /Users/bbaovanc/.local/share/cargo/registry/src/github.com-1ecc6299db9ec823/iced-0.8.0/src/application.rs:235:9
    |
235 | impl<A> crate::runtime::Application for Instance<A>
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `iced_winit::Renderer` is not implemented for `iced_graphics::Renderer<iced_wgpu::Backend, <A as application::Application>::Theme>`
    |
    = help: the following other types implement trait `iced_winit::Renderer`:
              iced_graphics::renderer::Renderer<B, T>
              iced_winit::renderer::Null
```

(error seems to happen on this PR regardless, and I don't get the other ones anymore, although maybe I changed something)

nicoburns:
Ok, this (when combined with https://github.com/iced-rs/winit/pull/9) is now passing CI and the `tour` example runs locally on my Mac (I am currently not able to test actually running on other platforms).

hecrj (approved):
Awesome! Looks good.

Thanks for taking care of this :partying_face:

PR #1742, opened 2023-03-02
[Fix] Fullscreen only works on primary monitor

bungoboingo:
This fixes a bug where `window::change_mode` Command was using only the primary monitor, not the current monitor. Also added a fullscreen/windowed hotkey to the `Todos` example


hecrj (approved):
Thanks!

Since we do not rely on `iced_web` anymore, we can expose the `window` commands for Wasm builds.

Issue #1066, opened 2021-09-29
Disabled state for widgets


sourcebox:
I want to have some buttons and a pick list grayed out and disabled depending on some application state. Ideally, a whole Row, Column or Container could be set to this state, disabling all widgets in it. Is this possible?

thenlevy:
Buttons can be grayed out by not assigning them an `on_press` message. When creating your button you could do something like 

```rust
let mut button = Button::new(/*... */) ;
if button_must_be_enabled {
   button = button.on_press(message)
}
```

As far as I know, `PickList`s cannot be disabled.

Also there is no method in a `Row` that allow you to disable all its buttons at once

theRookieCoder:
Could we include other types of widgets too, especially button-lilew widgets like radio buttons?

avsaase:
I would like to pick this up. I created a first PR for the checkbox widget (https://github.com/iced-rs/iced/pull/2109) and I also have a branch ready for the picklist widget.

https://github.com/iced-rs/iced/assets/880421/a11a563d-c37f-4342-8d40-418c30f2285f

jomaro110:
Still pick_list widget doesn't have a built-in disable mechanism?

PR #1744, opened 2023-03-03
Add functionality to make `TextInput` disabled 

JungleTryne:
People have been asking to add disabled state for several widgets in #1066. The most liked comment was about adding the state to the `TextEdit` widget. Indeed, it's quite useful to have functionalty to prevent user from changing the content of the text edit fields.

I added new logic and a theme for the widget to implement the functionality. 

Now text edit can be enabled/disabled like this:

```Rust
text_input(
  "Write something here",
  &self.data,
  Message::TextInputChanged,
).disabled(true);
```


JungleTryne:
> I would rather this be like the button. Have the text input by default be disabled like a button, but have an `on_change(MESSAGE)` method.

But that requires changing the API, which might break back-compatibility. It's not a problem to rewrite it, but I would like to know if it is okay to actually implement breaking changes at this point

JungleTryne:
I changed API, now it works like a button. It is not enabled if there is no callback for `on_change`.

I also change the example, I added an async 3 seconds timer to test an edge-case when the field is focused and gets disabled, suggested in discord by nicoburns#9458.

Finally, I updated all examples for the new API

JungleTryne:
It's been a week and this pull request didn't receive any review :(
Bump

hecrj (approved):
Thank your taking a shot at this! :bow: 

I have made some important changes:

- Renamed `on_change` to `on_input` in e6a93e960c3ac71a74f11555fd2d225c185f6e8c. I believe it conveys more intent.
- Added some diffing logic to unfocus the input if it suddenly becomes disabled in 7e69cb4b18cde29eceb57e3712c0630d279e7984.
- Fine-tuned the disabled styling a bit in 1de794aabfaa651f021ff2e5831397d7ce9fc53c.
- Changed the `mouse_interaction` to display a new `NotAllowed` mouse icon when hovering a disabled input in 7e7e66586d990788ffd77b17e98357e74252f497.

And that's it! Let's merge!

PR #1747, opened 2023-03-05
Require and implement debug on Stylesheet types

13r0ck:
Simplifies my use case where I have a type that holds one of the upstream iced theme types. ( currently `iced_style::theme::Button` )


13r0ck:
This is not longer necessary for me.

PR #1749, opened 2023-03-08
Configurable `width` for `Scrollable`

### LINKED ISSUE #1704: Scrollable supports `height` but not `width`

### What happened?

In attempting to size a `Scrollable`, the `height` method is shown as implemented, however the `width` method results in the following error:
```rs
error[E0599]: no method named `width` found for struct `iced_native::widget::scrollable::Scrollable` in the current scope
   --> src\main.rs:68:18
    |
68  |                 .width(Length::Fill)
    |                  ^^^^^ method not found in `iced_native::widget::scrollable::Scrollable<'_, _, _>`
    |
   ::: C:\Users\user\.cargo\registry\src\github.com-1ecc6299db9ec823\iced_native-0.8.0\src\widget.rs:124:8
    |
124 |     fn width(&self) -> Length;
    |        ----- the method is available for `iced_native::widget::scrollable::Scrollable<'_, _, _>` here
    |
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
1   | use iced_native::widget::Widget;
    |
```

Width to be implemented the same as height, given they are from the same `Widget` impl.

### Version

0.7

### Operative System

Windows

_No response_

FinnPerry:
Fix #1704.

Add width to scrollables so you can do stuff like `.width(Length::Fill)` to make the scroll bar appear on the edge of the screen, previously as far as I'm aware you had to set the width of the scrollables child element instead.

I've tested by running the scrollable example project and my own project which uses a scrollable and it appears to work correctly with no unexpected side effects but I'm not sure if there is a more rigorous way to test these changes.

### DIFF
--- a/native/src/widget/scrollable.rs
+++ b/native/src/widget/scrollable.rs
@@ -405,15 +405,7 @@ pub fn layout<Renderer>(
     horizontal_enabled: bool,
     layout_content: impl FnOnce(&Renderer, &layout::Limits) -> layout::Node,
 ) -> layout::Node {
-    let limits = limits
-        .max_height(f32::INFINITY)
-        .max_width(if horizontal_enabled {
-            f32::INFINITY
-        } else {
-            limits.max().width
-        })
-        .width(width)
-        .height(height);
+    let limits = limits.width(width).height(height);
 
     let child_limits = layout::Limits::new(
         Size::new(limits.min().width, 0.0),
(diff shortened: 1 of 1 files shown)


hecrj:
I believe we should strive for widgets to have very specialized purposes and have fewer responsibilities in the long run.

If this behavior can already be achieved with a `container` with `width(Length::Fill)` as the child of the `scrollable`, then let's keep it that way! No need to have many different ways to achieve the same thing.

hecrj:
I forgot that scrollables now can scroll horizontally as well! A `width` attribute makes sense then.

hecrj (approved):
Looks great! Thanks :bow:

PR #1762, opened 2023-03-17
Radio example

ahoneybun:
Hello! This PR adds an example for using radio buttons which I found a bit difficult to understand from the tour example. Thanks to @13r0ck for the help with understanding it to make this example.


hecrj (approved):
Thanks! I have moved the example to the `Radio` docs since I feel it's a better place for simpler examples like this one.

PR #1765, opened 2023-03-19
winit: Fix replacement of node in wasm

traxys:
Replacing a node ends up with the following error: Node.replaceChild: Child to be replaced is not a child of this node

It seems that Node.replaceChild is not recommended, and instead Element.replaceWith should be preferred. Using it avoids the panic.


hecrj (approved):
Good catch! Thanks.

PR #1777, opened 2023-04-01
Re-expose winit features for window servers in Linux

Xaeroxe:
`winit` allows users to opt out of x11 or wayland in their applications via cargo features. This PR propagates that choice through `iced_winit`.

The new default features do nothing on non-Linux platforms.


hecrj (approved):
Seems reasonable! Thanks :bow:

PR #1779, opened 2023-04-01
remove colons from shader labels

Davidster:
when using the DXC compiler feature from wgpu, iced's shaders fail to compile due to a weird bug I discovered that causes something to go wrong if the shader's label contains any colons. So this PR removed the colons (they can alternatively be escaped if you prefer that, like `"iced\\:\\:image"`. I checked it and it works).

### DIFF
--- a/wgpu/src/image.rs
+++ b/wgpu/src/image.rs
@@ -141,7 +141,7 @@ impl Pipeline {
 
         let shader =
             device.create_shader_module(wgpu::ShaderModuleDescriptor {
-                label: Some("iced_wgpu::image::shader"),
+                label: Some("iced_wgpu image shader"),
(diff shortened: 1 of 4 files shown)


hecrj (commented):
I feel this should be automatically escaped by `wgpu`. Has this issue been reported there?

Davidster:
I mentioned it in Element, but from what I saw it's a limitation of their dxc dependency hassle, or potentially even the directX compiler itself. I'll take a closer look to try to pin down the right spot where the change should be made, just thought I'd open the PR since I'm now depending on an iced fork with this change.

hecrj:
Sure! We can merge this. Just pointing out that `wgpu` should probably be escaping these if they can cause issues.

hecrj (approved):
Thanks! :partying_face:

PR #1782, opened 2023-04-05
Updating Roadmap.md

bq-wrongway:
I collapsed Animations because the underlying issue was resolved, and i added issue #343 to 3D Canvas widget.


hecrj (approved):
Thanks!

The `ROADMAP` needs almost whole rewrite, but a nice change in the meanwhile :bow:

PR #1783, opened 2023-04-06
Feat: Offscreen Rendering & Screenshots

bungoboingo:
## This PR adds support for offscreen rendering in Iced! 🎉 

https://user-images.githubusercontent.com/4241774/230238292-03b8d796-c8cb-4870-adc1-3dbac36c3a8d.mov

### 📸 Users can now capture screenshots 🖼️ using the `window::screenshot()` command.

```rust
    match message {
        Message::Screenshot => {
            return iced::window::screenshot(Message::ScreenshotData); //new!
        }
        //...
```

The `window::screenshot` command requires a `Message` which takes a `Screenshot`. This is a new struct exposed in the `iced_runtime` crate which contains the RGBA bytes of the screenshot, in addition to its size.

For now this command is limited to the entire viewport, with an optional `crop()` method available which will crop the bytes to the specified `Rectangle`.

https://user-images.githubusercontent.com/4241774/230238237-a9b3ddfa-cb79-463f-85d9-18621585dd79.mov

```rust
let cropped = screenshot.crop(Rectangle::<u32> {
    x: 0,
    y: 0,
    width: 500,
    height: 500,
});
```

This method returns a `Result<Screenshot, CropError>` for situations where the cropped region is out of bounds, or is not visible.

### 🎉 Offscreen rendering capabilities are now added to both `wgpu` and `tiny_skia`!

Rendering offscreen is now exposed in the `Compositor` interface with a `render_offscreen` function. The outputted texture data is always guaranteed to be in RGBA format in the SRGB color space for compatibility with image libraries & iced images, and to simplify the `Screenshot` struct with a predetermined byte order & pixel size.

### 🔢 Implementation details

For `wgpu`, performing offscreen rendering was fairly straightforward. I create a texture based on the viewport size & the chosen texture format, and then pass that into the backend with a new `backend.offscreen` method. This method performs a normal render pass on the new texture, and optionally runs a tiny compute shader after drawing all primitives that converts it to `wgpu::TextureFormat::Rgba8Unorm` with a manual srgb conversion due to `wgpu::TextureFormat::Rgba8UnormSrgb` not being a supported format for a storage texture. The whole process on my (admittedly very powerful!) test suite of GPUs (AMD, NVIDIA, M1) took about ~2µs in the test example. The only added overhead over a normal render pass is the compute shader to convert to rgba.

For `tiny-skia` this was _very_ straightforward, the same as a normal `present` call but rendering to a new buffer vs the surface buffer. Our backend returns `ARGB` format, so just had to do some bit shiftin' and we're all good.

### 🤔 Outstanding API questions

1. It was discussed whether it's a cleaner API to do a `window::screenshot` command which optionally can be cropped, or a `window::screenshot(area)` command which can either be `Fullscreen` or `Region`, which would be a rectangle. Obviously there is a performance hit when rendering the whole window to an offscreen buffer vs a potentially very small region if only a small portion is desired, but the execution is so fast it's almost negligible. I'm curious as to people's thoughts on this!
 
 2. The offscreen render capabilities are only exposed at the moment through this `window::screenshot` command. The intent was to implement the capabilities for offscreen rendering in this PR and expand upon it further in later iterations if new applications of offscreen rendering are needed. Curious to see if anyone has any thoughts on how else this might be exposed!
 
 ### 🧪 Testing
Tested wgpu & tiny-skia implementations on MacOS + M1, Linux + PopOS + AMD, Windows 10 + Nvidia combos. Further testing would be greatly appreciated!

Feedback very much welcome! This was my first time working this deeply with textures on the GPU so might have made some rookie mistakes 😉 I also know that there is an open draft PR for offscreen rendering but it hasn't been updated in a few years so thought I'd take a crack at it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,11 +41,6 @@ palette = ["iced_core/palette"]
 system = ["iced_winit/system"]
 # Enables the advanced module
 advanced = []
-# Enables chrome tracing support for Iced
-chrome-trace = [
-    "iced_winit/chrome-trace",
-    "iced_renderer/tracing",
-]
 
 [badges]
 maintenance = { status = "actively-developed" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,11 @@ palette = ["iced_core/palette"]
 system = ["iced_winit/system"]
 # Enables the advanced module
 advanced = []
+# Enables chrome tracing support for Iced
+chrome-trace = [
+    "iced_winit/chrome-trace",
+    "iced_renderer/tracing",
+]
 
 [badges]
 maintenance = { status = "actively-developed" }
--- a/examples/screenshot/Cargo.toml
+++ b/examples/screenshot/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "screenshot"
+version = "0.1.0"
+authors = ["Bingus <shankern@protonmail.com>"]
+edition = "2021"
+publish = false
+
+[dependencies]
+iced = { path = "../..", features = ["debug", "image", "advanced"] }
+image = { version = "0.24.6", features = ["png"]}
+env_logger = "0.10.0"
+wasm-bindgen = "0.2"
--- a/examples/screenshot/Cargo.toml
+++ b/examples/screenshot/Cargo.toml
@@ -9,4 +9,3 @@ publish = false
 iced = { path = "../..", features = ["debug", "image", "advanced"] }
 image = { version = "0.24.6", features = ["png"]}
 env_logger = "0.10.0"
-wasm-bindgen = "0.2"
--- a/graphics/src/compositor.rs
+++ b/graphics/src/compositor.rs
@@ -59,6 +59,18 @@ pub trait Compositor: Sized {
         background_color: Color,
         overlay: &[T],
     ) -> Result<(), SurfaceError>;
+
+    /// Renders the current [`Renderer`] primitives to an offscreen texture, and returns the bytes of
+    /// the texture.
+    ///
+    /// [`Renderer`]: Self::Renderer;
+    fn render_offscreen<T: AsRef<str>>(
+        &mut self,
+        renderer: &mut Self::Renderer,
+        viewport: &Viewport,
+        background_color: Color,
+        overlay: &[T],
+    ) -> Vec<u8>;
 }
 
 /// Result of an unsuccessful call to [`Compositor::present`].
@@ -82,7 +94,7 @@ pub enum SurfaceError {
     OutOfMemory,
 }
 
-/// Contains informations about the graphics (e.g. graphics adapter, graphics backend).
+/// Contains information about the graphics (e.g. graphics adapter, graphics backend).
 #[derive(Debug)]
 pub struct Information {
     /// Contains the graphics adapter.
--- a/graphics/src/compositor.rs
+++ b/graphics/src/compositor.rs
@@ -60,11 +60,11 @@ pub trait Compositor: Sized {
         overlay: &[T],
     ) -> Result<(), SurfaceError>;
 
-    /// Renders the current [`Renderer`] primitives to an offscreen texture, and returns the bytes of
-    /// the texture.
+    /// Screenshots the current [`Renderer`] primitives to an offscreen texture, and returns the bytes of
+    /// the texture ordered as `RGBA` in the sRGB color space.
     ///
     /// [`Renderer`]: Self::Renderer;
-    fn render_offscreen<T: AsRef<str>>(
+    fn screenshot<T: AsRef<str>>(
         &mut self,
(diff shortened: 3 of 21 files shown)


graphics/src/compositor.rs
```diff
@@ -59,6 +59,18 @@ pub trait Compositor: Sized {
...
+    /// Renders the current [`Renderer`] primitives to an offscreen texture, and returns the bytes of
+    /// the texture.
+    ///
+    /// [`Renderer`]: Self::Renderer;
+    fn render_offscreen<T: AsRef<str>>(
+        &mut self,
+        renderer: &mut Self::Renderer,
+        viewport: &Viewport,
+        background_color: Color,
+        overlay: &[T],
```
tarkah, line 72: Do we care about rendering the debug overlay via this?

tarkah: Yeah I think it makes sense. It's as simple as toggling it off if you don't want it included

tarkah:
> Lint wants me to change `0xFF_000000` to `0xFF00_0000` but imo it reads less clearly and is a silly lint. Propose to ignore `unusual-byte-ordering`?

How about `0b11111111_00000000_00000000_00000000` :P

Cargo.toml
```diff
@@ -41,6 +41,11 @@ palette = ["iced_core/palette"]
 system = ["iced_winit/system"]
 # Enables the advanced module
 advanced = []
+# Enables chrome tracing support for Iced
+chrome-trace = [
+    "iced_winit/chrome-trace",
+    "iced_renderer/tracing",
+]
```
hecrj, line 48: Let's remove the tracing for now.

graphics/src/compositor.rs
```diff
@@ -59,6 +59,18 @@ pub trait Compositor: Sized {
         background_color: Color,
         overlay: &[T],
     ) -> Result<(), SurfaceError>;
+
+    /// Renders the current [`Renderer`] primitives to an offscreen texture, and returns the bytes of
+    /// the texture.
+    ///
+    /// [`Renderer`]: Self::Renderer;
+    fn render_offscreen<T: AsRef<str>>(
```
hecrj, line 67: We can just call it `screenshot`!

wgpu/src/backend.rs
```diff
@@ -122,6 +124,99 @@ impl Backend {
...
+            target_size,
+            &layers,
+        );
+
+        self.quad_pipeline.end_frame();
+        self.text_pipeline.end_frame();
+        self.triangle_pipeline.end_frame();
+
+        #[cfg(any(feature = "image", feature = "svg"))]
+        self.image_pipeline.end_frame();
```
hecrj, line 187: Can we just call `present`?

runtime/src/screenshot.rs
```diff
@@ -0,0 +1,78 @@
+use iced_core::{Rectangle, Size};
+use std::fmt::{Debug, Formatter};
+
+/// Data of a screenshot, captured with `window::screenshot()`.
+///
+/// This screenshot will always be in RGBA format.
```
hecrj, line 6: Probably important to specify it's sRGBA or RGBA in the sRGB color space.

examples/screenshot/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+authors = ["Bingus <shankern@protonmail.com>"]
+edition = "2021"
+publish = false
+
+[dependencies]
+iced = { path = "../..", features = ["debug", "image", "advanced"] }
+image = { version = "0.24.6", features = ["png"]}
+env_logger = "0.10.0"
+wasm-bindgen = "0.2"
```
hecrj, line 12: `wasm-bindgen` doesn't seem to be used.

hecrj: Nit: File is missing a newline at the end.

wgpu/src/backend.rs
```diff
@@ -24,10 +24,10 @@ pub struct Backend {
     quad_pipeline: quad::Pipeline,
     text_pipeline: text::Pipeline,
     triangle_pipeline: triangle::Pipeline,
+    offscreen_pipeline: offscreen::Pipeline,
```
hecrj, line 27: Could we create the `Pipeline` locally in the `screenshot` method only if necessary? How much of an overhead would that be?

hecrj (changes requested):
Looking good! Some suggestions here and there :monocle_face: 

Love your detailed PRs! :100:

hecrj:
> Lint wants me to change 0xFF_000000 to 0xFF00_0000 but imo it reads less clearly and is a silly lint. Propose to ignore unusual-byte-ordering?

Quite subjective, in my opinion. I think it's fairly standard to group evenly. Does `0xFF_00_00_00` work?

PR #1788, opened 2023-04-12
Don't publish redundant `on_scroll` offsets

tarkah:
This helps improve scrollable performance by suppressing redundant messages from getting published when the `RelativeOffset` hasn't changed.

I noticed the scrolling is very sluggish on debug mode, even with few elements. When quickly scrolling, _dozens_ of redundant messages were coming across `update` causing unnecessary state changes / re-renders. 

This improves the snappiness of scrollables in debug mode a lot.

### DIFF
--- a/native/src/widget/scrollable.rs
+++ b/native/src/widget/scrollable.rs
@@ -895,7 +895,7 @@ pub fn draw<Renderer>(
 }
 
 fn notify_on_scroll<Message>(
-    state: &State,
+    state: &mut State,
     on_scroll: &Option<Box<dyn Fn(RelativeOffset) -> Message + '_>>,
     bounds: Rectangle,
     content_bounds: Rectangle,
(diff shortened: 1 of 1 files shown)


native/src/widget/scrollable.rs
```diff
@@ -510,7 +510,9 @@ pub fn update<Message>(
                 state.scroll(delta, bounds, content_bounds);
 
                 notify_on_scroll(
-                    state,
+                    state.offset_x,
+                    state.offset_y,
+                    &mut state.last_notified,
```
hecrj, line 515: Could we pass `&mut State` instead? Harder to pass the wrong arguments that way.

hecrj (changes requested):

native/src/widget/scrollable.rs
tarkah: Yup :+1:

hecrj (approved):
Awesome! Thanks :tada:

PR #1789, opened 2023-04-12
Update `wgpu` to `0.15`

ElhamAryanpur:
Hey there!

This is my first time contributing to this amazing project! I was actually trying to add iced as a plugin to my [engine][https://github.com/AryanpurTech/BlueEngine] and I noticed that the wgpu was a version behind, even though it's been a while since it was released, so thought to bump the versions and update the code related to them.

Please review and let me know if it's ok or if I need to update some parts,
thanks,

### DIFF
--- a/examples/integration_wgpu/src/main.rs
+++ b/examples/integration_wgpu/src/main.rs
@@ -33,7 +33,8 @@ pub fn main() -> Result<(), Box<dyn std::error::Error>> {
         web_sys::window()
             .and_then(|win| win.document())
             .and_then(|doc| doc.get_element_by_id("iced_canvas"))
-            .and_then(|element| element.dyn_into::<HtmlCanvasElement>().ok())?
+            .and_then(|element| element.dyn_into::<HtmlCanvasElement>().ok())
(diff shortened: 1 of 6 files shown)


examples/integration_wgpu/src/main.rs
```diff
@@ -91,11 +93,16 @@ pub fn main() {
...
+        let capabilities = surface.get_capabilities(&adapter);
+
         (
-            surface
-                .get_supported_formats(&adapter)
-                .first()
+            capabilities
+                .formats
+                .iter()
+                .filter(|format| format.describe().srgb)
```
hecrj, line 102: This is cool! Hopefully it fixes inaccurate colors for some!

hecrj (approved):
Thank you! :tada: 

I changed some things here and there, but everything looks good to go!

examples/integration_wgpu/src/main.rs
ElhamAryanpur: Haha yeah the API is improving a lot, I was surprised too! Thank you.

PR #1791, opened 2023-04-13
Add missing operations for component

tarkah:
We're missing a couple mapped operations for component


hecrj (approved):
Whoops! Thanks. Will release a patch!

PR #1796, opened 2023-04-14
Add `scroll_to` operation

tarkah:
Supersedes #1795 

Exposes a new `scroll_to` operation to allow scrolling a scrollable to a desired "absolute" offset.

### DIFF
--- a/native/src/widget/operation/scrollable.rs
+++ b/native/src/widget/operation/scrollable.rs
@@ -64,15 +64,6 @@ pub fn scroll_to<T>(target: Id, offset: AbsoluteOffset) -> impl Operation<T> {
     ScrollTo { target, offset }
 }
 
-/// The current absolute & relative offset of a [`Scrollable`]
-#[derive(Debug, Clone, Copy, PartialEq, Default)]
-pub struct CurrentOffset {
-    /// The [`AbsoluteOffset`] of a [`Scrollable`]
-    pub absolute: AbsoluteOffset,
-    /// The [`RelativeOffset`] of a [`Scrollable`]
-    pub relative: RelativeOffset,
-}
-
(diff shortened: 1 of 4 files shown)


native/src/widget/operation/scrollable.rs
```diff
@@ -34,7 +37,52 @@ pub fn snap_to<T>(target: Id, offset: RelativeOffset) -> impl Operation<T> {
...
+}
+
+/// The current absolute & relative offset of a [`Scrollable`]
+#[derive(Debug, Clone, Copy, PartialEq, Default)]
+pub struct CurrentOffset {
+    /// The [`AbsoluteOffset`] of a [`Scrollable`]
+    pub absolute: AbsoluteOffset,
+    /// The [`RelativeOffset`] of a [`Scrollable`]
+    pub relative: RelativeOffset,
+}
```
hecrj, line 74: There's some redundancy here.

I think it'd be better to have a `Window` or `Viewport` type with the already existing private `Offset`, the content bounds, and the scrollable bounds.

Then we can expose `relative_offset` and `absolute_offset` methods in this type that compute the actual values from a single source of truth.

hecrj (changes requested):

native/src/widget/operation/scrollable.rs
tarkah: Great suggestion  6ad5e03d]

hecrj (approved):
Awesome!

I have made a change to compare the absolute offsets as well in `notify_on_scroll`, since it's theoretically possible that the relative offset stays the same but the absolute offset doesn't (e.g. if the bounds of the `Scrollable` change while scrolling). See 8a711408de26c53cf6b18045ea19c677e8edbd95.

I think we can merge! :ship:

PR #1797, opened 2023-04-14
Make mouse::Button::Other take u16 instead of u8

bbb651:
On wayland keys correspond to <input-event-codes.h>, and they are past the limit of u8, causing the
back and forward buttons to be 20 and 19 which definitely isn't right (they should all be around 0x110..=0x117).

This is a breaking change but will probably only affect apps on wayland relying on these wrong codes,
I also have [a winit pr that will also necessitate a breaking change to this enum](https://github.com/rust-windowing/winit/pull/2770),
maybe this should wait so both of these should be done at the same time.

### DIFF
--- a/core/src/mouse/button.rs
+++ b/core/src/mouse/button.rs
@@ -11,5 +11,5 @@ pub enum Button {
     Middle,
 
     /// Some other button.
-    Other(u8),
+    Other(u16),
 }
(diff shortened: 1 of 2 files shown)


i509VCB:
Technically for Wayland the `Other` variant would need to be a full u32.

hecrj (approved):
Thanks!

PR #1799, opened 2023-04-18
Invalidate responsive layout when shell layout is invalidated

tarkah:
`Responsive` layout was only being recomputed if `size` changed. This caused a panic if a downstream widget changed between `layout` calls of the same `view` / User Interface root element. This is technically possible with `Component` and panic's can be reliably recreated by nesting a `Component` inside a `Responsive` and returning totally different elements from it's `view` after processing updates.

### DIFF
--- a/lazy/src/responsive.rs
+++ b/lazy/src/responsive.rs
@@ -58,7 +58,7 @@ impl<'a, Message, Renderer> Content<'a, Message, Renderer>
 where
     Renderer: iced_native::Renderer,
 {
-    fn layout(&mut self, renderer: &Renderer) -> layout::Node {
+    fn layout(&mut self, renderer: &Renderer) {
         if self.layout.is_none() {
             self.layout =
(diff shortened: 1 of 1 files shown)


lazy/src/responsive.rs
```diff
@@ -191,10 +200,18 @@ where
...
                     clipboard,
-                    shell,
+                    &mut local_shell,
                 )
             },
-        )
+        );
+
+        if local_shell.is_layout_invalid() {
+            let _ = content.layout.take();
```
hecrj, line 209:
```suggestion
            content.layout = None;
```

hecrj (changes requested):

hecrj (approved):
Looks good! Thanks :bow: 

I made a small change that allows us to avoid cloning the `layout::Node` altogether in ecc1a4509ab25611397386c9f9c85fe7b269fda3.

PR #1803, opened 2023-04-20
Add env var to default to software renderer

ids1024:
This allows a user for whom wgpu initializes but doesn't work properly to switch to the software renderer without having to re-compile with different features.

This could be done in a more complicated way with a `;` separated list of renderer backends in the order to try them, but there's no need for that when there are only two options.

### DIFF
--- a/renderer/src/compositor.rs
+++ b/renderer/src/compositor.rs
@@ -1,3 +1,5 @@
+use std::env;
+
 use crate::core::Color;
 use crate::graphics::compositor::{Information, SurfaceError};
 use crate::graphics::{Error, Viewport};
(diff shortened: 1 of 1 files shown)


hecrj:
Thanks!

I have already introduced an `ICED_BACKEND` environment variable in the `incremental-rendering` branch that should land eventually. See https://github.com/iced-rs/iced/commit/42b2e9b007d49506c98d0242b8b061df3aa4228b.

ids1024:
Ah, I was just about to look at the implementation of incremental rendering, but didn't think to check if this was implemented there. Seems good.

PR #1811, opened 2023-04-26
Incremental rendering

hecrj:
This PR implements basic incremental rendering for the upcoming software renderer powered by `tiny-skia`.

Basically, we compare the graphical primitives of the last drawn frame with the current primitives and produce a list of rectangles representing the damaged regions of the frame. Then, this list is reduced by grouping rectangles in a certain distance threshold together. Finally, when drawing, we simply skip any primitive that does not intersect with any damaged region.

The following video shows the `tour` example but old primitives fade with time and the red rectangles are the grouped damage regions:

https://user-images.githubusercontent.com/518289/234604847-e9fc9c05-e7ca-4a4a-82c8-7b844bd3e6d3.mp4

The diffing logic of graphical primitives is generic enough and could be eventually reused in `iced_wgpu`.


tiny_skia/src/backend.rs
```diff
@@ -31,57 +34,127 @@ impl Backend {
 
             #[cfg(feature = "svg")]
             vector_pipeline: crate::vector::Pipeline::new(),
+
+            last_primitives: Vec::new(),
```
i509VCB, line 38: If a single Backend were able to be used multiple times (for multiple windows if that was intended) you would end up with a situation where the second window is fully redrawn because the primitives may be completely different on that other window.

ids1024:
https://github.com/rust-windowing/softbuffer/pull/99 now has APIs for buffer age and damage (where supported), so incremental rendering can be done with no-copy presentation. On some platforms, anyway.

On Wayland this uses double-buffering, so `age` can be `None`, `1`, or `2`, and you need to be able to update the parts of the surface that have changed in the past two frames. On MacOS currently there's a new buffer time so `age` is always `None`. (https://github.com/rust-windowing/softbuffer/pull/95 or https://github.com/rust-windowing/softbuffer/pull/96 would fix that). Other platforms currently use a single buffer so age is `None` or `1`.

For Wgpu, it looks like there's no API for damage or anything that exposes the buffer age, or a way to update only part of it? https://github.com/gfx-rs/wgpu/issues/2869 discusses damage, I don't see any mention of a feature like `EGL_EXT_buffer_age`.

hecrj:
@ids1024 Cool! It looks like all of those features are still being worked on. Let's wait a bit until everything settles!

PR #1814, opened 2023-04-27
Do not require `Copy` for text `StyleSheet::Style`

ids1024:
For most widgets, `Style` only requires `Default`. A few require `Clone`. Only this one requires `Copy`.

Some of the types in the default theme has a custom variant requiring `Box<dyn Trait>`, or `Rc<dyn Trait>` to provide `Clone`, but this isn't possible if `Copy` is required.

It would be good to also address the inconsistency of requiring `Clone` in some places and not others.

This removes `style/src/text.rs` which is unused in this branch and thus confusing. If there's a reason to keep it, that can be removed from the change.


style/src/theme.rs
```diff
@@ -980,7 +980,7 @@ impl scrollable::StyleSheet for Theme {
 }
 
 /// The style of text.
-#[derive(Clone, Copy, Default)]
+#[derive(Clone, Default)]
```
hecrj, line 983: No reason for this to not tbe `Copy` anymore.

hecrj (changes requested):

hecrj (approved):
Thanks!

The current styling system is something I'm still not completely convinced with. That's why some parts may be inconsistent. I'll do a once-over before the next release.

PR #1822, opened 2023-05-01
`text::Shaping` strategy selection

hecrj:
This PR introduces a new `Shaping` enum in the `text` module that allows users to control the shaping strategy used to lay out text.

The `Basic` strategy is considerably more efficient (specially in debug builds!) and works well for simple scripts (the use cases the library currently satisfies).

Because of this, `Basic` is the default strategy used in text-based widgets (except the `TextInput`). In other words, `Advanced` shaping must be explicitly enabled when necessary.


AustinMReppert:
As some one who knows nothing about fonts, I feel like this could result in some problems:
* Looks fine in dev/test, but user enters text requiring advanced shaping and app re-displays said text incorrectly
* App dev localizes their app's text and ships something that appears fine to them. 

I wonder if there is any room for warnings in debug mode to detect text that requires advanced shaping.

hecrj:
We will get there eventually.

Developers need to know what they are doing for now.

MoSal:
Would making the default `cfg`-able with a `default_advanced_shaping` feature be a bad idea?

Also, wouldn't it be better if the enum was marked as `#[non_exhaustive]`, with the assumption that `Auto` or some other shaping impl/strategy may get implemented in the future?

pomoke:
Simply make advanced shaping default, and this is how a lot of GUI frameworks work. i18n will heavily depend on fallback and shaping.

hecrj:
Eventually, maybe.

Issue #1810, opened 2023-04-23
Application freezes when hidden using wayland


sevenautumns:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

Under Hyprland (wayland), hidden windows are not drawn. This probably stalls the rendering and makes iced applications unusable as the applications update function is not called until the rendering finished. This happens with `glow` and `wgpu`. The provided log is from `wgpu` after the window was hidden. As it can be seen, "Presented. End of Frame" does not appear when the window is hidden.

### What is the expected behavior?

The applications stays usable because the `update` function is still regularly called with new message. 

### Version

master

### Operative System

Linux

### Do you have any log output?

```shell
DEBUG wgpu_core::present             > Presented. End of Frame
 DEBUG wgpu_core::device              > Buffer (161, 1, Vulkan) map state -> Waiting
 DEBUG wgpu_core::device              > Buffer (13, 1, Vulkan) map state -> Waiting
 DEBUG wgpu_core::device              > texture view (2, 291, Vulkan) is dropped
 DEBUG wgpu_core::device              > Create view for texture (1, 583, Vulkan) filters usages to COLOR_TARGET
 DEBUG iced_wgpu::backend             > Drawing
 DEBUG wgpu_core::device              > Buffer (159, 1, Vulkan) map state -> Idle
 DEBUG wgpu_core::device              > Buffer (156, 1, Vulkan) map state -> Idle
 DEBUG wgpu_core::device::life        > Texture view Valid((2, 291, Vulkan)) will be destroyed
 DEBUG wgpu_core::device::life        > Buffer Valid((161, 1, Vulkan)) map state -> Active
 DEBUG wgpu_core::device::life        > Buffer Valid((13, 1, Vulkan)) map state -> Active
 DEBUG wgpu_core::present             > Removing swapchain texture Valid((1, 583, Vulkan)) from the device tracker
```

ids1024:
Mesa's Wayland EGL implementation blocks on a `frame` callback in `eglSwapBuffers` if `eglSwapInterval` is set to `1` (basically, if vsync is enabled). The Vulkan WSI implementation should be similar. The frame callback isn't sent to occluded windows.

See https://gitlab.freedesktop.org/mesa/mesa/-/issues/4932 for some discussion about software behavior this has impacted.

[...]

PR #1823, opened 2023-05-02
Allow user defined renderer::Settings

sevenautumns:
This PR adds a `renderer` field to the `Settings` struct for overriding `renderer::Settings`.

When this field is used, `renderer::Settings` other than what is already defined in the `Settings` struct (e.g. text size, anti-aliasing, ..) are used.

This helps with #1810 because wpgu `display_mode` can be overwritten.

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -203,7 +203,9 @@ pub trait Application: Sized {
             } else {
                 None
             },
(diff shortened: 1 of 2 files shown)


hecrj:
We should avoid exposing internal details of the renderers in the root crate.

PR #1828, opened 2023-05-04
Support configurable `LineHeight` in text widgets

hecrj:
This PR adds a new configurable `line_height` attribute to text widgets, replacing the previous hardcoded value (`1.2`).

Line height can be set using the new `LineHeight` enum, which has `Relative(f32)` and `Absolute(Pixels)` variants.


core/src/text.rs
```diff
@@ -56,6 +60,59 @@ pub enum Shaping {
...
 
+/// The height of a line of text in a paragraph.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum LineHeight {
+    /// A factor of the size of the text.
+    Relative(f32),
+
+    /// An absolute height in logical pixels.
+    Absolute(Pixels),
+}
```
clarkmoody, line 71: 👌

clarkmoody (approved):
Great change. No red flags from a quick inspection.

PR #1838, opened 2023-05-09
Add Widget related base iterator functions 

my4ng:
This PR adds base functions that take iterators rather than slices as arguments to the following functions related to `Widget`:
- `layout::flex::resolve`
- `overlay::from_children`
- `widget::tree::diff_children`

Currently, these functions are unnecessarily restrictive in the sense that they only take slice references, as most non-vector/array data structures in a widget field would be expensive to convert to, e.g. `HashMap`, multiple vectors, optional fields, etc. (One such example can be found in my [table widget fork](https://github.com/my4ng/iced/tree/table-widget) where the widget contains an optional header and vector content rows fields.) Furthermore, these functions do **not** utilise any special property of a slice apart from the `len` method, since _most_ of the implementations are iterator-based anyway.

Therefore, the following base functions take a more general approach with iterator arguments while keeping the original versions as wrappers to avoid any breaking changes, for example:
```rust
pub fn from_children<'a, Message, Renderer>(
    children: &'a mut [crate::Element<'_, Message, Renderer>],
    tree: &'a mut Tree,
    layout: Layout<'_>,
    renderer: &Renderer,
) -> Option<Element<'a, Message, Renderer>>
where
    Renderer: crate::Renderer,
{...}
```
Becomes
```rust
pub fn from_children<'a, Message, Renderer>(
    children: &'a mut [crate::Element<'_, Message, Renderer>],
    tree: &'a mut Tree,
    layout: Layout<'_>,
    renderer: &Renderer,
) -> Option<Element<'a, Message, Renderer>>
where
    Renderer: crate::Renderer,
{
    from_children_iter(children.iter_mut(), tree, layout, renderer)
}

pub fn from_children_iter<'a: 'b, 'b, Message: 'b, Renderer: 'b>(
    children: impl IntoIterator<
        Item = &'b mut (impl std::borrow::BorrowMut<
            dyn crate::Widget<Message, Renderer> + 'a,
        > + 'b),
    >,
    tree: &'b mut Tree,
    layout: Layout<'_>,
    renderer: &Renderer,
) -> Option<Element<'b, Message, Renderer>>
where
    Renderer: crate::Renderer,
{...}
```
In order to facilitate this, further `Borrow` and `BorrowMut` trait implementations are added to `Element`, e.g. :
```rust
impl<'a, Message, Renderer> BorrowMut<dyn Widget<Message, Renderer> + 'a>
    for Element<'a, Message, Renderer>
{
    fn borrow_mut(&mut self) -> &mut (dyn Widget<Message, Renderer> + 'a) {
        self.widget.borrow_mut()
    }
}
```
In places where the length is required, a `*_len` argument is supplied instead. 

Currently, the base functions have a `pub(crate)` visibility modifier, so they are not exposed to outside the crate, but of course that can easily be the subject of any further changes. 

---

EDIT: I have changed the visibility modifiers to `pub` given 'official' custom widgets may be implemented outside the crate, e.g. in `iced_aw`. I have also added documentations to distinguish them from their original counterparts by emphasising the `IntoIterator` type of their argument.

### DIFF
--- a/native/src/element.rs
+++ b/native/src/element.rs
@@ -10,7 +10,7 @@ use crate::{
 };
 
 use std::any::Any;
-use std::borrow::Borrow;
+use std::borrow::{Borrow, BorrowMut};
 
 /// A generic [`Widget`].
 ///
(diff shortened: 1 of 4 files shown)


hecrj:
As we discussed on Discord, I believe these changes introduce a bunch of complexity and generics in the type signatures of these functions.

I'd like to understand better the use cases we are tackling here. Optional widget contents are already possible ([see `pane_grid::Content` and it's optional `TitleBar`](https://github.com/iced-rs/iced/blob/f6966268bb6d58b4b03ba61fc5732e1bf016e2a1/widget/src/pane_grid/content.rs#L20)). Has that approach been attempted?

Once we have a set of specific use cases, then we will be able to better evaluate the complexity introduced here. Let's wait until then!

PR #1841, opened 2023-05-09
fix: tiny-skia svg premultiply final filtered color

wash2:
If the color is pre-multiplied before grabbing the alpha of the pixel, then the filter color is leaked even what the pixel is completely transparent. This is noticeable if you disable `wgpu` and use the dark theme in the svg example.


hecrj (approved):
I see. Here I was hoping there would be a way to avoid premultiplying every pixel...

Thanks! :bow:

PR #1842, opened 2023-05-09
fix: border radius typo

wash2:


hecrj (approved):
Good catch! :sweat_smile:

PR #1843, opened 2023-05-10
fix: quad rendering including border only inside of the bounds

wash2:
It seems that border should not be drawn outside the quad bounds. For example, a slider handle with a large border radius will leave a trail of the border color when it is dragged, because it goes un-tracked in the damage. 

This change also makes tiny-skia rendering of quads more consistent with the wgpu backend, which didn't seem to draw outside the bounds as far as i could tell with the same slider test.

### DIFF
--- a/tiny_skia/src/backend.rs
+++ b/tiny_skia/src/backend.rs
@@ -159,7 +159,6 @@ impl Backend {
                 border_width,
                 border_color,
             } => {
-                // XXX border seems to be contained by the bounds of the primitive in wgpu and for damage regions, so we do the same here
                 let physical_bounds = (*bounds + translation) * scale_factor;
 
                 if !clip_bounds.intersects(&physical_bounds) {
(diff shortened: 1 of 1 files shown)


tiny_skia/src/backend.rs
```diff
@@ -172,7 +173,26 @@ impl Backend {
...
+                    height: bounds.height - 2.0 * border_width,
+                };
+                // fill border radius is the border radius minus the border width
+                let mut fill_border_radius = *border_radius;
+                for radius in &mut fill_border_radius {
+                    *radius = (*radius - border_width / 2.0)
+                        .min(path_bounds.width / 2.0)
+                        .min(path_bounds.height / 2.0);
+                }
+                let path = rounded_rectangle(path_bounds, fill_border_radius);
```
hecrj, line 195: I'd expect the border to be rendered on top of the background like a `border-box` in CSS.

wash2:
Hmm this might need some more testing. I'm going to mark it as a draft until i've tested it some more :sweat_smile:

wash2:
This seems to be working pretty well now.

hecrj (approved):
Thanks!

PR #1844, opened 2023-05-11
Remove min width 1 from scrollbar & scroller

tarkah:
I'm not sure if there's a reason to enforce this, but ideally the user can set it to `0.0` to "hide" the scrollbar / scroller.

The only way to "hide" the scrollbar / scroller currently is via the stylesheet, which is only aesthetic. This still allows them to be interacted with.

You can test this change by running `scrollable` example. If you set both to 0, the scrollable still works via scrollwheel / trackpad, but the scrollbar / scroller are hidden / don't work. You can even set scrollbar to 0 but keep scroller width to have a floating scroller w/ no scrollbar, which is a feature imo.

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -809,9 +809,11 @@ pub fn draw<Renderer>(
              style: Scrollbar,
              scrollbar: &internals::Scrollbar| {
                 //track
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Awesome! Thanks :bow:

PR #1845, opened 2023-05-11
Feat: Offscreen Rendering & Screenshots

bungoboingo:
## This PR adds support for offscreen rendering in Iced! 🎉 

https://user-images.githubusercontent.com/4241774/230238292-03b8d796-c8cb-4870-adc1-3dbac36c3a8d.mov

### 📸 Users can now capture screenshots 🖼️ using the `window::screenshot()` command.

```rust
    match message {
        Message::Screenshot => {
            return iced::window::screenshot(Message::ScreenshotData); //new!
        }
        //...
```

The `window::screenshot` command requires a `Message` which takes a `Screenshot`. This is a new struct exposed in the `iced_runtime` crate which contains the RGBA bytes of the screenshot, in addition to its size.

For now this command is limited to the entire viewport, with an optional `crop()` method available which will crop the bytes to the specified `Rectangle`.

https://user-images.githubusercontent.com/4241774/230238237-a9b3ddfa-cb79-463f-85d9-18621585dd79.mov

```rust
let cropped = screenshot.crop(Rectangle::<u32> {
    x: 0,
    y: 0,
    width: 500,
    height: 500,
});
```

This method returns a `Result<Screenshot, CropError>` for situations where the cropped region is out of bounds, or is not visible.

### 🎉 Offscreen rendering capabilities are now added to both `wgpu` and `tiny_skia`!

Rendering offscreen is now exposed in the `Compositor` interface with a `render_offscreen` function. The outputted texture data is always guaranteed to be in RGBA format in the SRGB color space for compatibility with image libraries & iced images, and to simplify the `Screenshot` struct with a predetermined byte order & pixel size.

### 🔢 Implementation details

For `wgpu`, performing offscreen rendering was fairly straightforward. I create a texture based on the viewport size & the chosen texture format, and then pass that into the backend with a new `backend.offscreen` method. This method performs a normal render pass on the new texture, and optionally runs a tiny compute shader after drawing all primitives that converts it to `wgpu::TextureFormat::Rgba8Unorm` with a manual srgb conversion due to `wgpu::TextureFormat::Rgba8UnormSrgb` not being a supported format for a storage texture. The whole process on my (admittedly very powerful!) test suite of GPUs (AMD, NVIDIA, M1) took about ~2µs in the test example. The only added overhead over a normal render pass is the compute shader to convert to rgba.

For `tiny-skia` this was _very_ straightforward, the same as a normal `present` call but rendering to a new buffer vs the surface buffer. Our backend returns `ARGB` format, so just had to do some bit shiftin' and we're all good.

### 🤔 Outstanding API questions

1. It was discussed whether it's a cleaner API to do a `window::screenshot` command which optionally can be cropped, or a `window::screenshot(area)` command which can either be `Fullscreen` or `Region`, which would be a rectangle. Obviously there is a performance hit when rendering the whole window to an offscreen buffer vs a potentially very small region if only a small portion is desired, but the execution is so fast it's almost negligible. I'm curious as to people's thoughts on this!
 
 2. The offscreen render capabilities are only exposed at the moment through this `window::screenshot` command. The intent was to implement the capabilities for offscreen rendering in this PR and expand upon it further in later iterations if new applications of offscreen rendering are needed. Curious to see if anyone has any thoughts on how else this might be exposed!
 
 ### 🧪 Testing
Tested wgpu & tiny-skia implementations on MacOS + M1, Linux + PopOS + AMD, Windows 10 + Nvidia combos. Further testing would be greatly appreciated!

Feedback very much welcome! This was my first time working this deeply with textures on the GPU so might have made some rookie mistakes 😉 I also know that there is an open draft PR for offscreen rendering but it hasn't been updated in a few years so thought I'd take a crack at it.

(reopened from [1783](https://github.com/iced-rs/iced/pull/1783) which was auto closed due to `advanced-text` merge)


wgpu/src/shader/offscreen_blit.wgsl
```diff
@@ -0,0 +1,22 @@
...
+fn main(@builtin(global_invocation_id) id: vec3<u32>) {
+    // texture coord must be i32 due to a naga bug:
+    // https://github.com/gfx-rs/naga/issues/1997
+    let coords = vec2(i32(id.x), i32(id.y));
+
+    let src: vec4<f32> = textureLoad(u_texture, coords, 0);
+    let srgb_color: vec4<f32> = vec4(srgb(src.x), srgb(src.y), srgb(src.z), src.w);
+
+    textureStore(out_texture, coords, srgb_color);
+}
```
hecrj, line 22: This won't work with the `web-colors` feature for similar reasons to #1885.

I think we should avoid doing any color conversions in shader code. These should be handled by the GPU automatically.

I believe we can rewrite this using a render pipeline so we can easily control the output format from Rustland.

hecrj (changes requested):

wgpu/src/shader/offscreen_blit.wgsl
bungoboingo: I will look into it, pretty sure I couldn't use a srgb texture format for a storage texture so it will have to be a render pipeline vs compute, yeah.

hecrj: Not sure I follow. I am fairly certain I have sampled textures with a different format than the output format in a render pipeline.

`glyphon` uses a masked texture with a single channel for most glyphs, for instance.

bungoboingo: might have a different way to do it using a render pipeline, I'm poking around with it this morning.

bungoboingo: Nvm i was just being a pepeg, should be good now 👍

bungoboingo: Taking another look, I think now I can just re-use the existing `blit.wgsl` shader we have for multisampling

bungoboingo:
Updated in d955b3444da19e24bf0de6d6f432f06623ed5db2 to just use the existing `blit.wgsl` that we use for msaa. Nice and simple!

hecrj (approved):
Awesome! Thank you!

Just moved a couple things around to simplify some APIs, but everything seems to work! Let's merge! :partying_face:

PR #1849, opened 2023-05-11
WIP: Iced accessibility

wash2:
This is a work in progress implementation of https://github.com/iced-rs/rfcs/pull/21. 

I have added widget implementations where I could, but they need more work. Some widget implementations may need changes to be more compatible with accessibility.


DataTriny:
Hello @wash2, @hecrj, are you currently working on this? Can I help you?

wash2:
Yes, I need to move over some of my updates for this PR. I'd love some help after I've done that. I will ping you after pushing the updates.

> Hello @wash2, @hecrj, are you currently working on this? Can I help you?

RastislavKish:
@wash2 thanks for your work on this! Do you think the pop-os iced fork is ready for some experimental real-world a11y testing?

I've tried to compile the iced counter example with the winit and a11y features, but the accessibility layer does not seem to be present, Orca is unable to focus the window. Ubuntu Mate 24.04.

I'm a blind programmer, I would be glad to provide relevant feedback when you're ready, as this will help not just iced, but Cosmic too.

wash2:
> @wash2 thanks for your work on this! Do you think the pop-os iced fork is ready for some experimental real-world a11y testing?
> 
> I've tried to compile the iced counter example with the winit and a11y features, but the accessibility layer does not seem to be present, Orca is unable to focus the window. Ubuntu Mate 24.04.
> 
> I'm a blind programmer, I would be glad to provide relevant feedback when you're ready, as this will help not just iced, but Cosmic too.

Hi, you will probably have more luck testing on the COSMIC alpha than on Ubuntu 24.04, because it includes experimental support for accessibility APIs on wayland via https://github.com/pop-os/at-spi2-core and cosmic-comp. I've done some testing, and it seems to be working ok to me, but I'd be happy to hear if there are some key features missing in your opinion.

ids1024:
Or without at-spi2-core patches, it may be best to test on X11. (Or on Windows or macOS.) Maybe Mate is still on X11, at least by default, though. I'm not familiar with it's current state.

DataTriny:
I just tested on Windows pop-os fork master branch (iced-accessibility cannot fetch winit from git).

Had to add raw-window-handle as a dependency of iced_winit (it's currently only present on Linux but it is a hard requirement on all platforms for AccessKit).

I tried a few examples and none of them work. I can clearly see that an AccessKit tree is generated but keyboard focus does not work at all. Neither does mouse exploration.

Is there a specific program that is known to work outside of the Cosmic suite?

wash2:
> Is there a specific program that is known to work outside of the Cosmic suite?

By keyboard focus, do you mean tabbing through the application to focus different widgets, or the Focus action? Tab and Shift + Tab won't work for a lot of the iced examples. I've primarily done the testing with the cosmic suite. Widget handling of accessibility events is implemented for things like the default action on buttons, but the Focus action is actually not handled in the winit shell as of right now, though it *should* be, and seems to have been an oversight on my part. The reader does read off widget names, but it is somewhat limited for sure.

DataTriny:
> By keyboard focus, do you mean tabbing through the application to focus different widgets

Yes, that's what I meant.

> I'd be happy to hear if there are some key features missing in your opinion.

I can't see anywhere calls to the following methods, so I guess text input is not yet implemented: `NodeBuilder::set_character_lengths`, `NodeBuilder::set_character_positions`, `NodeBuilder::set_character_widths`, `NodeBuilder::set_word_lengths`. There are ovbiously many more missing features, but I think this is probably the most important.

Also, please don't set the live property unconditionally on all labels, this will make for a very annoying experience. This property should only be used for live regions such as a message list in a chat application.

roboteng:
@wash2, What's the status on this? I saw that this hasn't had much going on recently. I've spent some time trying to merge pop-os/master in this branch, but it didn't seem trivial to do. I've got a bit of time, and I'd like to help as much as I can! I haven't worked on pop-os or iced before, but I've got a background in accessibility, and I've got a fair bit of Rust experience. I was able to run this branch on MacOs, and got some basic screen reader interaction going!

wash2:
I've unfortunately not had time to update this PR yet. It would require significant changes because Iced moves quite fast, and there have been additional changes to the pop branch that need to be added. I guess it would probably be at least a month or two before I get a chance to look into updates. If you want to take a shot in the meantime, I don't mind.

roboteng:
If anyone is interested, I've done some work in #3111. It's not complete, but a POC

PR #1853, opened 2023-05-13
Implement math operators for Point

st0rmbtw:


liigo:
what's the use cases? i don't think it's very useful.

hecrj:
These operations do not actually make sense for `Point`. You normally want to operate with a `Point` and a `Vector`.

PR #1854, opened 2023-05-15
fix: when clearing damaged surface with background color blend mode should be Source only

wash2:
This is most noticeable when using a background color which is fully transparent, but should also cause issues with any partially transparent background color.


hecrj (approved):
Thanks!

PR #1855, opened 2023-05-15
wip: animations for default widgets

lazytanuki:
Hi!

This WIP implements default animations for default widgets. The goal here is to make Iced application appear smooth when hovering / clicking things, without breaking the API. It is also possible to keep the actual behavior (using a simple enum), although I do not know which should be default.
This heavily uses the new frame subscription API.
This is one thing that I was missing since transitioning from GTK, where you don't really need to worry about animations but can still end up with something that looks nice and smooth.

# Current progress

## Buttons

Buttons have smooth color transitions when hovered and pressed.

https://github.com/iced-rs/iced/assets/43273245/4d000c71-3eca-4b0a-8326-e647ed1bf42d

Things to discuss and fix:

- [ ] Buttons look for cursor movement events to know whether they're being hovered or not. Could this be replaced by a new `Hovered` event in the `update()` function?
- [ ] When clicking a button that is moved on click, it is possible that it is hovered anymore. However, if the mouse does not move, it will not be aware of it and will remain highlighted until the cursor moves again.

## Text input (cursor)

In the text input, the cursor will fade-blink. It will however remain a full opacity while typing

https://github.com/iced-rs/iced/assets/43273245/101e9de9-55a2-49eb-8836-1b5d2d3b72f6

## Radio buttons

Radio buttons have smooth color transitions when hovered and pressed.

https://github.com/iced-rs/iced/assets/43273245/571a9a7c-451e-4e46-9591-fbc1825bbe96

## Togglers

Togglers have smooth color and movements transitions when pressed.

https://github.com/iced-rs/iced/assets/43273245/0af06664-739f-4c04-91dd-ef061f3530a6

# To do

- [ ] checkboxes
- [ ] scrollables
- [ ] sliders
- [ ] ...

# Interferences

I stumbled upon [cosmic-time](https://github.com/pop-os/cosmic-time), which implements a framework for animations in iced. From what I understand, it is made for more complex animations to be implemented by the user, whereas my goal is to bring default animations to Iced. However, I do not know if it would be better to use what has been made for cosmic-time for this or not. Or maybe this WIP is better suited to be used in cosmic-time after all.

I think default animations would be great, at least to be on par with GTK, but that's just IMO!
Thanks!

### DIFF
--- a/core/src/color.rs
+++ b/core/src/color.rs
@@ -131,6 +131,22 @@ impl Color {
     pub fn inverse(self) -> Color {
         Color::new(1.0f32 - self.r, 1.0f32 - self.g, 1.0f32 - self.b, self.a)
     }
+
+    /// Mix with another color with the given ratio (from 0 to 1)
+    pub fn mix(&mut self, other: Color, ratio: f32) {
+        if ratio >= 1.0 {
+            self.r = other.r;
+            self.g = other.g;
+            self.b = other.b;
+            self.a = other.a;
+        } else if ratio > 0.0 {
+            let self_ratio = 1.0 - ratio;
+            self.r = self.r * self_ratio + other.r * ratio;
+            self.g = self.g * self_ratio + other.g * ratio;
+            self.b = self.b * self_ratio + other.b * ratio;
+            self.a = self.a * self_ratio + other.a * ratio;
+        }
+    }
 }
 
 impl From<[f32; 3]> for Color {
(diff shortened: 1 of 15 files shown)


max397574:
just stumbled upon this and it looks really promising
Any plans on how to continue with this @hecrj @lazytanuki ?

nerocui:
Any plan to merge this one?

lazytanuki:
Hi ! I don't have a lot of time to work on this right now, but I believe we should have a discussion with @hecrj in order to figure out what is the best way to implement this. I know folks over at System76 have their own Cosmic time animation system and I don't want to step on their feet. However their system makes animations "external" to the widgets (at least from what I saw a while back, may have changed now), but I think it would be nice to have the animations system "internal" to widgets, at least for the ones that don't modify the layout. That way, users wouldn't have any extra work to do to have simple fade in/out animations for buttons, cursors, etc.
This is still on my list of todos.

genusistimelord:
The Idea behind this looks Solid so its just a matter of if Hector wants to take a look at it or if he rather you Post a topic about it first.

lazytanuki:
Superseded by #2483.

PR #1856, opened 2023-05-16
[Feature] Enhance PaneGrid to split panes by drag & drop

jhff:
## Summary 

Previously, dropping a pane onto another would just swap them. This leads users of `PaneGrid` needing to manually create splits first and then swap panes (and most probably closing unnecessary ones) in order to change the layout of the grid.

This change adds functionality to `PaneGrid` in order to be able to split a pane with another by simply dragging and dropping the pane in a certain region of the target pane.

## Implementation

Introduced `Region` parameter to `pane_grid::DragEvent::Dropped` that is the target pane area (top, left, bottom, right, center) where the dragged pane is hovering on. Panes were divided in thirds. Left and right sides have priority over top and bottom.

Users of `PaneGrid` can call `pane_grid::State::split_with(target, pane, region)` in order to split the target with a given pane on a given region of the target. 

The `Region` of the hovered target dropping area is highlighted with a given `style` in order for users to easily see how the panes will be displayed.

## Demo

https://github.com/iced-rs/iced/assets/51237625/edd4a8da-fd14-418d-ac28-b1f3782b1827

### DIFF
--- a/widget/src/pane_grid/state.rs
+++ b/widget/src/pane_grid/state.rs
@@ -2,7 +2,9 @@
 //!
 //! [`PaneGrid`]: crate::widget::PaneGrid
 use crate::core::{Point, Size};
-use crate::pane_grid::{Axis, Configuration, Direction, Node, Pane, Split};
+use crate::pane_grid::{
+    Axis, Configuration, Direction, Node, Pane, Region, Split,
+};
 
 use std::collections::HashMap;
 
@@ -163,6 +165,53 @@ impl<T> State<T> {
         let _ = self.maximized.take();
 
         Some((new_pane, new_split))
+    }
+
+    /// Split a target [`Pane`] with a given [`Pane`] on a given [`Region`].
+    ///
+    /// Panes will be swapped by default for [`Region::Center`].
+    pub fn split_with(
+        &mut self,
+        target: &Pane,
+        pane: (T, &Pane),
+        region: Region,
+    ) {
+        match region {
+            Region::Center => {
+                let (_, pane) = pane;
+                self.swap(pane, target);
+            }
+            Region::Top => {
+                self.split_and_swap(Axis::Horizontal, target, pane, true)
+            }
+            Region::Bottom => {
+                self.split_and_swap(Axis::Horizontal, target, pane, false)
+            }
+            Region::Left => {
+                self.split_and_swap(Axis::Vertical, target, pane, true)
+            }
+            Region::Right => {
+                self.split_and_swap(Axis::Vertical, target, pane, false)
+            }
+        }
+    }
+
+    fn split_and_swap(
+        &mut self,
+        axis: Axis,
+        target: &Pane,
+        pane: (T, &Pane),
+        invert: bool,
+    ) {
+        let (state, pane) = pane;
+
+        if let Some((new_pane, _)) = self.split(axis, target, state) {
+            if invert {
+                self.swap(target, &new_pane);
+            }
+
+            let _ = self.close(pane);
+        }
     }
 
     /// Swaps the position of the provided panes in the [`State`].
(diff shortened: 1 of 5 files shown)


tarkah:
This is great! Love the simple implementation.

Stretch goal or follow-up: It'd be super cool if we could also drop on the "splits" between panes to distribute the source pane evenly between the two target panes on either side of the split. I think that'd round out this pane grid implementation very nicely and give people all of the options they want for laying things out without having to manually drag split regions.

jhff:
> This is great! Love the simple implementation.

Thanks! 🙌 

> Stretch goal or follow-up: It'd be super cool if we could also drop on the "splits" between nodes to place the source pane between the two target nodes on either side of the split. I think that'd round out this pane grid implementation very nicely!
> 
> Open question for behavior when dropping on a split: How do the nodes get resized to "distribute" across the one source and 2 target nodes? I'd assume they keep the same ratio of their current width / total width of all 3 nodes * available width.

That's interesting. I've tried to find some good examples that implement something like that, so I could have a better understanding on how that would "feel", but didn't find any. Usually apps just allow the user to place a pane either on the left side of a pane (if on the right side of the split) or vice-versa. Check out VS Code or https://golden-layout.com/ for example. This way also avoids having to figure out how to handle the radio of their width/height and leave that entirely to the user.

Do you know any example that does that?

tarkah:
> Do you know any example that does that?

I do not. I just had the thought when watching your video. In the 2x2 grid, it seemed like it'd be nice to drag to the top right pane over the bottom middle split and end up with 3 evenly distributed panes on the bottom.

However now that I'm thinking about this more, I don't think it'd always work that intuitively for 2 reasons:

- Splits aren't always setup as one would think. Some splits are between single pane nodes and others against more nested nodes.
- Even distribution is only possible when all nodes are at the same width. This starts to become less useful without a perfect starting grid like the 2x2 example.

So it's hard to say how useful this would actually be in practice without implementation or another example to reference.

jhff:
> > Do you know any example that does that?
> 
> I do not. I just had the thought when watching your video. In the 2x2 grid, it seemed like it'd be nice to drag to the top right pane over the bottom middle split and end up with 3 evenly distributed panes on the bottom.
> 
> However now that I'm thinking about this more, I don't think it'd always work that intuitively for 2 reasons:
> 
> * Splits aren't always setup as one would think. Some splits are between single pane nodes and others against more nested nodes.
> * Even distribution is only possible when all nodes are at the same width. This starts to become less useful without a perfect starting grid like the 2x2 example.
> 
> So it's hard to say how useful this would actually be in practice without implementation or another example to reference.

Exactly, that might introduce a lot of complexity or weird behaviour to the user itself. Anyway, I think this can be enhanced even more by being able to drag & drop at the edges of the whole grid in order to create a new row at top/bottom or new column at left/right side of the grid with the dragged pane. Gonna add that here.

widget/src/pane_grid/state.rs
```diff
@@ -165,6 +167,53 @@ impl<T> State<T> {
         Some((new_pane, new_split))
     }
 
+    /// Split a target [`Pane`] with a given [`Pane`] on a given [`Region`].
+    ///
+    /// Panes will be swapped by default for [`Region::Center`].
+    pub fn split_with(
+        &mut self,
+        target: &Pane,
+        pane: (T, &Pane),
```
hecrj, line 176: Could we take `&Pane` here and then use `HashMap::remove` from `self.panes` to get the state `T` internally?

hecrj (changes requested):
Looks fantastic! Just an idea!

hecrj:
@tarkah @jhff We can focus on merging this bit of functionality first. We can add further improvements in a different PR.

widget/src/pane_grid/state.rs
jhff: Ah yes! We can get state `T` by closing it first: bc590e2

Much cleaner now and no visual changes to the user

hecrj (approved):
Fantastic! :tada:

Thanks :bow:

PR #1859, opened 2023-05-20
Add Nix instructions to README

RGBCube:
We should add instructions for Ubuntu, Arch, Windows and MacOS too, just NixOS on its own feels weird.


RGBCube:
I think this is ready to be merged?

hecrj (approved):
Cool! Thanks :bow:

PR #1860, opened 2023-05-21
feat(text_input): For ignored events, pass Ignored

titoeb:
For a [small side-project of mine](https://github.com/titoeb/mrc-workout-creator) I wanted to implement the following feature: On my GUI, I have three text inputs on top and I wanted that pressing F1 would focus the first text input, F2 would focus the second text input ect, even if another text input is focused. 

That was working fine, when no other text input was focused, but as soon as one was focused the Event of pressing F1, F2, F3 was just caught by the text-input without doing anything with them or passing them as ignored. Currently [only Tab, Up and Down](https://github.com/iced-rs/iced/blob/640e13943c04754ef74e11db470b6c9470640623/widget/src/text_input.rs#L881-L884)-events are not caught by the text-input. 

In this PR, for all Events of type `Event::Keyboard(keyboard::Event::KeyReleased{...})` or `Event::Keyboard(keyboard::Event::KeyPressed{...})` that are not explicitly handled, the text input will return `event::Status::Ignored`. 

For my local example this worked fine without impacting the functionality of the `TextInput` but I don´t know if that collides with the overall architecture of the crate.

### DIFF
--- a/widget/src/text_input.rs
+++ b/widget/src/text_input.rs
@@ -878,12 +878,9 @@ where
                         state.keyboard_modifiers =
                             keyboard::Modifiers::default();
                     }
-                    keyboard::KeyCode::Tab
-                    | keyboard::KeyCode::Up
-                    | keyboard::KeyCode::Down => {
+                    _ => {
                         return event::Status::Ignored;
                     }
-                    _ => {}
                 }
 
                 return event::Status::Captured;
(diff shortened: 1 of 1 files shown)


flosse:
I have nothing against this PR, but I find the motivation interesting: why do you use three text inputs instead of one?

titoeb:
> I have nothing against this PR, but I find the motivation interesting: why do you use three text inputs instead of one?

Do you mean my side project? There I want to design workouts for my bike-trainer. A workout consists of individual efforts, which is basically holding a wattage for a specific time-span. So for an effort you need potentially three inputs: length in minutes, wattage in the beginning, wattage in the end (in between it is linearly interpolated). So I naively used hree text-inputs :shrug:

hecrj:
The implementation here doesn't seem correct. There are definitely keyboard events related to text input that we are now marking as ignored. See #652 for the reasoning behind the current implementation.

You can listen to captured events with [the `events_with` subscription](https://docs.rs/iced/latest/iced/subscription/fn.events_with.html). Does that not work for you?

PR #1861, opened 2023-05-21
Support conversion from Fn trait to custom theme

marienz:
...instead of just from function pointers.

I'm making this change not because I actually want to pass a closure, but to make passing a single fixed function work. This commit also simplifies the scrollable example slightly, and without the other half of this change that simplified example fails to compile with:

```
error[E0277]: the trait bound `iced::theme::ProgressBar: From<for<'a> fn(&'a Theme) -> iced::widget::progress_bar::Appearance {progress_bar_custom_style}>` is not satisfied
   --> examples/scrollable/src/main.rs:292:28
    |
292 |                     .style(progress_bar_custom_style)
    |                      ----- ^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<for<'a> fn(&'a Theme) -> iced::widget::progress_bar::Appearance {progress_bar_custom_style}>` is not implemented for `iced::theme::ProgressBar`
    |                      |
    |                      required by a bound introduced by this call
    |
    = help: the trait `From<for<'a> fn(&'a Theme) -> iced::widget::progress_bar::Appearance>` is implemented for `iced::theme::ProgressBar`
    = note: required for `for<'a> fn(&'a Theme) -> iced::widget::progress_bar::Appearance {progress_bar_custom_style}` to implement `Into<iced::theme::ProgressBar>`
note: required by a bound in `iced::widget::ProgressBar::<Renderer>::style`
   --> /home/marienz/src/iced/widget/src/progress_bar.rs:77:21
    |
77  |         style: impl Into<<Renderer::Theme as StyleSheet>::Style>,
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `ProgressBar::<Renderer>::style`
```

This happens because `progress_bar_custom_style` by itself is a function item, which is typically coerced to a function pointer when one is needed, but not in this case. It is possible to work around this on the caller's side, but especially since the compiler diagnostic for this is a bit rough (see https://github.com/rust-lang/rust/issues/100116) let's try to make it work out of the box.

### DIFF
--- a/examples/scrollable/src/main.rs
+++ b/examples/scrollable/src/main.rs
@@ -289,18 +289,13 @@ impl Application for ScrollableDemo {
             }
             Direction::Horizontal => {
                 progress_bar(0.0..=1.0, self.current_scroll_offset.x)
-                    .style(theme::ProgressBar::Custom(Box::new(
-                        ProgressBarCustomStyle,
-                    )))
+                    .style(progress_bar_custom_style)
                     .into()
             }
             Direction::Multi => column![
                 progress_bar(0.0..=1.0, self.current_scroll_offset.y),
-                progress_bar(0.0..=1.0, self.current_scroll_offset.x).style(
-                    theme::ProgressBar::Custom(Box::new(
-                        ProgressBarCustomStyle,
-                    ))
-                )
+                progress_bar(0.0..=1.0, self.current_scroll_offset.x)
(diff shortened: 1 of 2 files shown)


hecrj (approved):

PR #1865, opened 2023-05-22
[Feature] Pane Grid: drag & drop panes to the edges

jhff:
## Summary

Previously, in order to create a new column or row on a `PaneGrid`, a client would necessarily have to create and remove panes, as it was only able to split existing panes.

Following #1856, this change adds the ability to drag & drop panes of a `PaneGrid` to its edges in order to enhance even more an easy layout customisation of the grid.

## Implementation

`DragEvent::Dropped` instead of `Pane` and `Region` parameters, now holds a `Target` that can either be `Target::PaneGrid` (new variant, that wraps an `Edge`) or `Target::Pane` (which holds the previous parameters).

Users of `PaneGrid` can call `pane_grid::State::move_to_edge(pane, edge)` in order to create a new row/column depending on the `edge` with that `pane`'s `state`.  This will split the major `Node` of the grid in order to create said row/column.

Hovered `PaneGrid` edges are detected and highlighted with the minimum thickness of either the total grid height or width ratio of `1/25`. I feel it is a good balance after some experimentation, but I'm up to change this.

## Demo

https://github.com/iced-rs/iced/assets/51237625/d5944df5-fd63-44ac-8dfb-2960b64e1018

### DIFF
--- a/examples/pane_grid/src/main.rs
+++ b/examples/pane_grid/src/main.rs
@@ -108,10 +108,15 @@ impl Application for Example {
             Message::Dragged(pane_grid::DragEvent::Dropped {
                 pane,
                 target,
-                region,
-            }) => {
-                self.panes.split_with(&target, &pane, region);
-            }
+            }) => match target {
+                pane_grid::Target::PaneGrid(edge) => {
+                    self.panes.move_to_edge(&pane, edge)
+                }
+                pane_grid::Target::Pane {
+                    pane: target,
+                    region,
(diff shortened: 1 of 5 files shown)


jhff:
Question: should these added features be disableable, i.e. user of a `PaneGrid` be able to not allow drag & drop split/move-to-edge?

hecrj (approved):
Awesome stuff! Thanks :partying_face: 

Made some very minor changes here and there and added a new `drop` helper method to `State`.

Let's ship this! :ship:

PR #1869, opened 2023-05-23
Extended border radius on relevant widgets

casperstorm:
This PR exposes the internal `BorderRadius([f32; 4])` and replaces it with the current type (`f32`).
This means we now have the possibility to create buttons, containers and so on with different border radii.

**Breaking changes**
This is a breaking change, however to migrate to this new BorderRadius we have the following `impl`
So in most cases a simple `.into()` will solve it.

```rust
impl From<f32> for BorderRadius {
    fn from(w: f32) -> Self {
        Self([w; 4])
    }
}

impl From<[f32; 4]> for BorderRadius {
    fn from(radi: [f32; 4]) -> Self {
        Self(radi)
    }
}
```

### DIFF
--- a/examples/scrollable/src/main.rs
+++ b/examples/scrollable/src/main.rs
@@ -289,18 +289,13 @@ impl Application for ScrollableDemo {
             }
             Direction::Horizontal => {
                 progress_bar(0.0..=1.0, self.current_scroll_offset.x)
-                    .style(theme::ProgressBar::Custom(Box::new(
-                        ProgressBarCustomStyle,
-                    )))
+                    .style(progress_bar_custom_style)
                     .into()
             }
             Direction::Multi => column![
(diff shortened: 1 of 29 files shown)


examples/scrollable/src/main.rs
```diff
@@ -289,13 +289,18 @@ impl Application for ScrollableDemo {
             }
             Direction::Horizontal => {
                 progress_bar(0.0..=1.0, self.current_scroll_offset.x)
-                    .style(progress_bar_custom_style)
```
hecrj, line 292: Can we keep the cleaner approach with a `fn` pointer?

hecrj (changes requested):

hecrj (approved):
Awesome! Thanks :tada:

PR #1871, opened 2023-05-24
[Fix] Make gradient pack fn public for iced_graphics::Gradient

bungoboingo:
**Changes:**
- Moved `pack_gradient` to a pub method of `iced_graphics::Gradient` named `pack`. This allows users to pack gradient data when creating their own `Primitive::GradientMesh` with `GradientVertex2D`.

The `pack_gradient` fn for `iced::core::Gradient` is unchanged as it's internal.

### DIFF
--- a/graphics/src/gradient.rs
+++ b/graphics/src/gradient.rs
@@ -4,7 +4,7 @@
 //!
 //! [`Gradient`]: crate::core::Gradient;
 use crate::core::gradient::ColorStop;
-use crate::core::{Color, Point};
+use crate::core::{self, Color, Point, Rectangle};
 use std::cmp::Ordering;
 
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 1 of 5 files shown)


graphics/src/gradient.rs
```diff
@@ -85,4 +94,28 @@ impl Linear {
 
         self
     }
+
+    /// Packs the [`Gradient`] for use in shader code.
+    pub fn pack(&self) -> [f32; 44] {
```
hecrj, line 99: Let's make an opaque `Packed` type in this module and use that instead of `[f32; 44]` everywhere. We'll keep the internal details hidden that way.

hecrj (changes requested):

graphics/src/gradient.rs
bungoboingo: Updated in 413526ad09d006853eb9659efabee168f4a0e0a4

graphics/src/gradient.rs
```diff
@@ -96,26 +96,33 @@ impl Linear {
...
-        pack
+        Packed { data }
     }
 }
+
+/// Packed [`Gradient`] data for use in shader code.
+#[derive(Debug)]
+pub struct Packed {
+    /// The packed [`Gradient`] data.
+    pub data: [f32; 44],
```
hecrj, line 127: Ideally, we should make this struct `repr(C)` and the `data` completely private.

hecrj (changes requested):
Cool! Just one more iteration!

graphics/src/gradient.rs
bungoboingo: Good idea! Changed in e843b4f045a5a825e991346b1aaffe8eece3ffbe. The imports will be moved to more appropriate places in #1873

bungoboingo:
Looks good! 👍

PR #1873, opened 2023-05-25
[Fix] Quads of different background types not ordered

bungoboingo:
So caught up in the elation of lerping, I completely forgot to order quads of different background types 🤦

This will fix any issues with using quads of different background types on top of one another.

I made a separate `order` field of the `layer::Quads` struct to consolidate quads for rendering, similar to how we're doing it with triangle meshes already.


hecrj:
I have some ideas to improve layering overall in the backlog.

For now, having gradients always be on top of solid seems consistent with the way other primitives work for a specific layer. I don't think we should introduce ordering (primitive layering) at a layer level.

bungoboingo:
This is just for grouping quads together to be rendered actually in order of when they're supposed to be. Right now you can't put a container with a solid background over a container with a gradient background or else the solid will be rendered over. I'm not sure where else to put ordering besides at the layer level since that's where they're being grouped, unless we want to introduce like a depth value or something to primitives.

Current master:
Screenshot 2023-05-25 at 2 14 49 PM

Fixed:
Screenshot 2023-05-25 at 2 13 54 PM

```rust
fn view(&self) -> Element<'_, Self::Message, Renderer<Self::Theme>> {
        let small_center_box = container(":-)")
            .center_x().center_y()
            .height(100)
            .width(100)
            .style(theme::Container::Custom(Box::new(SolidBackgroundStyle)));

        let center_box = container(small_center_box)
            .center_x()
            .center_y()
            .height(300)
            .width(300)
            .style(theme::Container::Custom(Box::new(SolidBackgroundStyle)));

        container(center_box)
            .width(Length::Fill)
            .height(Length::Fill)
            .center_x()
            .center_y()
            .style(theme::Container::Custom(Box::new(GradientBackgroundStyle)))
            .into()
    }
```

bungoboingo:
Adjusted in 10a68acdda9d74488594b7c28ae797b47316e683 👍

hecrj (approved):
Awesome!

Moved some code around to avoid the `quad` pipeline details from leaking into `layer` by removing `layer::quad` altogether! We should probably do the same thing with the other `layer` submodules; but no need to do that here.

Also simplified some minor things! Hopefully I didn't break anything.

Let's merge! :ship:

hecrj:
Found the right way to keep `solid` and `gradient` details also private!

bungoboingo:
Looks good, we can move the other `layer` submodule code into its respective pipeline implementation files later.

PR #1875, opened 2023-05-25
Upgrade `palette` dependency

clarkmoody:
- Upgrades `palette` to 0.7 across the library
- Fixes breakages
  - Use `rgb::Rgb` instead of `Srgb` directly
  - `RelativeContrast` trait has been deprecated in favor of `color_difference::Wcag21RelativeContrast`
  - `Hue::to_positive_degrees()` -> `Hue::into_positive_degrees()`

### DIFF
--- a/core/src/color.rs
+++ b/core/src/color.rs
@@ -1,5 +1,5 @@
 #[cfg(feature = "palette")]
-use palette::rgb::{Rgb, Rgba};
+use palette::rgb::{Srgb, Srgba};
 
 /// A color in the sRGB color space.
 #[derive(Debug, Clone, Copy, PartialEq, Default)]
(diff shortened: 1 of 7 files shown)


core/src/color.rs
```diff
@@ -183,48 +183,48 @@ macro_rules! color {
...
 
 #[cfg(feature = "palette")]
-/// Converts from palette's `Srgba` type to a [`Color`].
-impl From<Srgba> for Color {
-    fn from(srgba: Srgba) -> Self {
-        Color::new(srgba.red, srgba.green, srgba.blue, srgba.alpha)
+/// Converts from palette's `Rgba` type to a [`Color`].
+impl From<Rgba> for Color {
+    fn from(rgba: Rgba) -> Self {
+        Color::new(rgba.red, rgba.green, rgba.blue, rgba.alpha)
```
hecrj, line 189: Why has this been changed? `Color` is in the sRGB color space.

hecrj (changes requested):

hecrj:
Thanks! Not sure what's up with CI. Let me take a look.

core/src/color.rs
hecrj: Ah, I see. It's a type definition! Still, it seems more descriptive to use `Srgb` directly.

hecrj: Reverted it in b741893013eda90045e5a3950c37ba21d073b351 for now.

hecrj (approved):
Looks good to me! Let me know if you agree with the `Srgb` change.

clarkmoody:
Sticking with `Srgb` is fine with me :shipit:

PR #1878, opened 2023-05-27
Minor Scrollable Improvements

AustinMReppert:
This PR provides minor improvements for the Scrollable widget.
* Align vertical and horizontal properties. Vertical properties are now optional and None by default. 
* Prevent mouse wheel scrolling when vertical/horizontal properties are None.

### DIFF
--- a/examples/scrollable/src/main.rs
+++ b/examples/scrollable/src/main.rs
@@ -1,12 +1,11 @@
-use iced::widget::scrollable::{
-    Properties, ScrollbarProperties, Scrollbar, Scroller,
-};
+use iced::widget::scrollable::{Properties, Scrollbar, Scroller};
 use iced::widget::{
     button, column, container, horizontal_space, progress_bar, radio, row,
     scrollable, slider, text, vertical_space,
 };
 use iced::{executor, theme, Alignment, Color};
 use iced::{Application, Command, Element, Length, Settings, Theme};
+
 use once_cell::sync::Lazy;
 
 static SCROLLABLE_ID: Lazy<scrollable::Id> = Lazy::new(scrollable::Id::unique);
@@ -201,7 +200,7 @@ impl Application for ScrollableDemo {
                     .spacing(40),
                 )
                 .height(Length::Fill)
-                .scrollbar_properties(ScrollbarProperties::Vertical(
+                .direction(scrollable::Direction::Vertical(
                     Properties::new()
                         .width(self.scrollbar_width)
                         .margin(self.scrollbar_margin)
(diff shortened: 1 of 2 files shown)


tarkah:
> Prevent mouse wheel scrolling when vertical/horizontal properties are None.

Good catch on this!

AustinMReppert:
What is the purpose of changing the mouse coordinates like this?
https://github.com/iced-rs/iced/blob/fcb1b454368638209862aeb5db41bc5f7d6d51a7/widget/src/scrollable.rs#L467

This will cause strange behaviors when trying scroll with the mouse outside of the window bounds with nested scrollables.
```rust
use iced::widget::scrollable::{Properties, Viewport};
use iced::widget::{Container, Scrollable};
use iced::{
    executor, widget, Application, Background, Color, Command, Element, Length, Settings, Size,
    Theme,
};

    HelloWorld::run(Settings::default())
}

#[derive(Debug, Clone, Copy)]
pub enum Message {
    Scrolled(Viewport),
}

struct HelloWorld {}

impl Application for HelloWorld {
    type Executor = executor::Default;
    type Message = Message;
    type Theme = Theme;
    type Flags = ();

    fn new(_flags: Self::Flags) -> (Self, Command<Self::Message>) {
        (HelloWorld {}, Command::none())
    }

    fn title(&self) -> String {
        String::from("Hello World!")
    }

    fn update(&mut self, _message: Message) -> Command<Self::Message> {
        Command::none()
    }

        Scrollable::new(
            Scrollable::new(
                Container::new("Foo")
                    .style(|theme: &iced::Theme| widget::container::Appearance {
                        text_color: None,
                        background: Some(Background::Color(Color::new(1.0, 0.2, 0.3, 1.0))),
                        border_radius: Default::default(),
                        border_width: 0.0,
                        border_color: Default::default(),
                    })
                    .width(Length::Fixed(5000.0))
                    .height(Length::Fixed(5000.0)),
            )
            .on_scroll(Message::Scrolled)
            .width(Length::Fill)
            .height(Length::Fill)
            .vertical_scroll(Properties::default())
            .horizontal_scroll(Properties::default()),
        )
        .into()
    }

    fn theme(&self) -> Theme {
        Theme::default()
    }
}
```

bungoboingo:
> Align vertical and horizontal properties. Vertical properties are now optional and None by default.

I also messed with making vertical scroll optional when I was adding horizontal support, but couldn't see the usefulness of a situation where a `Scrollable` couldn't scroll. Originally I made an enum that was like:

```rust
enum ScrollDirection {
    #[default]
    Vertical,
    Horizontal,
    Both,
}
```

which might be a better representation of how it's actually being used?

hecrj (changes requested):
@bungoboingo I agree. A `scrollable` that does not scroll does not make sense. Use another widget instead.

Also, we definitely do not want `scrollable` to disable scrolling on both axes by default, which is what this PR is currently doing.

hecrj (approved):
Awesome! Thanks :bow: 

I have just renamed `ScrollbarProperties` to simply `Direction` and simplified some stuff here and there.

We can merge this! :ship:

PR #1888, opened 2023-05-31
Introduce `web-colors` feature flag to enable "sRGB linear" blending

hecrj:
This is how browsers perform color management. They treat gamma-corrected sRGB colors as if they had linear components.

Correctness aside, this mode is introduced for legacy reasons. Most UI/UX tooling uses this color management as well, and many developers and designers have created an intuition about how color should behave from interacting with a browser.

This feature flag should facilitate application development with `iced` in those cases.

More details: https://webcolorisstillbroken.com/

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,8 @@ smol = ["iced_futures/smol"]
 palette = ["iced_core/palette"]
 # Enables querying system information
 system = ["iced_winit/system"]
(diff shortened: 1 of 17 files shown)


hecrj:
Normal vs Web colors

imageimage

bungoboingo:
A necessary evil I think. LGTM

PR #1892, opened 2023-06-01
Add `border_radius` to slider rail

casperstorm:
This small PR adds the ability to adjust the border radius for the slider rail.  
I also updated the default colors of the rail slightly so it use two different colors rather than just one.

Screenshot 2023-06-01 at 10 46 05


hecrj (approved):
Looks good!

Can we make those sliders the default styling too? They look nice.

hecrj:
Changed it myself!

I also fixed this rendering issue in 66d4decc0c5c6ea787ec482ff6841829dfc7ffba:

image

Apparently, our `quad` pipeline doesn't produce the best result when only some borders are being rounded.

casperstorm:
> Changed it myself!
> 
> 
> 
> I also fixed this rendering issue in 66d4decc0c5c6ea787ec482ff6841829dfc7ffba:
> 
> 
> 
> image
> 
> 
> 
> Apparently, our `quad` pipeline doesn't produce the best result when only some borders are being rounded.

Problem with the solution you have is if the handle is very slim, say 1px. Then you see the rounding in the middle where the rectangles meet.

bungoboingo:
I think if quad width is <= 2.0 we should not try to do a border radius, just use the normal quad border algorithm; border-radius of like 5 with a quad width of 2 doesn't make sense imo. Maybe just if quad width < border radius we render as a normal squared border? Or progressively shrink it until it's < 2.0 (in which case there will only be 1 px available for each corner)?

hecrj:
@casperstorm Let's see how common that is. Maybe by the time someone complains we have already fixed the quad glitches.

PR #1902, opened 2023-06-07
Example loading spinners

nicksenger:
This PR adds an example of how to implement animated loading spinner widgets. I went with the material design style because the animation easing curves etc are described in detail, but the colors can be changed through the theme, & the speed+sizing through the example widgets' APIs.

https://github.com/iced-rs/iced/assets/16542714/26cc2763-ad20-4f40-b74a-7f42fcdb408c

TODO:

- [x] I need to investigate whether requesting a redraw unconditionally like this is a viable approach, or if it would be better to determine an appropriate frame-rate and use `RedrawRequest::At()`, which may allow for some throttling/debouncing if necessary.


13r0ck:
Iced aw has some spinners https://github.com/iced-rs/iced_aw/blob/main/src/native/spinner.rs

examples/loading_spinners/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+version = "0.1.0"
+authors = ["Nick Senger <dev@nsenger.com>"]
+edition = "2021"
+publish = false
+
+[dependencies]
+flo_curves = "0.7"
+iced = { path = "../..", features = ["canvas"] }
+iced_core = { path = "../../core" }
+iced_widget = { path = "../../widget" }
```
hecrj, line 12: You should be able to rely solely on `iced` if you enable the `advanced` feature, which will expose the `advanced` module.

hecrj (changes requested):

hecrj:
> I need to investigate whether requesting a redraw unconditionally like this is a viable approach, or if it would be better to determine an appropriate frame-rate and use RedrawRequest::At(), which may allow for some throttling/debouncing if necessary.

Requesting `NextFrame` every time will render at the maximum framerate. Currently, this is quite inefficient since we don't support incremental rendering in `iced_wgpu` yet.

We should be able to efficiently incrementally re-render eventually, however!

examples/loading_spinners/Cargo.toml
nicksenger: Oh cool, I also forgot to remove the flo_curves dependency now that it is instead using lyon for the easing curves. Updated in 56eacdb3583e701cde8cca074e151b24c3fd8df3

nicksenger:
> Requesting `NextFrame` every time will render at the maximum framerate. Currently, this is quite inefficient since we don't support incremental rendering in `iced_wgpu` yet.
> 
> [We should be able to efficiently incrementally re-render eventually](https://whimsical.com/roadmap-iced-7vhq6R35Lp3TmYH4WeYwLM), however!

I see, I've updated to use `RedrawRequest::At` and assume 60fps. Still probably not ideal, but at least it shouldn't go beyond that

examples/loading_spinners/src/circular.rs
```diff
@@ -0,0 +1,470 @@
...
+pub struct Circular<'a, Theme>
+where
+    Theme: StyleSheet,
+{
+    size: f32,
+    bar_height: f32,
+    style: <Theme as StyleSheet>::Style,
+    easing: &'a Easing,
+    cycle_duration: Duration,
+    rotaion_duration: Duration,
```
IohannRabeson, line 34: There is an error here `rotaion` should be `rotation`

nicksenger: Oops, fixed it in 9d69af10ccb4317f01bff32e2e1e58d498b2fc89

hecrj (approved):
Really cool example! Thanks :partying_face:

PR #1907, opened 2023-06-10
Update `resvg` to `0.35` and `tiny-skia` to `0.10`

alec-deason:
This PR upgrades resvg to 0.34 and tiny_skia to 0.10. Most of the changes are minor API adjustments. The one significant change is that resvg removed FitTo so I have rewritten iced's use of that based on what resvg does in it's command line renderer.

Tests pass and all the examples look right to me with a caveat: I'm testing with the wgpu feature turned off because that is not working on my system for reasons I haven't figured out yet. So someone who has that working should validate that things look right in wgpu.

### DIFF
--- a/tiny_skia/Cargo.toml
+++ b/tiny_skia/Cargo.toml
@@ -34,5 +34,5 @@ version = "1.6.1"
 features = ["std"]
 
 [dependencies.resvg]
-version = "0.34"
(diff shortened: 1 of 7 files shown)


hecrj (approved):
Awesome! Thank you :bow:

PR #1910, opened 2023-06-12
Only scroll w/ wheel if over scrollable

tarkah:
Fixes a regression from #1904 where scrollable should only scroll via wheel if mouse is over it.


hecrj (approved):
Good catch! Thanks 🙇

PR #1911, opened 2023-06-12
Add Layout operation to `Container`

13r0ck:
An operation that will return a message to the application with the dimensions of the widget as a `Rectangle`.

`container::get_layout(CONTAINER_ID.clone(), Message::GotLayout)`

The only slightly unusual think about this operation is that creating the operation requires the return message. But this API is the same as `Command::Perform`, and operations are just `Commands`, so I believe this should be fine.

### DIFF
--- a/core/src/widget/operation.rs
+++ b/core/src/widget/operation.rs
@@ -1,5 +1,6 @@
 //! Query or update internal widget state.
 pub mod focusable;
+pub mod layout;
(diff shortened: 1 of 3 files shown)


13r0ck:
I can add the `layout` operation to other widgets before this is merged. Just pushing this now to make sure implementation is ok, before doing the extra work.

PR #1919, opened 2023-06-16
RFC: Improve text measuring in tiny_skia

StratusFearMe21:
It appears that the `Backend.measure` function in `iced_tiny_skia` measures the width of a given `Buffer` correctly, but not it's height. It doesn't take into account characters that could overflow, or characters that are not perfectly within the given font size.

I *believe* that making the changes proposed in this PR would remedy this issue, but I'm not entirely sure. Hence the RFC.

Thank you for your consideration

### DIFF
--- a/tiny_skia/src/text.rs
+++ b/tiny_skia/src/text.rs
@@ -163,14 +163,11 @@ impl Pipeline {
             },
         );
 
-        let (total_lines, max_width) = paragraph
+        paragraph
             .layout_runs()
-            .enumerate()
-            .fold((0, 0.0), |(_, max), (i, buffer)| {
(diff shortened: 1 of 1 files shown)


hecrj:
`line_y` represents the baseline of the line, so that would still not get us the glyph bounds of the paragraph. `cosmic-text` needs to expose this somehow.

Generally, the line height should be high enough to fit all the glyphs displayed. Most fonts should work with a relative line height of `1.2`.

The issue with the software renderer currently is that text isn't being clipped to its bounds, which breaks incremental rendering in some circumstances.

Let's close this and revisit once we figure out how to expose glyph bounds in `cosmic-text`.

PR #1926, opened 2023-06-21
Made Color struct serializable

dsrich:
I needed to Serialize the color struct for configuration files. Not sure why the forbid rust 2018 idioms thing caused it to fail, because I changed that to warn and nothing came up, but that is what I had to do to make it compile


hecrj:
There's no need for us to pull `serde` as a dependency.

Instead, you can leverage [the `from` and `into` container attributes](https://serde.rs/container-attrs.html).

PR #1928, opened 2023-06-23
Fix iced build when using vendored dependencies

tarkah:
Iced can't be built as a vendored dependency (`cargo vendor`) since there is a cross crate include_bytes reference. This fixes that issue by duplicating the font file and making it accessible from each crate needing it.

This relates to https://github.com/iced-rs/winit/pull/11 in my efforts to successfully build an iced app w/ Flatpak. Both of these patches unblock and I can get it successfully working. A new PR will be opened to update the winit dep once that is merged.


mtkennerly:
Out of curiosity, what's your Flatpak build definition? I haven't had any issues with this in [Ludusavi](https://github.com/mtkennerly/ludusavi) (Flatpak: [com.github.mtkennerly.ludusavi.yaml](https://github.com/flathub/com.github.mtkennerly.ludusavi/blob/master/com.github.mtkennerly.ludusavi.yaml)), so I'm wondering what's different.

tarkah:
> Out of curiosity, what's your Flatpak build definition? I haven't had any issues with this in [Ludusavi](https://github.com/mtkennerly/ludusavi) (Flatpak: [com.github.mtkennerly.ludusavi.yaml](https://github.com/flathub/com.github.mtkennerly.ludusavi/blob/master/com.github.mtkennerly.ludusavi.yaml)) - granted that I'm not using `cargo vendor`, but it wasn't necessary in my case.

Looks like it's essentially the same. This issue however is from the recent refactor of all the iced crates. So if you're targeting 0.9 or earlier and not master, that's probably why it works for you.

hecrj (approved):
Thanks!

PR #1931, opened 2023-06-25
Minor grammar correction

GyulyVGC:
This morning I was improving the issue templates of my repo, taking inspiration from your bug report template.
While doing so, I noticed a minor grammar error.
The correct version in English is _Operating_ System.


hecrj (approved):
Thanks! My Spanish is leaking 😅

GyulyVGC:
In the past I used to do the same error, since also in Italian it is "Sistema Operativo" 😄

PR #1932, opened 2023-06-26
Backend-specific primitives

hecrj:
This PR introduces a new `Custom` generic type and variant to `Primitive` in `iced_graphics`.

Graphical backends can now define their custom primitive type. As a result, we can properly decouple the `tiny-skia` dependency from `iced_graphics`.

This should also lay the foundations for choosing a completely custom, unofficial renderer at runtime.

### DIFF
--- a/graphics/src/backend.rs
+++ b/graphics/src/backend.rs
@@ -10,6 +10,7 @@ use std::borrow::Cow;
 ///
 /// [`Renderer`]: crate::Renderer
 pub trait Backend {
+    /// The custom kind of primitives this [`Backend`] supports.
     type Primitive;
 }
 
--- a/graphics/src/backend.rs
+++ b/graphics/src/backend.rs
@@ -6,6 +6,13 @@ use iced_core::{Font, Point, Size};
 
 use std::borrow::Cow;
 
+/// The graphics backend of a [`Renderer`].
+///
+/// [`Renderer`]: crate::Renderer
+pub trait Backend {
+    type Primitive;
+}
+
 /// A graphics backend that supports text rendering.
 pub trait Text {
     /// The icon font of the backend.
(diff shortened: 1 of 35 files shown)


graphics/src/backend.rs
```diff
@@ -6,6 +6,13 @@ use iced_core::{Font, Point, Size};
 
 use std::borrow::Cow;
 
+/// The graphics backend of a [`Renderer`].
+///
+/// [`Renderer`]: crate::Renderer
+pub trait Backend {
+    type Primitive;
```
bungoboingo, line 13: Not sure if we want documentation on this associated type for those implementing their own `Backend`

graphics/src/lib.rs
```diff
@@ -8,8 +8,8 @@
     html_logo_url = "https://raw.githubusercontent.com/iced-rs/iced/9ab6923e943f784985e9ef9ca28b10278297225d/docs/logo.svg"
 )]
 #![deny(
-    missing_debug_implementations,
-    missing_docs,
+    //missing_debug_implementations,
+    //missing_docs,
```
bungoboingo, line 12: Meant to leave this commented?

hecrj: Nope, couple things left I want to do in this PR still!

Thanks for the heads up.

bungoboingo: Gotcha, will hold off on final comments, though it's looking great so far! Will definitely make things less awkward for Custom Shader [RFC](https://github.com/iced-rs/rfcs/pull/23) 🥳 Having custom backends be able to define their own primitive types definitely makes the most sense.

hecrj:
Alright! I have decoupled the `Mesh` primitives in fa5650cfd1115e6ccec2ad795cf58fd970d5b43c and wrote the missing docs in 6921564c9f66e8103e19ec658099c5f5c32e8cc5.

I think this is ready to stail! :ship:

PR #1940, opened 2023-07-01
Fix/scrollable touch input stuck

AustinMReppert:
Should fix this scrollable touch input bug. The main "problem" seems to be exiting the event handler on a un-touch/touch loss returning immediately after handling it, even though later in the code the event is handled again. I have defaulted the return value to ignore and only return immediately when absolutely necessary, so that the various portions of the event handlers do not have to "fight" over the event.

https://github.com/iced-rs/iced/assets/29180043/19879875-ed63-437c-817f-311947c3f6be


hecrj (approved):
Thanks!

PR #1945, opened 2023-07-08
add platform specific settings for x11

addap:
I use the i3 window manager and wanted to make an iced window floating by default by using X11 window type hints.
I added a `PlatformSpecific` struct for unix-like systems so that you can pass the window type to the Application::run function like this
```
MyApp::run(Settings {
            platform_specific: iced::window::PlatformSpecific {
                x11_window_type: vec![
                    winit::platform::x11::XWindowType::Notification,
                    winit::platform::x11::XWindowType::Utility,
                ],
            },
        },
        ..Settings::default()
    })
```
This makes the image floating (with some extra settings not shown to set the size and position)
image

I would also add the rest of the fields of winit's [PlatformSpecificWindowBuilderAttributes](https://github.com/rust-windowing/winit/blob/4652d4810548cb48c89964e7b81f8fc59686a4d3/src/platform_impl/linux/mod.rs#L88) for linux. But before I continue I wanted to get some clarification about conditional compilation.

### DIFF
--- a/winit/src/settings.rs
+++ b/winit/src/settings.rs
@@ -7,6 +7,16 @@ mod platform;
 #[path = "settings/macos.rs"]
 mod platform;
 
+#[cfg(any(
+    target_os = "linux",
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "openbsd",
+))]
+#[path = "settings/unix.rs"]
+mod platform;
+
 #[cfg(target_arch = "wasm32")]
 #[path = "settings/wasm.rs"]
 mod platform;
@@ -14,6 +24,11 @@ mod platform;
 #[cfg(not(any(
     target_os = "windows",
     target_os = "macos",
+    target_os = "linux",
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "openbsd",
     target_arch = "wasm32"
 )))]
 #[path = "settings/other.rs"]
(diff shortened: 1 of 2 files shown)


winit/src/settings.rs
```diff
@@ -158,12 +173,16 @@ impl Window {
         ))]
         {
             // `with_name` is available on both `WindowBuilderExtWayland` and `WindowBuilderExtX11` and they do
-            // exactly the same thing. We arbitrarily choose `WindowBuilderExtWayland` here.
-            use ::winit::platform::wayland::WindowBuilderExtWayland;
+            // exactly the same thing. We arbitrarily choose `WindowBuilderExtX11` here.
+            use winit::platform::x11::WindowBuilderExtX11;
```
addap, line 177: iced_winit defines features `x11` and `wayland` but it assumes that they are always set. Afaik they will always be set if iced_winit is used as an indirect dependency of iced. 
But is that intended or should users of iced be able to disable one of these features? If so, we'd also need to put the use here behind a #[cfg()]

winit/src/settings.rs
```diff
@@ -7,13 +7,28 @@ mod platform;
 #[path = "settings/macos.rs"]
 mod platform;
 
+#[cfg(any(
+    target_os = "linux",
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "openbsd",
+))]
```
addap, line 16: I used this to restrict the target to unix-like systems. This is what winit used until version 0.27. Now they are using a [`free_unix`](https://github.com/rust-windowing/winit/blob/4652d4810548cb48c89964e7b81f8fc59686a4d3/build.rs#L16) which I think includes more systems but `free_unix` is implied by this configuration predicate.
I guess it is okay to use a more restricted configuration predicate than winit?

hecrj:
Since `window::Settings` now lives in `iced_core`, we cannot re-export `winit` types there.

I think we should tackle the use case directly (i.e. floating windows). Using `Notification` and `Utility` types seems like a hack, however.

PR #1949, opened 2023-07-09
Forward redraw requests in component overlay

nicksenger:
Currently the redraw request is not forwarded from a component's overlay to the containing shell, which prevents use of animations in component overlays. This PR resolves the issue by forwarding the redraw request.


hecrj (approved):
Good catch, as always! Thanks!

PR #1952, opened 2023-07-11
[Feature] 🌫️ Gaussian Blur

bungoboingo:
## 🌫️ This PR aims to bring blur capabilities to Iced

https://github.com/iced-rs/iced/assets/4241774/3d01288b-3342-4fb6-9f3f-d44ff234e35d

This PR also introduces very basic **sub-surface compositing support** for the `iced_wgpu` backend. Note that tiny-skia appears to have no support for blurring, so the plan is to not support it at this time.

## Implementation details
Quite a few changes were needed in our rendering backend, as it was set up with only the viewport's texture as a render attachment in mind!

### The `blur` widget
A `Blur` widget can be used to wrap any `Element`, like so:

```rust
let blurred_content = blur(10, content);
```

Where `10` here represents the blur "radius". I chose to implement it as its own widget vs an appearance attribute (like how it's done on web) to allow for the most flexibility. Feedback welcome!

### A new `iced_wgpu::layer::Kind` **
This could be done any number of ways; first I tried just adding a blur option to the existing layer which worked fine, but thought this approach was more readable.

I've adjusted `iced_wgpu::Layer` to have a `kind` field, `Immediate` or `Deferred` (naming suggestions welcome :P). A `layer::Kind::Immediate` is rendered to the viewport's surface immediately. A `layer::Kind::Deferred` must first render everything within the layer to its own `Surface`, and that will be composited onto the viewport's surface later.

** This is very likely to be completely reworked after layering changes!

### The `Composite` pipeline (name TBD)
This is just a general purpose pipeline which can be used to composite any sub-surface onto the viewport's texture. The general idea is that if we want to do any sort of layer `Effect` we first must render it to a `composite::Surface`, and that will be used to sample from in the `Composite` pipeline. In terms of blur, this is necessary as it is much faster to do each blur pass (horizontal & vertical) on a downsampled texture and then upsample it back up to the proper size.

I abstracted it out into its own pipeline as it's more general purpose than just for blur, but originally had it as just part of the blur pipeline. If we don't think this is useful could move it back again! 🤷 

That's about it in terms of details. Max blur time for large images caps out at around ~150µs on my various rigs which I believe should be fast enough for its use case in Iced 😉

### Outstanding TODOS
- [ ] move Primitive::Effect to be a custom primitive of `iced_wgpu` only since it can't be used with `tiny-skia`; just need to figure out how best to do this 🧐 
- [ ] Different blur types? Do we think this is necessary? Could do a kawaise option, or box blur. Didn't really see the use-case for a single-direction motion blur but that could also be done.
- [ ] Test changes I had to make to Glyphon to see if there's a better way to do this.
- [ ] Revisit transformation math to avoid some unnecessary multiplication, of which I'm sure there's a few.

### Outstanding Questions
- [ ] Does it make sense for content which is blurred beyond a certain amount to be interactable? If I blur a button by like 100 radius for example, and then try to click the vaguely blurred outline of it, is sort of a weird experience.

Let me know what you think! Note again that the internals of this implementation will probably change quite a bit after upcoming layering changes to Iced.


hecrj:
I think we can close this since it seems we may be able to implement blur on top of the `shader` widget relatively easily (see https://github.com/derezzedex/iced_blur) in the future.

PR #1953, opened 2023-07-13
Expose methods to change viewport alignment

tarkah:
As discussed in #1912, consumers need to be able to calculate new viewport offsets when swapping between alignments.

This exposes two new `with_<direction>_alignment` methods. If user wants to keep the scrollable's viewport anchored in the same position while swapping alignments, they can call `viewport.with_alignment(new_alignment)` and return the new viewport's `relative_offset` to `scroll_to`.

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -554,7 +554,14 @@ pub fn update<Message>(
 
             state.scroll(delta, direction, bounds, content_bounds);
 
-            notify_on_scroll(state, on_scroll, bounds, content_bounds, shell);
+            notify_on_scroll(
+                state,
+                on_scroll,
+                bounds,
+                content_bounds,
+                direction,
+                shell,
+            );
 
             return event::Status::Captured;
         }
@@ -592,6 +599,7 @@ pub fn update<Message>(
                             on_scroll,
                             bounds,
                             content_bounds,
+                            direction,
                             shell,
                         );
                     }
@@ -637,6 +645,7 @@ pub fn update<Message>(
                         on_scroll,
                         bounds,
                         content_bounds,
+                        direction,
                         shell,
                     );
 
@@ -672,6 +681,7 @@ pub fn update<Message>(
                         on_scroll,
                         bounds,
                         content_bounds,
+                        direction,
                         shell,
                     );
                 }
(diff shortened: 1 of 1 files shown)


hecrj (commented):
I have simply introduced an `absolute_offset_reversed` method instead. Will that work?

widget/src/scrollable.rs
```diff
@@ -1095,6 +1095,22 @@ impl Viewport {
...
+    /// alignment reversed.
+    ///
+    /// This method can be useful to switch the alignment of a [`Scrollable`]
+    /// while maintaining its scrolling position.
+    pub fn absolute_offset_reversed(&self) -> AbsoluteOffset {
+        let AbsoluteOffset { x, y } = self.absolute_offset();
+
+        AbsoluteOffset {
+            x: ((self.content_bounds.width - self.bounds.width).max(0.0) - x)
+                .max(0.0),
```
tarkah, line 1108: Nit: you shouldn't need both max since x will never be more than content - bounds.

tarkah (commented):
Looks good, @hecrj. Thanks!

hecrj:
Eventually, we should probably expose a variant of the `Scrollable` widget that lets the user completely manage the offset.

hecrj (approved):
Cool! Thanks again :bow:

tarkah:
> Eventually, we should probably expose a variant of the `Scrollable` widget that lets the user completely manage the offset.

At what point should we prefer widget operations over managed app state?

The same can probably be said for text input widget.

As for composing widgets, a parent widget can still do widget operations within on_event right? It's not just the application that can do these.

hecrj:
> At what point should we prefer widget operations over managed app state?

When commands are regularly needed to keep a widget in sync with some application state, then that's a sign there are multiple sources of truth.

With managed app state you can derive any scrolling state from your application state without necessarily storing offsets at all.

> The same can probably be said for text input widget.

Yes, a text input with a fully managed cursor has a bunch of valid use cases.

> As for composing widgets, a parent widget can still do widget operations within on_event right? It's not just the application that can do these.

Yes, although I'm not sure that'd be very useful. You'd want the unmanaged widget to depend on the managed one, and the managed one won't expose operations ideally.

tarkah:
Agreed! The need for managed state satisfies both use cases.

The text input seems like a good test case for this since it really can go either way.

Would you prefer two separate widgets, or one widget that exposes a `managed` builder for taking state provided by the application and internally it can switch between that or tree state?

hecrj:
Two separate widgets; one built on top of the other.

PR #1955, opened 2023-07-13
Dont blink input cursor when window loses focus

tarkah:
The text input cursor animation should not run and the cursor should not be rendered when the window loses focus. This is standard behavior across all other applications.


hecrj (approved):
Sweet! Good catch!

PR #1956, opened 2023-07-15
Add viewport to `Widget::on_event`

tarkah:
As a widget creator, I want to be able to access viewport in `on_event` to understand if the widget is within view or not. This will allow me to cull expensive event logic for widgets not visible. It will also allow me to create new widgets that can emit messages when they become visible in the viewport (during `RedrawRequested`).

This specifically allows for creating some nice optimizations and behavior around elements inside a `scrollable`.


hecrj (approved):
Cool! I just needed this!

Thanks :partying_face:

PR #1960, opened 2023-07-20
Update `CHANGELOG`

hecrj:
`0.10` is almost here :partying_face: 

We are currently waiting for a new [`glyphon`](https://github.com/grovesNL/glyphon) release, which should happen soon!


lufte:
That change log looks yummy, I've been waiting for many of those improvements. Thanks for the hard work!

PR #1963, opened 2023-07-21
Add app id setting for linux

tarkah:
This adds a new platform specific window setting on linux to set `application_id`. This allows the user to reliably tie their application to the XDG spec for `.desktop` file, where icon can be specified. 

Without this, the application id (on X11, this is WMClass) is set to the executable name so renaming the application breaks proper `.desktop` usage.

### DIFF
--- a/winit/src/settings.rs
+++ b/winit/src/settings.rs
@@ -7,6 +7,10 @@ mod platform;
 #[path = "settings/macos.rs"]
 mod platform;
 
+#[cfg(target_os = "linux")]
+#[path = "settings/linux.rs"]
+mod platform;
+
 #[cfg(target_arch = "wasm32")]
 #[path = "settings/wasm.rs"]
 mod platform;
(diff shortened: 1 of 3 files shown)


winit/src/settings.rs
```diff
@@ -192,6 +196,28 @@ impl Window {
...
+            {
+                use winit::platform::wayland::WindowBuilderExtWayland;
+
+                window_builder = window_builder.with_name(
+                    &self.platform_specific.application_id,
+                    &self.platform_specific.application_id,
+                );
+            }
+        }
+
```
tarkah, line 220: This technically sets the same field twice by default since both features are enabled. I confirmed they both set the same field internally, but I felt this was easier to read than 3 different sets of `cfg` attributes to determine which import is used.

hecrj (approved):
Looks good! Thanks :bow:

PR #1964, opened 2023-07-21
[Feature] 🪟 Multi Window 🪟 .. redux!

bungoboingo:
This PR introduces support for multiple windows!

Branched off of @derezzedex's old PR (closes #1439), which implements [RFC #8](https://github.com/iced-rs/rfcs/pull/8).

**Note** that much of the code in this PR is from his branch, updated to Iced master from 0.4, with some added bits on top and small tweaks/changes.

https://github.com/iced-rs/iced/assets/4241774/78f68725-8996-4c83-9528-600363e90840

### 🆕  A new `multi_window::Application` trait
The goal is to be as non-invasive as possible when adding support for multiple windows for existing Iced users. Users can now implement the `multi_window::Application` trait for their applications if they want the capability of having more than one window.

This new trait adds a `window::Id` parameter to several methods of the familiar `Application` trait, e.g.

```rust
fn view(&self, window: window::Id) -> Element<Message>;
```

Here a user can choose which `Element` to return for each window, based on its `window::Id`. A `window::Id` is a unique identifier for a window, created when the window is first spawned. Users can store this ID and use it to reference specific windows. This is different from winit's `WindowId`. Internally, we reserve `window::Id::MAIN` to represent the first window spawned.

### 👐  Creating a window
Users can create a new window with the `window::spawn` command, which takes a `window::Id` and `window::Settings`.

```rust
window::spawn(window::Id::new(my_window), window::Settings::default());
```

### 🌂  Closing a window
A user can close a specific window with the `window::close(id)` command. I've utilized the existing `exit_on_close_request` field of `iced::Settings` to determine whether or not to close the entire application when the "main" or first window is closed. Not sure if this is wanted or needed behavior, but this `exit_on_close_request` doesn't make much sense in multi window otherwise. Could do a different set of settings, or just ignore it!

### 🤔 New Window Events
Users can choose to listen to new window events implemented for multi window, such as `window::Event::Created` which is emitted after the window is finished being created by winit, or `window::Event::Destroyed` which is emitted after the window is fully destroyed. There is also `window::Event::CloseRequested`, which can be subscribed to to close the window with the `window::close` command.

**🚤   Known areas to improve performance:** 
- Right now I am making one instance of a `Renderer` per-window, each with their own `Backend`. This is very bad! We should be sharing a single `Backend` across multiple `Renderer`s. This will take some refactoring to our `Compositor` and our `Backend` to make sure that all the caching is refactored with multiple surfaces in mind.
- There is currently no way to know what window needs to be updated based on the result of an application update. The plan is later to introduce some way to get a list of windows that need to be updated from `multi_window::Application::update`. This requires us to rebuild all user interfaces when processing any message, regardless of if that message affects all windows or just one
- We currently have no way of knowing what window needs to be redrawn, so we redraw all windows every frame. With upcoming changes to widgets to manually request to be redrawn on state changes, this problem should be solved.

Other suggestions very welcome and appreciated!

**👻  Commit history..**
Yes it's a mess. I tried to rebase from the original PR, but that was done before the great Purification which changed nearly all of the codebase, and it was a real struggle. If we care enough I will get back in the trenches and try to sort out the commit history, but for now it's a bit of a frankenstein.

**TODOs:**
- [ ] Test multi-window `Program` without a `multi_window::Application` trait. Not sure if we even want this behavior though.
- [ ] Does it make sense to limit this to non-wasm targets..? I assume yes but not sure if there is a use case for it.

Tested on latest MacOS, PopOS & Windows 10.


examples/events/src/main.rs
```diff
@@ -55,8 +55,9 @@ impl Application for Events {
                 Command::none()
             }
             Message::EventOccurred(event) => {
-                if let Event::Window(window::Event::CloseRequested) = event {
-                    window::close()
+                if let Event::Window(id, window::Event::CloseRequested) = event
+                {
+                    window::close(id)
```
bungoboingo, line 60: This is currently the only way to close a window if a user manually closes it with whatever OS-specific close button there is; listen to the `CloseRequested` event, and then manually return the `window::close` command in application update. Not sure if we want a more "opt-in" way to respond to `CloseRequested` events. Maybe closing the window automatically should be the default? Could repurpose the `exit_on_close_requested` setting to indicate this behavior maybe.

bungoboingo:
With 83c7870c569a2976923ee6243a19813094d44673 I made `exit_on_close_request` a field of `window::Settings` instead of an application-level setting. I think this makes more sense in the context of multi-window; now every window can choose whether or not it closes on CloseRequests, or must be closed manually. I can see the benefit of this in an application where let's say you have a few windows, and a settings window. You don't care about the other windows, but you do care that a user saves their settings before close.

The only bit of weirdness is that on MacOS, doing a force quit fires no events besides `winit::event::Event::LoopDestroyed`, which is the last guaranteed event to be fired, so there's no change of propagating a `CloseRequested` down to the user. On windows, alt + f4 does fire a `CloseRequested` on the window the keyboard input was recorded in, but sometimes when pressing alt+f4 in the MW example this causes the application to stall/crash. Still debugging this.

Wondering what behavior we want when a user force quits an app. Should we just ignore `exit_on_close_requested: false` settings and break the main application loop?

examples/events/src/main.rs
bungoboingo: Changed to make more sense in 83c7870c569a2976923ee6243a19813094d44673

core/src/window/id.rs
```diff
@@ -0,0 +1,21 @@
...
+/// Internally Iced reserves `window::Id::MAIN` for the first window spawned.
+pub struct Id(u64);
+
+impl Id {
+    /// The reserved window [`Id`] for the first window in an Iced application.
+    pub const MAIN: Self = Id(0);
+
+    /// Creates a new unique window [`Id`].
+    pub fn new(id: impl Hash) -> Id {
+        let mut hasher = DefaultHasher::new();
```
ids1024, line 16: I am somewhat concerned about how Iced is using hashes like this. Generally to avoid collisions you'd want a 128-bit hash like MD5 (which still isn't robust against malicious inputs; but that may generally not be an issue for window IDs anyway).

At least for window IDs there are likely to not be very many of them.

Using an incrementing counter would be much more reliable, but wouldn't offer any easy way to statically allocate IDs. (Though 0 could still be reserved for the main window.)

bungoboingo: Can you think of a use case where this would actually be an issue? iirc the `DefaultHasher` is siphash 1-3, which would need an average of 2^32 hashes before having a collision, unless I'm missing something. I will admit my knowledge of hashing is not the best.

ids1024: Probably it's more a concern with something like the hashing used for image `Handle`s than here. In that case, presumably it would be quite possible for a malicious actor to generate images that collide, which may be an issue in some applications.

Maybe it's okay here.

bungoboingo: I'm fine with switching to an incrementing internal window ID count as well, but I believe there were some problems when it was originally tried (..? might be misremembering). I can look into it more 👍 Also fine with switching to a 128 bit hashing output if there are security concerns since storage here is pretty negligible.

genusistimelord: to get away from hashes you would need a generational array that would store the actual data and give a GenerationalIndex. This does work quiet well but it a much different approach from how the hashing version works. In all likely hood the hashing inside of the application shouldn't be susceptible to outside interference, unless They already have full control of your system or for WASM its loading some things from other websites that could become compromised. 

Hashing can have overlaps but based on how many and how the hashes are being stored and at what representation level. This in a UI library should not have to many issues. Though moving to a GenerationalIndex would solve some other types of issues but would lack in the "reuse of" issues as you would still need to Hash things in order to know if it already exists or not. 

Now if your worried about a bad actor Getting your installer and implementing Compromised images and then re-uploading for people to download. there is not much we can do here since they can reverse engineer and change aspect of the software And any image they do replace will just be as if it exists since we don't Store the hashes until runtime when we load everything. Even if we did he could just replace said stores with reverse engineering the source.

hecrj:
Alright! Finally got this to the finish line! Great stuff, @bungoboingo! :partying_face: 

I have refactored a bunch of stuff here and there. Most important changes are:

- The `Event` in the `multi_window` shell has been split into `Event` and `Control` in d34bc4e4a251bb28854770575d379d4a53f2db12. We avoid the awkward `unsafe impl Send` and some `unreachable!`, as well as the weird nesting inside the `Proxy` type.
- `window::Id` has been changed to use an opaque `AtomicU64` in ea42af766f345715ff7a7168182d3896ee79cfbc. `window::spawn` initializes and returns a brand new unique `window::Id` alongside the `Command`. This way, we avoid weird scenarios where a user could spawn a window with the same identifier twice.
- The `Windows` abstraction has been completely refactored into a `WindowManager` in 5c5e7653bed248ba63faa6563e4d673e4441415e. Instead of having the state distributed in multiple `Vec`, the new abstraction uses a single `BTreeMap` to store all the state for each `Window` and another one for aliasing `winit` identifiers.

And I think that's it! I have most definitely changed some of the behavior (i.e. `Closed` event returned right after a `CloseRequested` is handled), but I believe it's still somewhat intuitive and consistent.

Let me know what you think! And let's ship this! :ship:

bungoboingo:
LGTM, let's 🚢 it!!!

varbhat:
@hecrj @bungoboingo ,

> We currently have no way of knowing what window needs to be redrawn, so we redraw all windows every frame. With upcoming changes to widgets to manually request to be redrawn on state changes, this problem should be solved.

Will all windows redraw on state change? I think this is one thing i dislike and which i think can be improved. I hoped for selective window redraw.

bungoboingo:
> @hecrj @bungoboingo ,
> 
> > We currently have no way of knowing what window needs to be redrawn, so we redraw all windows every frame. With upcoming changes to widgets to manually request to be redrawn on state changes, this problem should be solved.
> 
> Will all windows redraw on state change? I think this is one thing i dislike and which i think can be improved. I hoped for selective window redraw.

Currently all windows are being redrawn due to existing limitations; there are plans in the near-ish future to have widgets explicitly request to be redrawn on state changes, which will fix this behavior. Lots of iterations will be made for multi-window performance increases, like shared rendering resources. They will all be added in time!

PR #1969, opened 2023-07-25
Access proxy

wiiznokes:
This could be useful to execute something which already have a runtime. This don't works and this is just a poc.
I spent a lot of time fighting against the type system but I wonder if it's even possible to do it this way. 
If anyone is interested, I could use a little help.


genusistimelord:
could you maybe make a working example and place it in the Example directory to display what this feature would be useful for?

wiiznokes:
@genusistimelord i made the example. I also made a little explanation in [readme](https://github.com/iced-rs/iced/blob/0a8d3643aabbf3d57bace88a0e693563e5369e4f/examples/proxy/README.md)

stappersg:
image

Run `cargo fmt --all -- --check --verbose`

image

Run `cargo clippy --workspace --all-features --all-targets --no-deps -- -D warnings`

wiiznokes:
thanks but this doesn't even compile.

PR #1971, opened 2023-07-26
`visible_bounds` operation for `Container`

hecrj:
This PR introduces a brand new `visible_bounds` operation to `container` which can be used to obtain the visible screen bounds of a `Container` widget with an `Id`.

This operation can be very useful when combined with the `window::screenshot` command.

https://github.com/iced-rs/iced/assets/518289/e2280009-fdd0-447b-929b-cb1ee4b37843

Closes #1911.


bungoboingo:
Looks good, tested it in an example with screenshots and bounds all seem correct, in scrollables too! :partying_face: Only issue with the workflow is that `Screenshot`'s size is in physical pixels bc it's based off the underlying texture size, and this operation returns logical. I think the best way to handle it would be to include the `scale_factor` in the `Screenshot` data, and then for the `crop` method just do the scaling for the user. I'll pop up a PR for it now.

hecrj:
I think including `scale_factor` in `Screenshot` makes sense. Everything in `iced` is in logical units.

PR #1989, opened 2023-07-31
Fix wrong color channels

Joylei:
`tiny-skia` renderer does not work as expected due to wrong color channels.


Joylei:
Well, it's not color issue,.
Need to find out why canvas does not work properly to draw lines or backgrounds.

PR #1991, opened 2023-07-31
cycle `combo_box` with keybinds

casperstorm:
A few keybinds optimizations for `combo_box`:

1. When pressing <kbd>↑</kbd> at the very first entry you cycle to bottom
2. When pressing <kbd>↓</kbd> at the very last entry you cycle to top
3. Add support for <kbd>Tab</kbd> and <kbd>Shift</kbd>+<kbd>Tab</kbd> which does the same as  <kbd>↓</kbd> and <kbd>↑</kbd>.

### DIFF
--- a/widget/src/combo_box.rs
+++ b/widget/src/combo_box.rs
@@ -529,7 +529,7 @@ where
                         {
                             if let Some(index) = &mut menu.hovered_option {
                                 if *index
-                                    == state
+                                    >= state
                                         .filtered_options
                                         .options
                                         .len()
--- a/widget/src/combo_box.rs
+++ b/widget/src/combo_box.rs
@@ -468,11 +468,13 @@ where
 
                 if let Event::Keyboard(keyboard::Event::KeyPressed {
                     key_code,
+                    modifiers,
                     ..
                 }) = event
                 {
-                    match key_code {
-                        keyboard::KeyCode::Enter => {
+                    let shift_modifer = modifiers.shift();
+                    match (key_code, shift_modifer) {
+                        (keyboard::KeyCode::Enter, _) => {
                             if let Some(index) = &menu.hovered_option {
                                 if let Some(option) =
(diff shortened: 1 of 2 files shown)


widget/src/combo_box.rs
```diff
@@ -511,15 +523,28 @@ where
...
-                        keyboard::KeyCode::Down => {
+                        (keyboard::KeyCode::Down, _)
+                        | (keyboard::KeyCode::Tab, false)
+                            if !modifiers.shift() =>
+                        {
                             if let Some(index) = &mut menu.hovered_option {
-                                *index = index.saturating_add(1).min(
-                                    state
+                                if *index
+                                    == state
```
hecrj, line 532:
```suggestion
                                if *index
                                    >= state
```

Maybe?

hecrj (approved):
Awesome! Thanks 🎉

mtkennerly:
> Add support for Tab and Shift+Tab which does the same as ↓ and ↑.

Would this interfere with using tab to cycle through elements in a form, when that form includes a combobox?

hecrj:
> > Add support for Tab and Shift+Tab which does the same as ↓ and ↑.
> 
> Would this interfere with using tab to cycle through elements in a form, when that form includes a combobox?

As long as the form handler honors the `event::Status`, it should be fine. These interactions are capturing the events.

mtkennerly:
> As long as the form handler honors the event::Status, it should be fine. These interactions are capturing the events.

The fact that they're captured is why I was concerned 😅 If you honor the status, but you generally use tab for `focus_next`, then how would you cycle focus out of a combobox?

One solution might be:

* click in combobox -> triggers overlay -> tab cycles through options in overlay
* escape closes overlay -> tab cycles focus past combobox

However, at the moment, escape also clears the cursor in the combobox and removes focus, so `focus_next` would lose its place.

PR #1998, opened 2023-08-02
fix: Check cursor in bounds when scrolling image::Viewer

### LINKED ISSUE #1997: Image Viewer scrolls with cursor anywhere in the window

### What happened?

Currently the image::Viewer widget captures all vertical scroll events, regardless of where the cursor is within the application. When combined with a touchpad and a horizontal scrolling widget, the slight vertical components within the horizontal scrolling adjust the image::Viewer along with the horizontal scroll widget.

I would expect that the cursor has to be over the widget for scroll events to affect it.

### Version

master

### Operating System

Linux

_No response_

malramsay64:
Ensure that the cursor is within the bounds of the image::Viewer when
performing the scrolling.

Fixes #1997


hecrj (approved):
Nice catch! Thanks 🙇

PR #2000, opened 2023-08-03
Fix `Command<T>::perform` to return a `Command<T>`

ids1024:
This seems like clearly the correct thing to do here. If the type bound on `Command` isn't specified, it makes no difference, since the generic is inferred in a way that works with either definition. But this is important if `Command<T>` is aliased with a concrete type.


wash2 (approved):

hecrj (approved):
Makes sense! Thanks!

PR #2003, opened 2023-08-04
Fix overlay group event handling

Remmirad:
Currently the overlay group runs ` on_event` for all its children without checking if one of them already handled the event. 

This PR simply fixes that by checking the result of each ` on_event` call before making the next. I am not sure if this is intentional or not, but since `nested.rs` seems to care about checking the returned status for overlays I thought it should be the same here.

https://github.com/iced-rs/iced/blob/cb8b70bec3e8bbf809e7d8ffc559adb712f45e14/runtime/src/overlay/nested.rs#L210)

The children are iterated in reverse drawing order so that overlays in the front get the first chance to handle the event.

### DIFF
--- a/core/src/layout.rs
+++ b/core/src/layout.rs
@@ -54,7 +54,9 @@ impl<'a> Layout<'a> {
     }
 
     /// Returns an iterator over the [`Layout`] of the children of a [`Node`].
-    pub fn children(self) -> impl Iterator<Item = Layout<'a>> {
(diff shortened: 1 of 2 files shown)


hecrj:
`Group` is supposed to work like a `Column` or `Row`, which don't currently have this behavior. It's unclear which use case this is exactly tackling, so let's not add unnecessary complexity!

PR #2008, opened 2023-08-07
Update arc.rs with improved documentation

ua-kxie:
added units and improved description to field comments of Arc and Ellipse structs.


hecrj (approved):
Thanks!

Maybe we should think about creating an opaque `Radians` type eventually, analogous to our `Pixels` type.

hecrj:
We actually have [a `Radians` type](https://docs.rs/iced/latest/iced/struct.Radians.html) already!

PR #2009, opened 2023-08-07
wgpu: set `layers` to `Layer::Full`

yukibtc:
Hi, updating `iced` to `v0.10` and I noticed that `glow` was removed. Using `wgpu` I'm experiencing issues with svg/images rendering: I see a black box instead of the image.
The solution that I found is to replace `Layer::Empty` with `Layer::Full`. It seems working, but I'm not sure if this can cause any other issues.

### DIFF
--- a/wgpu/src/image/atlas.rs
+++ b/wgpu/src/image/atlas.rs
@@ -55,7 +55,7 @@ impl Atlas {
         Atlas {
             texture,
             texture_view,
-            layers: vec![Layer::Empty],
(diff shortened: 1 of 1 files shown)


genusistimelord:
this seems a bit unorthodox as a fix.  As full is basically saying the image is full so we either Resize the texture or error out. so if full is working for you It could cause it to resize but then shows that your GPU might not support a certain size of textures. which would be weird.

hecrj (commented):
This isn't a proper fix!

PR #2014, opened 2023-08-09
Add font loader for Sandbox, Program and Application.

genusistimelord:
This will add a Sync based font loader that will load all custom fonts placed inside the settings under custom_fonts. 

#Example
```rust
fn main() -> iced::Result {
    let mut settings = Settings::default();
    settings = settings.add_custom_fonts([iced_aw::graphics::icons::ICON_FONT_BYTES]);
    CardExample::run(settings)
}
```


hecrj:
API changes need an RFC or, at the very least, some coordination!

I have other plans for this.

PR #2017, opened 2023-08-11
Bounds Contains update.

### LINKED ISSUE #2010: Two buttons placed next to each other can both be clicked at the same time

### What happened?

If you place two buttons next to each other, you can place the cursor in between the two, and both will be highlighted. If you click, both buttons will receive the `on_press` command. This happens in both GPU and CPU rendering.

This is problematic for my minesweeper game, composed of many buttons touching each other.

My current workaround is to place a space between each button.

It should not be possible for the cursor to be touching both buttons at the same time. It should be either one or the other.

### Version

crates.io release

### Operating System

Windows

_No response_

genusistimelord:
this update should fix the bounds issues when 2 objects are near each other but not actually overlapping each other. 
this fixes #2010

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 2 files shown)


hecrj (approved):
Thanks!

PR #2019, opened 2023-08-12
Support focus/unfocus events for TextInput

lufte:
I've been using this patch in my app to support custom reactions to a text input being focused/unfocused. I'm sharing in it here with the hope it's useful to others. Please let me know if I should update tests or documentation somewhere else.


hecrj:
The issue here is that many things can cause an input to lose focus. One of them being the loss of continuity from one `view` call to another one!

We need to talk use cases first here to understand if these downsides are fine.

lufte:
Fair enough, I started a discussion in #2030 to talk about it.

PR #2020, opened 2023-08-13
Add a way to set the winit window theme

khassar00:
The following text is generated by automatic translation.
Add a way to set the winit window theme, and then update the styling example to make the window theme follow the iced theme.

image
image
I only tested it on Windows 10, but this calls the winit function directly.
related links:[winit pull](https://github.com/rust-windowing/winit/pull/2553) or [docs](https://docs.rs/winit/0.28.6/winit/window/struct.Window.html#method.set_theme)
There are a few areas that could be improved:

1.Setting window themes via winit functions does not take effect immediately on Windows 10.I solved this problem by reenabling window decorations. Looking at the source code for winit is beyond my ability .Users should rarely change the theme, I think it is acceptable.
```
window.set_theme(theme);
window.set_decorations(false);
window.set_decorations(true);
```
2.The window theme needs to be set manually by the developer, which may cause inconsistencies with the iced theme.
Maybe it should be set by application::StyleSheet?
Actually, I'm not familiar with rust and contribution codes.Please tell me what's wrong.

### DIFF
--- a/core/src/window.rs
+++ b/core/src/window.rs
@@ -1,6 +1,7 @@
 //! Build window-based GUI applications.
 pub mod icon;
 
+mod window_theme;
 mod event;
 mod level;
 mod mode;
@@ -9,6 +10,7 @@ mod user_attention;
 
 pub use event::Event;
 pub use icon::Icon;
+pub use window_theme::WindowTheme;
 pub use level::Level;
 pub use mode::Mode;
 pub use redraw_request::RedrawRequest;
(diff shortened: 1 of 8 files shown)


hecrj (changes requested):
This should probably be exposed as an additional field in [`application::Appearance`](https://docs.rs/iced/latest/iced/application/struct.Appearance.html).

Get in touch first before writing any code!

PR #2021, opened 2023-08-13
Softbuffer update 0.2 -> 0.3

payload:
I saw that iced-tiny-skia still uses softbuffer 0.2. Update to 0.3 is a breaking softbuffer API change. To test this I copied the system_information example, renamed it to system_information_tiny_skia and disabled default features to force to build it with iced-tiny-skia softbuffer CPU rendering.

### DIFF
--- a/examples/system_information_tiny_skia/Cargo.toml
+++ b/examples/system_information_tiny_skia/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "system_information"
+version = "0.1.0"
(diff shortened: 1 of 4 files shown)


hecrj:
This patch breaks incremental rendering.

In `softbuffer 0.3`, buffers have a variable [`age`](https://docs.rs/softbuffer/latest/softbuffer/struct.Buffer.html#method.age).

Please, get in touch first before writing any code!

PR #2024, opened 2023-08-14
fix dead link to `iced_native` in `README.md`

NyxAlexandra:


NyxAlexandra:
Looked at the verify test. Is the changelog supposed to be updated for trivial commits like this one?

hecrj (approved):
Thanks!

PR #2025, opened 2023-08-14
Fixed clippy warnings/errors

41Leahcim:


genusistimelord:
maybe also run the nightly Clippy and see if there are any important things to also complete. I know it has some that was missed in the standard clippy.

41Leahcim:
I have nightly as default compiler. So I always get the nightly clippy warnings and errors, unless a project overrides it. Which isn't the case for this project. There are a few warnings I left in, as I didn't know what to do with the private modules in the public Widget module. All other warnings and errors of the nightly version I have updated to on 2023-08-13 or 2023-08-14, have been solved. Some changes may cause problems with other pull-requests as I have changed some mutable references and mutable parameters to be immutable.

hecrj:
We have a `lint` workflow that is currently passing.

We do not keep up with `nightly`, as that is unreasonable.

PR #2034, opened 2023-08-16
[docs] Don't use &mut self for view()

PolyMeilex:
Using `&mut self` for view method could be misleading for people not familiar with ELM


hecrj (approved):
Thanks!

PR #2039, opened 2023-08-18
Update Integration to work on WASM fully.

genusistimelord:
This will add and load a Font for the Integration project which will allow Text to render when running as WASM. This has to be shown since WASM cant load system fonts.


hecrj:
This doesn't seem like the proper fix. If Wasm is truly broken by default, the fix should happen at a different abstraction level.

Discussions before code and all that!

genusistimelord:
generally I did discuss this with @jackpot51 (jermey and  dfrg. https://discord.com/channels/628993209984614400/1027584473631555604/1142122898140696698 The browser doesn't have access to fonts so they can not be loaded the normal way. This is the only way to fix this issue. So adding it to the example is a way to show them how to load the fonts. unless you are thinking of another internal way of preloading select fonts inside of iced. but this would mean iced would have to store these inside the library for WASM.

PR #2041, opened 2023-08-18
Make the style attribute available on Font

### LINKED ISSUE #2040: After latest text refactor it is no longer possible to load an italic font

### What happened?

Back when `Font::External` was used to load fonts it was possible to load italic fonts, but now with the plain `Font` struct there is now way to specify the style of the font. 

It should be possible to load fonts with Italic and Oblique styles

### Version

crates.io release

### Operating System

Linux

_No response_

lufte:
Just a suggested patch to get started.

Fixes #2040


hecrj (approved):
Looks good! Thanks.

PR #2046, opened 2023-08-21
Support automatic style type casting for Buttons.

dtzxporter:
Changes the signature of Button::style to use `impl Into<...>` instead of taking the style sheet itself. (Matches other widgets).


hecrj (approved):
Thanks!

PR #2055, opened 2023-08-29
Compute gradients in Oklab color space

matze:
This is the simple port of https://www.shadertoy.com/view/ttcyRS to compute quad gradients in Oklab as discussed in https://github.com/iced-rs/iced/discussions/2043. I have very little experience with iced internals nor wgsl in general, so don't hesitate to redo as you see fit.


matze:
I added a small example to experiment with the outcome. What I also noticed is that the stop order is somewhat confusing because the first stop at 0.0 does not denote the left but the right side. Not sure if that is intentional or a bug …

GyulyVGC:
> What I also noticed is that the stop order is somewhat confusing because the first stop at 0.0 does not denote the left but the right side. Not sure if that is intentional or a bug …

It depends on the specified angle.
If you want 0.0 to denote the left, you can add/subtract 180 degrees.
I'm agree with you that an angle of zero should have 0.0 on the left though.

hecrj (commented):
We need to enable OkLab gradients only if the `web-colors` feature is enabled.

This can be checked at compile-time with [the `GAMMA_CORRECTION` constant](https://docs.rs/iced_graphics/latest/iced_graphics/color/constant.GAMMA_CORRECTION.html). If it's `true`, the new OkLab shader should be used. If it's `false`, the current shader should be used instead.

hecrj:
Let me give that a shot.

hecrj (approved):
I have split the shader code and used a bit of macro magic to conditionally reassemble it based on the `color::GAMMA_CORRECTION` value.

I have also simplified the example a bit!

Hopefully I didn't break anything. Give it a test, let me know, and we can merge this!

hecrj:
@bungoboingo ^

bungoboingo:
LGTM, tested on MacOS (m1) & windows 10, no issues.

matze:
> Hopefully I didn't break anything. Give it a test, let me know, and we can merge this!

Also from my side, thanks for picking it up. Since you also do the phase shift by pi to have start on the left side, do you consider it … weird as well?

hecrj:
@matze You are right. In the web, it seems an angle of 0 "points up".

@bungoboingo I have fine-tuned/simplified the `Radians::to_direction` logic to reproduce this behavior in 90cbab18b95a2a90a6a527280a6ca461203ea1b3. Does the new logic make sense? Or have I misunderstood the purpose of the helper? I _think_ it works.

bungoboingo:
I'm fine with adjusting it to be more consistent with expectations, I went off this 👇  diagram for current behavior, although maybe OP means it should be flipped? Math appears to be correct to me!

image

hecrj:
@bungoboingo Makes sense! We have to keep in mind that the Y axis is inverted in `iced`, so the angle increasing clock-wise seems correct.

I think the angle defining the `end` position of the gradient makes more sense than the `start`. Now it's just a matter of choosing whether a `0` angle denotes the `(1, 0)` position or the `(0, 1)`. Since it seems trivial, let's just do what CSS does; for familiarity.

PR #2058, opened 2023-08-30
Explicit text caching

hecrj:
This PR enables explicit text caching at the widget level by introducing a new `Paragraph` associated type to `core::text::Renderer`.

Until now, widgets could draw text indirectly by issuing a `Renderer::fill_text` command. The `Renderer` then took care of layout and caching behind the scenes. This caused a bunch of challenges; specifically, hashing, aliasing, and cache invalidation.

With caching, memory management also becomes tricky and less intuitive. If a specific frame needs to measure a lot of intermediate text, then the internal map used for caching may allocate a bunch of memory unnecessarily, as most of the entries may not be reused.

After this PR, a widget can choose to create and manage a `Paragraph`. A `Paragraph` represents a bunch of text that has been laid out and is ready to be rendered at minimum cost. Since it's already shaped and laid out, its bounds are already computed and, therefore, measuring it has no cost. Internally, a `Paragraph` is just a thin wrapper of a `cosmic_text::Buffer`.

The tricky part is that a `Paragraph` needs to be managed. Widgets are responsible to keep it in sync manually with their state. This normally consists in calling `Renderer::update_paragraph` during `Widget::layout`, which will try to reconcile any differences as efficiently as possible (e.g. a change in paragraph bounds can skip shaping altogether!).

`Text`, `TextInput`, `Checkbox`, `Radio`, and `Toggler` use the new managed approach, which I expect should improve performance and memory usage in most cases.

There are some cases where caching is superior, however. Specifically, given that paragraphs are stored in the widget state tree, a small change in the widget tree can cause a bunch of text to be relaid out (potentially all of it!). The cached approach doesn't have this problem because caching happens on a per-frame basis, independently of the widget tree.

However, I have some ideas to improve the diffing strategy to handle most common cases more efficiently. And it's also important to note that a `text::Renderer` does still support the cached strategy, so we could eventually give users some control over the strategy used.

This also sets up some of the foundations necessary for multi-line text editing, as it is now possible to easily query and manipulate text buffers without recalculating (nor hashing) everything.

# Related changes

Besides the `Paragraph` additions, there have been some other changes here and there:

- A new `layout::next_to_each_other` helper. No more awkwardly using `Row` inside widget code!
- `Widget::layout` now takes a mutable `widget::Tree`. This is useful to cache layout data or update managed paragraphs, for instance.
- All text sizes are represented by the `Pixels` type now.
- `core::Text` does no longer have `color`. Instead, the `color` is provided to the `Renderer` when drawing.
- The `bounds` in `core::Text` are now a `Size` instead of a `Rectangle`. The final `position` of the `Text` must be provided when drawing.

# TODOs
- [x] ~~Rename `Paragraph`. Technically, `Paragraph` isn't the right name because the idea it currently represents can contain multiple paragraphs. I'm considering renaming it to something different like `text::Allocation` or `text::Buffer`.~~ I can't find a more descriptive name, so `Paragraph` it is!
- [x] Implement `Paragraph` drawing in `iced_tiny_skia`.
- [x] Implement `Icon` support for `TextInput`.
- [x] Invalidate paragraphs when a new font is loaded.


graphics/Cargo.toml
```diff
@@ -23,6 +23,9 @@ log = "0.4"
 raw-window-handle = "0.5"
 thiserror = "1.0"
 bitflags = "1.2"
+cosmic-text = "0.9"
+rustc-hash = "1.1"
+unicode-segmentation = "1.6"
```
Imberflur, line 28: I can't find where `unicode-segmentation` is used in this crate, maybe it just isn't used yet, pending the TODO with glyph vs graphmeme indices.

widget/src/lazy/component.rs
```diff
@@ -254,11 +254,12 @@ where
 
     fn layout(
         &self,
+        tree: &mut Tree,
         renderer: &Renderer,
         limits: &layout::Limits,
     ) -> layout::Node {
         self.with_element(|element| {
-            element.as_widget().layout(renderer, limits)
+            element.as_widget().layout(tree, renderer, limits)
```
tarkah, line 262: We need to use the same logic for accessing `tree` as in `on_event`, this currently panics when using `Component`.

graphics/src/text/paragraph.rs
```diff
@@ -0,0 +1,246 @@
...
+            horizontal_alignment: self.0.horizontal_alignment,
+            vertical_alignment: self.0.vertical_alignment,
+        }
+    }
+
+    pub fn resize(&mut self, new_bounds: Size, font_system: &FontSystem) {
+        if let Some(internal) = Arc::get_mut(&mut self.0) {
+            // If there is no strong reference holding on to the paragraph, we
+            // resize the buffer in-place
+            internal.buffer.set_size(
```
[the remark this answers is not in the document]
hecrj: Good catch! We definitely want `Arc::into_inner`.

`Weak` references should only be held by the `Primitive` tree, and they will only be invalidated right before creating a new one.

This makes the `Primitive` trees much cheaper to clone for incremental rendering and, since a `paragraph::Weak` holds its original bounds and alignment, it is still possible to figure out its damage region.

Widgets should never be handling a `Weak` primitive directly.

widget/src/lazy/component.rs
hecrj: Good catch! Thanks!

graphics/src/text/paragraph.rs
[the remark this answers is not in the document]
hecrj: `try_unwrap` needs ownership, so we need to obtain it from a mutable reference.

I have implemented an unsafe version with `MaybeUninit` in 935c722a278071d3bcd0bf499a0ef8f687ed51ad, and another safe version with `Option` in 51e69d7040c943aad8453bc03031c75b4cb302bb.

I think the `MaybeUninit` is sound because the memory is only uninitialized while we have unique access to it, but I don't feel very comfortable with unsafe yet.

Let's go with the `Option` approach; I don't think the overhead will be noticeable in practice.

widget/src/lazy/component.rs
hecrj: I have a feeling that the lazy widgets may be able to be simplified a bunch now that we have access to a mutable `Tree` in `layout` (`Overlay::layout` also takes a mutable self now).

widget/src/keyed.rs
```diff
@@ -0,0 +1,55 @@
...
+//! stage, but they involve comparing trees at certain depths and
+//! backtracking... Quite computationally expensive.
+//!
+//! One approach that is cheaper consists in letting the user provide some hints
+//! about the identities of the different widgets so that they can be compared
+//! directly without going deeper.
+//!
+//! The widgets in this module will all ask for a "hint" of some sort. In order
+//! to help them keep continuity, you need to make sure the hint stays the same
+//! for the same items in your user interface between `view` calls.
```
hecrj, line 39: Hey! I wrote some docs for once!

graphics/Cargo.toml
hecrj: Turns out it isn't necessary, since the glyphs that `cosmic-text` gives us already contain string indices representing the clusters. We can just find the right glyph: 956512338905bac0b156fdaf16fe3c3e07e97a84.

Issue #2056, opened 2023-08-30
`column!` not working in styling example


Barugon:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

Here's the result of running the styling example:
![Screenshot from 2023-08-29 17-08-33](https://github.com/iced-rs/iced/assets/16503728/57f49370-c17d-475e-aafa-f41273a93e1d)

Note the missing checkbox and toggler to the right of the scroll area.

[This code](https://github.com/iced-rs/iced/blob/c9bd48704dd9679c033dd0b8588e2744a3df44a0/examples/styling/src/main.rs#L138) doesn't seem to be working properly.

### What is the expected behavior?

Checkbox and toggler should be visible to the right of the scroll area.

### Version

master

### Operating System

Linux

akshayr-mecha:
```
        let scrollable = scrollable(
            column!["Scroll me!", vertical_space(800), "You did it!"]
                .width(Length::Fill),
        )
        .height(100);
```
        
        Length::Fill of column inside scrollable is causing this issue
        as scrollable is not having width defined so column inside it is taking entire width
        that is why checkbox and togger are getting hidden
        Can be patched by adding width or removing Length::Fill
        
```
        let scrollable = scrollable(
            column!["Scroll me!", vertical_space(800), "You did it!"]
                .width(Length::Fill),
        )
        .width(300)
        .height(100);
```

akshayr-mecha:
Created PR #2062, Let's see if they will merge.

Barugon:
This also seems to work:
```Rust
        let scrollable = scrollable(
            column!["Scroll me!", vertical_space(800), "You did it!"]
                .width(Length::Fill),
        )
        .width(Length::Fill)
        .height(100);
```

PR #2062, opened 2023-09-01
fix(examples-styling): fixed checkbox and toggler getting hidden behind scrollbar

akshayr-mecha:
Fix for this issue https://github.com/iced-rs/iced/issues/2056
Column inside scrollable has Length::Fill so it is taking the entire width thus hiding check box and toggler. Added fixed width to scrollable so Lenth::Fill will be relative to fixed width.

### DIFF
--- a/examples/styling/src/main.rs
+++ b/examples/styling/src/main.rs
@@ -108,6 +108,7 @@ impl Sandbox for Styling {
             column!["Scroll me!", vertical_space(800), "You did it!"]
                 .width(Length::Fill),
         )
+        .width(300)    
         .height(100);
 
         let checkbox = checkbox(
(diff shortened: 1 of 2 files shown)


examples/styling/src/main.rs
```diff
@@ -108,6 +108,7 @@ impl Sandbox for Styling {
             column!["Scroll me!", vertical_space(800), "You did it!"]
                 .width(Length::Fill),
         )
+        .width(300)
```
hecrj, line 111: Passing `Length::Fill` here should work as well. Let's do that.

```suggestion
        .width(Length:Fill)    
```

This kind of layout inconsistency will be fixed soon.

hecrj (changes requested):

hecrj (approved):

hecrj:
Thanks!

Issue #2061, opened 2023-09-01
Can't use backend-agnostic third party libraries since version 0.10


Davidster:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

I am writing a wgpu app with an iced overlay. In order to achieve this, I followed the same setup as the integration example. This worked fine under iced 0.9 and I was able to integrate the `plotters-iced` crate, but since 0.10 this integration failed.

I have already discussed this with `plotters-iced` as can be seen here: https://github.com/Joylei/plotters-iced/issues/29.

I created a [repo](https://github.com/Davidster/plotters-iced-repro-case) that reproduces the issue by simply copying the integration example and adding `plotters-iced`, just uncomment [line 65 in chart.js](https://github.com/Davidster/plotters-iced-repro-case/blob/82531bbed53c6d93a61c9907fdd117abcb5ad39a/src/chart.rs#L65) to see the compiler error that is produced.

Since the `impl Program` setup used in the integration example requires the user to specify which backend you're using (wgpu in my case), it will naturally fail with any third party libraries that rely on traits that they implement for the general `iced::Renderer` type, rendering them unusable.

Obviously, this issue is preventing me from upgrading iced in my project, which also means I cannot upgrade winit nor wgpu until it is fixed unless I drop `plotters-iced`.

### What is the expected behavior?

We should be able to use such libraries with the `impl Program` flow

### Version

crates.io release

### Operating System

Linux

MartensCedric:
I am affiliated with this individual and experience the same issue

Davidster:
I was able to get it working by using this approach: https://github.com/Joylei/plotters-iced/issues/29#issuecomment-1703388377

PR #2063, opened 2023-09-02
use general Renderer type in integration example

Davidster:
This allows applications that are based off the integration example to use the generic `iced_renderer::Renderer` type in the rest of their codebase and to work with third party libraries.

I couldn't compile the example with the `plotters-iced` lib without this change. See #2061 for more details.

### DIFF
--- a/examples/integration/src/controls.rs
+++ b/examples/integration/src/controls.rs
@@ -1,4 +1,4 @@
-use iced_wgpu::Renderer;
+use iced_renderer::Renderer;
 use iced_widget::{slider, text_input, Column, Row, Text};
 use iced_winit::core::{Alignment, Color, Element, Length};
 use iced_winit::runtime::{Command, Program};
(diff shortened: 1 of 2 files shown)


hecrj:
`iced_renderer::Renderer` is _less_ generic than `iced_wgpu`.

The fact that you can easily change the types like this shows that there is no inherent limitation to `iced` when it comes to handling different renderers.

Davidster:
Yeah, I was wrong in the original issue, there is no limitation in iced. 

But I believe the PR is still valid: writing the integration example with the backend-agnostic renderer type makes it easier to use with other iced code. Sorry I didn't discuss this with the team first, I figured it was a no-brainer.

hecrj:
`iced_renderer::Renderer` isn't backend-agnostic, however!

It is a `Renderer` enum that supports runtime switching between `iced_tiny_skia ` and `iced_wgpu`. It is completely unnecessary if you only plan on using `iced_wgpu`.

Instead, library code like `plotters-iced` should rely only on the `Renderer` traits in `iced_core` to allow users to use any renderer implementation, like [the `Program` trait](https://docs.rs/iced_runtime/latest/iced_runtime/program/trait.Program.html).

Davidster:
Thanks for the example with the Program trait. It's starting to make more sense now

PR #2067, opened 2023-09-02
Introduce `theme::Custom::with_fn` to generate completely custom themes

hecrj:
As discussed in https://discourse.iced.rs/t/is-it-possible-to-control-the-extended-palette/40/4.


AustinMReppert:
Fantastic!

PR #2070, opened 2023-09-04
Added a Frame::scale_nonuniform method 

ripytide:
I see no reason why not to make `Frame::scale` take a `Vector` rather than a single `f32` to allow non-same xy scaling.

I have tried to update all references I could find in the code-base but may need to rely on the CI to see if I missed any.

### DIFF
--- a/examples/game_of_life/src/main.rs
+++ b/examples/game_of_life/src/main.rs
@@ -550,7 +550,7 @@ mod grid {
                     frame.translate(center);
                     frame.scale(self.scaling);
                     frame.translate(self.translation);
-                    frame.scale(Cell::SIZE as f32);
+                    frame.scale(Cell::SIZE);
 
                     let region = self.visible_region(frame.size());
 
@@ -576,7 +576,7 @@ mod grid {
                         frame.translate(center);
                         frame.scale(self.scaling);
                         frame.translate(self.translation);
-                        frame.scale(Cell::SIZE as f32);
+                        frame.scale(Cell::SIZE);
 
                         frame.fill_rectangle(
                             Point::new(cell.j as f32, cell.i as f32),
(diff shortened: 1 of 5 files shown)


renderer/src/geometry.rs
```diff
@@ -170,7 +170,7 @@ impl Frame {
 
     /// Applies a scaling to the current transform of the [`Frame`].
     #[inline]
-    pub fn scale(&mut self, scale: f32) {
+    pub fn scale(&mut self, scale: Vector) {
```
hecrj, line 173: Uniform scaling seems to be the most common use case. Instead of making the simple use case more complicated, I think a new `scale_nonuniform` method for advanced usage would be better.

We can also make `scale` take an `impl Into<f32>` and `scale_nonuniform` an `Into<Vector>`. This way, we'll be able to call both nicely:

```rust
frame.scale(2);
frame.scale_nonuniform([2.0, 3.0]);
```

hecrj (changes requested):

examples/game_of_life/src/main.rs
```diff
@@ -548,9 +548,15 @@ mod grid {
 
                 frame.with_save(|frame| {
                     frame.translate(center);
-                    frame.scale(self.scaling);
+                    frame.scale(Vector {
+                        x: self.scaling,
+                        y: self.scaling,
+                    });
```
hecrj, line 554: Make sure to run `cargo fmt`!

renderer/src/geometry.rs
ripytide: I agree with adding `scale_nonuniform` rather than changing `scale`.

I'm not sure this PR is the best place for changing the parameters to `impl Into<_>` though since that seems like more of a global iced API issue rather than something to add ad-hoc in a vaguely related PR.

hecrj (approved):
Made those changes myself! Thanks!

PR #2071, opened 2023-09-04
use @interpolate(flat) attribute as per the WebGPU spec

Davidster:
I wasn't able to get iced working with webgpu without this

User-defined vertex outputs and fragment inputs of scalar or vector integer type must always be specified as @interpolate(flat)

https://www.w3.org/TR/WGSL/#interpolation


hecrj (approved):
Looks good! Thanks!

wgpu/src/shader/quad.wgsl
```diff
@@ -147,12 +147,12 @@ fn solid_fs_main(
...
-    @location(3) colors_3: vec4<u32>,
-    @location(4) colors_4: vec4<u32>,
-    @location(5) offsets: vec4<u32>,
-    @location(6) direction: vec4<f32>,
+    @location(1) @interpolate(flat) colors_1: vec4<u32>,
+    @location(2) @interpolate(flat) colors_2: vec4<u32>,
+    @location(3) @interpolate(flat) colors_3: vec4<u32>,
+    @location(4) @interpolate(flat) colors_4: vec4<u32>,
+    @location(5) @interpolate(flat) offsets: vec4<u32>,
+    @location(6) @interpolate(flat) direction: vec4<f32>,
```
hecrj, line 155: This one isn't an integer type. I imagine we can revert it.

Davidster: Oops, yeah that was unintentional

PR #2072, opened 2023-09-06
Add access to bounds/content bounds from a scrollable viewport.

dtzxporter:
Added access to the bounds and content bounds of a scrollable's viewport. This is necessary to perform some scrollable optimizations such as virtual lists without creating your own widget, and piggybacking on the scrollable.

Added 2 methods for accessing the corresponding bounds from `Viewport`.

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -1137,16 +1137,6 @@ impl Viewport {
         }
     }
 
-    /// Returns the bounds of the current [`Viewport`].
-    pub fn bounds(&self) -> Rectangle {
-        self.bounds
-    }
-
-    /// Returns the content bounds of the current [`Viewport`].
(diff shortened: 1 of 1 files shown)


widget/src/scrollable.rs
```diff
@@ -1137,6 +1137,16 @@ impl Viewport {
...
 
+    /// Returns the bounds of the current [`Viewport`].
+    pub fn bounds(&self) -> Rectangle {
+        self.bounds
+    }
+
+    /// Returns the content bounds of the current [`Viewport`].
+    pub fn content_bounds(&self) -> Rectangle {
+        self.content_bounds
+    }
```
hecrj, line 1148: Let's move these after `relative_offset`.

hecrj (changes requested):

PR #2076, opened 2023-09-09
Get web examples running again

### LINKED ISSUE #1522: tour example fails on the web: RuntimeError: unreachable

### What happened?

I want to compile a web version of the tour example. The method you provided was used:
`
cd examples/tour
`
`trunk serve`

The browser cannot display any graphics and reports an error: RuntimeError: unreachable

The same behavior as the desktop

### Version

master

### Operative System

Windows

```shell
Uncaught (in promise) RuntimeError: unreachable
    at __rust_start_panic (tour-dc667960ab6beea7_bg.wasm:0x1293a3a)
    at rust_panic (tour-dc667960ab6beea7_bg.wasm:0x127fe98)
    at std::panicking::rust_panic_with_hook::h70a0e195f4db2a29 (tour-dc667960ab6beea7_bg.wasm:0xf77890)
    at std::panicking::begin_panic_handler::{{closure}}::hdcfc819ce836829e (tour-dc667960ab6beea7_bg.wasm:0x1044ce8)
    at std::sys_common::backtrace::__rust_end_short_backtrace::h53cabafab5b09ada (tour-dc667960ab6beea7_bg.wasm:0x1275e8e)
    at rust_begin_unwind (tour-dc667960ab6beea7_bg.wasm:0x11ad1cc)
    at core::panicking::panic_fmt::h751be80779d42b53 (tour-dc667960ab6beea7_bg.wasm:0x1244a86)
    at std::time::SystemTime::now::hc0a63dc16c77bca3 (tour-dc667960ab6beea7_bg.wasm:0x1244a03)
    at env_logger::fmt::humantime::imp::<impl env_logger::fmt::Formatter>::timestamp_seconds::haf7734f28160ab95 (tour-dc667960ab6beea7_bg.wasm:0x113e2f0)
    at env_logger::fmt::DefaultFormat::write_timestamp::hb067d7130f409a9c (tour-dc667960ab6beea7_bg.wasm:0xc7e273)
```

Nisatru:
Heya!

First things first, thanks for developing this great library!

I recently started experimenting with integrating iced into an existing wgpu application that targets native as well as the web.

Im well aware that web is not the main focus of development, but since there are some examples that have some setup for running on the web, I updated them to build and run again.

The following (web) examples have been updated:
- Counter
- Integration
- Todos
- Tour

With getting the `tour` example running again this should fix #1522 .
The cause for the `RuntimeError: unreachable` as described in the issue #1522 back then is likely not the same as now, since the example changed in the meantime.

Anyways the issue description still fits the current situation "When starting the tours app it crashes on load (only web)".

The reason for the `RuntimeError: unreachable` is the line `tracing_subscriber::fmt::init();` which panics when running on the web.

I added `console_log` for logging and a panic hook to also log panics in the browser console, similar to the integration example, so next time we encounter a `RuntimeError: unreachable` we should get a more descriptive error message.

<br>

I have also updated the `README`s where necessary.

<br>

Note: Text is not rendered on the web because no font is loaded, which is ignored here because it is another already known topic.

### DIFF
--- a/examples/tour/Cargo.toml
+++ b/examples/tour/Cargo.toml
@@ -18,4 +18,3 @@ iced.features = ["image", "debug", "webgl"]
 
 console_error_panic_hook = "0.1"
 console_log = "1.0"
-log.workspace = true
--- a/examples/tour/Cargo.toml
+++ b/examples/tour/Cargo.toml
@@ -10,3 +10,7 @@ iced.workspace = true
 iced.features = ["image", "debug"]
 
 tracing-subscriber = "0.3"
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+iced.workspace = true
+iced.features = ["image", "debug", "webgl"]
--- a/examples/tour/Cargo.toml
+++ b/examples/tour/Cargo.toml
@@ -9,8 +9,13 @@ publish = false
 iced.workspace = true
 iced.features = ["image", "debug"]
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tracing-subscriber = "0.3"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 iced.workspace = true
 iced.features = ["image", "debug", "webgl"]
+
+console_error_panic_hook = "0.1"
+console_log = "1.0"
+log.workspace = true
--- a/examples/counter/Cargo.toml
+++ b/examples/counter/Cargo.toml
@@ -7,3 +7,7 @@ publish = false
 
 [dependencies]
 iced.workspace = true
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+iced.workspace = true
+iced.features = ["webgl"]
--- a/examples/counter/README.md
+++ b/examples/counter/README.md
@@ -13,4 +13,12 @@ You can run it with `cargo run`:
 cargo run --package counter
 ```
 
+The web version can be run with [`trunk`]:
+
+```
+cd examples/counter
+trunk serve
+```
+
 [`main`]: src/main.rs
+[`trunk`]: https://trunkrs.dev/
--- a/examples/integration/Cargo.toml
+++ b/examples/integration/Cargo.toml
@@ -19,6 +19,5 @@ iced_wgpu.features = ["webgl"]
 
 console_error_panic_hook = "0.1"
 console_log = "1.0"
-log.workspace = true
 wasm-bindgen = "0.2"
 web-sys = { version = "0.3", features = ["Element", "HtmlCanvasElement", "Window", "Document"] }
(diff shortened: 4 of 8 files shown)


examples/tour/Cargo.toml
```diff
@@ -9,4 +9,13 @@ publish = false
...
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tracing-subscriber = "0.3"
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+iced.workspace = true
+iced.features = ["image", "debug", "webgl"]
+
+console_error_panic_hook = "0.1"
+console_log = "1.0"
+log.workspace = true
```
hecrj, line 21: The `log` dependency here is most likely unnecessary.

examples/tour/Cargo.toml
```diff
@@ -9,4 +9,13 @@ publish = false
 iced.workspace = true
 iced.features = ["image", "debug"]
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tracing-subscriber = "0.3"
```
hecrj, line 13: `tracing-subscriber` not working on Wasm is quite annoying.

There should be a crate that puts everything together so we can avoid writing the conditional dependencies here. I may create that myself later.

hecrj (changes requested):
Thanks! Looks good. Just a small suggestion!

Hopefully we can remove the `webgl` feature soon!

examples/tour/Cargo.toml
Nisatru: It is (almost) unnecessary here if I am not mistaken.
`log` is needed to specify the `log::Level` when initializing the `console_log` logger.
```rust
console_log::init_with_level(log::Level::Debug)
```
Correct me if Im missing something tho.

hecrj: Can we use just `init`?

examples/tour/Cargo.toml
Nisatru: Yeah I was quite surprised when I found out that tracing-subscriber is actually causing the error.

While looking for a solution with `tracing-subscriber` I also found [tracing-web](https://github.com/WorldSEnder/tracing-web) which can be used in addition to get tracing_subscriber working on Wasm.

This still requires us to differentiate between Wasm or not, so I went the same route as the `integration` example did and simply put `console_log` there.

examples/tour/Cargo.toml
Nisatru: Oh yeah sure `init` uses `log::Level::Info` I think that should be fine too.

I wasn't sure what log level would be sensible here in that example.

Should I also change
```rust
console_log::init_with_level(log::Level::Debug)
```
to 
```rust
console_log::init()
```

And remove the `log` dependency in the integration example too?

hecrj: Go for it!

Nisatru: Done

hecrj (approved):
Awesome! Thanks :bow:

PR #2077, opened 2023-09-09
Fix majority of unresolved documentation links

matze:
I noticed that links either don't work at all (see `LineHeight` [here](https://docs.rs/iced/latest/iced/widget/radio/struct.Radio.html#method.text_line_height)) or point to invalid URLs (see link to `PaneGrid` [here](https://docs.rs/iced/latest/iced/widget/pane_grid/struct.Pane.html)). Most could be fixed by pointing to the actual location of an entity however a few are unresolvable because sub-crates cannot link to entities that are not dependencies.

### DIFF
--- a/.github/workflows/document.yml
+++ b/.github/workflows/document.yml
@@ -1,8 +1,5 @@
 name: Document
-on:
-  push:
-    branches:
-      - master
+on: [push, pull_request]
 jobs:
   all:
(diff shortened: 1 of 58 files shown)


hecrj (approved):
Awesome! Thanks.

I have set up CI to check for these; so that they don't break again!

PR #2078, opened 2023-09-10
Fix `ComboBox` widget panic on wasm

Nisatru:
The `ComboBox` widget always uses `std::time::Instant` which panics on Wasm.

-> switch to `crate::core::time::Instant`


hecrj (approved):
Good catch!

PR #2080, opened 2023-09-12
Migrate twox-hash -> xxhash-rust. Switch to Xxh3 for better performance.

dtzxporter:
xxhash-rust is more maintained, built against `::core`, so no workaround for wasm is necessary. Switch to Xxh3 for better performance, which shows when loading/hashing image buffers.


hecrj (approved):

Issue #2066, opened 2023-09-02
Skia panicing rendering widgets


sdgoihew:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

Panic with meaningless explanation (using tiny_skia backend)

Minimized example:
```rust
    fn view(&self) -> Element<'_, self::Message> {
        let regex_field = TextInput::new("qwe", "asd")
            .on_input(Message::RegexText)
            .on_submit(Message::RegexProcess);
        let controls = Row::new().push(regex_field);
        let column = Column::new().push(controls);
        let spacer = Text::new("spacer");
        let row: Element<'_, Message> = Row::new().push(column).push(spacer).into();
        let content = Scrollable::new(row/*.explain(Color::WHITE)*/)
            .direction(iced::widget::scrollable::Direction::Both {
                vertical: widget::scrollable::Properties::new(),
                horizontal: widget::scrollable::Properties::new() });
        Container::new(content).into()
    }
```

### What is the expected behavior?

Compile time error, or sane explanation of what happened

### Version

crates.io release

### Operating System

Windows

### Do you have any log output?

```shell
PS C:\workspace\iced-tour> cargo run --release
    Finished release [optimized] target(s) in 0.35s
     Running `target\release\tour.exe`
thread 'main' panicked at C:\Users\Nikolay\.cargo\registry\src\index.crates.io-6f17d22bba15001f\iced_tiny_skia-0.1.0\src\backend.rs:706:22:
Build rounded rectangle path
stack backtrace:
   0:     0x7ff78bd57eca - std::sys_common::backtrace::_print::impl$0::fmt
                               at /rustc/03a119b0b0e310d22d94399b24ed030056050f13/library\std\src\sys_common\backtrace.rs:44
   1:     0x7ff78bd7089b - core::fmt::rt::Argument::fmt
                               at /rustc/03a119b0b0e310d22d94399b24ed030056050f13/library\core\src\fmt\rt.rs:138
   2:     0x7ff78bd7089b - core::fmt::write
                               at /rustc/03a119b0b0e310d22d94399b24ed030056050f13/library\core\src\fmt\mod.rs:1094
   3:     0x7ff78bd54281 - std::io::Write::write_fmt<std::sys::windows::stdio::Stderr>
[...]

PR #2082, opened 2023-09-14
[Issue 2066] Fix Tiny Skia panic on infinite length quads

joshuamegnauth54:
Fixes issue 2066.

Iced may produce primitives with an infinite length or width. The WGPU renderer handles those somewhere along the pipeline. Tiny Skia's [PathBuilder::finish](https://docs.rs/tiny-skia/latest/tiny_skia/struct.PathBuilder.html#method.finish) returns `None` if non-normal floats are passed to the builder. This patch adds a check to prevent that.

I only fixed the immediate issue since it didn't seem like the other primitives needed the check. Plus, I didn't want to preemptively add checks where they're not needed. Let me know if I should make any changes.

See:
* https://github.com/RazrFalcon/tiny-skia/blob/master/path/src/path_builder.rs#L421
* https://github.com/RazrFalcon/tiny-skia/blob/master/path/src/rect.rs#L344


tiny_skia/src/backend.rs
```diff
@@ -154,7 +154,26 @@ impl Backend {
                 border_width,
                 border_color,
             } => {
-                let physical_bounds = (*bounds + translation) * scale_factor;
+                // Tiny Skia's `PathBuilder::finish` returns `None` on infinites and NaNs which causes iced to panic.
+                // Certain widget combinations can lead to an infinite in either dimension hence the need for this check.
```
hecrj, line 158: What kind of _valid_ widget combinations could lead to an infinite quad? It's important we don't try to fix invalid use cases.

joshuamegnauth54: That's a good point! I mainly tried to fix the panic as reported in #2066, but the code is likely wrong since the submitter is sticking a `TextInput` into a `Scrollable` which is what directly leads to the panic. It also doesn't do what they _want_ it do.

With that said, the WGPU renderer doesn't panic whereas the Tiny Skia renderer panics when it unwraps `None`. The message confused the submitter as mentioned in their original issue since it doesn't say anything about what caused the panic.

Would it make sense to merge this so neither renderer panics? Or should the Tiny Skia renderer still panic for the incorrect code?

hecrj (approved):
I just changed this to a `debug_assert!` with a more descriptive message, since it's an invalid primitive.

PR #2085, opened 2023-09-14
[Feature] Custom Shader Widget

bungoboingo:
### 🌈  Custom Shader Widget

https://github.com/iced-rs/iced/assets/4241774/e752d35b-4b1d-4369-869b-bc662c355eb6

This PR implements [RFC #23](https://github.com/iced-rs/rfcs/pull/23) and adds support for a new type of widget which allows for the use of custom pipelines & shaders embedded directly into your Iced application, using Iced's existing `iced_wgpu` crate! 🥳 

## 🎨 Creating a custom shader
If you have the `wgpu` feature enabled for Iced, a custom shader can be created very similarly to a `Canvas`:

```rust
iced::widget::Shader::new(program);
```
Each shader widget requires a `widget::shader::Program` which holds internal state, provides ways to update the widget based on input events, and defines a special associated type called `Primitive`.

```rust
pub trait Program<Message> {
    type State;

    type Primitive: shader::Primitive;

    fn update(..) -> (event::Status, Option<Message>);

    fn draw(..) -> Self::Primitive;

    fn mouse_interaction(..) -> mouse::Interaction;
}
```

## 🔺 Primitives
A `Primitive` must implement a new trait, `shader::Primitive`. This trait defines how a the `Primitive` will be uploaded to the GPU, and how it will be rendered.

```rust
pub trait Primitive {
    fn prepare(
        &self,
        format: wgpu::TextureFormat,
        device: &wgpu::Device,
        queue: &wgpu::Queue,
        target_size: Size<u32>,
        scale_factor: f32,
        transform: Transformation,
        storage: &mut Storage,
    );

    fn render(
        &self,
        storage: &Storage,
        bounds: Rectangle<u32>,
        target: &wgpu::TextureView,
        target_size: Size<u32>,
        encoder: &mut wgpu::CommandEncoder,
    );
}
```

This trait gives unique widget-level access to necessary wgpu reference needed to render primitives! Note that due to not knowing the nature of what a custom shader might contain, every `render` call gets access to the `CommandEncoder` and must start its own render pass.

The `Primitive` trait has access to the underlying custom pipeline storage, appropriately named `Storage`, where users can store & fetch their pipelines to prepare & render their primitives, referenced by `TypeID`.

```rust
//for example..
fn prepare(
        &self,
        format: wgpu::TextureFormat,
        device: &wgpu::Device,
        queue: &wgpu::Queue,
        target_size: Size<u32>,
        _scale_factor: f32,
        _transform: Transformation,
        storage: &mut shader::Storage,
    ) {
        if !storage.has::<MyPipeline>() {
            storage.store(MyPipeline::new(device, queue, format, target_size))
        }

        let pipeline = storage.get_mut::<MyPipeline>().unwrap();

        //upload data to GPU!
        pipeline.update(
            device,
            queue,
            target_size,
            &self.uniforms,
            self.cubes.len(),
            &self.cubes,
        );
    }
```

Let me know what you think! I've tried to work through few different use cases and hope I've covered all bases. Go forth and shade! 🔢 🌈


core/src/rectangle.rs
```diff
@@ -183,6 +183,17 @@ impl From<Rectangle<u32>> for Rectangle<f32> {
     }
 }
 
+impl From<Rectangle<f32>> for Rectangle<u32> {
+    fn from(rectangle: Rectangle<f32>) -> Self {
+        Rectangle {
+            x: rectangle.x as u32,
+            y: rectangle.y as u32,
+            width: rectangle.width as u32,
+            height: rectangle.height as u32,
```
clarkmoody, line 192: Should we `.round()` here? What's the risk of converting directly with `as`?

hecrj: There is already a `snap` method in `Rectangle` that does exactly this.

hecrj: Rounding is tricky because it may not honor the original dimensions of the rectangle (e.g. `x` may be floored but `width` may be ceiled). This can cause annoying graphical glitches.

0x7CFE:
Amazing progress! Any estimates on when this can be merged?

hecrj:
@0x7CFE Today or tomorrow, most likely!

renderer/src/lib.rs
```diff
@@ -292,3 +274,23 @@ impl<T> crate::graphics::geometry::Renderer for Renderer<T> {
         }
     }
 }
+
+#[cfg(feature = "wgpu")]
+impl<T> iced_wgpu::primitive::pipeline::Renderer for Renderer<T> {
+    fn draw_pipeline_primitive(
```
bungoboingo, line 280: Much nicer than `draw_custom`; nice change!

PR #2089, opened 2023-09-19
Chore: Inline format args for ease of reading

nyurik:
A minor cleanup to inline all simple cases of format arguments. Makes the format strings just a bit easier to read.


hecrj (approved):
Thanks!

PR #2090, opened 2023-09-19
Chore: Apply clippy map transformations

nyurik:
Convert `.map().unwrap_or()` to `.map_or()` and similar transformations.


hecrj (approved):
Thanks!

PR #2095, opened 2023-09-20
A few minor clippy optimizations

nyurik:
* Use `let-else` in a few places
* make `match` explicit to avoid accidentally missing new variants when they are added later
* `Default::default` is not recommended as it is harder to read
* Avoid `if !...` by inverting it for readability

### DIFF
--- a/core/src/element.rs
+++ b/core/src/element.rs
@@ -402,7 +402,7 @@ where
             viewport,
         );
 
-        shell.merge(local_shell, &self.mapper);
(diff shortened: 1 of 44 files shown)


hecrj:
Thanks, but some of these are quite subjective.

nyurik:
@hecrj could you tell which are ok and which I bad? I can adjust the PR if you want

hecrj:
No worries! I can fix them myself. I'll also make sure to enable the proper lints on CI. Thanks again! :bow:

PR #2099, opened 2023-09-24
Compute vertex positions in the shader

jim-ec:
Compute the vertex positions of solid and gradient quads in the vertex shader instead of passing explicit vertex and index buffers.

The solid, gradient, and image quad pipeline now do not need to create and bind vertex and index buffers, only per-instance vertex buffers are required.


wgpu/src/shader/vertex.wgsl
```diff
@@ -0,0 +1,7 @@
+// Compute the normalized quad coordinates based on the vertex index.
+fn vertex_position(vertex_index: u32) -> vec2<f32> {
+    // #: 0 1 2 3 4 5
+    // x: 1 1 0 0 0 1
+    // y: 1 0 0 0 1 1
+    return vec2<f32>((vec2(1u, 2u) + vertex_index) % 6u < 3u);
```
bungoboingo, line 6: Cool bit of code here! Didn't think about calculating a vec2<bool> and letting it be coerced to 1.0/0.0.

jim-ec:
Is there a problem with the PR? It's been a while...

MichalLebeda:
@hecrj why did the tests not run? This looks like a cool PR

hecrj (approved):
Graphics wizardry! Awesome! Thank you :partying_face:

Issue #900, opened 2021-05-31
Additional functionality for sliders


sourcebox:
I'm working on an application which mainly consists for sliders for parameter control. For better user experience, it would be nice to offer two features: reset to default and fine control.

On VST plugins this is normally done via double-clicks and/or modifier keys. E.g. double-clicking resets the slider to a default (mostly 0), alternatively Ctrl-/Cmd-Klick also does it. For fine control, often Shift-dragging is used. Fine control is useful in cases where the slider covers a large value range with only little screen space.

Is there a way to add such functionality with the existing code via traits or would that require an update to the widgets?

13r0ck:
An event would have to be added to the widget. The text_input widget is a good one to see example of how to implement more complex event detection. 

Without making a modification to iced, you MAY be able to get the same result by putting the slider inside a button and then watch for double clicks as a custom widget

13r0ck:
An event would have to be added to the widget. The text_input widget is a good one to see example of how to implement more complex event detection. 

Without making a modification to iced, you MAY be able to get the same result by putting the slider inside a button and then watch for double clicks as a custom widget

JL710:
Is this issue just forgotten or is there something that was not added in #2100?

PR #2100, opened 2023-09-24
Enhance `Slider` and `VerticalSlider` functionality

jpttrssn:
This attempts to address #900.

* Add optional default behavior
  * Add a `default` field
  * Add a `default()` method to set the `default` field
  * ~~A double-click, ctrl-click or command-click will set the slider to the default value~~
* Add optional fine-grained control
  * Add an optional `step_fine` field
  * Add a `step_fine()` method to set the `step_fine` field
  * Use `step_fine` in place of `step` while shift is pressed
* Add increment/decrement via up/down keys
* Update `Slider` and `VerticalSlider` examples


hecrj (approved):
These seem reasonable. Thanks! :bow: 

I made some changes here and there, but we can merge!

PR #2103, opened 2023-09-27
Fix theme palette name being ambiguous

ElhamAryanpur:
I had issues running the examples because the `theme.rs` had the palette module import being flagged as "ambiguous" by the compiler.

> error[E0659]: `palette` is ambiguous
> --> style/src/theme.rs:4:9
> |
> 4 | pub use palette::Palette;
> |         ^^^^^^^ ambiguous name
> |
> = note: ambiguous because of multiple potential import sources
> = note: `palette` could refer to a crate passed with `--extern`
> = help: use `::palette` to refer to this crate unambiguously
> note: `palette` could also refer to the module defined here
> --> style/src/theme.rs:2:1
> |
> 2 | pub mod palette;
> | ^^^^^^^^^^^^^^^^
> = help: use `self::palette` to refer to this module unambiguously

the `styling` example was also relying on the import structure so had to update that as well.

### DIFF
--- a/examples/styling/src/main.rs
+++ b/examples/styling/src/main.rs
@@ -53,13 +53,15 @@ impl Sandbox for Styling {
                 self.theme = match theme {
                     ThemeType::Light => Theme::Light,
                     ThemeType::Dark => Theme::Dark,
-                    ThemeType::Custom => Theme::custom(theme::Palette {
-                        background: Color::from_rgb(1.0, 0.9, 1.0),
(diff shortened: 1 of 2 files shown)


hecrj:
This should be fixed if you update your Rust toolchain with `rustup update`.

I'll create a PR to set the MSRV.

ElhamAryanpur:
Got it! thanks

PR #2105, opened 2023-09-29
A few minor improvements

JohnScience:


Cargo.toml
```diff
@@ -137,6 +137,8 @@ resvg = "0.35"
 rustc-hash = "1.0"
 smol = "1.0"
 softbuffer = "0.2"
+strum = "0.25"
```
max397574, line 140: why adding those dependencies?

JohnScience: By its nature, `ComboBox` is meant for selection of an item. When the set of choices is fixed, it's natural to represent the choices with an enum. The [`strum`](https://crates.io/crates/strum) crate has over 53M downloads and in this case it's used to get an iterator over enum variants. I suspect that it's a very common use case.

Maybe the dependency should be pushed to the `Cargo.toml` of the example but my concern is that keeping it updated can be problematic.

examples/i_choose_you/src/main.rs
```diff
@@ -0,0 +1,101 @@
...
+    Squirtle,
+}
+
+impl Sandbox for PokemonPicker {
+    type Message = Message;
+
+    fn new() -> Self {
+        PokemonPicker {
+            pokemon: Pokemon::default(),
+            choices: combo_box::State::new(Pokemon::iter().collect()),
```
JohnScience, line 41: Here, we create [`combo_box::State`](https://docs.rs/iced/latest/iced/widget/combo_box/struct.State.html) using the iterator that we obtained using the `strum` crate.

Cargo.toml
max397574: I know exactly what those things do and I use them myself
no matter how many downloads it has imo there shouldn't be any dependencies added to the crate that aren't needed just because they seem useful to you.

examples/i_choose_you/src/main.rs
max397574: okay
then add it as dependency to that example

hecrj (commented):
Hey! I'm afraid we have reached a point where we can't just keep adding more and more examples to the repository. Maintenance cost and all that!

Feel free to open a repository of your own and share the examples there. You can also submit a PR to https://github.com/iced-rs/awesome-iced to have them listed!

The other changes seem unrelated or unnecessary. CI should yell at us if something doesn't compile.

PR #2115, opened 2023-10-19
Bump `cosmic-text` to `0.10` & glyphon to rev = 6aaaf

### LINKED ISSUE #2005: Most text characters not rendering

### What happened?

Most characters do not render when trying to render text. It happens with text widgets, button widgets, etc. The only thing I've been able to find that correctly renders text is the text input. I tried to render just a simple text widget:
```rust
    fn view(&self) -> iced::Element<'_, Self::Message> {
        text("asd123+-=';*").into()
    }

```
, and here is what I get:
image

When running the todos example, I get this:
image
As you can see, the button has no text and the spacing looks weird for the text input. If I type into the text input, I see this:
image
Suspiciously, those same characters that actually render for me show up weird here, with what looks like a different font, colour, and spacing. I tried playing around with changing the font, but to no avail.

I expected text to render correctly with no problems. I have successfully used iced 0.9 in the past and it worked fine, but 0.10 is broken for me.

### Version

crates.io release

### Operating System

Linux

_No response_

dunxen:
Fixes #2005.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -114,10 +114,10 @@ iced_winit = { version = "0.12", path = "winit" }
 async-std = "1.0"
 bitflags = "1.0"
 bytemuck = { version = "1.0", features = ["derive"] }
-cosmic-text = "0.9"
+cosmic-text = "0.10"
 futures = "0.3"
 glam = "0.24"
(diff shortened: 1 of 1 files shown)


hecrj:
Superseded by #2122.

PR #2120, opened 2023-10-23
`Transform` primitive

hecrj:
This PR replaces the `Translate` primitive in `iced_graphics` with a more generic `Transform` primitive that can take any non-skewing affine `Transformation` (i.e. scaling and translation).

Additionally, a `transform` method has been implemented for `Geometry` in `iced_renderer`. This enables efficient translation and scaling of cached geometry with a `canvas::Cache`, since `transform` can be called on the returned cached geometry!

But not only that, this new primitive can be used to effectively scale any part of the UI. This can be very useful for building node-based editors, for instance.

There are still some quirks to iron out here. Mainly, scaling text does not work currently, but that should be fairly easy to fix.

### DIFF
--- a/graphics/src/primitive.rs
+++ b/graphics/src/primitive.rs
@@ -8,6 +8,7 @@ use crate::core::{
 };
 use crate::text::editor;
 use crate::text::paragraph;
+use crate::Transformation;
 
 use std::sync::Arc;
 
@@ -104,12 +105,12 @@ pub enum Primitive<T> {
         /// The content of the clip
         content: Box<Primitive<T>>,
     },
-    /// A primitive that applies a translation
-    Translate {
-        /// The translation vector
-        translation: Vector,
+    /// A primitive that applies a [`Transformation`]
+    Transform {
+        /// The [`Transformation`]
+        transformation: Transformation,
 
-        /// The primitive to translate
+        /// The primitive to transform
         content: Box<Primitive<T>>,
     },
     /// A cached primitive.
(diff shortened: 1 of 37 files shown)


graphics/src/transformation.rs
```diff
@@ -30,6 +29,22 @@ impl Transformation {
...
+
+    pub fn scale_x(&self) -> f32 {
+        self.0.x_axis.x
+    }
+
+    pub fn scale_y(&self) -> f32 {
+        self.0.y_axis.y
+    }
+
+    pub fn translation_x(&self) -> f32 {
```
0x7CFE, line 41: nit: probably worth name both sets accordingly. Either scale/translate or scaling/translation.

graphics/src/primitive.rs
```diff
@@ -105,23 +106,34 @@ pub enum Primitive<T> {
...
-        Self::Translate {
-            translation,
+        Self::Transform {
+            transformation: Transformation::translate(
+                translation.x,
+                translation.y,
+            ),
+            content: Box::new(self),
+        }
+    }
```
0x7CFE, line 131: Probably worth checking if the topmost primitive is `Transform` then just merge the transformations. Otherwise we could end up in a long chain of transformations.

alex13sh:
What about the rotation method?

tiny_skia/src/backend.rs
```diff
@@ -563,11 +558,11 @@ impl Backend {
                     clip_mask,
                     clip_bounds,
                     scale_factor,
-                    translation + *offset,
+                    transformation * *new_transformation,
                 );
             }
             Primitive::Clip { bounds, content } => {
-                let bounds = (*bounds + translation) * scale_factor;
+                let bounds = (*bounds * transformation) * scale_factor;
```
0x7CFE, line 565: This line is incorrect. Since transformation already can contain scaling, `scale_factor` here is redundant and causes clipping issues (see the P.S in https://discourse.iced.rs/t/rendering-way-too-many-rectangles-on-a-canvas/113/12?u=0x7cfe).

PR #2123, opened 2023-10-27
`TextEditor` widget (or multi-line text input)

hecrj:
This PR introduces a multi-line text input widget called `TextEditor`.

A `TextEditor` can be created with some `Content` and produces an `Action` when the user interacts with it. This `Action` can be performed on the `Content` during `update` to apply any cursor movements, clicks, and edits.

A `TextEditor` uses the new `text::Highlighter` API to potentially highlight text. A new subcrate `iced_highlighter` contains a new basic syntax highlighter implementation on top of [`syntect`](https://docs.rs/syntect/latest/syntect/).

A new `editor` example has been implemented which leverages the `TextEditor` to build a simple text editor program with syntax highlighting. The latest video-tutorial in the `iced-rs` YouTube channel walks through the creation of the example: https://www.youtube.com/watch?v=gcBJ7cPSALo


widget/src/text_editor.rs
```diff
@@ -0,0 +1,708 @@
...
+                            mouse::ScrollDelta::Lines { y, .. } => {
+                                if y.abs() > 0.0 {
+                                    (y.signum() * -(y.abs() * 4.0).max(1.0))
+                                        as i32
+                                } else {
+                                    0
+                                }
+                            }
+                            mouse::ScrollDelta::Pixels { y, .. } => {
+                                -y.signum() as i32
```
blazra, line 631: Scrolling behavior with touchpad is weird. If you scroll slowly it actually scrolls more lines more quickly than if you scroll a lot. 

Because the editor scrolls by whole lines, I think you will need to keep some state like "fraction of a line already scrolled" to add it to the scroll distance on the next event. Or something like that.

cnwzhu:
"TextEditor is cumbersome to use. I only need a multiline input, not a text editor. Is there a simpler solution?"

wildwestrom:
> "TextEditor is cumbersome to use. I only need a multiline input, not a text editor. Is there a simpler solution?"

Seconding this. Really frustrating.

PR #2130, opened 2023-11-05
fix: Total and Used Ram in Example and Add Videos for Examples

Tahinli:
Total and used ram was showing like TB and GB level in system information example, but the actual values are GB and MB.
I added videos about examples that don't have visual showcase.

### DIFF
--- a/examples/arc/README.md
+++ b/examples/arc/README.md
@@ -12,3 +12,7 @@ cargo run --package arc
 ```
 
 [`main`]: src/main.rs
+
+
+https://github.com/Tahinli/iced/assets/96421894/c784a1b4-895e-41bd-a535-ad08faf01c36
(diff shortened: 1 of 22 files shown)


max397574:
I think we definitely don't want to link to assets in your fork
you should just use github and drag and drop the videos e.g. into an issue or into the readmes to get them uploaded and get a link to them like this

Tahinli:
> I think we definitely don't want to link to assets in your fork
> you should just use github and drag and drop the videos e.g. into an issue or into the readmes to get them uploaded and get a link to them like this

Actually i just dragged and dropped to readme files. But if still something wrong, i try to fix

hecrj:
I will merge the first commit here with the work in #2223.

PR #2140, opened 2023-11-17
Improve `TextEditor` slow scrolling behavior with touchpads

blazra:
If you scroll really slowly, then the text editor does not scroll, because the scrolling events which scroll by less than a single line don't do anything.

I have made a change so that the TextEditor stores the fractional part of the number of lines to scroll and adds it on the next scroll event.
For example, Kate or Konsole seems to behave like this.


hecrj (approved):
Looks reasonable. Thanks! :bow:

PR #2141, opened 2023-11-21
Use the correct GIF for the progress bar example

arslee07:
The Pokédex GIF does not seem to be relevant for this example. 😅
This little PR fixes the URL to the correct GIF.


hecrj (approved):
Thanks! :bow:

PR #2146, opened 2023-11-24
feat: somewhere to place extra actions by platform

Decodetalkers:
I have view iced-sckt forked iced, and add the extra actions. and there do are some extra actions, like set margin for layer-shell, set lock for ext-session-shell. I think add an any will be of help maybe

I do not know if it is right..


hecrj (approved):
This seems reasonable.

The alternative is introducing a new `Custom` generic type in `Action`, but it inevitably ends up affecting the `Command` type and the `Program` trait; increasing complexity considerably.

I believe sacrificing a bit of type-safety for this rare use case is fine. Thank you! :bow:

Issue #582, opened 2020-10-27
Draw raster images in `Canvas`


savente93:
I've been trying to make an app that can make simple image annotations. As far as I can see `Image` doesn't `impl Layer`, meaning that I can't add it to a `Canvas`.  This seems like a good thing to me, especially since a "paint clone" is mentioned in the Roadmap? I'm wondering how easy this would be to implement, and whether this would be the correct way to go about it, or whether there is an easier way. If it's not too complicated, I'd be willing to try and make a PR for it. Thoughts?

tinaun:
I would like a Frame.draw_image(imageRect, canvasRect) function, much like the function in js's canvas

shssoichiro:
Hi, I wanted to see if there was any update on this, or if someone had found a workaround for it? This is a blocker for the project I'm currently working on, where I want to render an RGB image as the base layer in a canvas, then draw annotations, grids, etc. on top of it.

hardesh:
Hi, I'm also looking for this. Is there any update on this?

justDeeevin:
resolved by #2537. Use `Frame::draw_image` or `Frame::draw_svg`.

PR #2148, opened 2023-11-28
Add Frame::draw_image

jtojnar:
Fixes: https://github.com/iced-rs/iced/issues/582


jtojnar:
Too early.

hecrj:
Main reason this isn't implemented is because of layering.

I have plans to tackle the layering limitations of `iced_wgpu` soon. This should happen after that!

jtojnar:
Right, I noticed that transforms from nested frames like the ones in `game_of_life` example currently do not apply to images.

PR #2154, opened 2023-12-01
Fix text clipping

hecrj:
This PR fixes a couple issues related to text rendering and layouting:

- Text layout sizing not being clamped to maximum available space.
- Text being clipped to layout boundaries. Now the `viewport` bounds are used instead.

Since widgets can now control the clipping viewport when drawing text, the `text_input` widget does not need to issue a new rendering layer when horizontal scrolling is needed.

Additionally, the `container`, `column`, and `row` widgets will only draw its contents when they are actually visible.

### DIFF
--- a/graphics/src/primitive.rs
+++ b/graphics/src/primitive.rs
@@ -32,8 +32,8 @@ pub enum Primitive<T> {
         vertical_alignment: alignment::Vertical,
         /// The shaping strategy of the text.
         shaping: text::Shaping,
-        /// The viewport of the text.
-        viewport: Rectangle,
+        /// The clip bounds of the text.
+        clip_bounds: Rectangle,
     },
     /// A paragraph primitive
     Paragraph {
@@ -43,8 +43,8 @@ pub enum Primitive<T> {
         position: Point,
         /// The color of the paragraph.
         color: Color,
-        /// The viewport of the paragraph.
-        viewport: Rectangle,
+        /// The clip bounds of the paragraph.
+        clip_bounds: Rectangle,
     },
     /// An editor primitive
     Editor {
@@ -54,8 +54,8 @@ pub enum Primitive<T> {
         position: Point,
         /// The color of the editor.
         color: Color,
-        /// The viewport of the editor.
-        viewport: Rectangle,
+        /// The clip bounds of the editor.
+        clip_bounds: Rectangle,
     },
     /// A quad primitive
     Quad {
--- a/graphics/src/primitive.rs
+++ b/graphics/src/primitive.rs
@@ -14,24 +14,26 @@ use std::sync::Arc;
 pub enum Primitive<T> {
     /// A text primitive
     Text {
-        /// The contents of the text
+        /// The contents of the text.
         content: String,
-        /// The bounds of the text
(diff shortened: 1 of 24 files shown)


graphics/src/primitive.rs
```diff
@@ -14,24 +14,26 @@ use std::sync::Arc;
...
-        /// The horizontal alignment of the text
+        /// The horizontal alignment of the text.
         horizontal_alignment: alignment::Horizontal,
-        /// The vertical alignment of the text
+        /// The vertical alignment of the text.
         vertical_alignment: alignment::Vertical,
         /// The shaping strategy of the text.
         shaping: text::Shaping,
+        /// The viewport of the text.
+        viewport: Rectangle,
```
Mingun, line 36: I think, this is not very useful documentation. What is viewport, what is it purpose? How different values will influence rendering? By the way, docs of many properties suffers from that problem, not only that.

hecrj: A quick Google search:

> A viewport is a region of the screen used to display a portion of the total image to be shown

I think the name itself is quite self-descriptive.

These are internal APIs and users are expected to be familiar with graphical concepts.

I do agree on documentation of many properties not being very useful. Documentation is and has not been a priority yet! In fact, I don't believe documentation is generally a good idea, ever.

If you are familiar with the problem, the API of a library should feel intuitive. Examples should also be self-explanatory. If you are not familiar with the problem, then you need to learn about the problem first; not the library.

If a library needs a lot of documentation for people familiarized with the problem, it is because it has created too much distance with the problem. Its concepts aren't relatable enough. Therefore, instead of focusing on documentation, we should focus on the code.

Mingun: > These are internal APIs and users are expected to be familiar with graphical concepts.

If API is internal that is not meant that it shouldn't be documented. Current documentation just does not explain nothing about expected behaviour. Is text would be clipped? Wrapped? Centered? Stretched? In what coordinate system the viewport? If text is rotated, should the rectangle be still aligned with screen bounds, as the Google answer assumes?

Yes, the name of the field is quite self-descriptive, we know that it contains some well-known thing. But how we should use it? The name doesn't answer to that.

How to solve conflicts if one developer would think that text should be clipped, another is wrapped and the third that viewport should be expanded to include all text? While all of them are familiar with graphical concepts, each of them suggest the solution that is equally could be considered as correct.

> Do you need documentation to use a well-designed application or tool? Why should libraries be any different?

For example, take git. I think, it is well-designed enough, right? How you would do a simple task without documentation: replace current master with broken history with a new one where the history is correct? At the same time all other developers should do not do any unusual things. Do you think, that documentation of such sort is unnecessary?

<details>
<summary>Answer</summary>

https://git-scm.com/docs/git-merge#Documentation/git-merge.txt-ours

</details>

I known, you are doing great work and do not suggest you to check all the documentation and making it corresponding to some high ideals. But from my personal experience, if you couldn't write something non-trivial, that means either
- the thing is really trivial (for example, as setter for a property)
- you do not understand well why you introduces this thing
- documentation could be improved
- the thing is temporary and will be replaced very soon

PR #1908, opened 2023-06-10
Run convert_text on svg trees so text renders correctly

alec-deason:
Resvg will not render text unless `.convert_text` has been run on the tree first. This PR does that. This takes one approach to sharing the font database between the text pipeline and the vector pipeline but I'm open to alternatives.

Fixes #870.

PR #2161, opened 2023-12-11
Implement a fix for the native renderer not rendering text in SVGs

### LINKED ISSUE #2160: Text in SVGs is not rendered when using wgpu

### What happened?

This is related to #870 except for the native renderer and the same fix works for wgpu. Please find an example program reproducing the issue at [https://github.com/cfrenette/iced-svg-text-issue](https://github.com/cfrenette/iced-svg-text-issue). Text nodes inside the SVG are not currently rendered by wgpu on the master branch.

The text inside `<text/>` tags should be rendered. If we convert the text to paths like in #1908 the text is rendered.

### Version

master

### Operating System

Linux

_No response_

cfrenette:
Fixes #2160 
Related to #870, #1908 

The native renderer doesn't render text unless the text is converted to paths when the SVG tree is parsed. This implements the same fix used with tiny-skia.

### DIFF
--- a/wgpu/src/image/vector.rs
+++ b/wgpu/src/image/vector.rs
@@ -2,8 +2,9 @@ use crate::core::svg;
 use crate::core::{Color, Size};
 use crate::image::atlas::{self, Atlas};
 
+use iced_graphics::text;
 use resvg::tiny_skia;
-use resvg::usvg;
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks! Good catch :bow:

PR #2163, opened 2023-12-12
added svg hover, for styles impl

hicaru:
hello dear developers this is PR by ask @hecrj  separate the svg and list_picker PR.

### DIFF
--- a/widget/src/svg.rs
+++ b/widget/src/svg.rs
@@ -145,7 +145,7 @@ where
         theme: &Renderer::Theme,
         _style: &renderer::Style,
         layout: Layout<'_>,
-        _cursor: mouse::Cursor,
+        cursor: mouse::Cursor,
         _viewport: &Rectangle,
     ) {
         let Size { width, height } = renderer.dimensions(&self.handle);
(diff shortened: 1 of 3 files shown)


hicaru:
Please someone check this PR, very important for me

widget/src/svg.rs
```diff
@@ -166,7 +167,11 @@ where
                 ..bounds
             };
 
-            let appearance = theme.appearance(&self.style);
+            let appearance = if is_mouse_over {
+                theme.appearance(&self.style)
+            } else {
+                theme.hovered(&self.style)
+            };
```
hecrj, line 174: These branches need to be switched.

hecrj (changes requested):

widget/src/svg.rs
hicaru: i already switched. there was ask me to separate 2 changes.

hecrj: I mean `hovered` should be called if `is_mouse_over`, not in the `else` branch.

hecrj:
```suggestion
            let appearance = if is_mouse_over {
                theme.hovered(&self.style)
            } else {
                theme.appearance(&self.style)
            };
```

PR #2168, opened 2023-12-15
[Fix] `PaneGrid` click interaction on the top edge

jhff:
## Issue

Clicking on the top edge of a `PaneGrid` expands the clicked pane and pushes all other panes on the top row to below. 

This occurs because clicking a pane directly leads to a `Dragging` action, that, once button is released, will `drop` the `pane` on the top `edge`.

issuee

## Solution

This PR adds a new `action`, `Clicking`, which will only jump to `Dragging` if cursor moves a given deadband distance from the original click position. I have defined this deadband distance as `10` pixels, but let me know if it is too short/wide.

## Result

fixx


hecrj (approved):
Beautiful! Thanks :bow:

PR #2172, opened 2023-12-22
Added `text_shaping` method to `Tooltip`

GyulyVGC:
The ability to change the `text::Shaping` strategy, introduced in version 0.10, was still missing for `Tooltip`s.
I simply introduced a `text_shaping` method to update this widget to the new capabilities of the library.

### DIFF
--- a/widget/src/tooltip.rs
+++ b/widget/src/tooltip.rs
@@ -64,6 +64,12 @@ where
         self
     }
 
+    /// Sets the [`text::Shaping`] strategy of the [`Tooltip`].
+    pub fn text_shaping(mut self, shaping: text::Shaping) -> Self {
+        self.tooltip = self.tooltip.shaping(shaping);
+        self
+    }
+
     /// Sets the font of the [`Tooltip`].
     ///
     /// [`Font`]: Renderer::Font


widget/src/tooltip.rs
```diff
@@ -64,6 +64,12 @@ where
         self
     }
 
+    /// Sets the [`text::Shaping`] strategy of the [`Tooltip`].
+    pub fn text_shaping(mut self, shaping: text::Shaping) -> Self {
```
hecrj, line 68: Just `shaping` for consistency?

hecrj (changes requested):

widget/src/tooltip.rs
GyulyVGC: `Radio`, `Toggler`, and `Checkbox` [call it `text_shaping` as well](https://github.com/search?q=repo%3Aiced-rs%2Ficed%20text_shaping&type=code).

Let me know if you prefer to rename it just `shaping` everywhere, but I think that it could be confusing to apply to non-purely-`Text` widgets.

I'm agree with you to call it the simple way on `Text`, but on other widgets maybe being more explicit is better?

hecrj (approved):
Thank you!

PR #2182, opened 2024-01-05
Remove unused backend.rs file in renderer crate

Imberflur:
It looks like the module was removed in https://github.com/iced-rs/iced/pull/1932 but the file wasn't deleted.

The confused me for a moment when grepping for where certain functions exist and are called.


hecrj (approved):
Thanks!

PR #2184, opened 2024-01-05
Fix docs to include missing feature tags

william-shere:
Helper functions behind `lazy` feature were missing the tag in the documentation.

PR adds `#[cfg(feature = "lazy")]` attributes to the helper functions in `widget/src/lazy/helpers.rs`. That file is only included if that feature is enabled so this does not change compilation but allows rustdoc to pick up the tag.


hecrj (approved):
Thanks! :bow:

PR #2185, opened 2024-01-05
Add Length::FillPercentage

Qwinci:
Add `Length::FillPercentage(f32)` fill strategy that calculates the space to use with `percentage * max`, this is useful if you want to use a specific percentage of the available space. Any feedback is appreciated, for an example I am not sure if the percentage should be in the range 0.0..=1.0 like it is now or should it be in 0.0..=100.0 or if it should use `fill` instead of `max` for the calculation.

### DIFF
--- a/core/src/layout/limits.rs
+++ b/core/src/layout/limits.rs
@@ -50,6 +50,10 @@ impl Limits {
             Length::Fill | Length::FillPortion(_) => {
                 self.fill.width = self.fill.width.min(self.max.width);
             }
+            Length::FillPercentage(percentage) => {
+                let amount = self.max.width * percentage;
+                self.fill.width = amount.min(self.max.width).max(self.min.width);
+            }
             Length::Fixed(amount) => {
                 let new_width = amount.min(self.max.width).max(self.min.width);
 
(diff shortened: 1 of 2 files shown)


hecrj:
`FillPortion` already satisfies this use case.

Qwinci:
> `FillPortion` already satisfies this use case.

How do you specify that you want to fill eg. 50% of the space using FillPortion if you only have one element inside a full window widget?

hecrj:
You'd use a `row` or a `column` and `horizontal_space`.

Qwinci:
Its just that doing that is kind of ugly if you need to do it in both directions. I still think it would be nice being able to specify percentages, but now that I tested this PR a bit more it looks like it doesn't do what I wanted to do in all cases either, for an example if you add multiple widgets to a Row where the leftmost one has a percentage 0.25, middle 0.5 and right 0.25 I'd have liked it to calculate the amount of used space based on the initial max width/height of the full window container without any widgets but instead it calculates it relative to the remaining space. I guess Ill just have to do it with cols/rows/FillPortion.

PR #2188, opened 2024-01-08
Add doc to direct developers to enable features for iced::time::every

fogarecious:
Online document of Iced has the function iced::time::every. However, this function cannot be called when using thread-pool backend.

This PR adds doc to explicitly direct developers to enable appropriate features. This does not change any compiled code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `size_hint` not being called from `element::Explain`. [#2225](https://github.com/iced-rs/iced/pull/2225)
 - Slow touch scrolling for `TextEditor` widget. [#2140](https://github.com/iced-rs/iced/pull/2140)
 - `Subscription::map` using unreliable function pointer hash to identify mappers. [#2237](https://github.com/iced-rs/iced/pull/2237)
+- Missing feature flag docs for `time::every`. [#2188](https://github.com/iced-rs/iced/pull/2188)
 
 Many thanks to...
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -159,6 +159,7 @@
     rustdoc::broken_intra_doc_links
 )]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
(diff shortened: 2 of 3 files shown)


fogarecious:
This comment provides a more detailed explanation of the PR.

The [online document](https://docs.iced.rs/iced/) has a function named [iced::time::every](https://docs.iced.rs/iced/time/fn.every.html). As outlined in the document, there are no constraints on using this function, particularly requiring any specific features. However, when calling the function as shown below:

```rust
struct MyApp;

impl iced::Application for MyApp {
    type Executor = iced::executor::Default;
    type Message = iced::time::Instant;
    type Theme = iced::Theme;
    type Flags = ();

    fn new(_flags: Self::Flags) -> (Self, iced::Command<Self::Message>) {
        (Self, iced::Command::none())
    }

    fn title(&self) -> String {
        "".into()
    }

    fn update(&mut self, _message: Self::Message) -> iced::Command<Self::Message> {
        iced::Command::none()
    }

    fn view(&self) -> iced::Element<'_, Self::Message, iced::Renderer<Self::Theme>> {
        "".into()
    }

    fn subscription(&self) -> iced::Subscription<Self::Message> {
        iced::time::every(std::time::Duration::from_secs(1))
    }
}
```

and including the following dependencies in the `Cargo.toml` file:

```toml
[dependencies]
iced = "0.10.0"
```

it leads to a compiler error:

```output
error[E0425]: cannot find function `every` in module `iced::time`
```

To address this issue, the simplest solution is to enable one of the following features: [tokio](https://docs.rs/crate/iced/latest/features#tokio), [async-std](https://docs.rs/crate/iced/latest/features#async-std), or [smol](https://docs.rs/crate/iced/latest/features#smol).

This PR includes documentation for the function, explicitly instructing developers to enable the mentioned features.

While this may potentially be a `cargo doc` issue, it proved to be quite confusing for me, requiring a significant amount of time to identify the source of the error. I hope this PR can assist other developers facing the same problem.

(If you believe the problem is not significant, please inform me, and I will close the PR.)

hecrj (approved):
Thanks! I made some changes to use `doc_cfg` instead.

PR #2189, opened 2024-01-09
Add functions to fetch if a window is maximized or minimized

Calastrophe:
This pull request is in hopes to solve an issue for another individual who wants to be able to accurately determine if the window was previously maximized. I introduced these functions since you could listen for the `CloseRequested` then query if the window was maximized and save then - rather than creating a custom event listening for when a `Resized` event that makes the window maximized.

If anything needs to be corrected to fit the styling, please let me know and I will fix it.

### DIFF
--- a/core/src/window/event.rs
+++ b/core/src/window/event.rs
@@ -58,7 +58,7 @@ pub enum Event {
     /// for each file separately.
     FileHovered(PathBuf),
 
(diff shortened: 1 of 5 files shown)


hecrj (approved):
Awesome! Thank you! :tada:

PR #2169, opened 2023-12-15
Update `winit` to `0.29`

hecrj:
This PR updates our `winit` fork with the latest changes in `0.29`.

Fixes #2113.
Fixes #2138.

Issue #2113, opened 2023-10-12
`invalid setDrawableSize` on MacOS Sonoma 14.0


AzurIce:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

My Iced project not working on MacOS Sonoma 14.0 with the following error:

```
2023-10-12 10:29:32.396 iced-demo[3476:39547] CAMetalLayer ignoring invalid setDrawableSize width=4294967295.000000 height=4294967295.000000
thread 'main' panicked at /Users/azurice/.cargo/git/checkouts/glyphon-bce1d8fb6a1f7cb1/20f0f8f/src/text_render.rs:224:33:
attempt to add with overflow
```

And I tested some examples on branch master:
- some of them get panicked as soon as the windows appear. (tour)
- some of them are initially a blank window, and only after I resize the window it turns normal. (counter)
- Some of them seems normal. (todos)

It seems like a `winit` problem? I guess upgrading `winit` to v0.28.7 should fix it.

See https://github.com/rust-windowing/winit/issues/2876

I pasted a complete log of running example `tour` below.

### What is the expected behavior?

Panic not happens, and everything behaves normal

### Version

master

### Operating System

macOS

### Do you have any log output?

```shell
➜  iced git:(master) ✗ cargo run --package tour
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/tour`
[...]

PR #2190, opened 2024-01-09
Update to `softbuffer` 0.3, tracking up to `age` sets of primitives

ids1024:
This should perform better on Windows, Wayland, and X11, and the same of macOS. (https://github.com/rust-windowing/softbuffer/pull/95 would allow macOS to also perform better here.)

I've done basic testing on Wayland and Windows. I haven't tested X11 or macOS yet. I'm not sure how best to check the profile the performance impact.

The latest version of softbuffer is 0.4, which uses raw-window-handle 0.6. That probably will need https://github.com/iced-rs/iced/pull/2169, and maybe an update to wgpu (which has updated to raw-window-handle 0.6, but hasn't released a change with that yet).

It should also be possible to make use of Softbuffer's `present_with_damage`,  if we can get damage regions relative to the last present (not necessarily the same buffer). Some backends can benefit from that.

I did run into an issue with buffer age in softbuffer, which I seem to have made a small mistake in the Wayland implementation of (https://github.com/rust-windowing/softbuffer/pull/191). If everything else is fine, I can probably do a patch release with that for both Softbuffer 0.3 and 0.4. But currently this uses a git dependency with that patch.


ids1024:
Well, it crashes when I try to run on macOS, but I guess https://github.com/iced-rs/iced/issues/2113 is the issue there.

ids1024:
MacOS can be tested with softbuffer 0.4 in https://github.com/iced-rs/iced/pull/2191, since winit 0.29 fixes that issue.

> and the same of macOS

It seems I was misremembering. It's a bit awkward to implement buffer swapping with how the macOS backend works, so `.age()` on macOS currently always returns 0. So this loses damage tracking on macOS. Though it does avoid a copy during presentation.

Softbuffer could internally copy from a buffer, but I guess that would add unnecessary overhead for applications not using incremental rendering. 

That may be fine for now though. In previous discussion it was noted that Metal and thus wgpu should be pretty widely supported on macOS, so the software rendering fallback isn't as important there. It it also still performs pretty well, on an M1 Mac at least.

ids1024:
I've released softbuffer 0.4.1 and 0.3.4 with that fix.

This seems to work correctly across platforms. Other than that limitation on macOS, I think this is an overall improvement.

Having Iced updated here should also be handy for testing relevant improvements in softbuffer.

PR #2191, opened 2024-01-10
Update `wgpu` to `0.19`, `glyphon` to `0.5`, `softbuffer` to `0.4`, `window-clipboard` to `0.4`, and `raw-window-handle` to `0.6`

ids1024:
Based on https://github.com/iced-rs/iced/pull/2169 and https://github.com/iced-rs/iced/pull/2190. This updates wgpu to the latest `trunk` (the last release doesn't use `raw-window-handle` 0.6).

This can't be merged until those two PRs are, and when a wgpu release is made. I just wanted to test Iced with newer Wgpu changes. And to see how the API changes here.

This should make the API here safe, while `RawWindowHandle`/`RawDisplayHandle` can be unsafe to use, since they can contain pointers and don't constrain the lifetime.

Problems:
- To create a softbuffer `Context` and `Surface` from `W`, we need `W: Clone`. So this adds an `Arc` around `winit::window::Window`.
- `wgpu::WasmNotSendSync` bounds are a bit awkward.
- It should be possible for a `Compositor` to handle non-`'static` `W`. Need to try getting that to work.
- `window_clipboard` will need to be updated for raw-window-handle 0.6

If the way softbuffer and wgpu use `WindowHandle`/`DisplayHandle` is problematic for iced, perhaps changes could be made there.

Closes #2190.


ids1024:
Wgpu 0.19 has just been released.

But this depends on https://github.com/grovesNL/glyphon/pull/80, and an update to `window_clipboard`.

ids1024:
I think this should basically be ready pending glyphon and https://github.com/hecrj/window_clipboard/pull/24. If everything is working properly.

hecrj (approved):
Looks great! Thanks! :partying_face: 

I have simplified some stuff here and there. Let me know what you think and we can merge.

ids1024:
Those changes seem reasonable. Out-of-order ages won't currently happen with softbuffer, but might in the future if it makes sense for any backend, so that seems reasonable. Using the max age that's been seen seems like a better idea than what I had. It means not having damage tracking for the first couple frames (which perhaps is why I didn't think of it), but that's reasonable.

hecrj:
@ids1024 Awesome!

Currently blocked by https://github.com/gfx-rs/wgpu/pull/5088.

ids1024:
We could probably detect if the dx12 backend is being used an then set `desired_maximum_frame_latency` to `3` (which perhaps is what a fix in wgpu is going to do?). Though if they release a fix fairly soon that isn't really necessary.

dtzxporter:
> We could probably detect if the dx12 backend is being used an then set `desired_maximum_frame_latency` to `3` (which perhaps is what a fix in wgpu is going to do?). Though if they release a fix fairly soon that isn't really necessary.

This seems a bit hacky, I've spoken with the wgpu team and they intend to release a patch release for this fix. I already have an approved PR waiting to be merged that fixes the issue.

thunderstorm010:
https://github.com/gfx-rs/wgpu/issues/5088 is fixed by https://github.com/gfx-rs/wgpu/pull/5091!

hecrj:
Let's merge then!

PR #2194, opened 2024-01-11
Allow fetching the raw window handle of a window by it's `iced::window::Id`.

dtzxporter:
This is useful for specific platform code that the user may want to integrate with iced.

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -19,4 +19,5 @@ iced_core.workspace = true
 iced_futures.workspace = true
 iced_futures.features = ["thread-pool"]
 
+raw-window-handle.workspace = true
 thiserror.workspace = true
--- a/runtime/src/window.rs
+++ b/runtime/src/window.rs
@@ -15,6 +15,8 @@ use crate::core::{Point, Size};
 use crate::futures::event;
 use crate::futures::Subscription;
 
+use raw_window_handle::RawWindowHandle;
+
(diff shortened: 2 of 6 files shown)


ids1024:
The ecosystem is moving to `raw_window_handle::WindowHandle<'a>` type, which has a lifetime parameter, analogous to `BorrowedFd<'a>`. (This is necessary since raw window handles contain pointers in some cases, and there's a desire to offer safe APIs.)

Not sure how this would work. It could still provide a raw handle, but that would make it unsafe to work with. Or it could return `Arc<winit::Window>` or some wrapper around things like that. https://github.com/rust-windowing/raw-window-handle/issues/158 is relevant.

(https://github.com/iced-rs/iced/pull/2191 has a WIP update to the latest raw-window-handle with the safe handle types.)

hecrj:
@ids1024 We could allow providing a callback `impl FnOnce(WindowHandle<'a>)` on `Command` construction.

dtzxporter:
Because winit is unsound anyway, either method is technically introducing unsound code, and if winit rectifies the issue you would be unlikely to be able to Arc the window since the lifetime is tied to the windows and windows outlive the event loop.

I'm open to any approach FWIW that allows me to get the native handle from the window and manipulate it as it's necessary to fill a lot of issues with winit in general.

dtzxporter:
Closing because wgpu 0.19 is out and it will be easier to use RWH 0.6.

PR #2197, opened 2024-01-11
wgpu: require `Send` on stored pipelines

tzemanovic:
hi, thanks for this awesome crate!

I'm using `wgpu::Renderer/Backend` with `Send`, but since `pipeline_storage: pipeline::Storage` was added to `Backend` it is no longer `Send` (#2085). I'd like to restore this by adding the extra bound, it still works with the `examples/custom_shader/src/scene.rs`.


hecrj (approved):
Seems reasonable! Thanks :bow:

PR #2200, opened 2024-01-12
Provide native window access, via "fetch_native_handle" method on window.

dtzxporter:
This should alleviate some of the concerns in: https://github.com/iced-rs/iced/pull/2194#issuecomment-1888517346 because it will force the callback to be ran in the event loop in iced, and gives us a reference to the window directly instead. While the user _could_ mess up the iced state by making changes here, I think that anyone using this method should know what they are doing and it's an advanced use case.

Links:
#2194


dtzxporter:
Will rework this once #2191 lands.

hecrj (approved):
Thanks!

hecrj:
You call `as_raw`, then pattern match on `Wayland`, and you keep going from there.

https://docs.rs/iced/latest/iced/widget/shader/wgpu/rwh/enum.RawWindowHandle.html#variant.Wayland

WhyNotHugo:
I looked at that before asking here, but that variant's associated type is:

```
pub struct WaylandWindowHandle {
    pub surface: NonNull<c_void>,
}
```

I'm not sure how to operate on a `c_void` (which is basically a type-erased pointer). Can't this actually be a reference with a proper type instead?

PR #2207, opened 2024-01-20
Add `Interaction` overriding to `MouseArea`

VAWVAW:
The widget `MouseArea` allows to create button-like behavior but gives more control than `Button`. Currently there is no obvious way to change the mouse `Interaction` with the widget. I feel like `MouseArea` would be the most straightforward place to implement this.

This change adds the option to override the `Interaction` of the contained widget and set a custom one using the `mouse_interaction` callback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,7 @@ Many thanks to...
 - @tarkah
 - @tzemanovic
 - @varbhat
+- @VAWVAW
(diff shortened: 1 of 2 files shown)


hecrj (approved):
Thanks! :tada:

Made some changes to ensure the content interaction has priority over the `MouseArea` one, since the content event logic takes priority.

PR #2210, opened 2024-01-22
tiny_skia: Add a capacity limit to `GlyphCache`

MoSal:
* Trim the cache if `recently_used` size reaches the limit, even if a trim interval hasn't passed.
 * Shrink `entries` and `recently_used` to the limit when trimming.


MoSal:
Reformatted.

hecrj (approved):
Looks reasonable! Thanks! :bow:

PR #2211, opened 2024-01-22
feat(window): adds skip_taskbar for windows

MrAntix:
Trivial to add into the current way PlatformSpecific is implemented.

It would allow me to write a task tray app for windows.


MrAntix:
My pleasure, and thank you too.

PR #2214, opened 2024-01-25
Workaround issue with winit on windows not resuming the event loop.

dtzxporter:
With this fix specific to windows, we can at least process events that are scheduled when the user is resizing the window on win32 platform targets. This workaround can be removed if winit fixes the issue.

Works around: https://github.com/rust-windowing/winit/issues/3272


hecrj (approved):
Thanks!

I changed the approach to better centralize the workaround. Hopefully that still works!

PR #2219, opened 2024-01-27
Add border widths to `Toggler` `StyleSheet`

GyulyVGC:
This PR adds the possibility to customise the border widths of the `Toggler`.
This is needed in some circumstances and makes it consistent with all other widgets, that already have customisable border widths.
In this way we also remove the need to use `Option`s for the toggler border colors.


hecrj (approved):
Awesome! Thanks :bow:

PR #2220, opened 2024-01-29
Use `core::ptr::eq` to speed up `PartialEq` on `image::Bytes`

DoomDuck:
## Why
When loading images from memory `iced` takes a long time comparing `image::Handle`s.
To improve its speed using `core::ptr::eq` checks if the slices are identical  both the start pointer and the length.

## Effectiveness
I have not seen the optimization happen in either the standard library nor the results given by the compiler.
I also think I have seen the cost of `image::Handle`'s comparison in some `perf` data.


hecrj (approved):
Makes sense! Thanks :partying_face:

PR #2221, opened 2024-01-29
Text editor shrinking to content

Dworv:
I wanted to have a column with a mix of `text-editor`s and other elements, but the current implementation meant I needed to  create editors with fixed heights so that they wouldn't consume the entire space (not ideal). This PR resolves this issue by giving the option to shrink. Editors still default to filling the limit height.

Example usage: 
```rs
let editor = TextEditor::new(content).shrink_to_content(true);
```

No shrinking:
https://vimeo.com/907639154?share=copy

With shrinking
https://vimeo.com/907639138?share=copy

I'm not super familiar with how general layouts tend to work in GUI development, its possible that there are some oversights in my code and I may have overcomplicated the solution (as far as I could with 20 lines of code).


hecrj (approved):
Thanks!

I have fixed an issue with wrapped lines and renamed the `TextEditor` property to simply `height`.

The fix should have a bit of overhead, but `Shrink` editors should not be used without a text limit of some sort anyways!

Issue #2222, opened 2024-01-31
Wrong system information


3119369616:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

My real memory size is 16GB, but it is displayed incorrectly in the sample program `system_information`.
![image](https://github.com/iced-rs/iced/assets/38552079/8705872a-688f-475a-9197-6cd21b5e9df0)

### What is the expected behavior?

Show correct total memory size and used memory size.

### Version

master

### Operating System

Windows

### Do you have any log output?

```shell
No console output
```

brianch:
From what I see, [here](https://github.com/iced-rs/iced/blob/5540ac07e4695cc4e268979eca4efeb604b7c77f/winit/src/system.rs#L36) we call sysinfo's [`total_memory()`](https://docs.rs/sysinfo/0.30.5/sysinfo/struct.System.html#method.total_memory), which returns the value in bytes and not kylobytes as the documentation in Iced's [`Information`](https://docs.rs/iced_runtime/0.1.1/iced_runtime/system/struct.Information.html) struct says.

And so, in the example, changing:
```rust
ByteSize::kb(information.memory_total).to_string();
```
to
```rust
ByteSize::b(information.memory_total).to_string();
```
Fix the issue.

I can send a PR later to changing the example and the docs text if that's ok.

PR #2223, opened 2024-01-31
Small fix to the Information struct docs and fix memory info. in the system_information example.

brianch:
This fixes https://github.com/iced-rs/iced/issues/2222.

The sysinfo [total_memory()](https://docs.rs/sysinfo/0.30.5/sysinfo/struct.System.html#method.total_memory) called by Iced [here](https://github.com/iced-rs/iced/blob/5540ac07e4695cc4e268979eca4efeb604b7c77f/winit/src/system.rs#L36) returns the value in bytes and not kylobytes as the documentation in the `Information` struct previously said (the same for the process' memory usage).


hecrj (approved):
Thanks! :bow:

PR #2228, opened 2024-02-02
Refactor tooltip function to accept a generic element as a tooltip

jhannyj:
Allows for a more flexible tooltip element that can be further customized.


hecrj (approved):
Sweet! Thank you :bow:

PR #2234, opened 2024-02-03
Always mask text in tiny_skia renderer

### LINKED ISSUE #2173: tiny-skia does not truncate long text outside of Element when it is inside a Column

### What happened?

I'm not sure if this applies to other elements, but I noticed that if a button contains really long, unwrapped text, and is inside another element, with the ``tiny-skia`` renderer, the text bleeds out.

It was a little tricky to create an sscce, but here's the code to reproduce it:
```rust
use iced::widget::{button, column, container, row, scrollable, Button, Space};
use iced::{Element, Length, Sandbox};

// Long text mustn't contain symbols otherwise the renderer will wrap the text.
static LONG: &'static str = "looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong";
static SHORT: &'static str = "short";

    Example::run(iced::Settings {
            size: iced::Size::from([500.0, 500.0]),
        },
        ..Default::default()
    })
}

struct Example;

#[derive(Debug, Clone, Copy)]
struct Message;

impl Sandbox for Example {
    type Message = Message;

    fn new() -> Self {
        Example
    }

    fn title(&self) -> String {
        "skia-bounding".into()
    }

    fn theme(&self) -> iced::Theme {
        iced::Theme::Dark
    }

    fn update(&mut self, _message: Message) {}

        let dummy_content = (0..100)
            .map(create_button)
            .map(|button| {
                row![
                    button
                        .width(Length::Fill) // Setting the width to anything that's NOT "Length::Shrink" will do.
                        // .width(Length::Fixed(10.0)) // Uncomment to witness abmomination
                    ,    
                    Space::with_width(1)// Also necessary, **MUST** be at least 1
                ]
            })
            .fold(column![], |row, elem| row.push(elem));

        let inner = container(scrollable(dummy_content))
            .padding(50); // set padding to see text bleeding outside

        // let inner = container(content).padding(50); // Doesn't have to be in a scrollable

        container(inner).into()
    }
}

fn create_button<'a>(i: usize) -> Button<'a, Message> {
    button(match i % 2 == 0 {
        true => LONG,
        false => SHORT,
    })
}
```

To add in ``cargo.toml``:
```toml
[features]
wgpu = ["iced/wgpu"]

[dependencies.iced]
git = "https://github.com/iced-rs/iced.git"
default-features = false
```
Result:
Screenshot 2023-12-23 122605

Putting it into a scrollable and moving it results in this: 
Screenshot 2023-12-23 122616

Compiling the sscce with ``features=wgpu`` produces the expected outcome: 
Screenshot 2023-12-23 122649

### Version

master

### Operating System

Windows

_No response_

Koranir:
Fixes #2173 
Not sure exactly why physical_bounds says it isn't in the clip_bounds, but looking into the tiny_skia code it seems that there isn't any performance hit if we add a mask or not.

### DIFF
--- a/tiny_skia/src/backend.rs
+++ b/tiny_skia/src/backend.rs
@@ -456,16 +456,13 @@ impl Backend {
                     return;
                 }
 
-                let clip_mask = (!physical_bounds.is_within(&clip_bounds))
-                    .then_some(clip_mask as &_);
-
                 self.text_pipeline.draw_paragraph(
                     paragraph,
                     *position,
                     *color,
                     scale_factor,
                     pixels,
-                    clip_mask,
+                    Some(clip_mask),
                     transformation,
                 );
             }
@@ -482,16 +479,13 @@ impl Backend {
                     return;
                 }
 
-                let clip_mask = (!physical_bounds.is_within(&clip_bounds))
-                    .then_some(clip_mask as &_);
-
                 self.text_pipeline.draw_editor(
                     editor,
                     *position,
                     *color,
                     scale_factor,
                     pixels,
-                    clip_mask,
+                    Some(clip_mask),
                     transformation,
                 );
             }
@@ -514,9 +508,6 @@ impl Backend {
                     return;
                 }
 
-                let clip_mask = (!physical_bounds.is_within(&clip_bounds))
-                    .then_some(clip_mask as &_);
-
                 self.text_pipeline.draw_cached(
                     content,
                     *bounds,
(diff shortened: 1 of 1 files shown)


hecrj:
> looking into the tiny_skia code it seems that there isn't any performance hit if we add a mask or not

How so? It seems to me that the presence of a `mask` does add a new blitting stage:

https://github.com/RazrFalcon/tiny-skia/blob/ce3fe9bab0659fb928d6c22c03b74e009a8a2b77/src/pipeline/blitter.rs#L86-L88

PR #2235, opened 2024-02-04
Make the svg function to accept a path

fogarecious:
Currently, to load an SVG file by the [svg](https://docs.iced.rs/iced/widget/fn.svg.html) function, we have to call [Handle::from_path](https://docs.iced.rs/iced/widget/svg/struct.Handle.html#method.from_path).

```rust
use iced::widget::{svg, svg::Handle};
// ...
svg(Handle::from_path("path-to-pic.svg"));
```

A more ideal way is to pass the path directly into the svg function.

```rust
svg("path-to-pic.svg");
```

This PR enhances the convenience by adding `From<T: Into<PathBuf>>` to the [Svg Handle](https://docs.iced.rs/iced/widget/svg/struct.Handle.html). The modified code achieves consistency with the [image](https://docs.iced.rs/iced/widget/image/index.html) counterpart.


hecrj (approved):
Thanks!

PR #2237, opened 2024-02-04
Use `TypeId` to identify `subscription::Map`

### LINKED ISSUE #2236: Subscritptions are run twice

### What happened?

I started to notice that the Subscription::channel closure in Cosmic Term run twice on startup, and started to track it backwards in the stack. This only happens with --release builds, and not an all platforms ( https://github.com/jackpot51 could not reproduce this on his hardware ), on my Thinkpad T15 Gen 1, it happens every time. 

So, I notice that in iced/winit/src/application.rs we actually have two calls to runtime.track(application.subscription(...)). One in run_instance and another in update. Both these calls will trigger a new call to cosmic terms subscription::channel closure. This should not happen since there is a check in tracker.rs to prevent this:
https://github.com/iced-rs/iced/blob/878134883e5298aa8e12261a3ebd85879eae1d19/futures/src/subscription/tracker.rs#L85

But for some reason, this check fails, the id between the first call is different from the calls in the update function. So, when I started to dig further I found that the hashing function for Map, is the one that is getting different from the first to the following calls, and more exact it is the `self.mapper.hash(state);` that differs. 

https://github.com/iced-rs/iced/blob/878134883e5298aa8e12261a3ebd85879eae1d19/futures/src/subscription.rs#L166

And I debug print the self.mapper, the function pointers have different values. So, for some reason, it seems that the compiler, when run in release mode, decides to perform some optimizations to either move the closure between the calls, or create two different closures (I'm no expert in compiler optimizations, I just see that the closure have different pointers for the call from run_instance and update ). 

This is a quite critical bug, since that causes Subscription::channel to be run twice, and if you don't count on that but assume it only run once, you may end up with a dead channel in the application. 
In Cosmic Term, this was worked around by commit 
https://github.com/pop-os/cosmic-term/commit/4401a7dbef3e1579af9694c5f6c38c1b788a86e2

The iced version is 
https://github.com/pop-os/iced/tree/6115280d5277c50b8539d4eff6ab61050c51b592
since that is what libcosmic uses. But, I don't see anything that would have fixed this. 

This is with rustc rustc 1.75.0 (82e1608df 2023-12-21)
some info from /proc/cpuinfo
```
model name  : Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz
flags       : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtsc
p lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx est tm2 ssse3 sdbg fm
a cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb ssbd ibrs ibpb
stibp ibrs_enhanced tpr_shadow flexpriority ept vpid ept_ad fsgsbase tsc_adjust sgx bmi1 avx2 smep bmi2 erms invpcid mpx rdseed adx smap clflushopt intel_pt x
saveopt xsavec xgetbv1 xsaves dtherm ida arat pln pts hwp hwp_notify hwp_act_window hwp_epp vnmi md_clear flush_l1d arch_capabilities
vmx flags   : vnmi preemption_timer invvpid ept_x_only ept_ad ept_1gb flexpriority tsc_offset vtpr mtf vapic ept vpid unrestricted_guest ple pml ept_mode_base
```

The real fix would be to make the `self.mapper.hash(state);` stable.

### Version

master

### Operating System

Linux

_No response_

hecrj:
This PR fixes #2236. The idea here is to use the `TypeId` of a closure instead of the hash of a function pointer in `subscription::Map`, which I am assuming is more stable.

~~Coincidentally, this also introduces closure support for `Subscription::map`!~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `size_hint` not being called from `element::Map`. [#2224](https://github.com/iced-rs/iced/pull/2224)
 - `size_hint` not being called from `element::Explain`. [#2225](https://github.com/iced-rs/iced/pull/2225)
 - Slow touch scrolling for `TextEditor` widget. [#2140](https://github.com/iced-rs/iced/pull/2140)
+- `Subscription::map` using unreliable function pointer hash to identify mappers. [#2237](https://github.com/iced-rs/iced/pull/2237)
 
 Many thanks to...
 
--- a/futures/src/subscription.rs
+++ b/futures/src/subscription.rs
@@ -10,6 +10,7 @@ use crate::{BoxStream, MaybeSend};
 
 use futures::channel::mpsc;
 use futures::never::Never;
+use std::any::TypeId;
 use std::hash::Hash;
 
 /// A stream of runtime events.
@@ -88,7 +89,10 @@ impl<Message> Subscription<Message> {
     }
 
     /// Transforms the [`Subscription`] output with the given function.
-    pub fn map<A>(mut self, f: fn(Message) -> A) -> Subscription<A>
+    pub fn map<A>(
+        mut self,
+        f: impl Fn(Message) -> A + MaybeSend + Clone + 'static,
+    ) -> Subscription<A>
     where
         Message: 'static,
         A: 'static,
@@ -97,8 +101,9 @@ impl<Message> Subscription<Message> {
             recipes: self
                 .recipes
                 .drain(..)
-                .map(|recipe| {
-                    Box::new(Map::new(recipe, f)) as Box<dyn Recipe<Output = A>>
+                .map(move |recipe| {
+                    Box::new(Map::new(recipe, f.clone()))
+                        as Box<dyn Recipe<Output = A>>
                 })
                 .collect(),
         }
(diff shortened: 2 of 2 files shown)


snaggen:
I assume using TypeId will change the semantics slightly, that will only update the subscription if the closure is changed in such a way that the type changes. If you change the closure in such a way that the type is still the same, the subscription will not be updated. So, it is desired to have this kind of functionality, maybe an explicit id like in the subscription::channel function is preferred. That would break the API though.  So, not sure what is better, change the semantics or break API.

hecrj:
> I assume using TypeId will change the semantics slightly, that will only update the subscription if the closure is changed in such a way that the type changes.

Technically yes, but it should work as expected with function item types; which is the intended way to use the API here.

It'd be great to be able to restrict the closure to function item types, but I don't think that's possible in Rust yet.

snaggen:
In that case, I say, just document it and ship it :)

PR #2238, opened 2024-02-05
Decouple `Key` from `keyboard::Modifiers` and apply them to `text` in `KeyboardInput`

wash2:
Input like "ctrl + [a-z]" produces text like "a" instead of control characters when not accounting for modifiers. For Character keys, in most cases the text from the winit `KeyEvent` should match the received character anyway, so maybe it would be better to pass the text with modifiers accounted for.


hecrj:
I see. What is the use case for this?

It also seems that `text_with_all_modifiers` is not available on Wasm.

hecrj (approved):
I think this makes sense. I have also decoupled the `Key` from the keyboard modifiers, since it seems `winit` doesn't do that by default.

Wasm keeps the old behavior, since the supplement trait isn't implemented there. We'll have to figure out how to keep the API consistent there eventually.

wash2:
> I see. What is the use case for this?
> 
> It also seems that `text_with_all_modifiers` is not available on Wasm.

Oh sorry, I did not see this until now :sweat_smile: . The control characters are important in a terminal application, like cosmic-term, as an example.

PR #2240, opened 2024-02-07
Add support for primary clipboard

snaggen:
Add support for primary clipboard by using https://github.com/hecrj/window_clipboard/pull/25

I didn't update Cargo.toml, so it will not compile until window_clipboard is released and iced depend on that. 

Also widget/src/text_editor.rs is not updated in this PR, since that is a highlevel widget which would require some rework to update primary on selection, and paste on middle click.


hecrj (approved):
Thanks! :bow: 

Made some minor changes, but everything looks good.

PR #2242, opened 2024-02-09
Implement win32 workaround for multi_window application as well. Also trigger AboutToWait for Moved as well.

dtzxporter:
Fixes for multi_window as well. Move on windows breaks the event loop as well.


dtzxporter:
Audit failure seems unrelated.

hecrj (approved):
Thanks!

I apparently reverted your changes and forgot to apply the patch to `multi_window`.

PR #2243, opened 2024-02-09
Add `show_window_menu` action

ids1024:
Winit currently supports this only on Windows and Wayland.

This requests that a context menu is shown at the cursor position, like the menu normally triggered by right clicking the title bar. This is important for implementing client side decorations with Iced widgets.

### DIFF
--- a/runtime/src/window/action.rs
+++ b/runtime/src/window/action.rs
@@ -81,11 +81,11 @@ pub enum Action<T> {
     GainFocus(Id),
     /// Change the window [`Level`].
     ChangeLevel(Id, Level),
-    /// Show window menu at cursor position.
+    /// Show the system menu at cursor position.
     ///
     /// ## Platform-specific
     /// Android / iOS / macOS / Orbital / Web / X11: Unsupported.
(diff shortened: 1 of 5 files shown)


runtime/src/window/action.rs
```diff
@@ -81,6 +81,11 @@ pub enum Action<T> {
     GainFocus(Id),
     /// Change the window [`Level`].
     ChangeLevel(Id, Level),
+    /// Show window menu at cursor position.
+    ///
+    /// ## Platform-specific
+    /// Android / iOS / macOS / Orbital / Wayland / Web / X11: Unsupported.
+    ShowWindowMenu(Id),
```
Mingun, line 88: You said, it is supported on Wayland?

ids1024: Ah, I copied that from winit. It looks like the comment there wasn't updated when the Wayland implementation was added.

hecrj (approved):
Thanks! :partying_face: 

I renamed the action to `show_system_menu` to avoid redundancy since `window` is already the name of the module.

PR #2244, opened 2024-02-09
Separate image widget support from image codecs.

dtzxporter:
This optimization with feature flags allows a couple of cases:

- You can now use the Image widget without having any codecs configured (Raw pixel buffers).
- A release + lto binary shaves off 1.7mb on average removing codecs and still supporting the image widget.

Now, the from_path and from_memory methods are gated and will only be present when one or more codec features are enabled.


dtzxporter:
Audit issue is unrelated.

dtzxporter:
Thinking about this it may make more sense to enable the image widget by default, and turn the "image" feature into actual codec support in general. That way it would be a non-breaking change and still a win-win size wise.

PR #2249, opened 2024-02-12
Enable image widget by default (without any image load/codec support)

dtzxporter:
This is a non-breaking change that allows using images loaded with a different library in your program, or raw pixel buffers without including all of the bloat that image-rs is.

It's another attempt at #2244 that is backwards compatible with existing applications because they will already use the image feature if they want the widget.


hecrj:
Superseded by #2244,

PR #2252, opened 2024-02-15
Add `clip` property to `Container`, `Column`, `Row`, and `Button`

hecrj:
This PR makes clipping the contents of a container opt-in, which should help with text being cut-off by default when line height isn't high enough.


GyulyVGC:
Being `false` the default does this mean that `Text` shouldn't be cut-off when line height is not suitable?

Since 0.10 I've experienced some texts vanishing and I was only able to solve the problem by _decreasing_ the line height of the corresponding text. 
Not increasing, but _decreasing_.
And this is still true after this PR.

PR #2259, opened 2024-02-18
[wgpu.image] Workaround WGPU OpenGL heuristics

### LINKED ISSUE #1774: svg example initially displays black rectangle on Ubuntu

### What happened?

This is my Framework laptop:

```
OS: Ubuntu Cinnamon 22.04.2 LTS x86_64 
Host: Laptop A6 
Kernel: 5.19.0-38-generic 
Uptime: 3 mins 
Packages: 2692 (dpkg), 13 (snap) 
Shell: bash 5.1.16 
Resolution: 3006x2004 
DE: Cinnamon 5.2.7 
WM: Mutter (Muffin) 
WM Theme: (Default) 
Theme: Adwaita [GTK2/3] 
Icons: gnome [GTK2/3] 
Terminal: gnome-terminal 
CPU: 11th Gen Intel i7-1165G7 (8) @ 4.700GHz 
GPU: Intel TigerLake-LP GT2 [Iris Xe Graphics] 
Memory: 2471MiB / 31889MiB 
```

1. `cargo run --package svg`
2. Note that console says `vulkan: No DRI3 support detected - required for presentation Note: you can probably enable DRI3 in your Xorg config`
3. See app launch with black rectangle instead of tiger.
4. Either resize the window, or toggle the color-filter checkbox.
5. See expected tiger.
6. Per advice on [web](https://askubuntu.com/a/817232), `nano /etc/X11/xorg.conf.d/20-intel.conf`, add `Option "DRI" "3"`, and reboot.
7. Repeat steps. Tiger displays on first try, as expected.

FWIW, it seems to work fine on my main machine, a Beelink SER4 on Ubuntu 20.04. That's a Ryzen 4700U. And I discovered the issue because my own Iced app, which displays SVGs, exhibits the same behavior (except the images are fixed size, so I can't poke them the way I can with the svg example app).

Screenshot from 2023-03-30 19-08-54

Screenshot from 2023-03-30 19-09-46

see above

### Version

master

### Operative System

Linux

_No response_

### LINKED ISSUE #2180: Black rectangles instead of images when using the wgpu GLES backend

### What happened?

image

You get black rectangles instead of images if the wgpu GLES backend is used.

Reproduced under Linux with:
- Apple M1 Pro (asahi) - GLES is the default, because there is no Vulkan driver yet
- Intel Xe - by setting `WGPU_BACKEND=gl`
- AMD Radeon Vega - by setting `WGPU_BACKEND=gl`

I have tracked it to this wgpu issue: https://github.com/gfx-rs/wgpu/issues/1614

I have tried this https://github.com/gfx-rs/wgpu/issues/4665#issuecomment-1806012977 workaround - forcing the Atlas to start with 2 layers - and it fixes the issue (at least for the sake of confirming that it's this specific wgpu problem).

Probably the same issues: 

- https://github.com/iced-rs/iced/issues/1898
- https://discord.com/channels/628993209984614400/1188197273339973662 
- https://discourse.iced.rs/t/images-arent-properly-loading/84

The images are shown the same way as with other backends.

### Version

master

### Operating System

Linux

_No response_

PolyMeilex:
wgpu OpenGL image rendering has been broken for years (every image is a black rectangle), but finally I accumulated enough willpower to debug and fix this.

So, it turns out that wgpu [has to use heuristics](https://github.com/gfx-rs/wgpu/blob/004e3efe84a320d9331371ed31fa50baa2414911/wgpu-hal/src/gles/mod.rs#L371) to guess that the texture we are creating will be used as `TEXTURE_2D_ARRAY` rather than `TEXTURE_2D`, and OpenGL needs to know that ahead of time, otherwise it will refuse to bind it because of type mismatch.

The real fix should be on wgpu side, but that seems to be known and not obvious problem.
So in the meantime, I just increased default layers allocation in the atlas to 2. That way wgpu will see layer depth 2 and correctly assume that we want to use this texture as `TEXTURE_2D_ARRAY`

EDIT: Found an issue that mentions the array layer limitations: https://github.com/gfx-rs/wgpu/issues/1574

Fixes #1774.
Fixes #2180.

### DIFF
--- a/wgpu/src/image/atlas.rs
+++ b/wgpu/src/image/atlas.rs
@@ -23,11 +23,15 @@ pub struct Atlas {
 }
 
 impl Atlas {
-    pub fn new(device: &wgpu::Device) -> Self {
-        // We start with 2 layers, to help wgpu figure out that this texture is `GL_TEXTURE_2D_ARRAY` rather
-        // than `GL_TEXTURE_2D`
-        // https://github.com/gfx-rs/wgpu/blob/004e3efe84a320d9331371ed31fa50baa2414911/wgpu-hal/src/gles/mod.rs#L371
-        let layers = vec![Layer::Empty, Layer::Empty];
+    pub fn new(device: &wgpu::Device, backend: wgpu::Backend) -> Self {
+        let layers = match backend {
+            // On the GL backend we start with 2 layers, to help wgpu figure
+            // out that this texture is `GL_TEXTURE_2D_ARRAY` rather than `GL_TEXTURE_2D`
+            // https://github.com/gfx-rs/wgpu/blob/004e3efe84a320d9331371ed31fa50baa2414911/wgpu-hal/src/gles/mod.rs#L371
+            wgpu::Backend::Gl => vec![Layer::Empty, Layer::Empty],
+            _ => vec![Layer::Empty],
+        };
+
         let extent = wgpu::Extent3d {
             width: SIZE,
             height: SIZE,
--- a/wgpu/src/image/atlas.rs
+++ b/wgpu/src/image/atlas.rs
@@ -24,10 +24,14 @@ pub struct Atlas {
 
 impl Atlas {
     pub fn new(device: &wgpu::Device) -> Self {
+        // We start with 2 layers, to help wgpu figure out that this texture is `GL_TEXTURE_2D_ARRAY` rather
+        // than `GL_TEXTURE_2D`
+        // https://github.com/gfx-rs/wgpu/blob/004e3efe84a320d9331371ed31fa50baa2414911/wgpu-hal/src/gles/mod.rs#L371
+        let layers = vec![Layer::Empty, Layer::Empty];
         let extent = wgpu::Extent3d {
             width: SIZE,
             height: SIZE,
-            depth_or_array_layers: 1,
+            depth_or_array_layers: layers.len() as u32,
         };
 
         let texture = device.create_texture(&wgpu::TextureDescriptor {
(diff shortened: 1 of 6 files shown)


hecrj (approved):
Thanks! We finally figured out the mystery :partying_face: 

Looks like #2009 was quite close!

PolyMeilex:
Great, when the amount of layers is a multiple of 6 we will lose images again, as wgpu hard-codes 6 to a cube map 🤦
I belive this is still better then nothing tho.

hecrj:
@PolyMeilex Looking at the code, if we are hitting the first branch of the `match` it means the texture isn't cube compatible; so we should be safe and never hit that problem.

wgpu/src/image/atlas.rs
```diff
@@ -23,11 +23,19 @@ pub struct Atlas {
...
 impl Atlas {
-    pub fn new(device: &wgpu::Device) -> Self {
+    pub fn new(device: &wgpu::Device, backend: wgpu::Backend) -> Self {
+        let layers = match backend {
+            // On the GL backend we start with 2 layers, to help wgpu figure
+            // out that this texture is `GL_TEXTURE_2D_ARRAY` rather than `GL_TEXTURE_2D`
+            // https://github.com/gfx-rs/wgpu/blob/004e3efe84a320d9331371ed31fa50baa2414911/wgpu-hal/src/gles/mod.rs#L371
+            wgpu::Backend::Gl => vec![Layer::Empty, Layer::Empty],
+            _ => vec![Layer::Empty],
+        };
```
hecrj, line 33: Made it so that the additional memory is only allocated when using OpenGL.

hecrj (approved):
Thank you! Long-standing issues gone :partying_face:

PR #2260, opened 2024-02-18
add helper fun for Padding

wiiznokes:
This add some new method to the Padding struct. 

I have some other ideas, like also provide some more method like

```rust 
/// Create a [`Padding`] with equally spaces of the right and left side
    pub fn vertical(padding: f32) -> Self {
        Self {
            left: padding,
            right: padding,
            ..Padding::ZERO
        }
    }
```
and the same for vertical, but that will involve breaking change.

And also provide a constructor API alternative:

```rust 
pub fn vertical(self, padding: f32) -> Self {
        self.left = padding;
        self.right = padding;
        self
    }
```

to allow writting this:

```rust 
let padding = Padding::horizontal(10f32)
            .bottom(5f32);
```

But this will involve, reanaming the method i written like such

```rust 
fn new_top()
fn new_bottom()
fn new_vertical()
```

which is fine imo.

I think it make more sense to provide this funcions rather than

```rust
pub fn vertical(self) -> f32 {
        self.top + self.bottom
    }
```

because i don't really see a lot of use case for this (user perspective).

### DIFF
--- a/core/src/padding.rs
+++ b/core/src/padding.rs
@@ -46,7 +46,7 @@ pub struct Padding {
 }
 
 impl Padding {
-    /// Padding of zero
+    /// [`Padding`] of zero
     pub const ZERO: Padding = Padding {
         top: 0.0,
         right: 0.0,
@@ -54,7 +54,7 @@ impl Padding {
         left: 0.0,
     };
 
-    /// Create a Padding that is equal on all sides
+    /// Create a [`Padding`] that is equal on all sides
     pub const fn new(padding: f32) -> Padding {
         Padding {
             top: padding,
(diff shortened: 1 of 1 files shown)


core/src/padding.rs
```diff
@@ -64,6 +64,38 @@ impl Padding {
         }
     }
 
+    /// Create a top Padding
```
hecrj, line 67: Let's use the link notation and end with a period:

```rust
/// Create a top [`Padding`].
```

hecrj (changes requested):

Koranir:
I think that instead of these new function being constructors, they should be builder-like functions. It makes the api a lot more flexible, at the small cost of turning `Padding::bottom(5.0)` into `Padding::ZERO.bottom(5.0)`.

Koranir:
E.g.
```rust
    /// Sets the top padding to `padding`.
    pub const fn top(self, padding: f32) -> Self {
        Self {
            top: padding,
            ..self
        }
    }
```

wiiznokes:
> I think that instead of these new function being constructors, they should be builder-like functions. It makes the api a lot more flexible, at the small cost of turning Padding::bottom(5.0) into Padding::ZERO.bottom(5.0).

Yes that what i was saying. But we can still use `Padding::new_bottom(5.0)` for constructor and `padding.bottom(5.0)` for builder.

What do you think @hecrj ?

Koranir:
> > I think that instead of these new function being constructors, they should be builder-like functions. It makes the api a lot more flexible, at the small cost of turning Padding::bottom(5.0) into Padding::ZERO.bottom(5.0).
> 
> Yes that what i was saying. But we can still use `Padding::new_bottom(5.0)` for constructor and `padding.bottom(5.0)` for builder.
> 
> What do you think @hecrj ?

The naming convention for constructors of this type is `with_<thing>`, so `Padding::with_bottom(5.0)` in this example. However, I don't think having separate constructors for each is required here and just adds unnecessary complexity.

wiiznokes:
> However, I don't think having separate constructors for each is required here and just adds unnecessary complexity.

I better than

```rust
Padding {
   bottom: 5.0,
   ..Padding::ZERO
}
```
(one line instead of four)

Koranir:
> > However, I don't think having separate constructors for each is required here and just adds unnecessary complexity.
> 
> I better than
> 
> ```rust
> Padding {
>    bottom: 5.0,
>    ..Padding::ZERO
> }
> ```
> 
> (one line instead of four)

With builder pattern it would be
```rust
Padding::ZERO.bottom(5.0)
```

wiiznokes:
> With builder pattern it would be
> 
> ```rust
> Padding::ZERO.bottom(5.0)
> ```

True, still prefer the `with_bottom` though.

wiiznokes:
we could also add `with_vertical`, `with_horizontal` and change the behavior of `vertical` and `horizontal`.

wiiznokes:
I think we could also remove the From impl

PR #2261, opened 2024-02-18
Add a colored variant for Container style

wiiznokes:
This is useful for debugging


dtzxporter:
Instead of adding a new variant, it may make more sense to implement From<Color> for the style like some of the other widgets do so you can just have a colored container that way without adding another branch here.

hecrj (approved):
Made some changes so that the following works:
```rust
container.style(container::background(color))
```

Thanks!

PR #2262, opened 2024-02-18
Allow disabled `TextInput` to still be interacted with

### LINKED ISSUE #1880: Disabled TextInput should still allow mouse interaction, selection and copying

### What happened?

```
impl iced::Application for MyApp {
    pub fn view(&self) -> iced::Element<Message> {
        TextInput::new("", "Test")
    }
}
```

image

Copying text from readonly/disabled text input is useful and, from perspective of building UI, it is one of the reasons to use readonly/disabled text input (`TextInput::new()`) instead of a plain text label (`Text::new`).

#### Comparison

Web: 

image

Native Windows:

image

I would expect:

1. Enabled mouse interaction
2. "Arrow" cursor
3. Being able to select text
4. Being able to copy selected text

### Version

0.7

### Operative System

Windows

_No response_

Koranir:
Resolves #1880

Follows Firefox's text selection usage (i.e. no blinking cursor, arrow keys don't do anything unless shift is pressed, cursor click position is remembered).

Draft for the following reasons:
- Should the placeholder text be selectable?
- Should this apply to `TextEditor`s as well?
- Currently, the allowed operations when disabled are copy and select all in addition to the selection primitives. Should there be more?

### DIFF
--- a/widget/src/text_input.rs
+++ b/widget/src/text_input.rs
@@ -395,11 +395,11 @@ where
                             position,
                         );
 
-                    let is_cursor_visible = ((focus.now - focus.updated_at)
-                        .as_millis()
-                        / CURSOR_BLINK_INTERVAL_MILLIS)
-                        % 2
-                        == 0;
+                    let is_cursor_visible = !is_disabled
+                        && ((focus.now - focus.updated_at).as_millis()
+                            / CURSOR_BLINK_INTERVAL_MILLIS)
+                            % 2
+                            == 0;
 
                     let cursor = if is_cursor_visible {
                         Some((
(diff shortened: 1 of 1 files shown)


habics:
> * Should this apply to `TextEditor`s as well?

Yes please, imo TextEditor is a TextInput on steroids, it should have all the behavior of TextInput plus more features.
There are many befits to have this behavior for TextEditor as well, for example using it to display long informational text that can be copied without editing it.

Koranir:
> > * Should this apply to `TextEditor`s as well?
> 
> Yes please, imo TextEditor is a TextInput on steroids, it should have all the behavior of TextInput plus more features. There are many befits to have this behavior for TextEditor as well, for example using it to display long informational text that can be copied without editing it.

Now that I've taken a closer look at the TextEditor architecture, it seems impossible to add this, as the TextEditor's state is held by the user. This means that the user has to provide a way to manipulate the state via the set_action method, but that would enable the text input instead.

What would be more appropriate would be to have a separate TextDisplay or similar widget that provides copy/paste capabilities and maybe some additional customisation options.

habics:
@Koranir 
I understand, thank you for taking a closer look.

hecrj (approved):
Thanks!

Made some changes here and there. Hopefully I didn't break anything.

PR #2269, opened 2024-02-19
Add a configuration to scrollable to always show the scrollbar. Adds a disabled style for scrollbar.

dtzxporter:
Disabled styles are used when a scrollbar is shown, but no content is overflowing.

### DIFF
--- a/style/src/theme.rs
+++ b/style/src/theme.rs
@@ -1249,6 +1249,34 @@ impl scrollable::StyleSheet for Theme {
             Scrollable::Custom(custom) => custom.dragging(self),
         }
     }
+
+    fn disabled(&self, style: &Self::Style) -> scrollable::Appearance {
+        if let Scrollable::Custom(custom) = style {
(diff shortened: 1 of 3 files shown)


style/src/theme.rs
```diff
@@ -1249,6 +1249,34 @@ impl scrollable::StyleSheet for Theme {
...
+                            ..active.scrollbar.scroller.border.color
+                        },
+                        ..active.scrollbar.scroller.border
+                    },
+                },
+                ..active.scrollbar
+            },
+            ..active
+        }
+    }
```
hecrj, line 1279: Let's make this the default implementation!

hecrj (changes requested):

dtzxporter:
CI Issues are unrelated to the PR.

hecrj (approved):
Thanks!

I ended up reimplementing the feature on top of `master` and made the scrollbars actually affect content layout when always shown.

PR #2279, opened 2024-02-21
Add `text::Wrapping` support

nrjais:
This exposes wrapping strategy for `text` widget. I have kept the default to `Wrapping::Word` which it was in `cosmic_text`,
There are other widgets which are using text, I am not sure if Wrapping should be exposed there.

Sample
image

### DIFF
--- a/core/src/text.rs
+++ b/core/src/text.rs
@@ -41,6 +41,9 @@ pub struct Text<Content = String, Font = crate::Font> {
 
     /// The [`Shaping`] strategy of the [`Text`].
     pub shaping: Shaping,
+
+    /// The [`Wrapping`] strategy of the [`Text`].
+    pub wrapping: Wrapping,
 }
 
 /// The shaping strategy of some text.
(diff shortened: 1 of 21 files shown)


core/src/text.rs
```diff
@@ -40,6 +40,23 @@ pub struct Text<'a, Font> {
...
+/// The wrapping strategy of some text.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
+pub enum Wrapping {
+    /// No wrapping.
+    ///
+    /// This is the default.
+    None,
+    /// Word wrapping.
+    #[default]
+    Word,
```
Mingun, line 57: It seems that default is `Word` if see at https://github.com/iced-rs/iced/pull/2279/files#diff-b8dc21f0859d1f4cc40ec8a7bc71112427a373b082e7d101a94fe807887b54b4R54

core/src/widget/text.rs
```diff
@@ -50,6 +51,7 @@ where
             vertical_alignment: alignment::Vertical::Top,
             shaping: Shaping::Basic,
             style: Default::default(),
+            wrapping: Wrapping::Word,
```
Mingun, line 54: Maybe use `Wrapping::default()`?

core/src/text.rs
hecrj: The documentation needs to be updated.

core/src/widget/text.rs
hecrj: `default` seems good.

core/src/text.rs
nrjais: Updated

habics:
Hi, will this be available for TextEditor widget too or only for Text widget #2291

habics:
Please consider adding this feature to TextEditor widget too

nrjais:
Added for text editor also

nrjais:
Rebased for latest theme changes

hecrj (approved):
Thanks! :tada:

PR #2280, opened 2024-02-22
`fetch_position` command for `window` module

n1ght-hunter:
add a fetch_location action in logical pixels.
this is needed to impl client side decorators as the location needs to be know when resizing the top or left side.


hecrj (approved):
Thanks!

Issue #2283, opened 2024-02-22
Add command for setting window mouse passthrough mode


jquesada2016:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

I am attempting to create an overlay application but cannot enable/disable hit testing/mouse passthrough for the mouse. In `winit`, this is `Window::set_hittest`.

### What is the expected behavior?

`iced` to provide a window command for setting mouse passthrough behavior.

### Version

crates.io release

### Operating System

Windows

PR #2284, opened 2024-02-22
Add command for setting mouse passthrough

jquesada2016:
This PR addresses #2283 by exposing `winit`'s ability to set mouse hit testing.

I extended the `window::Action` enum to include two new entries:

- `EnableMousePassthrough`: enables the window passing through mouse events to whatever's underneath
- `DisableMousePassthrough`: disables mouse passthrough, essentially allowing the window to capture mouse events

I also added two respective functions for creating these commands, `enable_mouse_passthrough`, and `disable_mouse_passthrough`.

### DIFF
--- a/runtime/src/window.rs
+++ b/runtime/src/window.rs
@@ -147,6 +147,18 @@ pub enum Action {
 
     /// Screenshot the viewport of the window.
     Screenshot(Id, oneshot::Sender<Screenshot>),
+
+    /// Enables mouse passthrough for the given window.
+    ///
+    /// This disables mouse events for the window and passes mouse events
+    /// through to whatever window is underneath.
+    EnableMousePassthrough(Id),
+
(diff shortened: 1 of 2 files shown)


dtzxporter:
May be a little verbose to have two actions when you could just make it iced::window::set_cursor_hittest(id, enabled) instead?

jquesada2016:
> May be a little verbose to have two actions when you could just make it iced::window::set_cursor_hittest(id, enabled) instead?

@dtzxporter  I do personally agree. I ended up splitting it into both because I tried sticking to the rest of the `window::Action` variants, and it felt more consistant this way. Also, using `Set` in the name implies to me there should be a mechanism for fetching the hit test state, which `winit` does not have.

hecrj (approved):
Thanks!

PR #2285, opened 2024-02-23
add option for undecorated_shadow on windows

n1ght-hunter:
adds an option to enable undecorated_shadow on windows. something that is needed for client decorations.
this could actually be done on the client side though. since it just makes the window smaller and adds a 1px border around the edge.


n1ght-hunter:
also there is something about shadow on macos but its not documented properly so have no idea what it is. if someone is able to test that would be great.

link to the code in winit
https://github.com/rust-windowing/winit/blob/a127bd6f6684909852000cdf32469bac3b611887/src/platform/macos.rs#L103

hecrj (approved):

PR #2303, opened 2024-02-27
[Minor Fix] Scrollbar style when cursor is over scrollable area

GyulyVGC:
Small PR to fix a minor issue introduced in v0.12.
With this change, scrollbars display correctly also when `cursor_over_scrollable` is some.


GyulyVGC:
Ping @hecrj since this is a good fit for v0.12.2

PR #2309, opened 2024-03-03
ProgressBar buffer, vertical, reverse

ThisIsRex:
Improved the `ProgressBar` widget:
Added a `buffer` property (additional bar like value)
Added a `vertical` and a `reverse` orientation (`width` and `height` are replaced with `length` and `size` for that)


hecrj:
Thanks! But I'd like to keep the widgets simple and with a single responsibility, if possible.

Also, I'm unsure about where vertical and reversed progress bars may be used, and whether they make for good UX or not.

Maybe these could be contributed to `iced_aw` instead?

hecrj:
Okay. Today is the day I needed some vertical progress bars.

I'm going to implement some of these changes!

PR #2315, opened 2024-03-08
Add Command + ArrowLeft/Right input behavior for macOS

Brady-Simon:
This PR adds macOS-specific behavior for the input and text editor for Command + ArrowLeft to behave like Home and Command + ArrowRight to behave like End to match [the native behavior](https://support.apple.com/en-us/HT201236#:~:text=Command–Left%20Arrow%3A%20Move%20the,end%20of%20the%20current%20line.). This behavior is particularly useful for users with MacBooks since they don't have Home/End keys.

I added a new `platform::is_macos_command_pressed` function instead of using `modifiers.command()` since the latter defaults to CTRL on other platforms and represents a jump instead of moving to the beginning/end of the line.

This PR leaves out other shortcuts like Command + ArrowUp -- which would also move to the cursor to the beginning of the line for single line inputs and the beginning of the document for multi-line inputs -- to limit the scope of changes.

### DIFF
--- a/widget/src/text_editor.rs
+++ b/widget/src/text_editor.rs
@@ -726,7 +726,7 @@ impl Update {
                     }
 
                     if let keyboard::Key::Named(named_key) = key.as_ref() {
-                        if let Some(motion) = motion(named_key) {
+                        if let Some(motion) = motion(named_key, modifiers) {
                             let motion = if platform::is_jump_modifier_pressed(
                                 modifiers,
                             ) {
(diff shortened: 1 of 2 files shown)


oldgalileo:
Could this also be extended to address `Option + <Arrow>`?

Brady-Simon:
`Option + <Arrow>` is a jump between words instead of moving to the beginning/end of the line like Command, so that should be working already on recent versions of Iced. What behavior are you seeing with option + arrow?

widget/src/text_editor.rs
```diff
@@ -789,14 +789,22 @@ impl Update {
     }
 }
 
-fn motion(key: key::Named) -> Option<Motion> {
+fn motion(key: key::Named, modifiers: keyboard::Modifiers) -> Option<Motion> {
     match key {
+        key::Named::Home | key::Named::ArrowLeft
+            if platform::is_macos_command_pressed(modifiers) =>
```
hecrj, line 795: I believe the `if` guard here will affect the `Home` keybinding too.

Brady-Simon: Ah, good catch, I misunderstood how the conditional would be applied and I didn't test on another platform. I've adjusted the match arms accordingly.

hecrj (approved):
Thanks!

PR #2322, opened 2024-03-10
Fix scrollbar dragged state after release left mouse button

LuisLiraC:
**Description:**
Ensure the correct handling of event_status after checking all possible events. Previously, it was returned before checking Mouse Events, resulting in the scrollbar being dragged continuously from the first click onwards.

**Before:**
- The scrollbar is still dragged even after releasing the left button.

ezgif-1-5d725d3d5f

**After:**
- The scrollbar is released when the left button is no longer clicked.

ezgif-1-cc0517bf70

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -390,6 +390,21 @@ where
             )
         };
 
+        if matches!(
+            event,
+            Event::Mouse(mouse::Event::ButtonReleased(mouse::Button::Left))
+                | Event::Touch(
+                    touch::Event::FingerLifted { .. }
+                        | touch::Event::FingerLost { .. }
+                )
+        ) {
+            state.scroll_area_touched_at = None;
+            state.x_scroller_grabbed_at = None;
+            state.y_scroller_grabbed_at = None;
+
+            return event_status;
+        }
+
         if let event::Status::Captured = event_status {
             return event::Status::Captured;
         }
@@ -479,10 +494,7 @@ where
                             );
                         }
                     }
-                    touch::Event::FingerLifted { .. }
-                    | touch::Event::FingerLost { .. } => {
-                        state.scroll_area_touched_at = None;
-                    }
+                    _ => {}
                 }
 
                 event_status = event::Status::Captured;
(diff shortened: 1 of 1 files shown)


widget/src/scrollable.rs
```diff
@@ -650,6 +646,10 @@ where
             }
         }
 
+        if let event::Status::Captured = event_status {
+            return event::Status::Captured;
+        }
+
         event_status
```
PgBiel, line 653: Didn't look much into the logic, but, assuming it is correct, is the early return still necessary? Perhaps we can just return `event_status` normally now, even if it's `Captured`.

LuisLiraC: I think that while the functionality of the scrollable widget remains unaffected by the returned state value, it's still crucial to discern which event is being processed by the widget. In this scenario, the widget is actively listening to events, adjusting values that directly influence its behavior, that's why the return state being marked as Captured.

```
/// The status of an [`Event`] after being processed.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum Status {
    /// The [`Event`] was **NOT** handled by any widget.
    Ignored,

    /// The [`Event`] was handled and processed by a widget.
    Captured,
}
```

PgBiel: Yeah, sorry, I should have been clearer. I mean that, here, you're returning `event::Status::Captured` if `event_status` is `event::Status::Captured`; otherwise, you return `event_status`. But that's the same as just returning `event_status`.

In other words, the following would likely have the same behavior (unless there are side effects I'm not seeing here):

```suggestion
        event_status
```

LuisLiraC: Oh yeah, you are right. I was thinking the same when I was checking the logic again some few seconds ago. Thanks for noticing. I'm going to make the commit.

hecrj:
This doesn't seem right. If the contents of the scrollable capture the event; the scrollable should not handle it.

This is most likely caused by some other widget capturing the mouse release when it shouldn't.

Do you have an SSCCE of the issue?

hecrj (approved):
The real issue here is that the `TextEditor` is incorrectly capturing `mouse::Event::ButtonReleased`.

However, I don't think an incorrect widget implementation should break the scrollable; so I changed it a bit to only prioritize the release events even if they are captured.

Thanks!

PR #2324, opened 2024-03-11
Implement setter for filter_method property of viewer

Gigas002:
This PR implements a simple setter for a [viewer struct](https://github.com/iced-rs/iced/blob/master/widget/src/image/viewer.rs) in a way, like image struct does. Sorry if this is so by design, don't really see a reason why not implement it.


hecrj (approved):
No reason! It was just missing.

Thanks! :bow:

PR #2326, opened 2024-03-12
Use closures for widget styling

hecrj:
This PR replaces the function pointers used for styling since #2312 with boxed closures. These closures can both capture and borrow application state.

We trade a bit of overhead for a lot of flexibility. No need to juggle `themer`s everywhere!


soucosmo:
Incredible, this is something very important, the styling of the components was something very complicated.

PR #2327, opened 2024-03-12
Fix wgpu examples link in issue template

fenhl:
The examples have moved from `/wgpu/examples` to `/examples`, and the default branch has been renamed from `master` to `trunk`. This PR updates the `BUG-REPORT` issue template accordingly.


hecrj (approved):
Thanks!

PR #2335, opened 2024-03-16
Focus handling for buttons

njust:
Hi, this PR adds focus handling to buttons: 

https://github.com/iced-rs/iced/assets/2380253/59a15716-497f-4097-8722-6257f13dcc11


hecrj:
Proper focus management and accessibility are planned for next releases.

These are core features that require deep knowledge of the codebase and vision of the library.

PR #2338, opened 2024-03-18
Introduce `text` macro

derezzedex:
This PR introduces a `text` macro in the `widget::helpers` module analogous to the [`std::format`](https://doc.rust-lang.org/std/macro.format.html) macro.

### Implementation
This is a simple macro similar to the `row` and `column` helper macros, it uses `format` internally.
The `macro_rules` definition:
```rust
#[macro_export]
macro_rules! text {
    ($($arg:tt)*) => {
        $crate::Text::new(format!($($arg)*))
    };
}
```

### Usage
Since it uses the [`std::format`](https://doc.rust-lang.org/std/macro.format.html) macro internally, it's usage is analogous and should be familiar to most users, but some examples are documented in the macro definition:
 ```rust
use iced::Element;
use iced::widget::text;

     let simple = text!("Hello, world!");

     let keyword = text!("Hello, {}", "world!");

     let planet = "Earth";
     let local_variable = text!("Hello, {planet}!");
     // ...
 }
 ```

As a result, the `text(format!(` pattern has been replaced with the new `text` macro throughout the `examples`.


romancitodev:
It would be nice to be able to receive as a second parameter (optional) a custom style.

romancitodev:
> I would say that `.style()` or `.color()` express the styling intent much better than a second parameter on this macro.

Well, that's make more sense!

hecrj (approved):
Looks good! Thanks :bow:

PR #2343, opened 2024-03-19
Introduce `auto-detect-theme` feature

hecrj:
This PR changes the `Default` implementation of `Theme` if the `auto-detect-theme` feature is enabled to return an appropriate theme based on the dark/light mode of the OS.

The runtime should eventually react to OS theme changes—but that is quite an edge case and this simple approach gets us a lot of value already.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ all-features = true
 maintenance = { status = "actively-developed" }
 
 [features]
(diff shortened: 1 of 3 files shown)


Rudxain:
frewsxcv/rust-dark-light#39

TLDR: please use an exact version, preferably `=1.0.0`, as I've tested it works

PR #2355, opened 2024-03-26
git ref readme_deps

helloimalemur:
Hey Core Team!

Approaching Iced I found myself stumbling with getting dependencies to cooperate when using the version of Iced published to Crates.io while also referencing the current round of documentation on iced.rs.
I wasn't able to just "copy and paste" any of the examples and expect it work without coaxing.

I merely suggest a small change to the ReadMe file to include the git reference version of the Iced dependency, making users aware of their ability to reference the master branch directly in their Cargo.toml

Thanks for considering the suggestion!
James

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,8 +79,8 @@ you want to learn about a specific release, check out [the release list].
 
 [the release list]: https://github.com/iced-rs/iced/releases
 
-The majority of documentation currently follows the master branch,
-it may be suggested to use git referenced dependency in your Cargo.toml file.
+As the majority of the documentation follows the development branch,
+the latest revision of iced should be used for maximal compatibility.
 
 ```toml
 [dependencies.iced]
--- a/README.md
+++ b/README.md
@@ -79,6 +79,17 @@ you want to learn about a specific release, check out [the release list].
 
 [the release list]: https://github.com/iced-rs/iced/releases
 
+The majority of documentation currently follows the master branch,
+it may be suggested to use git referenced dependency in your Cargo.toml file.
+
+```toml
+[dependencies.iced]
+git = "https://github.com/iced-rs/iced.git"
+branch = "master"
+features = ["canvas", "tokio", "debug", "async-std", "lazy", "webgl"]
+```
+
+
 ## Overview
 
 Inspired by [The Elm Architecture], Iced expects you to split user interfaces


Koranir:
`docs.rs` documentation should be used for the `crates.io` releases. `docs.iced.rs` is for the latest master. On `iced.rs`, there are links to `docs.rs` for documentation on the latest release and to `docs.iced.rs` (which is explicitly referred to as the development branch).

I'm not quite sure how you ended up on the development branch's documentation, perhaps that is the problem we should be focusing on.

Koranir:
> The "Get Started" button on [iced.rs](https://iced.rs/) takes you to `book.iced.rs` which coincides with the development branch (mentions the usage of new patterns like iced::run(), which seems to be new to 0.13, instead of implementing Application, and running the run method on it). Perhaps the "Get Started" button on the main page should take the user to `https://docs.iced.rs/iced/index.html` until the book is complete and coincides with a release?

Yeah, that should probably be the case. Looking at the book, it does have a header at the top that does say it is intended for the latest master branch, but that could easily be missed.

README.md
```diff
@@ -79,6 +79,17 @@ you want to learn about a specific release, check out [the release list].
 
 [the release list]: https://github.com/iced-rs/iced/releases
 
+The majority of documentation currently follows the master branch,
+it may be suggested to use git referenced dependency in your Cargo.toml file.
```
Koranir, line 83:
```suggestion
As the majority of the documentation follows the development branch,
the latest revision of iced should be used for maximal compatibility.
```

README.md
```diff
@@ -79,6 +79,17 @@ you want to learn about a specific release, check out [the release list].
...
+
+```toml
+[dependencies.iced]
+git = "https://github.com/iced-rs/iced.git"
+branch = "master"
+features = ["canvas", "tokio", "debug", "async-std", "lazy", "webgl"]
```
Koranir, line 89: Should these be the set of dependencies we recommend? Might be better to not include them at all.

Koranir (changes requested):

README.md
helloimalemur: Thanks for the edit!
I do believe it would be helpful for anyone else who may be approaching the lib completely green to it, for them to be aware of this nuance, at least until 0.13 is released and all the docs align with a release. It could potentially help with adoption in the meantime.

hecrj:
Thanks! Closing given we have since removed the "Installation" section from the main `README`.

PR #2356, opened 2024-03-27
Utilize bytes::Bytes for images

Bajix:
This PR is to replace image::Bytes with bytes::Bytes as it's more optimized/feature rich and is basically a drop in replacement.

### DIFF
--- a/core/src/image.rs
+++ b/core/src/image.rs
@@ -1,10 +1,11 @@
 //! Load and draw raster graphics.
+pub use bytes::Bytes;
+
 use crate::{Rectangle, Size};
 
 use rustc_hash::FxHasher;
 use std::hash::{Hash, Hasher as _};
 use std::path::PathBuf;
-use std::sync::Arc;
 
 /// A handle of some image data.
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -29,12 +30,12 @@ impl Handle {
     pub fn from_pixels(
         width: u32,
         height: u32,
-        pixels: impl AsRef<[u8]> + Send + Sync + 'static,
+        pixels: impl Into<Bytes>,
     ) -> Handle {
         Self::from_data(Data::Rgba {
             width,
             height,
-            pixels: Bytes::new(pixels),
+            pixels: pixels.into(),
         })
     }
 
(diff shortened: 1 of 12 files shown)


core/src/image.rs
```diff
@@ -1,9 +1,10 @@
 //! Load and draw raster graphics.
+use bytes::Bytes;
```
hecrj, line 2: We should re-export the type.

hecrj (changes requested):

core/src/image.rs
Bajix: Updated.

Bajix:
Advantages of using bytes::Bytes:

- Vec<u8> and other types can be converted from without there being extra indirection and without reallocating
- Bytes can be loaded, cloned and converted into an image without reallocating. This makes it easier to author data loading behavior that converts into images or is otherwise deserialized with serde

hecrj (approved):
Looks great!

I made some further changes to use `Bytes` for the `ImageBuffer` data containers as well. This way, we avoid reallocating completely when `Handle::from_pixels` is used!

I have also removed the eager `Data` hashing and instead decided to use the `Bytes` pointer as the identifier of a `Handle`.

hecrj:
> decided to use the Bytes pointer as the identifier of a Handle.

Terrible idea. It's common for allocators to reuse memory causing false cache hits in some edge cases. Changed it to use an auto-incremented unique `Id` with `AtomicU64` in b52c7bb610f593fffc624d461dca17ac50c81626.

core/src/image.rs
```diff
@@ -80,90 +101,46 @@ where
...
+pub struct Id(_Id);
 
-impl PartialEq for Bytes {
-    fn eq(&self, other: &Self) -> bool {
-        let a = self.as_ref();
-        let b = other.as_ref();
-        core::ptr::eq(a, b) || a == b
-    }
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+enum _Id {
```
Bajix, line 121: Instead of introducing `_Id`, another option is to keep `Id` as is but to make Unique/Hash `#[doc(hidden)]`. Your call though

hecrj: We don't want to expose the variants. Just hiding the docs isn't a proper way to do that.

Bajix:
New changes look fantastic!

I think this work will tie in really nicely into a generalized asset loading solution whereby a generation arena stores load state, updates create new requests by initializing a loading state and getting an index, views read the current load state using the index, and the arena updates load state with bytes when processing responses. This will be super efficient both for loading images and for deserialization with borrows

PR #2358, opened 2024-03-29
Expose line_height & text_size fields for the text_editor widget

bungoboingo:
^

### DIFF
--- a/widget/src/text_editor.rs
+++ b/widget/src/text_editor.rs
@@ -110,6 +110,21 @@ where
         self
     }
 
+    /// Sets the text size of the [`TextEditor`].
+    pub fn size(mut self, size: impl Into<Pixels>) -> Self {
+        self.text_size = Some(size.into());
+        self
+    }
+
+    /// Sets the [`text::LineHeight`] of the [`TextEditor`].
+    pub fn line_height(
+        mut self,
+        line_height: impl Into<text::LineHeight>,
+    ) -> Self {
+        self.line_height = line_height.into();
+        self
+    }
+
     /// Sets the [`Padding`] of the [`TextEditor`].
     pub fn padding(mut self, padding: impl Into<Padding>) -> Self {
         self.padding = padding.into();


bungoboingo:
Added one more commit which "fixes" this issue:

https://github.com/iced-rs/iced/assets/4241774/85e13e2a-3503-4813-8523-170c2f228ebf

(Issue is known: https://github.com/pop-os/cosmic-text/issues/219)

After explicitly changing the wrap mode to "Glyph" we no longer overflow text buffer bounds in certain cases.

https://github.com/iced-rs/iced/assets/4241774/a0843a6d-ded9-4022-bbb3-5988ffdaad3a

I believe this is better than using `Wrap::Word` since it removes some jank. Can go back to using default `Word` when this issue is fixed on cosmic_text side.

hojjatabdollahi:
How does Wrap::Glyph wrap a normal sentence? Doesn't it just break the last word?

graphics/src/text/editor.rs
```diff
@@ -75,6 +75,12 @@ impl editor::Editor for Editor {
             cosmic_text::Shaping::Advanced,
         );
 
+        // There is no graceful fallback from word -> glyph wrapping in cosmic_text, so if you type
+        // a bunch of letters in a row they will overflow the text bounds. For now, we explicitly
+        // set glyph wrapping to avoid overflow issues.
+        //see: https://github.com/pop-os/cosmic-text/issues/219
+        buffer.set_wrap(font_system.raw(), cosmic_text::Wrap::Glyph);
```
hecrj, line 82: We should set this before `set_text` so we can avoid a relayout.

hecrj (changes requested):

hecrj:
It seems a fix may have been released in the latest `cosmic-text` version (https://github.com/pop-os/cosmic-text/issues/219#issuecomment-1924566617).

I plan to update soon.

bungoboingo:
I've reverted the last commit, just now just exposing those two methods 👍  Hopefully this is fixed in cosmic-text!

hecrj (approved):
Thanks! :partying_face:

PR #2360, opened 2024-04-01
Use generic `Content` in `Text` to avoid reallocation in `fill_text`

hecrj:
`Renderer::fill_text` asks for a `&str` just to turn it into a `String`... Amazing.

Let's get rid of that, shall we?

### DIFF
--- a/widget/src/overlay/menu.rs
+++ b/widget/src/overlay/menu.rs
@@ -526,7 +526,7 @@ where
 
             renderer.fill_text(
                 Text {
-                    content: &option.to_string(),
+                    content: option.to_string(),
                     bounds: Size::new(f32::INFINITY, bounds.height),
                     size: text_size,
                     line_height: self.text_line_height,
--- a/core/src/renderer/null.rs
+++ b/core/src/renderer/null.rs
@@ -67,7 +67,7 @@ impl text::Renderer for () {
 
     fn fill_text(
         &mut self,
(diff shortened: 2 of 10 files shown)


widget/src/overlay/menu.rs
```diff
@@ -526,7 +526,7 @@ where
 
             renderer.fill_text(
                 Text {
-                    content: &option.to_string(),
+                    content: option.to_string(),
```
hecrj, line 529: No more `(&option.to_string()).to_string()`! I will sleep tight today.

alex-ds13: Hey, since you're looking into this `.to_string` reallocations couldn't the `text` helper function avoid reallocation as well here:
[Text::new(text.to_string())](https://github.com/iced-rs/iced/blob/b40db569a2369dbdc79e40dc700974c2f7ac8d4b/widget/src/helpers.rs#L154)
I always tend to use `Text::new(&self.some_str)` instead of `text(&self.some_str)` because of this.

hecrj: Not if we want to keep the `ToString` compatibility (e.g. `text(10)`). Specialization hasn't landed yet in Rust.

alex-ds13: Didn't think of that. I guess I can always make my own helper! 😄

PR #2363, opened 2024-04-01
Specialize `widget::text` helper

hecrj:
This PR specializes the arguments of `widget::text` by manually implementing a new `text::IntoFragment` trait for the primitive types. No more extra allocations when calling `text("Hello!")` or `text(some_string)`! :tada: 

The downside is that the `text` helper cannot be used with all the implementors of `ToString` anymore. However, it is always possible to implement the `text::IntoFragment` trait for local types.

Furthermore, a new `widget::value` helper has been introduced with the same functionality as the old `text` helper.

cc @alex-ds13

### DIFF
--- a/core/src/widget/text.rs
+++ b/core/src/widget/text.rs
@@ -21,7 +21,7 @@ where
     Theme: Catalog,
     Renderer: text::Renderer,
 {
-    content: Cow<'a, str>,
+    content: Content<'a>,
     size: Option<Pixels>,
     line_height: LineHeight,
     width: Length,
@@ -39,9 +39,9 @@ where
     Renderer: text::Renderer,
 {
     /// Create a new fragment of [`Text`] with the given contents.
-    pub fn new(content: impl Into<Cow<'a, str>>) -> Self {
+    pub fn new(content: impl IntoContent<'a>) -> Self {
         Text {
-            content: content.into(),
+            content: content.into_content(),
             size: None,
             line_height: LineHeight::default(),
             font: None,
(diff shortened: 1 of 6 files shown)


alex-ds13:
Thank you! This is pretty good. I use a lot of `Arc<str>` on my apps which allow me to have one allocation with multiple references and I was avoiding using `text()` helper precisely because it would create a new allocation when it wasn't needed.

> manually implementing a new `text::IntoFragment` trait for the primitive types

Did you leave out `usize` and `isize` on purpose or is it just missing?

In case of custom local types is there really any need to implement `IntoFragment`? Can't you simply call `.to_string()` on it? Won't `IntoFragment` have the same allocations as `.to_string` in those cases?

Isn't this:
```rust
struct MyStruct;
impl ToString for MyStruct {
     fn to_string(&self) -> String {
         "MyStruct".into()
     }
}
//...
App {
    my_struct: MyStruct
}

impl App {
      text(self.my_struct.to_string())
      .into()
  }
}
```

Pretty much the same thing as this:
```rust
struct MyStruct;
impl ToString for MyStruct {
     fn to_string(&self) -> String {
         "MyStruct".into()
     }
}
impl<'a> IntoFragment<'a> for MyStruct {
     fn into_fragment(self) -> Fragment<'a> {
         Fragment::Owned(self.to_string())
     }  
}
impl<'a> IntoFragment<'a> for &MyStruct {
     fn into_fragment(self) -> Fragment<'a> {
         Fragment::Owned(self.to_string())
     }  
}
//...
App {
    my_struct: MyStruct
}

impl App {
      text(&self.my_struct)
      .into()
  }
}
```
Both cases will allocate a new `String` won't they?

This means that for custom local or remote types that implement `ToString` you should simply call `.to_string()` on them.

hecrj:
@alex-ds13 Say `MyStruct` has a `String`, then you can:

```rust
struct MyStruct {
    some_text: String,
}

impl<'a> IntoFragment<'a> for &'a MyStruct {
    fn into_fragment(self) -> Content<'a> {
        Cow::Borrowed(&self.some_text)
    }
}

let my_struct = MyStruct { some_text: String::from("Some text!") };

text(&my_struct)
```

It saves you from typing `text(&my_struct.some_text)` everywhere and lets you hide implementation details better.

hecrj:
Also, `ToString` is never implemented directly—but instead it's implemented through `Display`, which invokes the `fmt` machinery. `IntoFragment` lets you skip that if you want.

alex-ds13:
That's true. Overall I think this is a good change and if anyone needs to use a remote type that they can't implement `IntoFragment` they can always fallback to the `value` helper function. Or if they need to access an internal value like your example they can always wrap that type on a local type and implement `IntoFragment` for that.

alex-ds13:
Was there any technical reason for you to alias `Cow` to `Fragment` or was it just so it would fit better with the trait name?

hecrj:
@alex-ds13 This is so you don't need yet an additional import (`use std::borrow::Cow`) to implement `IntoFragment` if you are already importing `iced::widget::text`.

PR #2374, opened 2024-04-05
add `Command::dispatch` as a shortcut for `Command::perform` without action

Fogapod:
This function is useful when you want to send event without state. I did not find other way to do it with current api other than `Command::perform(async {}, |_| Message::MyMessage)` which is inconvenient and creates a useless extra future.

New api is `Command::dispatch(Message::MyMessage)` where

```rust
enum Message {
    MyMessage,
}
```

### DIFF
--- a/runtime/src/command.rs
+++ b/runtime/src/command.rs
@@ -41,6 +41,14 @@ impl<T> Command<T> {
         Self::single(Action::Widget(Box::new(operation)))
     }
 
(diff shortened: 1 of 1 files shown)


hecrj:
This encourages an anti-pattern.

Messages are meant to represent real events that happened. You should not use them to manually run your application logic. Simply define helper functions instead, and call them!

PR #2376, opened 2024-04-07
Add stronger guarantee of readability/contrast for palette background/text color pairs

Davidster:
If we run the todo example with the SolarizedDark theme, we can see that the text color selected for display on top of a cyan background is white ('All' button):

image

which [brutally fails the WCAG contrast guidelines](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=35C9BE).

This change guarantees that the text color passes the test, and [passes the guidelines](https://webaim.org/resources/contrastchecker/?fcolor=000000&bcolor=35C9BE) by selecting black text for the example in question.

### DIFF
--- a/core/src/theme/palette.rs
+++ b/core/src/theme/palette.rs
@@ -612,11 +612,19 @@ fn mix(a: Color, b: Color, factor: f32) -> Color {
 
 fn readable(background: Color, text: Color) -> Color {
     if is_readable(background, text) {
-        text
-    } else if is_dark(background) {
+        return text;
+    }
+
+    let fallback = if is_dark(background) {
         Color::WHITE
     } else {
         Color::BLACK
+    };
+
+    if is_readable(background, fallback) {
+        fallback
+    } else {
+        fallback.inverse()
     }
 }
 


hecrj:
Ended up reverting this because the solution here is just trading some bad cases for others.

The `fallback.inverse()` case still does not guarantee readability and it is affecting the `Light` and `Dark` theme variants, which are the most common.

Davidster:
Darn, well surely if swapping to the complete other end of the luminance spectrum doesn't fix the contrast then there is no possible way to pass the WCAG test for that particular background color, right? Maybe if neither white nor black passes the test then we could take the one that's closest to passing?

Davidster:
we could also use the AA contrast ratio (3:1) instead of AAA (7:1) as a minimum

hecrj:
> Maybe if neither white nor black passes the test then we could take the one that's closest to passing?

@Davidster Good idea. Did that in https://github.com/iced-rs/iced/commit/72b975ec82660b39f27b6cb015b763caf20e6483 and seems to work well.

PR #2380, opened 2024-04-08
Remove extra Download struct from dowload_progress example

alex-ds13:
I've just noticed that on the `download_progress` example there is an extra `Download` struct on the `download.rs` file that is not being used. For some reason, `rust_analyzer` doesn't give any warnings not even clippy. But that struct has no references to it.

I guess that it might have first been created to be used to pass the `id` and `url` from `main` to the `download::file` function, but then it ended up passing both parameters instead.

I don't know why didn't `rust_analyzer` pick this up before, but I tried running the example without it and it runs just fine. So might as well remove it so people don't confuse it with the actual `Download` struct that is created on the `main.rs` file.

### DIFF
--- a/examples/download_progress/src/download.rs
+++ b/examples/download_progress/src/download.rs
@@ -12,12 +12,6 @@ pub fn file<I: 'static + Hash + Copy + Send + Sync, T: ToString>(
     })
 }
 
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks! :bow:

PR #2386, opened 2024-04-15
Allow checking whether a `TextEditor` is focused

mtkennerly:
My use case is that I have a widget for wrapping `TextInput` and handling undo/redo shortcuts if the widget is focused. I'd like to do the same for `TextEditor`, but it doesn't currently expose a way to check for focus. I added an `is_focused` method, like the one that `TextInput` has.

I tested both of these checks with these changes:

* `tree.state.downcast_ref::<iced::widget::text_editor::State<iced::advanced::text::highlighter::PlainText>>().is_focused()`
* `tree.state.downcast_ref::<iced::widget::text_editor::State<iced::highlighter::Highlighter>>().is_focused()`

### DIFF
--- a/widget/src/text_editor.rs
+++ b/widget/src/text_editor.rs
@@ -319,7 +319,9 @@ where
     }
 }
 
-struct State<Highlighter: text::Highlighter> {
+/// The state of a [`TextEditor`].
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks! We should eventually consider implementing `operation::Focusable` and `operation::TextInput` for it too.

PR #2390, opened 2024-04-17
Text input maybe

JL710:
Just a small addition for the `TextInput` widget.

The `TextInput` has currently no `*_maybe` methods like `Button` has the `on_press_maybe`. This PR adds them.

Added Methods:
- `on_input_maybe`
- `on_submit_maybe`
- `on_paste_maybe`

These methods, especially the `on_input_maybe`, are usefull in the same manner as `Button::on_press_maybe` and can simplify the building of uis containing the TextInput.


hecrj (approved):
Thanks!

PR #2392, opened 2024-04-18
Enable horizontal scrolling without shift modifier

### LINKED ISSUE #2359: Cannot use mousewheel without shift to scoll a horizontal-only scrollable.

### What happened?

I am working on an applicaiton where I have two splits, one scrolling horizontally and one vertically:

image

I can scroll the vertical one with the mousewheel just fine, but not the horizontal one. It demands I hold down shift, even though there's no vertical scrollbal present (the scrollable is Horizontal, not Both, so it can only scroll in the one direction anyway). I don't believe I am able to change this behaviour.

Shift should be needed only when both scrollbars are present. Otherwise, scrollwheel should just scroll.
Alternatively, I should be able to specify whether I want shift scrolling on a Horizontal scrollable.

### Version

crates.io release

### Operating System

Windows

_No response_

woelfman:
fixes #2359 

The current implementation strictly scrolls vertically without the shift modifier and horizontally with the shift modifier. This patch checks the modifier only when the direction is Both.

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -706,15 +706,29 @@ where
 
                 let delta = match delta {
                     mouse::ScrollDelta::Lines { x, y } => {
-                        // TODO: Configurable speed/friction (?)
-                        let movement = if !cfg!(target_os = "macos") // macOS automatically inverts the axes when Shift is pressed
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks!

Issue #972, opened 2021-07-29
Container Hovered Style


johannesvollmer:
Hi! Is it correct that the container currently does not have a hovered style? Would you accept a PR?

13r0ck:
You are correct, the is not a style for hovering a container. There are a few technical reasons for it
Here is Hecrj's input from the Zulip chat about events for entering a container: 
```
"
However, `on_mouse_enter` has implementation challenges. What happens if the `view` changes and a new widget ends up under the mouse cursor? The user would expect the event to trigger, but currently it won't.
"
```
Still not a no though. PRs are always welcome 😃 , just no promises that they will get merged (and will likely take a while even if they are)

yusdacra:
Why not just use a button?

johannesvollmer:
I'm trying to display a list of complex items. Each item is a Row. When the user hovers the row, the background color of the whole row should change. I didn't think this could be implemented with a button, especially since the row will contain buttons itself. Will it be possible to use a button for this? That would be nice

hecrj:
I think this use case should be tackled directly by some kind of `Table` widget.

tfachmann:
I've seen `iced` version `0.14` will include a `Table` widget. Will it be intended for such use-case?

> I'm trying to display a list of complex items. Each item is a Row.
>> Why not just use a button?

I am myself thinking about just using a button as there will be no button inside, but I am unsure whether that's the correct approach

plorefice:
I'm facing the same exact issue and a solution that is working fairly well for me and doesn't feel hack-ish like using a button is to surround the item container with a `MouseArea` and listening for `on_enter` and `on_exit` events to set an `hovered` flag in my row item. Then I can style the container differently depending on whether it is hovered or not. Something like this:

```
            mouse_area(
                container(row)
                    .style(|theme| match msg.is_hovered {
                        true => theme::hovered_row(theme),
                        false => container::Style::default(),
                    }),
            )
            .on_enter(Message::MouseEnter(msg.id))
            .on_exit(Message::MouseLeave(msg.id))
            .into()
```

PR #2395, opened 2024-04-20
Add alternative (right click) for buttons

Gremious:
My app needs buttons that are right-clickable as a secondary, alternative action.

I am adding a very simple `on_press_alt` to go alongside `on_press`:

1. A button must still have `on_press` or else be disabled, as usual. Whether `on_press_alt` is missing or present is irrelevant.
2. As such, no `on_press_alt_maybe` - only removing the `on_press` counts as disabling the button.
  a) Which is why it checks whether both are present before allowing `state.is_pressed_alt = true;`

I have tested this on the counter example.
It sends different messages ok and styles the button on ether click if present.
Disabling `on_press` re-styles as disabled and stops both from working as expected.

### DIFF
--- a/widget/src/button.rs
+++ b/widget/src/button.rs
@@ -53,6 +53,7 @@ where
 {
     content: Element<'a, Message, Theme, Renderer>,
     on_press: Option<Message>,
+    on_press_alt: Option<Message>,
     width: Length,
     height: Length,
     padding: Padding,
@@ -75,6 +76,7 @@ where
         Button {
             content,
             on_press: None,
+            on_press_alt: None,
             width: size.width.fluid(),
             height: size.height.fluid(),
             padding: DEFAULT_PADDING,
@@ -118,6 +120,15 @@ where
         self
     }
 
+    /// Sets the message that will be produced when the [`Button`] is pressed
+    /// with the right mouse button.
+    ///
+    /// Unless `on_press` is also called, the [`Button`] will be disabled.
+    pub fn on_press_alt(mut self, on_press: Message) -> Self {
+        self.on_press_alt = Some(on_press);
+        self
+    }
+
     /// Sets whether the contents of the [`Button`] should be clipped on
     /// overflow.
     pub fn clip(mut self, clip: bool) -> Self {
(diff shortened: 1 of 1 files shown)


hecrj:
Ideally, we want to keep widgets as simple as possible and with a single purpose. I think you could achieve something similar with [the `MouseArea` widget](https://docs.rs/iced/latest/iced/widget/struct.MouseArea.html).

Additionally, buttons with a right click handler are not very common and, therefore, not ideal for UX.

Gremious:
Mouse area is what I have right now, and the reason for this PR.

I cannot achieved similar things with a mouse area, as I cannot easily re-create a button: I cannot style a mouse area or the container around it on hover (https://github.com/iced-rs/iced/issues/972), and creating the on-click (active) styling also suddenly becomes cumbersome if I'm gonna be sending messaged and rebuilding elements just to restyle for a brief moment (compared to just having a button which already does that underneath).

(Plus, when i'm doing messges like this, the timing of that doesn't seem as smooth/nice as what buttons do? Might be a me issue.)

I didn't, then, want to copy-paste the entire button widget in my own code and only add those few lines.

So, I felt like contributing this simple case. I implore you to reconsider, but if not, I guess I have to go the annoying route of having my own button widget, or having the clutter of a bunch of mesaages bouncing back and forth with different near-identical containers switching places. I _do_ need right clickable buttons.

---

> buttons with a right click handler are not very common and, therefore, not ideal for UX.

They are good UX in my case..? Sorry, I don't think this is applicable for me.

EHPS-Oliver:
I see this had the 0.13 milestone on it, did it get implemented some other way? I could also use RMB buttons

airstrike:
> I see this had the 0.13 milestone on it, did it get implemented some other way? I could also use RMB buttons

You don't need the button to be added to the library to do this. 

It should be quite easy to just copy `button.rs`, change the imports to import from `iced::` instead of `crate::core` etc. and add the right button logic to `fn on_event` (in `0.13`) or `fn update` (in `0.14.0-dev`/`master`).

Issue #2381, opened 2024-04-08
Nested scrollable behaviour


x86y:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

Parent scrollable shouldn't have its focus taken away from children scrollable. 

![CleanShot 2024-04-08 at 23 50 09](https://github.com/iced-rs/iced/assets/68605763/519bdb03-74df-4bf4-9157-2609a78de97b)

Now, even the text_editor will steal the focus from the main scrollable

### What is the expected behavior?

If you start scrolling in any scrollable, the scroll focus shouldn't be stolen from any mouseovered area/widget.

Take a look at the browser behaviour: 
![CleanShot 2024-04-08 at 23 47 49](https://github.com/iced-rs/iced/assets/68605763/67af18bb-13eb-4de6-ad2d-c1fb00a53025)

### Version

crates.io release

### Operating System

macOS

PR #2397, opened 2024-04-21
Don't consume unused scroll events

Koranir:
If the viewport of a scrollable is not changed by an event, it did not consume it. This updates the code to reflect that.

(This means that scrollables can be nested and scroll the parent scrollable if an end is reached, and that horizontal scrollables can be put in vertical ones with no issue)

This is _not_ mutually exclusive with #2381, but that would need some sort of heuristic to determine whether a scroll was active recently.


hecrj (approved):
Awesome! Thanks :partying_face:

PR #2401, opened 2024-04-24
Scroll Transactions

### LINKED ISSUE #2381: Nested scrollable behaviour

Koranir:
Fixes #2381

This PR adds a system to scrollables that keeps scrolling to only one layer at a time, based on [the behaviour in Mozilla's Gecko engine](https://wiki.mozilla.org/Gecko:Mouse_Wheel_Scrolling#Mouse_wheel_transaction), which is more in line with what users expect.

Also adds a nested scrollable in the scrollable example.

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -7,6 +7,7 @@ use crate::core::layout;
 use crate::core::mouse;
 use crate::core::overlay;
 use crate::core::renderer;
+use crate::core::time::{Duration, Instant};
 use crate::core::touch;
 use crate::core::widget;
 use crate::core::widget::operation::{self, Operation};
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks!

I simplified some stuff here and there. Also, the `visible_bounds` example already features nested scrollables, so we can use that for testing.

PR #2402, opened 2024-04-24
Move to home/end when meta key is pressed with left/right

nrjais:


hecrj:
This looks like a duplicate of #2315.

PR #2405, opened 2024-04-24
`Stack` widget

hecrj:
This PR implements a simple `Stack` widget that can be used to display elements on top of each other.

The `Stack` uses the first element as its base layer and every consecutive element is displayed on top of the previous one. The base element defines the intrinsic size of the `Stack`, so the elements on top can easily position themselves using other existing widgets (e.g. `container`).

Like `column` and `row`, it comes with both its function and macro helpers.

The `bezier_tool` example showcases how we can use this new widget to show the "Clear" button at the top right corner of the `Canvas`:

```rust
container(stack![
    self.bezier.view(&self.curves).map(Message::AddCurve),
    container(
        button("Clear")
            .style(button::danger)
            .on_press(Message::Clear)
    )
    .padding(10)
    .width(Length::Fill)
    .align_x(alignment::Horizontal::Right),
])
.padding(20)
.into()
```

This produces (amazing art not included!):

image

### DIFF
--- a/widget/src/stack.rs
+++ b/widget/src/stack.rs
@@ -1,0 +1,327 @@
+//! Distribute content vertically.
+use crate::core::event::{self, Event};
+use crate::core::layout;
+use crate::core::mouse;
+use crate::core::overlay;
+use crate::core::renderer;
+use crate::core::widget::{Operation, Tree};
+use crate::core::{
+    Clipboard, Element, Layout, Length, Rectangle, Shell, Size, Vector, Widget,
+};
+
+/// A container that displays children on top of each other.
+///
+/// The first [`Element`] dictates the intrinsic [`Size`] of a [`Stack`] and
+/// will be displayed as the base layer. Every consecutive [`Element`] will be
+/// renderer on top; on its own layer.
+///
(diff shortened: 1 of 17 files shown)


widget/src/stack.rs
```diff
@@ -0,0 +1,328 @@
...
+            &mut tree.children[0],
+            renderer,
+            &limits,
+        );
+
+        let size = limits.resolve(self.width, self.height, base.size());
+        let limits = layout::Limits::new(Size::ZERO, size);
+
+        let nodes = std::iter::once(base)
+            .chain(self.children[1..].iter().zip(&mut tree.children[1..]).map(
```
tarkah, line 169: Possible index error? We only check is_empty

hecrj: It'll just [return an empty slice](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=073451f81a0f8c31ac4a1d973cb1b801), which is what we want.

tarkah: How did I not know this? I guess the spread operator here protects it

PR #2409, opened 2024-04-27
Retain caches in `iced_wgpu` as long as `Rc` values are alive

hecrj:
This allows reusing a `canvas::Cache` at no cost even if it is not presented every frame.


hecrj:
I've created a new example that showcases how reusing caches in different frames may be useful. It is called `the_matrix`:


https://github.com/iced-rs/iced/assets/518289/421d6c72-96da-4b55-ba94-eb79a16e5fb0

PR #2419, opened 2024-05-02
Support wasm target for `download_progress`

skygrango:
I think it works, but I did make some tricky modifications because
1. reqwest doesn't support `chunk` method for wasm, I have to replace chunk by stream
2. the futures generated by wasm_bindgen NO Send, so I have to pin stream

螢幕截圖_20240502_171755


hecrj (approved):
Thanks!

I unified the `bytes_stream` approach and simplified the example with the new `stream::try_channel` helper.

PR #2420, opened 2024-05-02
Fix: grabbing `image::Viewer` from outside the widget

myuujiku:
#1998 fixed out of bounds scrolling, but I noticed that *grabbing* the image from anywhere was still possible.


hecrj (approved):
Good catch! Thanks.

PR #2421, opened 2024-05-03
Fix `clock` example doesn't get the correct local time under unix system

skygrango:
There is a long-standing problem (https://github.com/time-rs/time/issues/293) that has not yet been solved by time-rs 

Switch to chrono as it seemed to solve the problem (https://github.com/chronotope/chrono/pull/677)


hecrj (approved):
Thanks! 🎉

PR #2434, opened 2024-05-10
Simple Hello World Example

s4Dt0y:
I was just looking through the examples and saw that there was no hello world example. Thought I'd see whether ygs would take one if I made it.

### DIFF
--- a/examples/hello_iced/src/main.rs
+++ b/examples/hello_iced/src/main.rs
@@ -1,0 +1,19 @@
+use iced::Element;
+
+fn main() -> iced::Result {
+    iced::program("Hello - Iced", HelloIced::update, HelloIced::view).run()
+}
+
+#[derive(Default)]
+struct HelloIced {}
+
+#[derive(Debug)]
+enum Message {}
+
+impl HelloIced {
+    fn update(&mut self, _message: Message) {}
+    
+    fn view(&self) -> Element<Message> {
+        "Hello, Iced".into()
+    }
+}
--- a/examples/hello_iced/Cargo.toml
+++ b/examples/hello_iced/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "hello_iced"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+iced.workspace = true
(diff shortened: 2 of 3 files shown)


dzienisz (approved):

examples/hello_iced/src/main.rs
```diff
@@ -0,0 +1,19 @@
...
+
+fn main() -> iced::Result {
+    iced::program("Hello - Iced", HelloIced::update, HelloIced::view).run()
+}
+
+#[derive(Default)]
+struct HelloIced {}
+
+#[derive(Debug)]
+enum Message {}
```
BornIncompetence, line 11: If your goal is a minimal Hello World instead of something others can start off of, you can elide both `HelloIced` and `Message` here and just use the unit type `()` for both. `update` and `view` can then be written as free functions
```rust
fn update(_: &mut (), _: ()) {}

fn view(_: &()) -> Element<()> { "Hello, Iced".into() }
```

tsuza: In my eyes, hello world is generally a starting point, some sort of template. You'd be stripping way too much off a "normal" Iced program that way. Just my two cents on it.

s4Dt0y: Yes, I agree with @tsuza...I want this example to be more of a starting point, so I don't think it would be wise to strip the functions that much.

examples/hello_iced/src/main.rs
```diff
@@ -0,0 +1,19 @@
+use iced::Element;
+
+fn main() -> iced::Result {
+    iced::program("Hello - Iced", HelloIced::update, HelloIced::view).run()
```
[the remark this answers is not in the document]
soucosmo: Simple examples are essential for any lib, it's extremely bad to get super complex examples when you're a beginner, it's cool to see something simple and actually understand how it works, when the example is complicated you need to spend a certain amount of time trying to understand how it works , what each thing does, etc., it's great to have more usable examples, but simple, straight-to-the-point examples are essential, especially for new audiences.

A while ago I sent a PR to keyed_column, still hoping to get it merged haha.

https://github.com/iced-rs/iced/pull/2361

hecrj:
Thanks! But the `counter` is what we consider a "Hello, world!" example since it showcases interactions; which are a key part of a user interface.

The API reference of `run` also [includes a simpler version of the counter](https://docs.iced.rs/iced/fn.run.html).

PR #2436, opened 2024-05-10
Suggestion: scrollable::scroll_by operation

lufte:
This new operation scrolls an scrollable an absolute amount from their current position, in contrast to `scroll_to` that scrolls to an absolute position.

I'm not entirely convinced my implementation is good, I just hacked away until it worked. Suggestions welcome.

My use case for this is that I want to scroll a scrollable with the keyboard, using custom keys, and by custom amounts of pixels. The content bounds are required in case the new AbsoluteOffset needs to be applied on top of a RelativeOffset.


hecrj (approved):
Thanks!

PR #2438, opened 2024-05-11
Notify `scrollable::Viewport` changes

### LINKED ISSUE #2437: `on_scroll` is not fired when scrolling via operations

### What happened?

When scrolling via `scroll_to` or `snap_to`, the callback passed through `on_scroll` is not fired, hiding thus vital information about the current offset of the scrollable that can't be queried through other means.

If I show a percentage indicator of some scrollable in my app and perform a `scroll_to` operation, I don't know what's the new relative offset.

`on_scroll` should fire on every scrolling event.

### Version

crates.io release

### Operating System

Linux

_No response_

Koranir:
Fixes #2437.
Closes #2490.

Due to some inherent limitations, the operation can't access the `on_scroll` function, and so it sets a flag to be consumed on a later event. A better solution might be to give operations access to a Shell, so a message could be triggered manually, but that might not be necessary in this case.

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -12,6 +12,7 @@ use crate::core::touch;
 use crate::core::widget;
 use crate::core::widget::operation::{self, Operation};
 use crate::core::widget::tree::{self, Tree};
+use crate::core::window;
 use crate::core::{
     self, Background, Clipboard, Color, Element, Layout, Length, Padding,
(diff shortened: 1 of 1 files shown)


lufte:
@Koranir I'm gonna thank you on my behalf because this is good enough for me. Really clever.

I've applied to my fork :)

hecrj (approved):
Thanks!

I have changed the approach to simply notify any `Viewport` change on `RedrawRequested`.

This should also supsersede #2490.

PR #2439, opened 2024-05-11
Update scrollable children with new cursor position while scrolling

### LINKED ISSUE #2433: MouseArea: enter and exit does not work inside a scrollable if the cursor is not moved

### What happened?

I tried tracking the currently selected emoji with a mousearea, but the selected emoji didn't update when scrolling with the touchpad.
I copied the mouse area file because libcosmic is not based on the latest version.
https://github.com/leb-kuchen/cosmic-applet-emoji-selector/tree/main

enter and exit updates if the cursor is not moved.

### Version

crates.io release

### Operating System

Linux

_No response_

Koranir:
Fixes  #2433

Applies an artificial MouseMoved event on child widgets when scrolling with a mouse or touchpad.

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -586,6 +586,43 @@ where
                     content_bounds,
                     shell,
                 ) {
+                    // We've changed the viewport but the child widget still thinks that the mouse is in the same relative place.
+                    if let Some(cursor_position) = cursor_over_scrollable {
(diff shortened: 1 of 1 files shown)


hecrj (commented):
This is a hack.

Widgets should not rely on `CursorMoved` events to track cursor positions. The contract does not guarantee that the `mouse::Cursor` won't change without a `CursorMoved`.

#2433 should be fixed in `mouse_area` itself.

PR #2443, opened 2024-05-18
Render border above active progress for progress_bar widget.

B0ney:
Before:
Screenshot 2024-05-18 140250

After:
Screenshot 2024-05-18 140125

### DIFF
--- a/widget/src/progress_bar.rs
+++ b/widget/src/progress_bar.rs
@@ -152,21 +152,13 @@ where
                         width: active_progress_width,
                         ..bounds
                     },
-                    border: border::rounded(style.border.radius),
+                    border: Border {
+                        color: Color::TRANSPARENT,
+                        ..style.border
+                    },
                     ..renderer::Quad::default()
                 },
                 style.bar,
-            );
-        }
-
-        if style.border.color != Color::TRANSPARENT {
-            renderer.fill_quad(
-                renderer::Quad {
-                    bounds: Rectangle { ..bounds },
-                    border: style.border,
-                    ..renderer::Quad::default()
-                },
-                Color::TRANSPARENT,
             );
         }
     }
--- a/widget/src/progress_bar.rs
+++ b/widget/src/progress_bar.rs
@@ -139,10 +139,7 @@ where
         renderer.fill_quad(
             renderer::Quad {
                 bounds: Rectangle { ..bounds },
-                border: Border {
-                    color: Color::TRANSPARENT,
-                    ..style.border
-                },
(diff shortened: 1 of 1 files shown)


widget/src/progress_bar.rs
```diff
@@ -157,6 +157,17 @@ where
...
+        if style.border.color != Color::TRANSPARENT {
+            renderer.fill_quad(
+                renderer::Quad {
+                    bounds: Rectangle { ..bounds },
+                    border: style.border,
+                    ..renderer::Quad::default()
+                },
+                Color::TRANSPARENT,
+            );
+        }
```
hecrj, line 170: We should probably set the border color of the base quad as transparent, or we may run into issues with translucent borders.

hecrj (changes requested):

widget/src/progress_bar.rs
B0ney: Yeah, I originally did that at first, but when resizing the window, I would sometimes see really thin gaps between the base quad and the border:
Screenshot 2024-05-20 131211

Though this is really an issue with `wgpu` since this doesn't happen on `tiny-skia`.

Nonetheless, I found a better solution to all this. I can instead include the border information to the active bar but make the border colour transparent:

Screenshot 2024-05-20 132308

Though annoying gap is still there, but this time it's between the active bar and background. But I'll take it:

Screenshot 2024-05-20 132313

B0ney:
Looks like I discovered a transparency bug in `tiny-skia`:
 
Screenshot 2024-05-20 133400

`wgpu`:
Screenshot 2024-05-20 132308

sscce:
```rust
use iced::widget::container;
use iced::widget::progress_bar::Style;
use iced::widget::progress_bar;
use iced::{Border, Color, Element, Length};

    iced::program("Progress Bar - Iced", Progress::update, Progress::view)
        .theme(|_| iced::Theme::Light)
        .run()
}

#[derive(Default)]
struct Progress;

impl Progress {

    fn view(&self) -> Element<()> {
        container(progress_bar(0.0..=100.0, 50.0).style(|_| Style {
            background: Color::BLACK.into(),
            bar: Color::from_rgb8(0, 255, 0).into(),

            border: Border {
                color: Color::from_rgba8(0, 0, 127, 0.5),
                width: 4.0,
                ..Default::default()
            },
        }))
        .padding(20)
        .center_x(Length::Fill)
        .center_y(Length::Fill)
        .into()
    }
}
```

hecrj (approved):
Thanks!

PR #2446, opened 2024-05-20
Implement `std::fmt::Display` for `iced::Radians`

SolidStateDj:
This PR implements `std::fmt::Display` for the `iced::Radians` struct.

This allows the given angle of a widget (for example, a gradient) to be displayed as text.


core/src/angle.rs
```diff
@@ -237,3 +238,9 @@ impl PartialOrd<f32> for Radians {
         self.0.partial_cmp(other)
     }
 }
+
+impl Display for Radians {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
```
hecrj, line 244: We should probably the unit of measurement here:

```suggestion
        write!(f, "{} rad", self.0)
```

hecrj (changes requested):

core/src/angle.rs
SolidStateDj: The issue I see here is whenever a user does not wish to include the unit of measurement as removing it is more complex than having it gone by default and adding it whenever it's explicitly used/needed.

SolidStateDj:
I just merged the latest commits, but the CI doesn't appear to be running. Could you re-approve it, please and thank you?

PR #2448, opened 2024-05-22
implement From<String> for Element

JL710:
This PR adds a `From<String>` implemenation for Element. 

This works:
```rs
widget::button("")
```
This did not work:
```rs
widget::button(String::new())
```
You needed this:
```rs
widget::button(widget::text(String::new()))
```

This PR makes it a lot less redundant if you want have `format!` generated text in a button.

### DIFF
--- a/core/src/widget/text.rs
+++ b/core/src/widget/text.rs
@@ -329,6 +329,27 @@ where
     }
 }
 
+impl<'a, Theme, Renderer> From<String> for Text<'a, Theme, Renderer>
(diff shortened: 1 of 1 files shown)


JL710:
Awesome macro. I will replace abount 100 occourences in my code with it.

hecrj:
It's unclear when this `From` implementation may be useful; so I am closing this in favor of the new `text!` macro introduced in #2338.

PR #2449, opened 2024-05-25
Add viewport scale to screenshot data

bungoboingo:
Addresses [this](https://github.com/iced-rs/iced/pull/1971#issuecomment-1652724543), which I thought I already merged 😅 

Fixes:
- When using the widget bounds operation, the bounds are returned in logical pixels
- When using the screenshot command, the screenshot is taken in physical pixels
- When cropping a screenshot, we need to know the scale factor of the viewport to properly screenshot an individual widget


hecrj (approved):
Sweet! Thanks :tada:

Issue #2375, opened 2024-04-06
MouseArea does not report mouse wheel events.


Gremious:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

I wanted to use MouseArea to capture scrolling. However, it only captures mouse button events.

### What is the expected behavior?

It should have `on_scroll` which sends gives you a `Scroll:Up`/`Scroll:Down`, or two separate up/down hooks.

Scrolling with a mouse is a mouse-related event so MouseArea should capture that.

### Version

crates.io release

### Operating System

Windows

breynard0:
I've been wanting to get into contributing to Iced, and I think this will make a good first issue. I'll update when I've submitted a PR.

breynard0:
PR submitted, time to wait see.

breynard0:
Merged!

hecrj:
Fixed by #2450.

PR #2450, opened 2024-05-25
Add `on_scroll` handler to `mouse_area` widget

breynard0:
After reading through #2375, I decided to add the ability for MouseAreas to detect scrolling. This is a very simple change, following the style of the rest of the code in the file.


hecrj (approved):
Thanks!

PR #2460, opened 2024-06-12
Add FromIterator for Row and Column

Andrew-Schwartz:
As per [this conversation on the discord](https://discord.com/channels/628993209984614400/1021828532189528094/1250187498185818206). Implement `FromIterator<Element<...>>` for `Row` and `Column`, so that they may be used without interrupting iterator chains. The implementations function equivalently to `with_children`.


hecrj (approved):
Thanks! :bow:

PR #2467, opened 2024-06-18
Add rotation for viewer widget

Gigas002:
This draft PR is a follow-up from #2334 and #2330, and it introduces a `rotation` property for `viewer` widget.
Right now it includes basic stuff and I think it's best behavior is when using together with `ContentFit::None`, however, all the other `ContentFit`'s are not working as good as this one, and that especially noticable when using `Fill` (try zooming in/out and changing window sizes when the image is rotated and you'll see).
Would be glad if @DKolter or anyone else could help with further implementation for this :)


Gigas002:
A post to keep a track on the state of PR, what's working and what's not.
A problem I came up with and can't overcome, is that the rotated image doesn't respect updated layout. This results in incorrect behavior on all `content_fit` modes, except for `none`. This is especially noticable when trying to change windows sizes with rotated image on a `fill` view:


https://github.com/user-attachments/assets/5d41a225-fa86-4f7b-ae1c-ac76a97e8628


I didn't notice other problems, it's the only blocker for this PR to be ready for review

PR #2470, opened 2024-06-20
feat: Add methods for window settings in Application

ryankopf:
This commit adds new methods to the `Application` struct for setting window settings:

- resizable
- decorations
- position
- level.

The most recent PRs removed "window" from "Settings", and so now these window settings can not be set without these functions, making them critical.


hecrj (approved):
Thanks!

PR #2472, opened 2024-06-21
Multiple Windows.

ryankopf:
Would you be interested in supporting multiple - or no - windows in application?

This may alleviate the need for a separate Daemon setup, which duplicates a lot of what "application" already does.

If you think this is worth pursuing, I can create a more complete proposal, but this is a start of what it would look like.

Incomplete:
- One window works and no windows works
- Adding more windows than that doesn't work yet
- Maybe each window can have its own task handler for open/close
- There's could be complex task-assigning based on what window that might need to be thought about

At minimum though, something like this, or even just converting window: window::Settings to an Option, would simplify?

### DIFF
--- a/examples/ferris/src/main.rs
+++ b/examples/ferris/src/main.rs
@@ -91,7 +91,7 @@ impl Image {
         }
     }
 
-    fn view(&self) -> Element<Message> {
+    fn view(&self, _window: Option<window::Id>) -> Element<Message> {
         let i_am_ferris = column![
             "Hello!",
             Element::from(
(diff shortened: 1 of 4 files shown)


max397574:
Perhaps I'm completely missunderstanding what you're doing here
But isn't this already implemented with https://github.com/iced-rs/iced/pull/1964?

ryankopf:
> Perhaps I'm completely missunderstanding what you're doing here But isn't this already implemented with #1964?

Maybe, let me reevaluate

hecrj:
> This may alleviate the need for a separate Daemon setup, which duplicates a lot of what "application" already does.

No. Notice how you broke all the examples. `Daemon` precisely exists to make the complexity of multi-window applications opt-in.

PR #2473, opened 2024-06-21
Add `stroke_rectangle` to `canvas::Frame`

vladh:
This method should be able to leverage performance improvements in lyon's `tessellate_rectangle` over `tessellate_path`.

Did some quick profiling, this seems to result in a 25% reduction in frame time.


hecrj (approved):
Sweet! Thanks :tada:

PR #2475, opened 2024-06-21
Implement Copy on Fill and Stroke

vladh:
In general, would it not be better to have functions such as `fill_rectangle` take `&Fill` to prevent unnecessary copies, especially when drawing the same item over and over?


hecrj (approved):
Premature optimization. Compilers are smart and dereferencing could potentially be slower.

In any case, the cost of a copy is negligible when compared to the actual tessellation. So, why bother with these questions?

PR #2476, opened 2024-06-22
Add `override_redirect` for X11 windows

alliby:
This commit add the `override_redirect` boolean field to the `PlatformSpecific` struct . This is a X11-specific flag allow bypassing window manager mapping for precise positioning of windows.
It's particularly useful for creating Application launchers (like [rofi](https://github.com/davatorium/rofi/blob/75ae2a41f01b77c2123ec773d193fa686a8ef6d9/source/view.c#L1146)) and Custom overlay UIs (e.g., notification apps, screen capture tools)


hecrj (approved):
Thanks!

PR #2486, opened 2024-06-30
Add `Color::parse`

vladh:
This function has certain advantages over creating hex colours with `color!`:
* It allows parsing of hex color strings
* It allows opacity to be specified in hex, such as `#ffffff00`
* It allows the shorthand `#fff` notation
* It has better error handling; you can, for example, do `color!(0xfafafa9)` and there won't be an error, despite the hex string being invalid

If this PR gets merged, I'll send another patch to have `color!` use `from_hex`.

### DIFF
--- a/core/src/color.rs
+++ b/core/src/color.rs
@@ -88,10 +88,35 @@ impl Color {
         }
     }
 
+    /// Creates a [`Color`] from its linear RGBA components.
+    pub fn from_linear_rgba(r: f32, g: f32, b: f32, a: f32) -> Self {
+        // As described in:
+        // https://en.wikipedia.org/wiki/SRGB
+        fn gamma_component(u: f32) -> f32 {
+            if u < 0.0031308 {
+                12.92 * u
+            } else {
+                1.055 * u.powf(1.0 / 2.4) - 0.055
+            }
+        }
+
+        Self {
+            r: gamma_component(r),
+            g: gamma_component(g),
+            b: gamma_component(b),
+            a,
+        }
+    }
+
     /// Parses a [`Color`] from a hex string.
     ///
-    /// Supported formats are #rrggbb, #rrggbbaa, #rgb, and #rgba.
+    /// Supported formats are `#rrggbb`, `#rrggbbaa`, `#rgb`, and `#rgba`.
     /// The starting "#" is optional. Both uppercase and lowercase are supported.
+    ///
+    /// If you have a static color string, using the [`color!`] macro should be preferred
+    /// since it leverages hexadecimal literal notation and arithmetic directly.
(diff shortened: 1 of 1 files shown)


VAWVAW:
You missed the format `#abcf` in your documentation.

VAWVAW:
Is there a reason you don't use the more self-explaining `#rgba`, `#rrggbb`, ... for documentation?

vladh:
@VAWVAW Good points, thank you! Fixed.

core/src/color.rs
```diff
@@ -88,6 +96,41 @@ impl Color {
         }
     }
 
+    /// Creates a [`Color`] from a hex string. Supported formats are #rrggbb, #rrggbbaa, #rgb,
+    /// #rgba. The “#” is optional. Both uppercase and lowercase are supported.
+    pub fn from_hex(s: &str) -> Result<Color, ColorError> {
+        let mut hex = s.replace('#', "").to_lowercase();
```
Mingun, line 102: Using `strip_prefix` is more safe choice, to not allow something like `###01#2##856`.

Also, the overall method seems overcomplicated, many unnecessary allocations. Just check length, then loop: take 1 or 2 characters (depending on the length check), use `from_str_radix()`

vladh: You're totally right, thank you! Fixed.

n1ght-hunter:
@vladh what the advantage of this over just using the color macro 
https://github.com/iced-rs/iced/blob/950bfc07d4b71016bf3e9d53709395e185420cec/core/src/color.rs#L199

vladh:
@n1ght-hunter Oops, I didn't know about `color!`, thank you. Maybe I should send a patch to better document `color!`.

That being said, I still think my PR has certain advantages:
* It allows parsing of hex color strings
* It allows opacity to be specified in hex, such as `#ffffff00`
* It allows the shorthand `#fff` notation

If this PR gets merged, I'll send another patch to have `color!` use `from_hex`.

nednoodlehead:
i would love this. Sort of annoying when I open a hex color editor, pick a color and I can't easily use it to color buttons and whatnot.

vladh:
@hecrj Let me know if you want any changes on this (or just don’t want it which is fine).

hecrj (approved):
Thanks!

I renamed the method to `parse`, which I believe conveys clearly that there is a runtime cost involved with this helper.

The `color!` macro should be preferred since it leverages hexadecimal literal notation and arithmetic directly. The compiler is able to optimize most of these operations away in most cases, so no parsing and runtime cost necessary.

I have also extended the macro to support short hand notation and added some debug assertions for good measure.

PR #2489, opened 2024-07-02
scrollable: add static_content

vladh:
Sometimes it's useful to use `Scrollable` to perform scroll calculations without having the content inside actually be transformed. For example, you might have a custom widget that uses the scroll offsets to draw its own graphics with non-standard scroll behaviour. I'm working on a spreadsheet program that has idiosyncratic scrolling behaviour, and I previously used [QAbstractScrollArea](https://doc.qt.io/qt-6/qabstractscrollarea.html) in Qt for this purpose. This PR enables similar behaviour in iced's `Scrollable` with minimal changes.

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -38,6 +38,7 @@ pub struct Scrollable<
     content: Element<'a, Message, Theme, Renderer>,
     on_scroll: Option<Box<dyn Fn(Viewport) -> Message + 'a>>,
     class: Theme::Class<'a>,
+    static_content: bool,
 }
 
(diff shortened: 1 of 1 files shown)


hecrj (commented):
This needs discussion.

Specifically, how does this enable use cases that are not currently possible right now and why is it the best possible solution for them?

hecrj:
I don't see why the translation needs to be undone. "Normal" widgets draw things manually with `fill_quad` too.

If your widget has certain bounds, then it can render things where they should be in its virtual boundaries and everything should work as expected.

vladh:
Thanks — I'll look into this and report back.

PR #2491, opened 2024-07-02
rounded rectangle graphics

gintsgints:
This is my attempt to add functionality for additional canvas shape, rounded rectangle.

### DIFF
--- a/graphics/src/geometry/path.rs
+++ b/graphics/src/geometry/path.rs
@@ -9,7 +9,8 @@ pub use builder::Builder;
 
 pub use lyon_path;
 
-use iced_core::{Point, Size};
+use crate::core::border;
+use crate::core::{Point, Size};
 
 /// An immutable set of points that may or may not be connected.
 ///
@@ -45,6 +46,16 @@ impl Path {
     /// corner coordinate and its `Size`.
     pub fn rectangle(top_left: Point, size: Size) -> Self {
         Self::new(|p| p.rectangle(top_left, size))
+    }
+
+    /// Creates a new [`Path`] representing a rounded rectangle given its top-left
+    /// corner coordinate, its [`Size`] and [`border::Radius`].
+    pub fn rounded_rectangle(
+        top_left: Point,
+        size: Size,
+        radius: border::Radius,
+    ) -> Self {
+        Self::new(|p| p.rounded_rectangle(top_left, size, radius))
     }
 
     /// Creates a new [`Path`] representing a circle given its center
--- a/graphics/src/geometry/path/builder.rs
+++ b/graphics/src/geometry/path/builder.rs
@@ -1,6 +1,7 @@
 use crate::geometry::path::{arc, Arc, Path};
 
-use iced_core::{Point, Radians, Size};
+use crate::core::border;
+use crate::core::{Point, Radians, Size};
 
 use lyon_path::builder::{self, SvgPathBuilder};
 use lyon_path::geom;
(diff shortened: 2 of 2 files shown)


vladh:
I don't think this approach is flexible enough. For example, I would expect drawing a square with a very high border radius to result in a circle. This is what happens in [`iced::Border::rounded`](https://docs.iced.rs/iced/struct.Border.html#method.rounded) and also in CSS. However, in your implementation, a buggy shape is drawn.

gintsgints:
Hmm, so what should flexible solution do instead? 
I can make radius relative to the width of the rectangle or disallow too big radius.

vladh:
> Hmm, so what should flexible solution do instead?

It looks like, at the very least, you want something like:

```rust
radius = min(radius, min(height / 2, width / 2))
``` 

See:
* https://www.w3.org/TR/css-backgrounds-3/#the-border-radius
* https://www.w3.org/Style/CSS/Tracker/issues/29

It would also be nice to be able to specify horizontal and vertical radius, and to specify individual radii for each of the 4 corners, by using or extending the existing [`iced::Border::Radius`](https://docs.iced.rs/iced/border/struct.Radius.html).

gintsgints:
One commit implement just min check, other use Radius as argument, so it is possible to specify different radius for each corner.

core/src/border.rs
```diff
@@ -54,7 +54,7 @@ impl Border {
 /// The border radii for the corners of a graphics primitive in the order:
 /// top-left, top-right, bottom-right, bottom-left.
 #[derive(Debug, Clone, Copy, PartialEq, Default)]
-pub struct Radius([f32; 4]);
+pub struct Radius(pub [f32; 4]);
```
vladh, line 57: I don't think this needs to be `pub` — the `from()` methods can be used.

vladh (changes requested):

core/src/border.rs
gintsgints: Sorry, you are right. Fixed.

graphics/src/geometry/path.rs
```diff
@@ -47,6 +47,12 @@ impl Path {
         Self::new(|p| p.rectangle(top_left, size))
     }
 
+    /// Creates a new [`Path`] representing a rounded rectangle given its top-left
+    /// corner coordinate, its `Size` and radius.
```
vladh, line 51: Maybe:

```
its [`Size`] and border [`Radius`]
```

graphics/src/geometry/path/builder.rs
```diff
@@ -160,6 +161,51 @@ impl Builder {
         self.close();
     }
 
+    /// Adds a rounded rectangle to the [`Path`] given its top-left
+    /// corner coordinate its `Size` and curve radius.
```
vladh, line 165: Maybe:

```
its [`Size`] and border [`Radius`]
```

vladh (changes requested):

vladh:
This looks okay to me right now, but I'm not a maintainer so you'll have to wait for one to look at this.

graphics/src/geometry/path.rs
gintsgints: Fixed

graphics/src/geometry/path/builder.rs
gintsgints: fixed

hecrj (approved):
Thanks!

I have removed the example since we already have too many.

PR #2496, opened 2024-07-10
Abortable `Task`

hecrj:
This PR introduces the `Task::abortable` method, which returns a `task::Handle` that can be used to easily cancel a `Task` that has already been handled to the runtime.

This is useful for long-running tasks that are not well-suited to become subscriptions (i.e. may eventually finish on their own) and that may need to be canceled by the user, like downloading a file (yes, this means our `download_progress` example is outdated).

### DIFF
--- a/runtime/src/task.rs
+++ b/runtime/src/task.rs
@@ -55,24 +55,6 @@ impl<T> Task<T> {
         Self::stream(stream.map(f))
     }
 
-    /// Creates a new [`Task`] that runs the given [`Future`] and produces
-    /// its output.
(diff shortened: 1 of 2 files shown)


JL710:
What would the key difference between Tasks and Subscriptions be?

hecrj:
@JL710 The main difference is the guarantees they provide. From an application's perspective:

- A `Task` is some workload that _may_ eventually finish on its own, independently of the application state.

- A `Subscription` is a request to listen to "something" that must always be there for the whole lifetime of the application (like a webserver connection, the iced runtime, or just time itself). A `Subscription` cannot end on its own; only the application can eventually unsubscribe from it.

PR #2499, opened 2024-07-11
Add `align_x` and `align_y` helpers to `Scrollable`

hecrj:
These helpers can come in handy when you only want to change the alignment of a `Scrollable` without changing its direction.

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -109,6 +109,32 @@ where
         self
     }
 
+    /// Inverts the alignment of the horizontal direction of the [`Scrollable`], if applicable.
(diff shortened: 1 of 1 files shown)


hecrj:
@alex-ds13 That wouldn't work because those enums have 3 variants, while `scrollable::Alignment` has only 2.

alex-ds13:
> @alex-ds13 That wouldn't work because those enums have 3 variants, while `scrollable::Alignment` has only 2.

Oh I didn't notice this `Alignment` was specific for the scrollable... I thought it was the global `Alignment`.

Couldn't you use the version with 3 variants anyway and simply ignore the variant that doesn't make sense? It's just that this many different `Alignment`s can become confusing...

PR #2502, opened 2024-07-12
Add `on_press_with` method for `Button`

hecrj:
This allows using a closure to produce the message only when the `Button` is actually pressed. Useful when generating the message may be expensive.


nrjais:
I think this can help with relaxing the clone requirement on the Message, If the current `on_press` take a message that is cloneable and uses closure to clone it on demand.

hecrj:
I don't see the point in adding an unnecessary heap allocation. Messages should be cloneable and other widgets demand that bound as well, like `TextInput`.

PR #2504, opened 2024-07-12
Improved `view` ergonomics

hecrj:
This PR improves some of the ergonomics of `view` code:

- The `Fill`, `FillPortion`, and `Shrink` variants of `Length` are re-exported by the root crate. Effectively, this means we can now import these directly:

```rust
use iced::widget::container;
use iced::Fill;

container("Yay!").width(Fill)
```

- The `Center`, `Left`, `Right`, `Top`, and `Bottom` variants of the different `alignment` enums are also re-exported. Thus, we can now write:

```rust
use iced::widget::text;
use iced::{Center, Fill};

text("I am centered!").width(Fill).align_x(Center)
```

- `align_items` in `Column` and `Row` have been renamed to `align_x` and `align_y` respectively.
- `horizontal_alignment` and `vertical_alignment` in `Text` have been renamed to `align_x` and `align_y`.
- The `padding` module exposes free functions to build `Padding` values and the type has methods of the same name that work as setters. Thus, we can write:

```rust
use iced::widget::container;
use iced::padding;

container("I have a padding of 10 all around, except my right side!")
    .padding(padding::all(10).right(0))
```
- The `border` module has received a similar treatment to `padding`:

```rust
use iced::widget::container;
use iced::border;

container("My top right corner isn't rounded!")
    .style(|theme: &Theme| container::Style {
        border: border::rounded(border::radius(10).top_right(0)).color(Color::BLACK).width(2),
        ..container::Style::default()
    })
```

Closes #2260.


hecrj:
I don't think those are that common, since they are only used for images.

PR #2508, opened 2024-07-18
`rich_text` and `markdown` widgets

hecrj:
This PR implements support for __*rich* text__ on top of `cosmic-text` and also introduces `markdown` helpers in the `widget` module.

# Rich Text
Rich text can be created on-the-fly with the `span` helper and the `rich_text!` macro (analogous to `column!` and `row!`):

```rust
use iced::widget::{rich_text, span};
use iced::Font;

rich_text!["Hello ", span("iced").font(Font::MONOSPACE).size(20)]
```

Or you can store the spans in your application state and borrow them directly with the `rich_text` function:

```rust
rich_text(&self.my_cached_spans)
```

This can be useful if you are generating rich text as a result of some expensive operation, like parsing.

# Markdown
Markdown support can be enabled with the new `markdown` feature.

Then, you can parse markdown with the `parse` function:

```rust
use iced::widget::markdown;

self.items = markdown::parse("# Some markdown!", Theme::TokyoNight.palette()).collect();
```

For now, colors need to be attached to spans during construction; therefore, a `theme::Palette` is needed.

Once parsed, you can render it:

```rust
use iced::widget::markdown;

markdown(&self.items)
```

Further configuration options will be added eventually.

-------------

To showcase all of this, I have created a simple Markdown editor example:

image

Code blocks will be automatically highlighted if the `highlighter` feature is enabled.

### DIFF
--- a/core/src/text.rs
+++ b/core/src/text.rs
@@ -10,6 +10,7 @@ pub use paragraph::Paragraph;
 use crate::alignment;
 use crate::{Color, Pixels, Point, Rectangle, Size};
 
+use std::borrow::Cow;
 use std::hash::{Hash, Hasher};
 
 /// A paragraph.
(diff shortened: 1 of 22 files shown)


core/src/text.rs
```diff
@@ -221,3 +221,162 @@ pub trait Renderer: crate::Renderer {
...
+    /// The [`Fragment`] of text.
+    pub text: Fragment<'a>,
+    /// The size of the [`Span`] in [`Pixels`].
+    pub size: Option<Pixels>,
+    /// The [`LineHeight`] of the [`Span`].
+    pub line_height: Option<LineHeight>,
+    /// The font of the [`Span`].
+    pub font: Option<Font>,
+    /// The [`Color`] of the [`Span`].
+    pub color: Option<Color>,
```
Mingun, line 237: I think, more detail documentation is needed. What `None` values mean?

hecrj: What do you think they could mean?

Mingun: Missing thing (and error during render). Default value for thing. Change in behaviour when render thing. Anything.

hecrj: That's ridiculous! Thanks for the good laugh.

Mingun: I think a user looking for a solution to his problem will not see anything funny in the fact that he needs to study the insides of the library used instead of realizing his goals. Meaning of `None` is unclear and should be documented explicitly. Can you say just looking at this code, what will render if `size` is `None`? Or what is the difference between `line_height: None` and  `line_height: Some(0f32.into())`? If it is hard for you to document this behaviour when you are author of this code why do you think that arbitrary user will understand that details?

hecrj: Nah, I simply prefer to write documentation for users that can understand that types do not exist in a vaccuum. Types exist in a specific API of a specific library with a certain philosophy.

If I need to explain every field in a vaccuum for you, then you have a problem. No amount of documentation can help you.

hecrj: Furthermore, we are discussing an internal API in the `iced_core` crate. End users will actually use the `span` widget helper—something you would know if you actually read the PR description.

Mingun: Even if it is an internal API, this is not a reason to leave it undocumented / poorly documented. I see only one reason to do this - not to allow other developers to understand how everything works inside and keep a warm place for themselves, however, I don't understand why do this in an Open-Source project.

The decision is yours, but I will never understand the reason why developers take with hostility advice to make the documentation more understandable _for other developers too_.

> since the whole purpose of a span is precisely to override things. 

This is written in the documentation anywhere?

> If you are missing that part of the picture, then you are completely lost to begin with.

Of course! That is what I trying to tell you! But instead of writing documentation explaining this picture, you find it funny. It makes me sad.

Just imagine, that you need to make your first PR to the project which has such incomplete documentation as here. For example, you need to fix some bug and traces lead you to this code. Would you be funny trying to guess what that `Option`s means? Or you prefer to read description and see the whole picture immediately?

I known what I'm talking about. I contributed to many projects, part of them written on languages that I don't known and do not plan to learn. That does not mean that I cannot fix bugs that strikes me, but this everything takes much more time to understand all this tricky relationships between undocumented or poorly documented datatypes or functions, especially those that need to modify for fix.

I'm not forcing you to write tons of documentation, but the most basic things can be described somewhere? If all these `Option`s do the same thing, then this can be written in the type doc if you don't want to repeat yourself.

I hope, you understand good intentions in my claims.

hecrj: The documentation is fine. I know because I am the author of this project. I get to define what good enough is. I know better than you what the purpose of the documentation is.

If you don't understand it, you are not the target audience. You are not the center of the world. This whole idea that every API needs to be documented and understandable by every developer is total bullshit.

If a developer can't quickly figure out what these fields do given all the context of the project and the API then I simply don't want them near any of the core libraries. I prefer to deal with capable people there.

So, excuse me if I feel irritated by your "I know better than you about how to run _your_ project and what open source ought to be" attitude. You can shove it. I should not have to defend myself for not writing documentation that suits the taste of a random stranger.

[Open Source is Not About You](https://gist.github.com/richhickey/1563cddea1002958f96e7ba9519972d9)

PR #2417, opened 2024-05-01
Update `wgpu` to `0.20`

hecrj:
This PR updates `wgpu` to `0.20`. Depends on #2416.

Unfortunately, this `wgpu` update doesn't seem to work very well on my end (Arch, X11, RTX 4090, NVIDIA drivers). `0.19` is smooth as butter for me, while in `0.20` the examples feel very sluggish and I constantly get a validation error:

```
2024-05-01T14:25:29.145657Z ERROR wgpu_hal::vulkan::instance: VALIDATION [VUID-vkAcquireNextImageKHR-semaphore-01779 (0x5717e75b)]
        Validation Error: [ VUID-vkAcquireNextImageKHR-semaphore-01779 ] Object 0: handle = 0x4fac1c0000000032, type = VK_OBJECT_TYPE_SEMAPHORE; | MessageID = 0x5717e75b | vkAcquireNextImageKHR():  Semaphore must not have any pending operations. The Vulkan spec states: If semaphore is not VK_NULL_HANDLE it must not have any uncompleted signal or wait operations pending (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkAcquireNextImageKHR-semaphore-01779)
2024-05-01T14:25:29.145682Z ERROR wgpu_hal::vulkan::instance:   objects: (type: SEMAPHORE, hndl: 0x4fac1c0000000032, name: ?)
```

I imagine there is something wrong with the release, although maybe I missed something. Let me know if it works for you!

PR #2510, opened 2024-07-19
Update `wgpu` to `22.0`

hecrj:
Closes #2417.
Closes #2507.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,8 @@ advanced = ["iced_core/advanced", "iced_widget/advanced"]
 fira-sans = ["iced_renderer/fira-sans"]
 # Enables auto-detecting light/dark mode for the built-in theme
 auto-detect-theme = ["iced_core/auto-detect-theme"]
+# Enables strict assertions for debugging purposes at the expense of performance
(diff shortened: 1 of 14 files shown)


hecrj:
This is waiting on a resolution of https://github.com/gfx-rs/wgpu/issues/5756.

kmoon2437:
I'm the person who opened #2507 and going to use this framework with `vello` crate. However, it requires version `0.3.69` of `web-sys` and makes conflict with `iced`. Can you solve it at first, please?

hecrj:
I have finally been able to [trace down the root cause of the `wgpu` "issue" blocking this PR](https://github.com/gfx-rs/wgpu/issues/5756#issuecomment-2362316576).

As I explain in my comment, it turns out it's not an issue at all and we can simply control the validation layers with a new feature flag: `strict-assertions`.

I see no reason for users of the library to run validation layers unless they are having issues, therefore the feature flag is disabled by default.

In my testing, disabling validation actually makes `debug` builds quite faster when using the Vulkan backend. So, the overhead has turned into an unexpected speedup!

I think we can finally merge this.

PR #2513, opened 2024-07-21
feat: add width setter for Text-Editor widget

ayeniswe:
Credits on find: nednoodlehead
https://discourse.iced.rs/t/no-width-function-for-widget-text-editor/488


hecrj (approved):
Thanks!

PR #2514, opened 2024-07-23
Fix priority of `Binding::Delete` in `text_editor`

### LINKED ISSUE #2509: Editor widget does not handle '.'/'del' key on numeric keypad correctly.

### What happened?

If you have an Editor widget, then the "decimal point" on the numeric keyboard acts as the "del" key whether numlock is on or off.

You can see this in the Editor example from https://github.com/iced-rs/iced/tree/master/examples/editor.

Run:  cargo run --package=editor and type anywhere using the numeric keypad to see this behaviour. This also occurs in my application.

The "decimal point" key on the numeric keypad should enter a "." when numlock is on.

### Version

master

### Operating System

Linux

_No response_

shartrec:
Fixes #2509.

### DIFF
--- a/widget/src/text_editor.rs
+++ b/widget/src/text_editor.rs
@@ -981,7 +981,9 @@ impl<Message> Binding<Message> {
             keyboard::Key::Named(key::Named::Backspace) => {
                 Some(Self::Backspace)
             }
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks!

PR #2516, opened 2024-07-23
Add background styling to span / rich text

tarkah:
Adds a new `text::Background` type which can be optionally specified on `Span`.

`rich_text` will use this to render background quads for spans where this is set.

A new `Paragraph::span_bounds` method is added to get the bounds of each span to enabled this.

https://github.com/user-attachments/assets/3d81d1bc-8d6d-4f66-8a4d-2440f5efa95a

I'm using this in Halloy:

image

### DIFF
--- a/core/src/text.rs
+++ b/core/src/text.rs
@@ -8,7 +8,7 @@ pub use highlighter::Highlighter;
 pub use paragraph::Paragraph;
 
 use crate::alignment;
-use crate::{Border, Color, Pixels, Point, Rectangle, Size};
+use crate::{Background, Border, Color, Pixels, Point, Rectangle, Size};
 
 use std::borrow::Cow;
 use std::hash::{Hash, Hasher};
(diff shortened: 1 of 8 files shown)


graphics/src/text/paragraph.rs
```diff
@@ -251,6 +253,59 @@ impl core::text::Paragraph for Paragraph {
         Some(glyph.metadata)
     }
 
+    fn span_bounds(&self) -> HashMap<usize, Vec<Rectangle>> {
+        let internal = self.internal();
+
+        let mut all_span_bounds = HashMap::new();
```
tarkah, line 259: I originally tried `Vec<Vec<Rectangle>>` but it's possible for a span / metadata to be missing from the layout run (spaces omitted due to wrapping) which messes up the length / order of the array.

core/src/text.rs
```diff
@@ -406,3 +425,21 @@ into_fragment!(isize);
 
 into_fragment!(f32);
 into_fragment!(f64);
+
+/// The background style of text
+#[derive(Debug, Clone, Copy)]
+pub struct Background {
```
hecrj, line 431: Would `Padding` here make sense?

tarkah: Hmm, unless we can have that horizontal padding considered by cosmic text, there'll be overlap with neighboring glyphs. There's not always whitespace on either side of the span with background.

graphics/src/text/paragraph.rs
```diff
@@ -251,6 +251,57 @@ impl core::text::Paragraph for Paragraph {
...
+            .flat_map(|run| {
+                let line_top = run.line_top;
+                let line_height = run.line_height;
+
+                run.glyphs
+                    .iter()
+                    .map(move |glyph| (line_top, line_height, glyph))
+            })
+            .skip_while(|(_, _, glyph)| glyph.metadata != index)
+            .take_while(|(_, _, glyph)| glyph.metadata == index)
```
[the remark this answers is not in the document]
Mingun: Not always. If you have `[_, _, _, a, a, a, _, _, a]` the original code returns `[a, a, a]`, but your returns `[a, a, a, a]`

hecrj (approved):
I simplified some stuff here and there, but looks great! :partying_face:

PR #2517, opened 2024-07-24
Fix division by zero in `multitouch` example

JustSoup312:
This PR fixes a runtime error while running `examples/multitouch` due to an instance of remainder division dividing by zero.


hecrj (approved):
Good catch! Thanks.

PR #2520, opened 2024-07-25
Refactor Style methods

WailAbou:
#### Changes:
1. **Bundled Style presets in `Style` Struct**:
   - All style presets are now organized within the `Style` struct.
2. **Introduced `standard` Method**:
   - Added the `standard` method to all styling presets for a consistent standard style.
   - The `default` style now serves as a blank, usable style.
3. **Added Background Helper for Scrollable**:
   - New helper added to simplify background styling for scrollable components.

#### Motivation:
- **Logical Organization**: Grouping styles in the `Style` struct centralizes and simplifies access.
- **Consistent Styling**: New users would expect `default` to match the standard styling (`primary`). The new standard method ensures that users have a clear and predictable way to access standard styling. This change ensures that when someone wants to change the background color of a standard button for example, they use standard `.style(|theme, status| button::Style::standard(theme, status).background(color!(255, 0, 0)))` and get the expected result.
- **Enhanced Readability**: The background helper for scrollable elements reduces repetitive code and improves code clarity.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "editor.inlayHints.enabled": "offUnlessPressed"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "editor.inlayHints.enabled": "offUnlessPressed"
-}
--- a/widget/src/button.rs
+++ b/widget/src/button.rs
@@ -462,8 +462,8 @@ pub struct Style {
 }
 
 impl Style {
-    /// Updates the [`Style`] with the given [`Background`].
-    pub fn with_background(self, background: impl Into<Background>) -> Self {
+    /// Updates the [`Background`] of the [`Style`].
+    pub fn background(self, background: impl Into<Background>) -> Self {
         Self {
             background: Some(background.into()),
             ..self
@@ -501,7 +501,7 @@ impl Catalog for Theme {
     type Class<'a> = StyleFn<'a, Self>;
 
     fn default<'a>() -> Self::Class<'a> {
-        Box::new(primary)
+        Box::new(Style::standard)
     }
 
     fn style(&self, class: &Self::Class<'_>, status: Status) -> Style {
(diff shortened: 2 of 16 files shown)


alex-ds13:
I don't get what is the point of these changes? You are basically renaming `default` to `standard` and put it behind another struct...

What is the difference between your implementation:
```rust
button("Press Me!")
.style(|theme, status| button::Style::standard(theme, status).background(color!(255, 0, 0)))
```
And the current one:
```rust
button("Press Me!")
.style(|theme, status| button::default(theme, status).background(color!(255, 0, 0)))
```

The current one seems quite straight forward and simpler. Also if it is because of some widgets not having `default` like the container, I think that `transparent` is more explicit than `default` or `standard` since the default container is a transparent container.

I agree that the scrollable helper is nice but you can make a PR with just that helper... But this is just my opinion. For me the current solution seems better than this proposed one...

WailAbou:
Using `default` applies the following settings, which do not match the standard styling:

```rust
fn default() -> Self {
    Self {
        background: None,
        text_color: Color::BLACK,
        border: Border::default(),
        shadow: Shadow::default(),
    }
}
```

The new `standard` method ensures that users have a clear and predictable way to access standard styling. This means when someone wants to change the background color of a standard button, they can use `standard`:

```rust
.style(|theme, status| button::Style::standard(theme, status).background(color!(255, 0, 0)))
```

and get the expected result without unintentionally altering other style attributes.

For example, using the following code:

```rust
button::Style::default().background(color!(255, 0, 0))
```

will produce this output:

image

As you can see in the image at the top button, simply changing the background color also changes the border color, border radius, and text color. 

By using the `standard` method:

```rust
button::Style::standard(theme, status).background(color!(255, 0, 0))
```

you ensure that only the background color is altered, and the rest of the styling remains consistent with the standard button style.

alex-ds13:
> Using `default` applies the following settings, which do not match the standard styling:

I think you are confusing something... When you use `button("Press Me!")` without any style defined, you are using the `default` style, but you are not using the `button::Style` `Default` implementation, you are using the `default` function, [this one](https://github.com/iced-rs/iced/blob/master/widget/src/button.rs#L503), which in turn is using the primary.

So you can do the same with:
```rust
.style(|theme, status| button::primary(theme, status).background(color!(255, 0, 0)))
```

alex-ds13:
Yes it is true that some widgets have a `default` style function while others don't, like the button which is `primary`, or the container which is `transparent`. Maybe there could be a `default` style function for every widget to make it consistent. But I don't see the need to put it behind the `Style` struct... But again, this is just my opinion and I'm no maintainer, so maybe wait for their review...

WailAbou:
Yes, a `default` implementation for all elements would be ideal. Essentially, they would replace the `standard` method I added with `default`, which is also a good approach as long as it's consistent.

The point of the whole PR is to provide developers with a consistent way to access the `standard`styling. Currently, developers need to know that `primary` and `transparent` are the standard styles for buttons and containers, respectively, or they have to dig into the source code to find this information.

Placing these methods within the Style struct makes sense because they are purely related to styling, not functionality. This organization also enhances autocomplete functionality, allowing developers to type `button::Style::` and immediately see all available styling options.

alex-ds13:
> Yes, a `default` implementation for all elements would be ideal. Essentially, they would replace the `standard` method I added with `default`, which is also a good approach as long as it's consistent and doesn't introduce other side effects or confusion.

So from what I've gathered (bear in mind I might be missing some widget), but the default style functions for widgets that have `Style` are:
- button: primary
- checkbox: primary
- container: transparent
- combo_box: (uses `menu` and `text_input` defaults which in turn have the `default` function)
- menu: default
- pane_grid: default
- pick_list: default
- progress_bar: primary
- qr_code: default
- radio: default
- rule: default
- scrollable: default
- slider: default
- svg: (has no function, uses `svg::Style::default()`, which is the `Default` implementation)
- text: default
- text_editor: default
- text_input: default
- toggler: default
- vertical_slider: (uses `slider::default` but publicly re-exports it so you can use `vertical_slider::default`)

So almost every widget already has the `default` style function. So if anything one could add this function to the widgets that don't have it, which are `button`, `checkbox`, `progress_bar`, `svg` and `container`. Although I still think that `transparent` makes more sense for the container and if you think about it the `primary` also makes sense for the widgets that use it since those are the widgets that will portray some kind of primary, secondary, success or danger status to the user.

> Placing these methods within the `Style` struct makes sense because they are purely related to styling, not functionality. This organization also enhances autocomplete functionality, allowing developers to type `button::Style::` and immediately see all available styling options.

I don't think this is that much of a problem, since when you are using any code editor with `rust_analyzer` if you are inside the `.style()` function and you write `button::` it will only show the functions that can be used as style functions...

hecrj (commented):
Discussion is necessary before code is written!

PR #2523, opened 2024-07-26
Expose physical keys

Exidex:
Expose winit physical keys, aka native key codes. Useful for shortcuts that a based on actual position of keys. 

The data model basically just a copy from winit with some changes:
- `unreachable!()` in conversion function because of non_exhaustive, there is non_exhaustive_omitted_patterns lint which could help, but it is nightly-only
- removed serde feature
- added `#![allow(missing_docs)]` on top of file
- run cargo fmt with several small styling changes to make ci/cd happy 

~~Alternative would be to re-export types from winit. Let me know which one do you prefer.~~ winit is abstracted away

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ pkg/
 Cargo.lock
 dist/
 traces/
+
+.idea/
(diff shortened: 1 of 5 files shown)


hecrj (approved):
Thanks!

Made some changes in #2576, since you didn't give me permissions to change your branch.

PR #2524, opened 2024-07-26
Expose macOS ActivationPolicy and activate_ignoring_other_apps

Exidex:
Needed for launcher-like applications that need to stay always on top of all windows


hecrj (commented):
I don't think these fields make sense in the `window::Settings` struct, since they are global event loop settings rather than specific to a window.

PR #2526, opened 2024-07-28
Underline support for `rich_text`

hecrj:
This PR introduces a new `underline` method to `text::Span` that can be used to underline text:

image

It is far from perfect, but it may be good enough in some circumstances!


tarkah:
Let's add strikethrough as well with this 😎

hecrj:
@tarkah New PR coming.

PR #2527, opened 2024-07-28
Implement `Widget` for `()`

B0ney:
Useful for creating composite widgets.

For example, I'm creating a grid of coloured buttons (like a colour palette in a paint program); creating a button looks like this:
```rust
/* snippet */
let button = button("")
    .style(move |_, status| button_color(current_color, status))
    .on_press(Event::Color(current_color))
    .width(cell_size)
    .height(cell_size);
```
Since "``""``" turns into a text widget, it seems a bit wasteful as it's just used to create lots of clicky things with with a background.

Implementing `Widget` for `()` can help in this case.

### DIFF
--- a/core/src/widget.rs
+++ b/core/src/widget.rs
@@ -4,6 +4,7 @@ pub mod text;
 pub mod tree;
 
 mod id;
+mod void;
 
 pub use id::Id;
 pub use operation::Operation;
--- a/core/src/widget/void.rs
+++ b/core/src/widget/void.rs
@@ -1,0 +1,48 @@
+use crate::layout;
+use crate::mouse;
+use crate::renderer;
+use crate::widget::{Widget, tree::Tree};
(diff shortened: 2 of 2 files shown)


B0ney:
After thinking about it, I think we should use a concrete `Void` instead of an implicit `()`.

There's a chance that someone would forget to return a widget, so the type system would infer `()`, and the programmer would be left confused as to why their code isn't showing.

```rust
let fancy_text = |txt| { text!("~~ {txt} ~~"); };
```

tarkah:
You can use Space to describe this pattern, which I think is more intuitive and explicit. A button which simply contains some space.

B0ney:
Hmm yeah you're right, that would do just fine actually.

I honestly can't really think of any other reason to have a void widget in place of Space, other than it doesn't allocate. That is hardly worth justifying imo.

I'll close this. Thanks for giving me another perspective.

PR #2531, opened 2024-07-29
Snap `Quad` lines to the pixel grid in `iced_wgpu`

hecrj:
This should ensure the `Quad` primitives representing lines (1px thick) are always aligned to the pixel grid.

May introduce other artifacts, however.

### DIFF
--- a/wgpu/src/shader/quad/solid.wgsl
+++ b/wgpu/src/shader/quad/solid.wgsl
@@ -30,6 +30,15 @@ fn solid_vs_main(input: SolidVertexInput) -> SolidVertexOutput {
 
     var pos: vec2<f32> = (input.pos + min(input.shadow_offset, vec2<f32>(0.0, 0.0)) - input.shadow_blur_radius) * globals.scale;
     var scale: vec2<f32> = (input.scale + vec2<f32>(abs(input.shadow_offset.x), abs(input.shadow_offset.y)) + input.shadow_blur_radius * 2.0) * globals.scale;
+    var snap: vec2<f32> = vec2<f32>(0.0, 0.0);
+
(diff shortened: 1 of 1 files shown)


wgpu/src/shader/quad/solid.wgsl
```diff
@@ -30,6 +30,15 @@ fn solid_vs_main(input: SolidVertexInput) -> SolidVertexOutput {
...
     var scale: vec2<f32> = (input.scale + vec2<f32>(abs(input.shadow_offset.x), abs(input.shadow_offset.y)) + input.shadow_blur_radius * 2.0) * globals.scale;
+    var snap: vec2<f32> = vec2<f32>(0.0, 0.0);
+
+    if input.scale.x == 1.0 {
+        snap.x = round(pos.x) - pos.x;
+    }
+
+    if input.scale.y == 1.0 {
+        snap.y = round(pos.y) - pos.y;
+    }
```
hecrj, line 41: Any graphical wizards have any ideas how to get rid of these conditionals?

vladh: @hecrj 

```
snap = (1.0 - abs(sign(input.scale - 1.0))) * (round(pos) - pos)
```

hecrj: Cool! Let's leave it like this for now. I imagine shader compilation already optimizes this for the most part.

vladh: Yep! If you need it in the future you can write e.g. a `when_eq(x, y)` function that returns `1.0 - abs(sign(x - y))`.

PR #2533, opened 2024-07-29
Shortcut export for Length::Fixed

WailAbou:
This PR addresses a minor oversight by allowing `Length::Fixed` to be imported and used as `use iced::Fixed`, consistent with the other length variants.


hecrj:
This isn't an oversight. `Fixed` should never be necessary, since you can use `From` implementations directly.

WailAbou:
Ah, that makes sense. I completely forgot about that.

PR #2535, opened 2024-07-30
Implement `align_x` for `TextInput`

bungoboingo:
- Added ability to right-align `text_input`
- Added small example for text_inputs


hecrj (approved):
I simplified the logic a bit by assuming the text is always left-aligned and only translating the click and render targets.

We also support `alignment::Horizontal` now, which means `Center` is an option as well! The `todos` example showcases it.

image

PR #2536, opened 2024-08-03
text_editor: Avoid rendering text outside the border

meithecatte:
If the height could fit slightly less than an extra line, said line would protrude beyond the border of the text editor. This PR fixes this.


meithecatte:
To make it clearer what this PR achieves:

Before:

image


After:

image

hecrj (approved):
Thanks!

Made some changes to account for the `padding` of the `text_editor`.

PR #2549, opened 2024-08-14
Remove out of date comment from custom-widget example

JL710:
This PR Removes an out of date comment from the custom_widget example.


hecrj (approved):
Thanks!

PR #2551, opened 2024-08-16
Expose additional `mouse::Interaction` cursors

airstrike:
This PR adds a few more common mouse interaction cursors to `interaction.rs`, namely:

- Resizing diagonally in both directions (pick the right name is probably the hardest part here)
- Zooming out (we already had a cursor for zooming in but not one for zooming out)
- Cell selector (used in Microsoft Excel and possibly other spreadsheet or table-like UI elements)
- Move

Since the `Cell` selector is a bit niche, here is a screenshot of what it looks like in Excel. My use case is building a spreadsheet app, so users would expect this cursor to be there.

image

I take it `ZoomOut` and `Move` are pretty standard and need no specific use case.

There's an argument to be made that `iced` could just expose all of the cursors from `winit` and let the users figure it out, even if some of the cursors are very unusual. I'm not making that argument so I have only added those cursors whose inclusion was more immediately defensible IMHO but open to suggestions.

Admittedly I have not tested the specific resizing cursors because I'm on MacOS and they don't work, but naturally everything compiles and runs fine. They just default to the standard cursor, but I tested other cursors from `winit` and they were fine (including, for example, sizing westward, which is not in this PR) so I'm confident nothing's wrong with the code per se.

FYI this `winit` issue is being tracked here: https://github.com/rust-windowing/winit/issues/3724


airstrike:
Also maybe we have to wait until the `winit` issue is fixed before merging buggy cursors into `iced`, but I didn't want to forget about this PR so I'm submitting it now as it's unlikely the code will break between now and when it gets merged and we can at least track it, debate names and debate including additional cursors.

stevenxxiu:
I also would love to add another cursor, being the `copy` cursor.

I'm porting a *Conky* like project of mine, from *GTK* to *Iced*. *Iced* feels like a more direct toolkit than using the *GTK* *Rust* wrapper.

The use case is when I hover my cursor over some labels, I show a `copy` cursor to indicate the text can be copied.

Since this is a dashboard, the UI should be kept very simple. So I felt like a cursor was the best way to go.

hecrj (approved):
Thanks!

PR #2553, opened 2024-08-17
Expose `Clipboard` to `canvas::Program`

airstrike:
This PR lets `canvas::Program` access the `Clipboard`.

My specific use case is an interactive canvas board with shapes where I can cut, copy and paste shapes around.


hecrj (commented):
Thanks!

I think I prefer to keep clipboard access relatively limited to custom widgets and the `clipboard` tasks; since it essentially is global external state.

A `canvas` can read / write from the clipboard by producing a `Message` that the application can use to run a `clipboard` task.

PR #2555, opened 2024-08-22
Add compact variant for pane grid controls

mtkennerly:
Right now, when the pane grid controls would overlap with the title content, we hide the title content altogether. That's good for avoiding visual issues, but can be confusing for a user when the title content is also interactable and disappears when they try to select it. This PR adds an alternative compact variant to the pane grid controls, which allows overlap.

> https://github.com/user-attachments/assets/02e631b2-9916-4a59-91ff-63e6297999b7

### DIFF
--- a/widget/src/pane_grid/controls.rs
+++ b/widget/src/pane_grid/controls.rs
@@ -1,0 +1,59 @@
+use crate::container;
+use crate::core::{self, Element};
+
+/// The controls of a [`Pane`].
+///
+/// [`Pane`]: super::Pane
+#[allow(missing_debug_implementations)]
+pub struct Controls<
+    'a,
+    Message,
+    Theme = crate::Theme,
+    Renderer = crate::Renderer,
+> where
+    Theme: container::Catalog,
+    Renderer: core::Renderer,
+{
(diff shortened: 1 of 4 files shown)


mtkennerly:
Not sure what the Mac failure is about. There's no specific error message that I can see:

> image

widget/src/pane_grid/controls.rs
```diff
@@ -0,0 +1,59 @@
...
+impl<'a, Message, Theme, Renderer> From<Element<'a, Message, Theme, Renderer>>
+    for Controls<'a, Message, Theme, Renderer>
+where
+    Theme: container::Catalog,
+    Renderer: core::Renderer,
+{
+    fn from(value: Element<'a, Message, Theme, Renderer>) -> Self {
+        Self::new(value)
+    }
+}
```
hecrj, line 59: Nice. No breaking changes!

hecrj (approved):
The `PaneGrid` logic is getting quite complicated.

I wonder if at some stage we could make a `Controls` trait and allow users to fine-tune this kind of logic to their liking. Maybe we can think about that if we ever need further control.

Overall looks good! Let's merge! :ship:

PR #2556, opened 2024-08-23
Document how the state of a `Component` can be managed

dtoniolo:
Hi,
first of all thanks for your work on this amazing library: I've completed some toy projects with it and it was a pleasure to use.

While going through some of the examples I encountered a few challenges in understanding how `Component`s work, particularly regarding:
- How are the methods of a `Component` called? Is the invocation handled automatically by the runtime, or does this responsibility lie with the parent component?
- Why is the state passed as an argument instead of being part of the `Component`?

After spending some more time experimenting and reasoning through these questions, I believe I've gained a better understanding. However, I realized that other newcomers might face similar difficulties, so I decided to add a couple of sections to the documentation of `Component`. I hope that these additions will help others grasp these concepts more quickly.

My contribution is based on my current understanding, so I'm completely open to feedback. If there are any inaccuracies or areas where improvements can be made, I'm more than willing to make the necessary revisions.

Lastly, if you think that expanding the documentation on this topic more would be valuable, I'd be happy to put together a pull request to add a dedicated page to the official book that explains how components work and how to use them effectively.

### DIFF
--- a/widget/src/lazy/component.rs
+++ b/widget/src/lazy/component.rs
@@ -31,6 +31,19 @@ use std::rc::Rc;
 ///
 /// Additionally, a [`Component`] is capable of producing a `Message` to notify
 /// the parent application of any relevant interactions.
+///
+/// # State
+/// A component can store its state in one of two ways: either as data within the
+/// implementor of the trait, or in a type [`State`][Component::State] that is managed
+/// by the runtime and provided to the trait methods. These two approaches are not
+/// mutually exclusive and have opposite pros and cons.
+///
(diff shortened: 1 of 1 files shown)


airstrike:
FYI `Component` has been deprecated: https://github.com/iced-rs/iced/commit/9426418adbaac40f584fe16b623521a3a21a1a4c

hecrj (approved):
Thanks!

I removed the lifetime section since it was not really accurate. It's unclear what you meant by "parent component"; there is no such thing.

PR #2560, opened 2024-08-29
Initialize `fontdb` for `svg` rendering

ivanceras:
This PR just load the system default fonts such that rendering svg that uses text on it will be rendered using the default system fonts.


hecrj (approved):
Thanks!

PR #2565, opened 2024-09-02
slider: handle mouse wheel events

kiedtl:
Make `slider` + `vertical_slider` handle mouse wheel events -- i.e. when hovered, scrolling the mouse wheel will modify the slider.

---

Question: currently the slider is incr/decremented by `self.step` for both `ScrollDelta::Pixels` and `ScrollDelta::Lines`. This seems appropriate for `Lines`, but I'm not sure about `Pixels`. (It felt fine to me when testing, but I'm not sure what the "expected" behaviour is.)

### DIFF
--- a/widget/src/slider.rs
+++ b/widget/src/slider.rs
@@ -288,22 +288,6 @@ where
         };
 
         match event {
-            Event::Mouse(mouse::Event::WheelScrolled { delta }) => {
-                if let Some(_) = cursor.position_over(layout.bounds()) {
-                    let delta = match delta {
-                        mouse::ScrollDelta::Lines { x: _, y } => y,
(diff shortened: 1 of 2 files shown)


hecrj (approved):
Thanks!

I made a change to only enable slider scrolling when the `Ctrl` modifier is pressed. This way, we avoid accidental slider changes when inside a `scrollable`.

kiedtl:
Oh. I was actually trying to emulate Windows-specific behavior here, specifically the app I was replacing with my Iced project. Of course, users can adjust, but the initial friction will be there.

So I suppose there's no way for a widget inside another to prevent the outer widget from capturing its events?

PR #2568, opened 2024-09-04
add on_down and on_release events to button widget

edwloef:
Adds on_down and on_release events to the button widget. These are necessary to be able to perform actions only while a button is depressed.

The button is now enabled despite `on_press` never having being called, if `on_down` or `on_release` have been called. No other existing functionality has been changed.

Example usage:
```rs
use iced::{
    widget::{button, row, text},
    window::frames,
    Element, Subscription,
};

struct Counter {
    counter: usize,
    is_increasing: bool,
}

#[derive(Debug, Clone)]
enum Message {
    IncreaseDown,
    IncreaseReleased,
    Tick,
}

impl Counter {
    fn new() -> Self {
        Self {
            counter: 0,
            is_increasing: false,
        }
    }

        match message {
            Message::IncreaseDown => {
                self.is_increasing = true;
            }
            Message::IncreaseReleased => {
                self.is_increasing = false;
            }
            Message::Tick => {
                if self.is_increasing {
                    self.counter += 1;
                }
            }
        }
    }

        row![
            button("+")
                .on_down(Message::IncreaseDown)
                .on_release(Message::IncreaseReleased),
            text(format!("{}", self.counter)),
        ]
        .into()
    }

    fn subscription(state: &Self) -> Subscription<Message> {
        if state.is_increasing {
            frames().map(|_| Message::Tick)
        } else {
            Subscription::none()
        }
    }
}

impl Default for Counter {
    fn default() -> Self {
        Self::new()
    }
}

fn main() -> iced::Result {
    iced::application("", Counter::update, Counter::view)
        .subscription(Counter::subscription)
        .run()
}
```

### DIFF
--- a/widget/src/button.rs
+++ b/widget/src/button.rs
@@ -53,6 +53,8 @@ where
     Theme: Catalog,
 {
     content: Element<'a, Message, Theme, Renderer>,
+    on_down: Option<Message>,
+    on_release: Option<Message>,
     on_press: Option<OnPress<'a, Message>>,
     width: Length,
     height: Length,
@@ -89,6 +91,8 @@ where
 
         Button {
             content,
+            on_down: None,
+            on_release: None,
             on_press: None,
             width: size.width.fluid(),
             height: size.height.fluid(),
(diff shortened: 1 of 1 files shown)


hecrj:
Use the [`mouse_area`](https://docs.rs/iced/latest/iced/widget/struct.MouseArea.html).

wiiznokes:
To be fair, emit the action on press is a common pattern, and a fair amount of people think it should be the default

PR #2571, opened 2024-09-08
Paint Example

EmmanuelDodoo:
This PR adds a paint application to the examples. 

PS: I didn't expect it to get this large but I might as well make a PR for this. 

image


hecrj:
Very cool! Code looks quite good as well.

I think this is worth a repository of its own, however. I can't commit to maintaining such a long example, and I believe it has the potential to become a fully fleshed out open-source application.

PR #2574, opened 2024-09-10
Fix `wasm32` deployments not displaying anything

### LINKED ISSUE #2480: Running `trunk serve` on examples results in blank web pages

### What happened?

When I run `cargo r` on the examples, a window is created and it works as expected. When I run `trunk serve` on the examples which are also supposed to work on the web, the page just has a title and blank canvas. I'm not getting an error with trunk itself:
```
test@test-Standard-PC-Q35-ICH9-2009:~/iced/examples/counter$ trunk serve
2024-06-25T17:39:37.269355Z  INFO \U0001f680 Starting trunk 0.20.2
2024-06-25T17:39:37.274371Z  INFO \U0001f4e6 starting build
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.44s
2024-06-25T17:39:45.058106Z  INFO applying new distribution
2024-06-25T17:39:45.060747Z  INFO \u2705 success
2024-06-25T17:39:45.061741Z  INFO \U0001f4e1 serving static assets at -> /
2024-06-25T17:39:45.061967Z  INFO \U0001f4e1 server listening at:
2024-06-25T17:39:45.061991Z  INFO     \U0001f3e0 http://127.0.0.1:8080/
2024-06-25T17:39:45.062031Z  INFO     \U0001f3e0 http://[::1]:8080/
```

To reproduce:
- Clone this repo (commit `b9eb86199afe0f2d936eb4ab90af5b2a2c32a87a`)
- `cd examples/counter`
- `trunk serve`

I tried it on NixOS and also in a Ubuntu VM and both had the same result.

The counter and TODO examples should have text and buttons showing up on the web page.

### Version

master

### Operating System

Linux

```shell
Console at http://localhost:8080:

WEBGL_debug_renderer_info is deprecated in Firefox and will be removed. Please use RENDERER. counter-91af9f8a594c1ec3.js:1366:37
WebGL warning: texSubImage: Uploads from a buffer with a final row with a byte count smaller than the row stride can incur extra overhead. 2
WebGL warning: drawArraysInstanced: Drawing to a destination rect smaller than the viewport rect. (This warning will only be given once) 
```
```

derezzedex:
This fixes a bug where `web` deployments of an `iced` application would silently fail, not displaying anything.

The problem can be tested by using `trunk serve` in any of the examples that support it.

It seems to have started after #2469, where changes were introduced to allow an application to be started without an window. Although an window is not necessarily shown at the start, it's still created silently, and this is necessary for the acquisition of the graphics device and it's initialization, after that the window is discarded and only created again when requested by the `runtime` (or quickly after that on "non-deamon" applications).

On the `wasm32` path of the code, the window contains the [`HTMLCanvasElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement), which in turn, has the [`getContext`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext) method and contains the [`WebGLRenderingContext`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext) which is necessary for properly displaying graphics on the `canvas`.
By discarding the window, we end up discarding the `HTMLCanvasElement` which is associated with the context and other graphics dependencies for proper rendering; creating a new window without using the correct associated `canvas` element which was used to initialize the context, then makes the `canvas` graphically display nothing.

There could multiple solutions to this, but the simplest seems to be to simply retain the `canvas` element, and re-use it to create the `window` through the [`with_canvas`](https://docs.rs/winit/latest/winit/platform/web/trait.WindowAttributesExtWebSys.html#tymethod.with_canvas) window attribute.

This PR changes the `wasm32` path to store the [`web_sys::HTMLCanvasElement`](https://docs.rs/web-sys/latest/web_sys/struct.HtmlCanvasElement.html) and use it on the `with_canvas` method when creating an window, I've also added logging for the window attributes to the relevant code, which could be helpful in future debugging of similar problems.

As for possible problems with this, the code doesn't make any guarantees for other windows, it'll certainly silently fail when asked to create new windows, since those are basically `canvas` elements on `wasm32` that are not properly created or maintained. Although I would add that this was already the case before the PR.

Closes #2480.

### DIFF
--- a/winit/Cargo.toml
+++ b/winit/Cargo.toml
@@ -44,5 +44,5 @@ winapi.workspace = true
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 web-sys.workspace = true
-web-sys.features = ["Document", "Window"]
+web-sys.features = ["Document", "Window", "HtmlCanvasElement"]
 wasm-bindgen-futures.workspace = true
--- a/winit/src/program.rs
+++ b/winit/src/program.rs
@@ -252,6 +252,8 @@ where
         is_booted: std::rc::Rc<std::cell::RefCell<bool>>,
         #[cfg(target_arch = "wasm32")]
         queued_events: Vec<Event<Action<Message>>>,
+        #[cfg(target_arch = "wasm32")]
+        canvas: Option<web_sys::HtmlCanvasElement>,
     }
 
     struct BootConfig<C> {
@@ -277,6 +279,8 @@ where
         is_booted: std::rc::Rc::new(std::cell::RefCell::new(false)),
         #[cfg(target_arch = "wasm32")]
         queued_events: Vec::new(),
+        #[cfg(target_arch = "wasm32")]
+        canvas: None,
     };
 
     impl<Message, F, C> winit::application::ApplicationHandler<Action<Message>>
@@ -306,6 +310,12 @@ where
                     return;
                 }
             };
+
+            #[cfg(target_arch = "wasm32")]
+            {
+                use winit::platform::web::WindowExtWebSys;
+                self.canvas = window.canvas();
+            }
 
             let finish_boot = async move {
                 let mut compositor =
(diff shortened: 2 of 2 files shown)


hecrj (approved):
Sweet!

Made some changes to discard the empty polling events during boot.

PR #2578, opened 2024-09-13
Fix unnecessary COM initialization on Windows

### LINKED ISSUE #2577: Unnecessary COM initialization with drag and drop disabled

### What happened?

After upgrading my application from `0.12` to the `master` branch, I noticed COM (required for file drag and drop on Windows) is initialized even when the window setting flag `drag_and_drop` is set to false. This causes unnecessary collisions with user code, which can use COM in a different way.

The following code works in `0.12` but not in `master`:
```rust
fn main() {
    unsafe {
        CoInitializeEx(None, COINIT_MULTITHREADED).unwrap();
    }

    let window = window::Settings {
        platform_specific: PlatformSpecific {
            drag_and_drop: false,
            ..PlatformSpecific::default()
        },
        ..window::Settings::default()
    };

    iced::application("A cool counter", Counter::update, Counter::view)
        .window(window)
        .run()
        .unwrap();
}

#[derive(Default)]
struct Counter;

impl Counter {

    fn view(&self) -> Button<()> {
        iced::widget::button("Hello, world!")
    }
}
```

> 

COM is not touched by `winit` and the provided example works as in `0.12`.

### Version

master

### Operating System

Windows

```shell
thread 'main' panicked at C:\Users\<>\.cargo\git\checkouts\winit-57d3141eaf559308\254d6b3\src\platform_impl\windows\window.rs:1179:17:
OleInitialize failed! Result was: `RPC_E_CHANGED_MODE`. Make sure other crates are not using multithreaded COM library on the same thread or disable drag and drop support.
```

Jinderamarak:
I tracked the issue #2577  back to this [commit](https://github.com/iced-rs/iced/commit/341c9a3c12aa9d327ef1d8f168ea0adb9b5ad10b) and the file `winit/src/program.rs:L273`, where the window is first initialized with `default()` window options instead of the provided ones.

I am not going to lie, I don't know why we need to create a window with default attributes first and then initialize it again with proper attributes later, but if I change the first initialization to be without drag-and-drop support, then COM is initialized later only when it is really required.

I tested this change with `examples/events` and the reproduction code from the issue and everything works as I would expect it to on Windows.

Fixes #2577.

### DIFF
--- a/winit/src/program.rs
+++ b/winit/src/program.rs
@@ -296,14 +296,22 @@ where
                 return;
             };
 
-            let window = match event_loop.create_window(
-                winit::window::WindowAttributes::default().with_visible(false),
-            ) {
-                Ok(window) => Arc::new(window),
-                Err(error) => {
-                    self.error = Some(Error::WindowCreationFailed(error));
-                    event_loop.exit();
-                    return;
+            let window = {
+                let attributes = winit::window::WindowAttributes::default();
+
+                #[cfg(target_os = "windows")]
+                let attributes = {
+                    use winit::platform::windows::WindowAttributesExtWindows;
+                    attributes.with_drag_and_drop(false)
+                };
+
+                match event_loop.create_window(attributes.with_visible(false)) {
+                    Ok(window) => Arc::new(window),
+                    Err(error) => {
+                        self.error = Some(Error::WindowCreationFailed(error));
+                        event_loop.exit();
+                        return;
+                    }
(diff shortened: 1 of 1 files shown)


Jinderamarak:
I would also consider using the window options provided when creating the application in this place, because it would be hard to track all of the flags causing similar side effects.

hecrj (approved):
Thanks!

PR #2579, opened 2024-09-14
Add rust-version.

dcampbell24:
This is a tiny addition. Add the rust-version. I checked it with `cargo msrv`.


hecrj (approved):
Thanks!

I added a check in CI to make sure we don't forget updating the value in the future.

PR #2585, opened 2024-09-19
Scrollbar disabled style status to allow custom styles for always visible, but not overflowing scrollbars.

dtzxporter:
Feel free to change the 'name' of it. I originally used 'disabled' to signal that it's not usable, but still visible. 'Overflowing' may be a better term.

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -906,14 +906,21 @@ where
                 is_vertical_scrollbar_dragged: state
                     .y_scroller_grabbed_at
                     .is_some(),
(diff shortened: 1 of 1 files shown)


hecrj (commented):
Is this `disabled` state necessary? Isn't it already obvious for a user that a scrollbar without a scroller isn't interactable?

dtzxporter:
In my app, I have a custom style that more clearly signifies that scrolling isn't available (vs scrolling being not a thing period) It's nice to be able to customize based on whether or not it's usable.

hecrj (approved):
Fair enough. Thanks! :bow:

PR #2590, opened 2024-09-20
Update button Catalog and Style documentation

MichelleGranat:
Added a small explanation of where the button style comes from.
Added an example of implementing button catalog for a custom theme.

I ran the example to make sure it works on 0.13

### DIFF
--- a/widget/src/button.rs
+++ b/widget/src/button.rs
@@ -471,6 +471,9 @@ pub enum Status {
 }
 
 /// The style of a button.
+/// 
+/// If not specified with [`Button::style`]
+/// the theme will provide the style.
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct Style {
     /// The [`Background`] of the button.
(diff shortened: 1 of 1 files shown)


widget/src/button.rs
```diff
@@ -505,6 +508,47 @@ impl Default for Style {
...
+///                 style.background = Some(Background::Color(Color::from_rgb(0.941, 0.502, 0.502)));
+///             },
+///         }
+/// 
+///         style
+///     }
+/// }
+/// ```
 pub trait Catalog {
     /// The item class of the [`Catalog`].
```
hecrj, line 553: May be interesting to mention here in the `Class` that `Button::style` can only be used if the associated type here implements: `From<StyleFn<'a, Self>>`.

widget/src/button.rs
```diff
@@ -505,6 +508,47 @@ impl Default for Style {
...
 
 /// The theme catalog of a [`Button`].
+/// 
+/// All themes that can be used with [`Button`]
+/// must implement this trait.
+/// 
+/// # Example
+/// ```no_run
+/// #[derive(Debug, Default)]
+/// pub enum ButtonStyles {
```
hecrj, line 518: Maybe `ButtonClass`?

widget/src/button.rs
MichelleGranat: Done

widget/src/button.rs
MichelleGranat: Done

widget/src/button.rs
```diff
@@ -505,6 +508,51 @@ impl Default for Style {
...
+///         }
+/// 
+///         style
+///     }
+/// }
+/// ```
+/// 
+/// Although, in order to use [`Button::style`]
+/// with `MyTheme`, [`Catalog::Class`] must implement
+/// `From<StyleFn<'_, MyTheme>`.
```
liigo, line 555: need one more `>`

MichelleGranat: Done

hecrj (approved):
Looks good! Thank you.

MichelleGranat:
Fixed it and now it should pass the tests

PR #2605, opened 2024-09-24
Include images and saved meshes when pasting `Frame`

Gobbel2000:
`tiny_skia::Frame` was ignoring images in `Frame::paste`, making images not show up when created in a `with_clip` context.

`wgpu::Frame` similarly did not pass through meshes in its paste method, that may have been saved from a nested `with_clip` call.

I'm not sure that nesting `with_clip` makes a lot of sense, because only the innermost clip region gets applied (should this be documented?), also it seems like images are not yet affected by clipping? Nevertheless, these fixes seem relevant.


hecrj (approved):
Thanks!

PR #2606, opened 2024-09-25
feat: set total size of QRCode

tvolk131:


widget/src/qr_code.rs
```diff
@@ -84,11 +84,17 @@ where
     }
 
     /// Sets the size of the squares of the grid cell of the [`QRCode`].
-    pub fn cell_size(mut self, cell_size: u16) -> Self {
+    pub fn cell_size(mut self, cell_size: f32) -> Self {
```
tvolk131, line 87: Note for reviewer: This is a breaking change to the public API. Let me know if that's a problem! We can easily keep the rest of the changes and revert this back to accepting a `u16`.

hecrj (approved):

PR #2607, opened 2024-09-25
Feat: Add warning color to palette

DavidAguilo:
I have added a warning field to the palette since I have found myself needing one. It uses yellow, unless otherwise specified by the theme page.

### DIFF
--- a/core/src/theme/palette.rs
+++ b/core/src/theme/palette.rs
@@ -17,10 +17,10 @@ pub struct Palette {
     pub primary: Color,
     /// The success [`Color`] of the [`Palette`].
     pub success: Color,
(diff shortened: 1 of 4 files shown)


tvolk131 (commented):
A bit of a nitpick, but it might make more sense to have warning be listed between success and danger rather than after danger. Otherwise LGTM!

DavidAguilo:
You are right, it makes more sense. I've made the commit to move warning lines to be between success and danger.

tvolk131 (approved):
Nice!

hecrj (approved):
Sweet! I've been feeling like the built-in palette is quite limited for a while.

Thanks!

PR #2609, opened 2024-09-26
Expose shell to canvas::Program to match shader::Program.

dtzxporter:
Allows a canvas program to schedule redraws as needed without managing it externally.

This matches the more powerful shader::Program update method now.

### DIFF
--- a/examples/bezier_tool/Cargo.toml
+++ b/examples/bezier_tool/Cargo.toml
@@ -7,4 +7,4 @@ publish = false
 
 [dependencies]
 iced.workspace = true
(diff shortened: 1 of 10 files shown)


bbb651:
For both of them, is there a reason `Option<Message>` is still needed into the return value if you have access to `Shell`? It's strictly more powerful because you can publish more than one message, and for other cases it seems unnecessary to have two ways to do the same thing. `Widget` already works like this, returning only `Status`

hecrj:
This has already been addressed by #2662.

PR #2610, opened 2024-09-26
Fix slider

wiiznokes:
attempt to fix https://github.com/pop-os/cosmic-settings/issues/628

currently, the grab handle seems a little off, but i'm not sure this is a layout  _bug_.

master
Capture d’écran du 2024-09-26 23-42-39
pr
Capture d’écran du 2024-09-26 23-43-42

### DIFF
--- a/widget/src/slider.rs
+++ b/widget/src/slider.rs
@@ -464,21 +464,6 @@ where
 
         let rail_y = bounds.y + bounds.height / 2.0;
 
-        // background bar
-        renderer.fill_quad(
-            renderer::Quad {
-                bounds: Rectangle {
-                    x: bounds.x + handle_width / 2.0,
-                    y: rail_y - style.rail.width / 2.0,
-                    width: bounds.width - handle_width,
-                    height: style.rail.width,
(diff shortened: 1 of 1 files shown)


wiiznokes:
Hum, i'm starting to think its a cosmic only issue. Will see when they rebase they fork

wiiznokes:
Initially, i wanted to fix a cosmic applet that use the slider from Iced
screenshot-2024-10-02-15-38-05
(only visible on hover)

but yeah, it will require more investigation from me. Sorry for the useless PR

hecrj:
No worries!

peterkrull:
I tried to bring this up [on discourse](https://discourse.iced.rs/t/making-slider-widget-follow-curser-more-accurately/521), but I guess not many people saw it. My suggestion is to have "saturation zones" at each end of the slider, where clicking that region would bring the slider to either 0% or 100%. The length of these regions would be equal to the half-width of the handle. So the slider would look the same as it does now, but the way it responds to events would be modified slightly.

PR #2611, opened 2024-09-27
Don't fill out of viewport text

tarkah:
Super smol optimization. There's no need to submit this rendering operation when text is out of viewport. In `halloy` we leverage this optimization and it has a huge impact on primitive generation timings when dealing w/ lots of text in a scrollable.

### DIFF
--- a/core/src/widget/text.rs
+++ b/core/src/widget/text.rs
@@ -334,10 +334,6 @@ pub fn draw<Renderer>(
 {
     let bounds = layout.bounds();
 
-    if !bounds.intersects(viewport) {
-        return;
-    }
-
     let x = match paragraph.horizontal_alignment() {
         alignment::Horizontal::Left => bounds.x,
         alignment::Horizontal::Center => bounds.center_x(),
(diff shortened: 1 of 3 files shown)


tsuza:
I've tested it with ~10k text widgets, containing `Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua` in one of my projects. 
Scrolling with upstream iced is much, much, laggier. This PR makes it buttery smooth.

I've attached some screenshots of the debug info, before ( 0.13.1 stable branch ) & after ( this PR's branch ), while scrolling or idling.

There shouldn't have been too many changes in the dev branch ( 0.14.0 ) in comparison to the stable branch ( 0.13.1 ), so it shouldn't impact the fps too much. Though if it's the case, I can re-do the tests.

tarkah:
Thanks for the detailed comparison. Render times are way down!

tamewild:
IIRC, this might be fixed when Iced's glyphon has this: https://github.com/grovesNL/glyphon/pull/108

hecrj (approved):
I think it's better to cull primitives in `column` and `row` instead of `text`. This way, we cull _any_ primitive and all widgets benefit.

Let me know if that works.

tarkah (commented):
LGTM!

PR #2620, opened 2024-09-30
WASM: Create canvas before first window in resumed(). Fixes #2618

### LINKED ISSUE #2618: ModifiersChanged not working on WASM

### What happened?

Using a `text_editor` on wasm, using a modifier + key just results in a regular keypress instead of a modified key. (`Ctrl-C` just types the letter `c`)

Modifiers are applied to keystrokes

### Version

master

### Operating System

macOS

_No response_

boondocklabs:
This creates a canvas and inserts into the document body, and attaches it to the initial window in Runner::resumed(). This fixes an issue where events were being routed to a WindowId that was not inserted into the window manager as reported in #2618.

With this change, duplicate winit messages to the rogue WindowId are no longer observed, and ModifiersChanged events are associated with the correct WindowId.

### DIFF
--- a/winit/src/program.rs
+++ b/winit/src/program.rs
@@ -286,6 +286,77 @@ where
         F: Future<Output = ()>,
         C: Compositor + 'static,
     {
+        #[cfg(target_arch = "wasm32")]
+        fn resumed(&mut self, event_loop: &winit::event_loop::ActiveEventLoop) {
+            let Some(BootConfig {
+                sender,
+                fonts,
+                graphics_settings,
+            }) = self.boot.take()
+            else {
+                return;
+            };
+
+            let window_attributes =
+                winit::window::WindowAttributes::default().with_visible(false);
+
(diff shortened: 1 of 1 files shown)


winit/src/program.rs
```diff
@@ -286,6 +286,77 @@ where
         F: Future<Output = ()>,
         C: Compositor + 'static,
     {
+        #[cfg(target_arch = "wasm32")]
+        fn resumed(&mut self, event_loop: &winit::event_loop::ActiveEventLoop) {
```
Mirauzo, line 290: **
```suggestion
        fn resumed(&mut self, event_loop: &winit::event_loop::ActiveEventLoop) {
```
**

Mirauzo (commented):
- [ ] # > # - [ ] **_**

hecrj:
I think #2722 should take care of this. Give it a shot.

PR #2626, opened 2024-10-04
chore: remove once_cell dependency

tvolk131:
The need for `once_cell` was [removed](https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html#lazycell-and-lazylock) with the release of Rust v1.80, which we upgraded to in #2579 Let's remove this dependency!


hecrj (approved):
Cool! Thanks :tada:

PR #2627, opened 2024-10-04
Add `maximized` and `fullscreen` to `window::Settings`

bbb651:


hecrj (approved):
Seems reasonable. Thanks! :bow:

PR #2628, opened 2024-10-04
Fix/pane grid continuity

tarkah:
Fixes issues w/ pane grid continuity of widget state when adding / removing panes, drag / dropping panes, or maximize / restoring panes. I've tested this fix with `halloy` and can confirm that scrollable offsets within each pane are perfectly retained when doing all the above operations. Previously the states would have gotten shuffled around relative to the panes which would cause the wrong scrollable offsets to get used for each pane (jump around).

For `maximize` / `restore`, we still need to build `view` for the hidden panes so diff'ing them doesn't break. This can be optimized to avoid calling `diff` on those hidden panes, allowing us to skip viewing them, but this adds additional complexity I didn't want to try and tackle here. The current approach allows us to naively zip all content, layout & state and then just applying a filter to return the maximized pane within each widget operation.

### DIFF
--- a/widget/src/lazy/responsive.rs
+++ b/widget/src/lazy/responsive.rs
@@ -83,7 +83,10 @@ where
         new_size: Size,
         view: &dyn Fn(Size) -> Element<'a, Message, Theme, Renderer>,
     ) {
-        if self.size == new_size {
+        let is_tree_empty =
+            tree.tag == tree::Tag::stateless() && tree.children.is_empty();
+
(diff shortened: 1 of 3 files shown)


hecrj (approved):
Awesome! :tada: 

I fixed some duplicated state and renamed some stuff. The changes also exposed a pesky bug in the `responsive` widget.

Let's merge!

tarkah:
Looks great, thanks!

PR #2629, opened 2024-10-05
Add Borrow trait bound on QRCode Data parameter, to allow smart pointers to be owned by QRCode instance

boondocklabs:
This allows QRCode to own smart pointers (`Rc`, `Arc`), and is a non-breaking change as it also accepts `&Data`. This helps with interior mutability patterns when Data is behind RefCell/Mutex/RwLock.


hecrj:
What is the specific use case here?

Interior mutability patterns should not be allowed in `view` code since `iced` cannot magically tell when the borrowed data changes (i.e. spooky action at a distance).

`iced` also provides no guarantees on the lifetime of the widget tree; which means the mutable cells must never be written to outside of `update`. And since you have mutability in `update` already, it defeats the purpose of interior mutability.

PR #2632, opened 2024-10-08
Fixed bug where delete key wasn't working in the text editor

### LINKED ISSUE #2631: Delete key not working in text editor

### What happened?

The delete key is not deleting characters in the text editor widget. The corresponding action is also not being printed when I log all actions and press the delete key. The delete works fine in the text input widget.

The next character after the cursor should be deleted when the delete key is pressed.

### Version

master

### Operating System

Linux

_No response_

leo030303:
Fixes #2631.
When the delete key was pressed, it would only run the Delete binding if the content of the text editor was None. This was the source of the bug, I switched it to an is_some() check.

### DIFF
--- a/widget/src/text_editor.rs
+++ b/widget/src/text_editor.rs
@@ -1045,7 +1045,9 @@ impl<Message> Binding<Message> {
             keyboard::Key::Named(key::Named::Backspace) => {
                 Some(Self::Backspace)
             }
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks!

Your fix would have broken the delete key in the numpad. The proper fix is to actually match on the `DELETE` UTF-8 character code instead.

PR #2633, opened 2024-10-11
more window tasks

JL710:
This PR adds more tasks to change the properties of a window such as resizable and min/max size.

For the future it might be an idea to make it possible to apply [`Settings`](https://docs.iced.rs/iced/window/struct.Settings.html) for and existing window.


hecrj (approved):
Cool! Thanks :partying_face: 

I have removed the `change_title` task since applications can already change the title at any time through the `title` function (e.g. [`Tour::title`](https://github.com/iced-rs/iced/blob/717b296cdb17f22b410e2985b58e73af2199bfcc/examples/tour/src/main.rs#L60)).

PR #2637, opened 2024-10-13
Add PartialEq derives for widget styles

Brady-Simon:
This PR adds `PartialEq` support to all widget styles. This will allow comparison between styles, which will help for some animation-related work of tracking whether a style has changed due to changes to a widget's properties.


hecrj (approved):
Thanks!

PR #2639, opened 2024-10-14
Introduce consecutive click distance check like other toolkit do such as gtk,qt, imgui.

l-const:
Prior art:

- Gtk: https://github.com/GNOME/mutter/blob/ab95a95a1d53c95028600132f947e7a1ebb57bbe/clutter/clutter/clutter-text.c#L2247C1-L2256C1
- Qt: https://github.com/qt/qtbase/blob/dev/src/gui/kernel/qguiapplication.cpp#L2371-L2378
- imgui: https://github.com/ocornut/imgui/blob/master/imgui.cpp#L9789-L9794

### DIFF
--- a/core/src/mouse/click.rs
+++ b/core/src/mouse/click.rs
@@ -82,7 +82,7 @@ impl Click {
             None
         };
 
-        self.position == new_position
+        self.position.distance(new_position) < 6.0
             && duration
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks!

PR #2642, opened 2024-10-16
feat: add a window drag resize task

tsuza:
When implementing a custom title bar, most things are straight forward: minimizing, maximizing, closing, dragging, and the context menu. 
However, there is one small issue: resizing. Since you have to turn off decorations, the border also gets disabled, and there is no easy way to specifically enable that through winit ( afaik ).

This PR exposes `winit`'s `drag_resize_window` to give an easier time to resize the window, which is especially useful for the case I mentioned above.

Hopefully everything is up to standard!

### DIFF
--- a/winit/src/conversion.rs
+++ b/winit/src/conversion.rs
@@ -1136,9 +1136,7 @@ pub fn user_attention(
     }
 }
 
-/// Converts some [`ResizeDirection`] into its `winit` counterpart.
-///
-/// [`ResizeDirection`]: window::ResizeDirection
+/// Converts some [`window::Direction`] into a [`winit::window::ResizeDirection`].
 pub fn resize_direction(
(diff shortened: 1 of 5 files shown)


winit/src/conversion.rs
```diff
@@ -1112,7 +1112,7 @@ pub fn native_key_code(
     }
 }
 
-/// Converts some [`UserAttention`] into it's `winit` counterpart.
+/// Converts some [`UserAttention`] into its `winit` counterpart.
```
tsuza, line 1115: This is technically out of scope, but I noticed this typo so I figured I'd fix it while I'm at it.

hecrj (approved):
Cool! Thanks 🥳 

I think this may be a good escape hatch for achieving drag resize functionality while we figure out #1542.

PR #2644, opened 2024-10-19
Modified clock example to make the clock more readable.

kgday:
Add numbers on the clock face of the clock example and take the portion of the hour passed into consideration for the hour hand.  It now looks like a reasonable clock. 

image


hecrj:
Thanks!

I made the numbers bigger and added some ticks as well:

image

PR #2647, opened 2024-10-24
Add a hello world example

hydra:
When I attempted to learn Iced I looked for a hello world example that doesn't have/need/use any kind of state, I just wanted a basic window with some text on it.  I did not find such an example, so I made this one.

What was interesting is that the framework forces you to define a `Message` type and an update method which uses the `Message` type, but the Hello World case does not have any messages or state.  You also have to return something from the `update` method, again forcing you to use additional types that I didn't expect to have to learn about for a "Hello world" example.  i.e. `Task`.

Hope this helps others looking for a simpler example than the 'Counter' example.

### DIFF
--- a/examples/hello_world/Cargo.toml
+++ b/examples/hello_world/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "hello_world"
+version.workspace = true
+authors = ["Dominic Clifton <me@dominicclifton.name>"]
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+homepage.workspace = true
+rust-version.workspace = true
+
+[dependencies]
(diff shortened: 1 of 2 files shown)


alex-ds13:
This was already suggested previously. Here is the response of the maintainer: https://github.com/iced-rs/iced/pull/2434#issuecomment-2341558069

hydra:
> This was already suggested previously. Here is the response of the maintainer: [#2434 (comment)](https://github.com/iced-rs/iced/pull/2434#issuecomment-2341558069)

Interesting, happy to note I wasn't the first person looking for one!

hecrj:
As pointed out, the lack of a `hello_world` example is deliberate!

PR #2654, opened 2024-10-31
Publish `DragEvent::Canceled` always when picked `Pane` is dropped outside valid target area

jsatka:
~~Currently a `pane_grid::DragEvent::Canceled` is not published when a picked `pane_grid::Pane` is dropped near the origin point where it was picked up (closer than `DRAG_DEADBAND_DISTANCE`)~~ or outside the `PaneGrid`. However in both cases in the pane grid state `pane_grid::Memory` the action state gets set from `Action::Dragging` to `Action::Idle` i.e. cancelling the dragging action.

~~As such, clicking a pane title bar produces `DragEvent::Picked` event but not subsequent `DragEvent::Canceled`, leaving a user who needs to track the dragging state thinking the pane is being dragged while it is not.~~

This change makes sure that dropping a picked pane will always produce either `Dropped` or `Canceled` variant, depending whether the drop point was a valid `pane_grid:::Target` area.

**EDIT:** https://github.com/iced-rs/iced/pull/2691 added publishing `Canceled` event when dropping within deadband distance but dropping outside `PaneGrid` bounds still does not produce a `Canceled` event, making reliable dragging state not possible for the user with current version.


jsatka:
Easy way to test this is to run the **pane_grid** example with the following changes to `Message` handling to print all drag events to console:
```rust
            Message::Dragged(pane_grid::DragEvent::Picked { pane }) => {
                println!("{:?} drag started", pane);
            }
            Message::Dragged(pane_grid::DragEvent::Dropped {
                pane,
                target,
            }) => {
                self.panes.drop(pane, target);
                println!("{:?} drag dropped", pane);
            }
            Message::Dragged(pane_grid::DragEvent::Canceled { pane }) => {
                println!("{:?} drag canceled", pane);
            }
```
and then testing to:
- clicking the pane title bars
- dragging and dropping panes outside the window

PR #2658, opened 2024-11-02
feat: add special platform event

Decodetalkers:
There are many different events in different platform, so they should not included in any of the events, like touch, window or etc. This maybe will make the production platform plugin more flexible

Because we are trying to make iced on different wayland platforms, so ..

### DIFF
--- a/core/src/event.rs
+++ b/core/src/event.rs
@@ -13,7 +13,7 @@ use crate::window;
 ///
 /// [open an issue]: https://github.com/iced-rs/iced/issues
 #[derive(Debug, Clone, PartialEq)]
-pub enum Event {
+pub enum Event<Custom = ()> {
     /// A keyboard event
     Keyboard(keyboard::Event),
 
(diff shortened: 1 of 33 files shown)


hecrj (commented):
What's the specific use case here?

Decodetalkers:
> What's the specific use case here?

I want to send some unique platform event, for example, wayland event. And I think other days if there is another platform, and they want to send their event in their platform plugin, I think add a dynamic part here will be of some help

hecrj:
What kind of event? What's the _specific_ use case you are tackling?

Decodetalkers:
> What kind of event? What's the _specific_ use case you are tackling?

Like some wayland events, I want to pass wayland output in Events

HigherOrderLogic:
With this patch, will it be possible to run another Wayland dispatcher alongside the main event loop using the same connection?

Decodetalkers:
> With this patch, will it be possible to run another Wayland dispatcher alongside the main event loop using the same connection?

It may be possible, as long as the wayland connection is passed out through the event

Decodetalkers:
ok, we should use broadcast. so it is not an problem anymore, closed

PR #2662, opened 2024-11-04
Reactive Rendering

hecrj:
This PR changes the runtime to only perform a redraw of the user interface when either an `update` occurs, or a widget explicitly requests it.

Currently, `iced` renders all the time unconditionally on every runtime event. Effectively, this means that when you move your mouse around in an `iced` application, the runtime can easily draw hundreds of identical frames—wasting plenty of resources.

The changes here make both the runtime and the built-in widgets aware of their last status when drawn, and only request a redraw when needed. This should normally translate to a lower CPU and (specially) GPU usage when interacting with most applications.

As a consequence, the `Widget` and `Overlay` contract now require implementors to explicitly request redraws when necessary. Most of the time, this entails some kind of "diffing" during event processing to figure out if the widget state has changed.

This new feature is enabled by default but, since it makes widget implementation more difficult and may not make sense for highly dynamic applications, a user can opt out by enabling the `unconditional-rendering` feature; which restores the old behavior.

Related important breaking changes include:
- `Widget::on_event` does not return an `event::Status` anymore. Instead, a widget can capture an event by calling `Shell::capture_event`. Since widget logic is generally quite imperative, this paradigm fits better and is also more consistent with other shell-related logic.
- `Widget::on_event` has been renamed to `update`.
- `Program` both for `canvas` and `shader` now return a `widget::Action`, which can be leveraged to either publish a message or request a redraw, and also event capturing.
- Custom `Event` types for `canvas` and `shader` have been removed. They now simply use the foundational `core::Event` type.


felixfeixu:
I see this PR from the latest 0.14 release, but may I see a **document** about the feature? this is a quite big feature.

PR #2663, opened 2024-11-05
Update `wgpu` to `23.0`

hecrj:
See https://github.com/gfx-rs/wgpu/releases/tag/v23.0.0.


hecrj:
A small ~3% regression in our `layered text (heavy)` benchmark:

image

Nothing to be worried about, I believe.

PR #2664, opened 2024-11-05
Implement find_focus() for text_input

Leonie-Theobald:
This PR adds support for `find_focused()` for `text_input`.
Basically, it is a wrapper function that allows the user to indirectly make use of `find_focused()`within the widget module.

I also added a `From<widget::Id> for Id` for text_input so that the returned widget::Id can be converted to the text_input ID that the enduser can use.

This topic came up in a discussion in Discourse: [here](https://discourse.iced.rs/t/use-focus-and-find-focused-with-text-input/671/4)

Find an example that uses `find_focused()` below:
```Rust
use iced::widget::{column, text_input, Column};
use iced::{Subscription, Task};

use iced::keyboard;
use keyboard::key::Named::ArrowUp;

fn main() -> iced::Result {
    iced::application("Example For find_focused()", App::update, App::view)
        .subscription(App::subscription)
        .run()
}

#[derive(Default)]
struct App {
    list: [Item; 5],
}

struct Item {
    id: text_input::Id,
    content: String,
}

impl Default for Item {
    fn default() -> Self {
        Item {
            id: text_input::Id::unique(),
            content: String::new(),
        }
    }
}

#[derive(Clone, Debug)]
enum Message {
    ContentUpdated(usize, String),
    SwapFocusedItemWithTop,
    SwapItemWithTop(text_input::Id),
}

impl App {
    fn update(&mut self, message: Message) -> Task<Message> {
        match message {
            Message::ContentUpdated(position, content) => {
                self.list[position].content = content;
            }
            Message::SwapFocusedItemWithTop => {
                return text_input::find_focused()
                    .map::<Message>(|id| Message::SwapItemWithTop(id.into()));
            }
            Message::SwapItemWithTop(id) => {
                if let Some(position) = self.list.iter().position(|item| item.id == id) {
                    self.list.swap(0, position);
                };
            }
        }

        Task::none()
    }

    fn view(&self) -> Column<Message> {
        column(self.list.iter().enumerate().map(|(position, item)| {
            text_input("", &item.content)
                .on_input(move |new_content| Message::ContentUpdated(position, new_content))
                .id(item.id.clone())
                .into()
        }))
        .spacing(10)
    }

    fn subscription(&self) -> Subscription<Message> {
        keyboard::on_key_press(|key, _modifiers| match key {
            keyboard::Key::Named(ArrowUp) => Some(Message::SwapFocusedItemWithTop),
            _ => None,
        })
    }
}
```


hecrj (approved):
Thanks!

I have created an `is_focused` selector and unified the `selector` API to simply have two operations: `find` and `find_all`.

Leonie-Theobald:
Didn't think that the PR would be relevant anymore^^
But nice that it made its way into the code base :)

PR #2672, opened 2024-11-22
Remove `surface` argument of `Compositor::screenshot`

ids1024:
This argument was completely ignored by the wgpu renderer, and used only for the `clip_mask` by the `tiny_skia` renderer. I believe creating a new clip mask is correct.

This way it's possible to render offscreen without needing a surface.


hecrj (approved):
Sweet! Thanks :tada:

PR #2673, opened 2024-11-22
`pin` widget

hecrj:
This PR introduces a new widget: `pin`.

A `pin` widget can be used to position a widget at some fixed coordinates inside its bounds. For instance:

```rust
container(pin("• (50, 50)").x(50).y(50))
    .width(500)
    .height(500)
    .style(container::bordered_box)
```

### DIFF
--- a/widget/src/helpers.rs
+++ b/widget/src/helpers.rs
@@ -267,8 +267,6 @@ where
 ///
 /// fn view(state: &State) -> Element<'_, Message> {
 ///     pin("This text is displayed at coordinates (50, 50)!")
-///         .width(Fill)
-///         .height(Fill)
 ///         .x(50)
 ///         .y(50)
 ///         .into()
--- a/widget/src/helpers.rs
+++ b/widget/src/helpers.rs
@@ -24,7 +24,7 @@ use crate::text_input::{self, TextInput};
 use crate::toggler::{self, Toggler};
 use crate::tooltip::{self, Tooltip};
 use crate::vertical_slider::{self, VerticalSlider};
-use crate::{Column, MouseArea, Row, Space, Stack, Themer};
+use crate::{Column, MouseArea, Pin, Row, Space, Stack, Themer};
 
 use std::borrow::Borrow;
 use std::ops::RangeInclusive;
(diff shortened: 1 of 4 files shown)


widget/src/helpers.rs
```diff
@@ -249,6 +249,38 @@ where
     container(content).center(Length::Fill)
 }
 
+/// Creates a new [`Pin`] widget with the given content.
+///
+/// A [`Pin`] widget positions its contents at some fixed coordinates inside of its boundaries.
```
Mingun, line 254: What you mean "inside of its boundaries"? Container will be moves and its boundaries as well, so coordinates also changed. Do you mean boundaries of the parent container?

hecrj: No, that isn't reasonable. Everything in `iced` works in relative coordinates. This is no exception.

airstrike: You position some child content within the bounds of the `pin` widget. In the provided example, a `Text` widget (really an Element) is placed at coordinates `(50, 50)` within the bounds of the container-like `pin` widget which has width and height of `500`.

Mingun: I think the misunderstanding was because word "its" can refer to different things depending on the interpretation. I read it as
```
A [`Pin`] widget positions its contents ... inside of its boundaries.
                                  ^--------------------'
```
while the author assumed
```
A [`Pin`] widget positions its contents ... inside of its boundaries.
             ^-----------------------------------------'
```
I'm not sure if this kind of misinterpretation is possible for native speakers, maybe from their point of view this is pretty straightforward sentence. I just like to highlight that people with other background may read it incorrectly.

airstrike: I'm not a native speaker either, but I think your original interpretation would only be appropriate for the sentence "A `Pin` widget positions its contents ... inside of _their_ boundaries."

Either way the example (and now this discussion) should clarify any future confusion.

hecrj: Since only one of the interpretations makes absolutely no sense whatsoever... I believe reasonable people will fall back to the sensical one.

PR #2679, opened 2024-11-28
improve argument names in `Widget` trait

edwloef:
This removes the underscores in front of the argument names in provided `Widget` methods, in order to make an IDE-autofilled method signature more useful. This was done in a similar manner to https://github.com/Smithay/smithay/blob/bc1d7320f95cdf17f9e7aa6867cccc5903548032/src/xwayland/xwm/mod.rs#L330-L356

It also renames all instances where the `Tree` argument was named `state`, as `Tree` has a `state` field. This has led to confusion on my part a few times, with the IDE-autofilled method signature clashing with my own naming, in which case I assumed `state` was the widget's (downcasted) state, not the widget's tree. I can't think of any reasoning as to why it was named `tree` vs `state` in different methods, so it makes sense to me to make it consistent everywhere.

### DIFF
--- a/core/src/widget.rs
+++ b/core/src/widget.rs
@@ -101,7 +101,7 @@ where
     /// Applies an [`Operation`] to the [`Widget`].
     fn operate(
         &self,
-        _state: &mut Tree,
+        _tree: &mut Tree,
         _layout: Layout<'_>,
         _renderer: &Renderer,
         _operation: &mut dyn Operation,
@@ -113,7 +113,7 @@ where
     /// By default, it does nothing.
     fn update(
         &mut self,
-        _state: &mut Tree,
+        _tree: &mut Tree,
         _event: Event,
         _layout: Layout<'_>,
         _cursor: mouse::Cursor,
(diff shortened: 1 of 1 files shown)


airstrike:
Assuming this change is desirable, would it not be better to favor `#[allow(unused_variables)]` over `let _ = ...`?

edwloef:
I don't really mind either way, I just copied how Smithay dealt with this issue. If `#allow`'s are preferred I can take care of it as well.

hecrj:
You will rarely need to use all the arguments, so the usefulness is quite subjective.

I prefer to have to opt-in into arguments rather than opt-out.

In any case, this seems like we are adding complexity to the code for better IDE support. Shouldn't the issue be directly addressed in the LSP layer? Maybe a setting in `rust-analyzer`?

edwloef:
> You will rarely need to use all the arguments, so the usefulness is quite subjective.

You do always need all of them if your widget has child widgets, which to be fair I'm not sure how common that is.

If that's not such a common scenario it would make sense to not remove the underscores, since this is just a small annoyance I noticed while working on a few widgets that had each other as children.

edwloef:
FWIW I've reverted the change with the underscores, only leaving the argument name changes, since that's the one that's caused me more confusion anyways.

edwloef:
I've reopened this to only include the fix for #2949

PR #2681, opened 2024-12-01
fix: add an initial state to the `system_information` example

xosxos:
**FYI:** The `system_information` example was eternally stuck at the loading screen (tested on macOS).

Fix on my system:
Create an initial state of `(Example::Loading, Task<Message::Refresh>)` using `run_with` on the `Application` builder.


hecrj (approved):
Thanks!

PR #2685, opened 2024-12-05
Mouse interaction to hide system cursor

thorn132:
Adds support for hiding the system cursor via `mouse::Interaction`.
- Adds `mouse::Interaction::Hidden` variant
- Returns `Option<CursorIcon>` from `conversion::mouse_interaction`, where `None` denotes the cursor should be hidden
- Calls `set_cursor_visible` only when the interaction changes to/from `mouse::Interaction::Hidden`


hecrj (approved):
Thanks!

PR #2686, opened 2024-12-05
Fixed rendering in chromium

Fili-pk:
I fixed issue of iced not rendering on chromium based browser. Previously chromium just showed error `Failed to create WebGPU Context Provider` even if webgl feature was enabled. Also the issue was similar to this wgpu issue https://github.com/gfx-rs/wgpu/issues/6490#issuecomment-2453016187


hecrj (approved):
Thanks!

PR #2688, opened 2024-12-06
Use "outer" positions in all window-related operations

hecrj:
See https://docs.rs/winit/latest/winit/window/struct.WindowAttributes.html#method.with_position.

I think this is the best we can do until https://github.com/rust-windowing/winit/issues/2645 is addressed.

### DIFF
--- a/core/src/window/event.rs
+++ b/core/src/window/event.rs
@@ -9,8 +9,8 @@ pub enum Event {
     /// A window was opened.
     Opened {
         /// The position of the opened window. This is relative to the top-left corner of the desktop
-        /// the window is on, including virtual desktops. Refers to window's "inner" position,
-        /// or the client area, in logical pixels.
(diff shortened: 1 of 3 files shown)


SoulSharer:
Thanks for the fix! We also need to correct comments for window events:

`core\src\window\event.rs`:
```rust
  /// A window was opened.
    Opened {
        /// The position of the opened window. This is relative to the top-left corner of the desktop
        /// the window is on, including virtual desktops. Refers to window's "outer" position,
        /// or the window area, in logical pixels.
        ///
        /// **Note**: Not available in Wayland.
        position: Option<Point>,
        /// The size of the created window. This is its "inner" size, or the size of the
        /// client area, in logical pixels.
        size: Size,
    },
```
Now I'm also wondering whether we should provide "outer" size for `Opened` event.

PR #2692, opened 2024-12-08
Make `iced_wgpu::Renderer` `Send` by using `Arc` in text/triangle caches

Azorlogh:
The [iced integration for bevy](https://github.com/tasgon/bevy_iced) requires the `iced_wgpu::Renderer` to be Send. It currently isn't due to the text & triangle caches relying on Rc.

This PR changes them to use `Arc`s instead.


hecrj (approved):
Thanks!

PR #2693, opened 2024-12-08
Allow the user of `iced_wgpu::Engine` to submit the queue themself

Azorlogh:
The [iced integration for bevy](https://github.com/tasgon/bevy_iced/pull/33) needs to let Bevy take care of submitting the queue, so it can't call `Engine::submit`. But it still needs to call `finish/recall` on the staging belt and `end_frame` on all the pipelines.
This adds two additional methods to do that.

Note: The naming might be a bit confusing, I couldn't think of a better one :thinking:

### DIFF
--- a/wgpu/src/engine.rs
+++ b/wgpu/src/engine.rs
@@ -71,8 +71,18 @@ impl Engine {
         queue: &wgpu::Queue,
         encoder: wgpu::CommandEncoder,
     ) -> wgpu::SubmissionIndex {
-        self.staging_belt.finish();
+        self.finish();
         let index = queue.submit(Some(encoder.finish()));
+        self.end_frame();
+
+        index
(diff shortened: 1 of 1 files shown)


srussvoll:
Bevy 0.15 support has been completed in tasgon/bevy_iced#34, and is blocked by this PR. What is the status of this PR? Are there any changes needed to get this through?

hecrj:
Why does `present` not work exactly? Why does Bevy need to control all `Queue` submissions?

Azorlogh:
@hecrj In order to integrate to bevy's render graph, we need to append our commands to the `Queue` it provides us. It only submit the queue after the whole graph finished running.  
If iced submits the work to its own queue, then everything else will be drawn on top of its UI

PR #2695, opened 2024-12-11
Replace `format!` with `concat!` for string literals

Integral-Tech:
- Although `format!` macro is very powerful, it introduces overhead of memory allocation on the heap compared to `&str`. Therefore, when `format!` macro is unnecessary, we should avoid using it.

### DIFF
--- a/examples/editor/src/main.rs
+++ b/examples/editor/src/main.rs
@@ -53,9 +53,9 @@ impl Editor {
             },
             Task::batch([
                 Task::perform(
-                    load_file(format!(
-                        "{}/src/main.rs",
-                        env!("CARGO_MANIFEST_DIR")
+                    load_file(concat!(
+                        env!("CARGO_MANIFEST_DIR"),
+                        "/src/main.rs",
                     )),
                     Message::FileOpened,
                 ),
--- a/examples/ferris/src/main.rs
+++ b/examples/ferris/src/main.rs
@@ -95,9 +95,9 @@ impl Image {
         let i_am_ferris = column![
             "Hello!",
             Element::from(
-                image(format!(
-                    "{}/../tour/images/ferris.png",
-                    env!("CARGO_MANIFEST_DIR")
+                image(concat!(
+                    env!("CARGO_MANIFEST_DIR"),
+                    "/../tour/images/ferris.png",
                 ))
                 .width(self.width)
                 .content_fit(self.content_fit)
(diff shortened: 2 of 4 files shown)


ricvelozo:
I think the format macro already optimizes static strings (like the `env!` ones).

See: https://blog.m-ou.se/format-args/#small-steps

runtime/src/debug/basic.rs
ricvelozo, line 1: You are overcomplicating things here and still allocating with the ToString trait.

ricvelozo (changes requested):

Integral-Tech:
> I think the format macro already optimizes static strings (like the `env!` ones).
> 
> See: https://blog.m-ou.se/format-args/#small-steps

They are different. `concat!()` constructs the `&str' during compilation, which is still faster.

ricvelozo:
> > I think the format macro already optimizes static strings (like the `env!` ones).
> > 
> > See: https://blog.m-ou.se/format-args/#small-steps
> 
> They are different. `concat!()` constructs the `&str` during compilation, which is still faster.

Read the link I wrote.

Integral-Tech:
@ricvelozo I have read your link, it says `format_args()!` can inline static strings.
However, `format!()` is equivalent to `fmt::format(format_args!())`. The function `fmt::format` still involves heap allocation.

ricvelozo:
> @ricvelozo I have read your link, it says `format_args()!` can inline static strings.
> However, `format!()` is equivalent to `fmt::format(format_args!())`. The function `fmt::format` still involves heap allocation.
> If you don't like using `concat!()`, `format_args!().as_str()` can be used instead to avoid heap allocation.

Like I said in the review, you are using the ToString trait, and that allocates.

Integral-Tech:
@ricvelozo I have just removed the change involves `.to_string()` call :)

ricvelozo:
> @ricvelozo I have just removed the change involves `.to_string()` call :)

Nice! 👍🏻

Integral-Tech:
@ricvelozo Could you please approve the workflow run?

ricvelozo (approved):

ricvelozo (approved):

ricvelozo:
> @ricvelozo Could you please approve the workflow run?

Needs Héctor's approval. Eventually he will review.

Koranir:
All these usecases convert the string into a `PathBuf` anyways, which _does_ allocate and copy static string slice data. PathBufs _don't_ allocate if created from a `String`, so this PR doesn't really change anything to do with allocation amounts.

PR #2697, opened 2024-12-16
Fix for a `tiny-skia` text rendering (float imprecision)

SoulSharer:
This commit fixes iced-rs/iced/issues/2665.


T-256:
it seems the issue has been closed.

PR #2698, opened 2024-12-17
Headless Mode Testing

hecrj:
This PR introduces a brand new crate for headless mode testing: `iced_test`.

This crate features a `Simulator` type that can be used to simulate user interactions and query any `Element` produced by `iced`.

# Basic Usage
Let's assume we want to test [the classical counter interface].

First, we will want to create a `Simulator` of our interface:

```rust,no_run
use iced_test::simulator;

let mut counter = Counter { value: 0 };
let mut ui = simulator(counter.view());
```

Now we can simulate a user interacting with our interface. Let's use `Simulator::click` to click the counter buttons:

```rust
use iced_test::selector::text;

let _ = ui.click(text("+"));
let _ = ui.click(text("+"));
let _ = ui.click(text("-"));
```

`Simulator::click` takes a `Selector`. A `Selector` describes a way to query the widgets of an interface. In this case, `selector::text` lets us select a widget by the text it contains.

We can now process any messages produced by these interactions and then assert that the final value of our counter is indeed `1`!

```rust,no_run
for message in ui.into_messages() {
    counter.update(message);
}

assert_eq!(counter.value, 1);
```

We can even rebuild the interface to make sure the counter _displays_ the proper value with `Simulator::find`:

```rust
let mut ui = simulator(counter.view());

assert!(ui.find(text("1")).is_ok(), "Counter should display 1!");
```

And that's it! That's the gist of testing `iced` applications!

`Simulator` contains additional operations you can use to simulate more interactions—like `tap_key` or `typewrite`—and even perform _snapshot testing_!

[the classical counter interface]: https://book.iced.rs/architecture.html#dissecting-an-interface


GyulyVGC:
This is a remarkable milestone for Iced.

GG Héctor 🙌

andrewdavidmackenzie:
WDYT about me taking some of the above text and examples and adding that to the repo as a README.md?

Do you see a future in this repo...I wanted to try it for some headless use with existing tests (that open a UI), then move on to proper UI testing...

andrewdavidmackenzie:
Also, a publish of current iced_test to crates.io would be nice:
https://crates.io/crates/iced_test is 0.0.0 not 0.1.0

PR #2699, opened 2024-12-17
Add `menu_height` method to `PickList` and `ComboBox` widgets.

GyulyVGC:
Add `menu_height` method to `PickList` and `ComboBox` widgets.

This is particularly useful to reduce the menu height in presence of a long list of options.

***

<div align="center">

_Before_

</div>

***

<div align="center">

_After_

</div>


emkkla:
Shouldn't we call it menu_length instead of height ?

GyulyVGC:
> Shouldn't we call it menu_length instead of height ?

I don't think so... iced always refers to vertical lengths as "height" and to horizontal lengths as "width".

hecrj (approved):
Thanks!

PR #2701, opened 2024-12-19
always increment quad, mesh, text and image layer counts in wgpu layer rendering

### LINKED ISSUE #2700: Layer rendering breaks if one is not drawn at any point

### What happened?

If you have a layer that is not drawn either because it is fully outside of the application window, or because its snapped bounds have an area of zero, this breaks subsequent layer draws. This is because at https://github.com/iced-rs/iced/blob/f2c9b6b2ffc50d67d9789e77cb55eeb2a0ebe470/wgpu/src/lib.rs#L271-L275 and https://github.com/iced-rs/iced/blob/f2c9b6b2ffc50d67d9789e77cb55eeb2a0ebe470/wgpu/src/lib.rs#L277-L279 rendering for those layers is skipped without updating the `quad_layer`, `mesh_layer` and `text_layer` indices.

Here is a video of this occuring in a real iced app:

https://github.com/user-attachments/assets/f0c0a6f2-c44f-4dea-88dc-ea8a0c4a9f31

and the relevant lines in the source code of this app: https://github.com/generic-daw/generic-daw/blob/b5593615b46625cfde0f194d16582f15cc672d7a/src/generic_front/widget/track.rs#L56-L59

Layer rendering doesn't break when a layer is not visible.

### Version

crates.io release

### Operating System

Linux

_No response_

edwloef:
Fixes #2700

### DIFF
--- a/wgpu/src/lib.rs
+++ b/wgpu/src/lib.rs
@@ -268,13 +268,37 @@ impl Renderer {
         let scale = Transformation::scale(scale_factor);
 
         for layer in self.layers.iter() {
-            let Some(physical_bounds) =
-                physical_bounds.intersection(&(layer.bounds * scale))
+            let Some(scissor_rect) = physical_bounds
+                .intersection(&(layer.bounds * scale))
+                .and_then(Rectangle::snap)
             else {
-                continue;
-            };
+                if !layer.quads.is_empty() {
+                    quad_layer += 1;
+                }
 
-            let Some(scissor_rect) = physical_bounds.snap() else {
+                if !layer.triangles.is_empty() {
+                    mesh_layer += layer
+                        .triangles
+                        .iter()
+                        .filter(|item| {
+                            matches!(item, triangle::Item::Group { .. })
+                        })
+                        .count();
+                }
+
+                if !layer.text.is_empty() {
+                    text_layer += layer
+                        .text
+                        .iter()
+                        .filter(|item| matches!(item, text::Item::Group { .. }))
+                        .count();
+                }
+
+                #[cfg(any(feature = "svg", feature = "image"))]
+                if !layer.images.is_empty() {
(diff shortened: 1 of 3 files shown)


hecrj (approved):
I just hit this bug. Pesky one!

Thanks :bow:

hecrj:
In fact, I believe we can just avoid preparing these layers altogether as a simpler fix:

https://github.com/iced-rs/iced/commit/3428a3d2afb766c264453a58e15d33953d438238

flakes:
@hecrj These changes apparently broke the `start_layer` clipping when using `draw_geometry` in a custom widget. The clipping still works for text, but not rectangles:

Screenshot 2025-02-14 at 22 34 22

Notice how the text is clipped to the viewport of the custom widget, but the rectangles are drawn outside the widget, on top of the button.


Latest master with changes from this PR reverted:

Screenshot 2025-02-14 at 22 33 20


The widget code is here: https://github.com/syndicodefront/infekt/blob/3f15caa6fdc8aa381422a5a28340984de589fbb4/infekt-v2/src/gui/widget/enhanced_nfo_view.rs#L231

Am I abusing a bug here as a feature or is it a regression?

edwloef:
It seems like some of the changes made here were later refactored in https://github.com/iced-rs/iced/commit/3428a3d2afb766c264453a58e15d33953d438238, could that commit also be affecting things?

edwloef:
The replacement of `scale` with `scale_factor` in the refactoring commit looks suspicious to me. Maybe you could try switching that out and see if it fixes your issue?

hecrj:
@flakes Fixed in 7c54bdef1062e4008582d8b1e215ffcc1c1ee016.

flakes:
@hecrj confirming it works, much appreciated!

PR #2707, opened 2024-12-27
Add `text_color` to toggler `Style`

xosxos:
Hey, looks like it's not possible to set `text color` for the `Toggler` widget. So, I copied the implementation over from `Checkbox`, which included `text_color` in its `Style`. This solved my issue with setting the color, and maybe this change is something to consider. Cheers


hecrj (approved):
Thanks!

PR #2714, opened 2025-01-01
Add text shaping setter on ComboBox for menu

mytdragon:
Hi, unlike PickList, ComboBox didn't have a setter for text_shaping so I added it and adapted the combo_box example with languages that would use Advanced text shaping.


hecrj (approved):
Thanks!

Although this shouldn't be as necessary since `iced` auto-selects the best shaping strategy by default now.

Issue #2339, opened 2024-03-19
Container shadows are wrong size with `tiny-skia` and custom theme


mtkennerly:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

I have a custom theme where the `Container` appearance includes:

```
            shadow: Shadow {
                color: Color::BLACK,
                offset: Vector::ZERO,
                blur_radius: 0.0,
            }
```

With `wgpu`, it looks fine (no visible shadow). With `tiny-skia`, it looks incorrect (black fills whole container).

I can work around it by using `Color::TRANSPARENT`.

Gist (using red shadow for illustration purposes): https://gist.github.com/mtkennerly/993c3fbdd7ea770c0c28d268cf063664

### `ICED_BACKEND=wgpu cargo run`
> ![Screenshot 2024-03-18 234327](https://github.com/iced-rs/iced/assets/4934192/98f4492e-5cdb-4436-af1b-5361f705423f)

### `ICED_BACKEND=tiny-skia cargo run`
> ![Screenshot 2024-03-18 234342](https://github.com/iced-rs/iced/assets/4934192/c4e3d979-7fca-4d47-97b0-1fa956710ff0)

### What is the expected behavior?

The `tiny-skia` example should look the same as the `wgpu` example.

### Version

crates.io release

### Operating System

Windows

njust:
Hi, I have the same Issue, but wanted to add, that for buttons shadows work fine also with the tiny-skia backend:

Wgpu backend
![image](https://github.com/iced-rs/iced/assets/2380253/62bc4bf8-9779-4ab4-a8d2-63931eb3c0e6)

Tiny-skia backend
![image](https://github.com/iced-rs/iced/assets/2380253/436f7fcb-a57a-4d93-9e7e-b8957706272c)

What do you mean by "can work around it by using Color::TRANSPARENT."? Using Color::TRANSPARENT for the shadow effectively means no visible shadow. So the workaround for the tiny-skia backend is not using shadows at all?

mtkennerly:
> So the workaround for the tiny-skia backend is not using shadows at all?

My use case is without any shadows, but yeah, that doesn't help if you actually want them 😅

SzilvasiPeter:
[...]

PR #2715, opened 2025-01-02
fix shadow rendering for tiny-skia backend

### LINKED ISSUE #2712: [tiny-skia] Changes affected to `shadows` needs to have full redraw

### What happened?

By using `tiny-skia` backend, try to apply a change to a shadow (e.g. hover mouse over button with shadow), it seems it renders incorrectly.
You can also run `custom_quad` from examples folder with `set ICED_BACKEND=tiny-skia`.

Note that if it would solve by redrawing whole window. try resizing window to trigger full redraw event.

It should be as same as `wgpu`.

### Version

master

### Operating System

Windows

_No response_

T-256:
Fixes #2712 

After commits bisecting, I found the issued came out from #2382 (commit 1e802e776cb591f3860d1bfbaa1423d356fc8456).
This PR adds `bounds_with_shadow` to `Quad`.

I verified it by running `cargo r -p custom_quad`.

### TODO
- [x] https://github.com/iced-rs/iced/issues/2339#issuecomment-2566356141
- [x] In `custom_quad` example, after moving shadow down, hovering on sliders shouldn't destroy shadow.
- [x] Prevent shadows beyond scrollable area.


B0ney:
I don't think this is an optimal fix. Since now the software renderer will redraw everything every update, leading to much higher CPU usage.

B0ney:
Looking at the issues you linked, I think it might be that the damage regions aren't taking the shadow's blur radius + offset into consideration.

This means that as the damaged areas are redrawn, because the regions don't include the shadows, they won't get cleared properly. 

Consequently, the renderer will draw over them (additively) after every update. This repeated stacking might explain why the shadows get darker.

dtzxporter:
Looking at this, it may make sense to abstract the bounds calculation to "layout bounds" and "render bounds" where one deals with collision/layout, and the other is the bounds that will be modified when rendering, for dirty tracking, etc.

T-256:
@hecrj I tested with your latest commit https://github.com/iced-rs/iced/commit/669ab2bf30576215a5aebcb2e0eb2199c22828b6 but issues still ther:


https://github.com/user-attachments/assets/af784e4f-1a73-4327-bd2b-b9349e938041

T-256:
OK, using `.abs()` on your commit solved first issue (clipping glitch):
```diff
--- a/tiny_skia/src/layer.rs
+++ b/tiny_skia/src/layer.rs
@@ -218,7 +218,7 @@ impl Layer {

                 vec![if quad.shadow.color.a > 0.0 {
                     bounds.expand(
-                        quad.shadow.offset.x.max(quad.shadow.offset.y) + quad.shadow.blur_radius,
+                        quad.shadow.offset.x.abs().max(quad.shadow.offset.y.abs()) + quad.shadow.blur_radius,
                     )
                 } else {
```

the second issue (shadow destruction when hover over scrollbars) still exists.

T-256:
@hecrj Thanks for your fix commit from the previous comment.

OK, now the second issue is also solved by this patch -- same approach in this PR but with math from previous comment:
```diff
--- a/tiny_skia/src/engine.rs
+++ b/tiny_skia/src/engine.rs
@@ -34,13 +34,20 @@ impl Engine {
         clip_mask: &mut tiny_skia::Mask,
         clip_bounds: Rectangle,
     ) {
-        let physical_bounds = quad.bounds * transformation;
-
-        if !clip_bounds.intersects(&physical_bounds) {
+        let local_clip_bounds = if quad.shadow.color.a > 0.0 {
+            quad.bounds.expand(
+                quad.shadow.offset.x.abs().max(quad.shadow.offset.y.abs())
+                    + quad.shadow.blur_radius,
+            )
+        } else {
+            quad.bounds
+        } * transformation;
+
+        if !clip_bounds.intersects(&local_clip_bounds) {
             return;
         }

-        let clip_mask = (!physical_bounds.is_within(&clip_bounds)).then_some(clip_mask as &_);
+        let clip_mask = (!local_clip_bounds.is_within(&clip_bounds)).then_some(clip_mask as &_);

         let transform = into_transform(transformation);

@@ -81,6 +88,9 @@ impl Engine {
                 shadow_bounds.width as u32,
                 shadow_bounds.height as u32,
             );
+
+            let physical_bounds = quad.bounds * transformation;
+
             let half_width = physical_bounds.width / 2.0;
             let half_height = physical_bounds.height / 2.0;

@@ -125,7 +135,7 @@ impl Engine {
                     pixmap.as_ref(),
                     &tiny_skia::PixmapPaint::default(),
                     tiny_skia::Transform::default(),
-                    None,
+                    clip_mask,
                 );
             }
         }
```

PR #2717, opened 2025-01-03
Allow maximizing window when app starts

rhysd:
This PR adds the capability to start an application window in maximized state.

[`winit::window::WindowAttributes`](https://docs.rs/winit/latest/winit/window/struct.WindowAttributes.html#structfield.maximized) provides the attribute to maximize a window when creating it. However iced doesn't expose the functionality. This PR just exposes it by adding the same attribute to `iced::window::Settings`.

### DIFF
--- a/core/src/window/settings.rs
+++ b/core/src/window/settings.rs
@@ -44,6 +44,9 @@ pub struct Settings {
     /// The maximum size of the window.
     pub max_size: Option<Size>,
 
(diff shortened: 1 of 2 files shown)


hecrj:
This has been already implemented by #2627, which was just merged.

rhysd:
Great. Thanks!

Issue #1002, opened 2021-08-19
rfd for file dialogs


sourcebox:
@PolyMeilex I saw that you're a frequent contributor here. Is there a chance that your [rfd](https://github.com/PolyMeilex/rfd) crate can be used for file dialogs?

yusdacra:
Yes it can be. I use `rfd` in my iced app https://github.com/harmony-development/Crust/blob/master/src/screen/mod.rs#L1336.

PolyMeilex:
If you are asking if it can be used at the moment, then sure, it can easily be done.

If you are asking about tighter integration, for example proper setting of parent window, that would probably require integration inside iced, because we need window handle for that. I'm open to working on something like that, if there is enough interest from users and from @hecrj himself

sourcebox:
I think file dialogs were requested several times in the past. To me, `rfd` is the best existing crate for this I'm aware of. This is because it's pure Rust and so allows cross-compilation (e.g. for ARM Macs) in an easy way.

hecrj:
`rfd` looks great!

We could expose a `file-dialog` feature to expose additional actions related to file dialogs once the new `Command` implementation lands (see #1019).

Since the new `Command` actions are processed in the event loop, at the shell level, it should be easy to obtain the window handle for proper integration, without actually exposing it in `iced`.

PolyMeilex:
#1019 sounds great 👍

sourcebox:
@PolyMeilex Since winit 0.26 was integrated to `egui`, the dialogs work fine there even on macOS (they were freezing before). What is the situation regarding this here with `iced`? Do I have to use the async versions to make them work? Would it be better to wait until `winit` is also updated?

PolyMeilex:
> @PolyMeilex Since winit 0.26 was integrated to `egui`, the dialogs work fine there even on macOS (they were freezing before). What is the situation regarding this here with `iced`? Do I have to use the async versions to make them work? Would it be better to wait until `winit` is also updated?

Good to know that my fix was published already 🎉
But winit has its own fork of winit, so it has to be rebased. The update of winit is also blocking my wgpu PR  #1164

Alch-Emi:
Working on designing an API for this [in my fork](https://github.com/Alch-Emi/iced/tree/rfd).  It might turn into a PR once I'm a little happier with the shape of it

jpercyasnet:
[...]

PR #2718, opened 2025-01-04
Add `iced::window::run_with_window_handles` for usage e.g. with rfd 0.15+

flakes:
Based on

- several discussions on Discord and #1002
- the latest update to rfd which took away the auto-parent selection on MacOS

I decided to give this a chance. There's a few rough edges, specifically:

- naming and placement of the new structure `WindowHandles`
- there's no way to start another task from within the closure, so we can't use `async` code without another intermediate `Message`
- usage of `unwrap()`
- removal of `run_with_handle` (?)

All comments welcome :-)


nik-rev:
Neat PR! I am using it. I ran into a tricky situation and the way to solve it wasn't very straightforward. I'll post this here in case anyone has the same problem

Essentially, when you try to use `AsyncFileDialog::save_file` (which returns a `Future`) inside of an `iced::window::run_with_window_handles` (from this PR), you have something like this:

```rs
Message::OpenDialog => return iced::window::run_with_window_handles(window_id, move |window_handle| {
    Task::future(async move {
        let dialog = AsyncFileDialog::new().set_parent(window_handle).save_file().await;
        /* ... */
    })
})
```

This is supposed to return `Task<Message>`. Unfortunately, it actually returns `Task<Task<Message>>`
The annoying part is that `window_handle` has a lifetime, we can't store it anywhere. It doesn't implement something like `Clone` either.

Instead what I did is split this into two parts: a `Message::OpenDialogPart1` and a `Message::OpenDialogPart2`.

```rs
Message::OpenDialogPart1 => {
  return iced::window::run_with_window_handles(window_id, move |window_handle| {
    Message::OpenDialogPart2(AsyncFileDialog::new().set_parent(window_handle))
})
}
Message::OpenDialogPart2(file_dialog) => {
    Task::future(async move { file_dialog.save_file().await; /* ... */ })
},
```

This way, the firstbranch returns a `Task<Message>` where `Message = Message::OpenDialogPart2` and the 2nd task is the one that actually does the async work.

hecrj (approved):
Thanks!

I introduced a `Window` trait and changed `run_with_handle` to simply `run`.

Issue #2713, opened 2025-01-01
`image::viewer::Viewer` zoom in/out with mouse scrolling doesn't work


rhysd:
### Is your issue REALLY a bug?

- [X] My issue is indeed a bug!
- [X] I am not crazy! I will not fill out this form just to ask a question or request a feature. Pinky promise.

### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

`iced::widget::image::viewer::Viewer` provides the functionality to zoom in/out by scrolling up/down. However it has stopped working from the latest of `master` branch.

Here is the reproduction:

```rust
use iced::widget::image::viewer::Viewer;
use iced::widget::image::Handle;
use iced::widget::{container, Container};
use iced::Length::Fill;
use std::env;

struct App(Handle);

impl Default for App {
    fn default() -> Self {
        Self(Handle::from_path(env::args_os().nth(1).unwrap()))
    }
}

impl App {
    fn update(&mut self, _: ()) {}

    fn view(&self) -> Container<()> {
        container(Viewer::new(&self.0).width(Fill).height(Fill)).center(Fill)
    }
}

fn main() -> iced::Result {
    iced::run("Test", App::update, App::view)
}
```

Run this program with passing an image file path to the first command line argument. Then try to zoom in/out the image in the window by scrolling up/down.

This example works fine with iced v0.13.1, however it doesn't work on the latest master branch (e4d4c0849e7d4b1447b77869f82a981084e41083).

### What is the expected behavior?

Scrolling up/down on `Viewer` widget makes the image zoom in/out same as v0.13.1.

### Version

master

### Operating System

macOS

PR #2726, opened 2025-01-11
Fix image viewer not zooming

leo030303:
Fix for issue #2713 
The image viewer update method wasn't requesting redraws, it now requests a redraw every time an event is captured.
I've tested this with the example app from the linked issue and it fixes the issue.


edwloef:
The `ButtonPressed` and `ButtonReleased` events probably don't need to request a redraw, since they don't change what needs to be drawn as far as I can tell.

leo030303:
I tried not redrawing on those events and it all renders the same except the cursor icon doesn't change to the "grabbed" version when clicking, or it gets stuck as the grabbed version, so thought it was worthwhile leaving them in, see the videos for what I mean, I'm on Wayland with Gnome 43

**No Redraw**
no_redraw

**With Redraw**
redraw

edwloef:
Ah right, oops.

rhysd:
I confirmed this patch fixed #2713.

leo030303:
This was fixed by https://github.com/iced-rs/iced/pull/2795 so closing this

PR #2728, opened 2025-01-13
Add option to make the transparent window blurry

rhysd:
This PR adds `blur` option to the `iced::window::Settings` struct. When enabling it, the window has a blurry effect. The option value will be passed to [winit](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_blur).

Here is a screenshot of example:

image

```rust
use iced::theme::Style;
use iced::widget::{container, text};
use iced::window::Settings;
use iced::{Color, Element, Length::Fill, Theme};

    iced::application("Test", Test::update, Test::view)
        .window(Settings {
            transparent: true,
            blur: true,
            ..Default::default()
        })
        .style(Test::style)
        .run()
}

#[derive(Default)]
struct Test;

impl Test {

    fn view(&self) -> impl Into<Element<()>> {
        container(text("This is test")).center(Fill)
    }

    fn style(&self, theme: &Theme) -> Style {
        Style {
            background_color: Color::from_rgba(0.0, 0.0, 0.0, 0.3),
            text_color: theme.palette().text,
        }
    }
}
```


rhysd:
My use case for this feature is creating a visual effect for macOS apps. It is common for the side menu of macOS apps to have a blurry appearance. For example, Finder (the default file manager in macOS) has a similar effect, as shown below. I would like to achieve the same effect for my macOS app to provide a natural look and feel.

image

PR #2731, opened 2025-01-16
Remove a cast

will-lynas:
I'm not sure what the point of this cast is. If I remove it then the code still compiles and runs fine.


hecrj (approved):
Thanks!

PR #2732, opened 2025-01-17
implement shrinkable scrollbar

JL710:
This PR adds the ability to let the scrollbar shrink if the scrollbar is neither hovered nor dragged. This behavior only affects the drawing and not the layout or anything else. So the hover area is not affected.

By default, this is disabled but can be enabled with the `shrinking_scrollbar` function. The default behavior is defined in the `scrollbar_shrink_factor` field of the `Scrollable`.

An advantage of a shrinking scrollbar is, beside subjective design, that the scrollbar hides less content.

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -81,6 +81,7 @@ pub struct Scrollable<
     on_scroll: Option<Box<dyn Fn(Viewport) -> Message + 'a>>,
     class: Theme::Class<'a>,
     last_status: Option<Status>,
+    scrollbar_shrink_factor: f32,
 }
 
 impl<'a, Message, Theme, Renderer> Scrollable<'a, Message, Theme, Renderer>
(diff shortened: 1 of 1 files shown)


widget/src/scrollable.rs
```diff
@@ -232,6 +234,17 @@ where
         self
     }
 
+    /// Will set the shirking factor of the scrollbar that is applied if the scrollbar is not hovered or dragged.
+    ///
+    /// If you want the scrollbar to shrink a good factor is about `0.5`.
+    pub fn shrinking_scrollbar(mut self, factor: f32) -> Self {
+        if !(0.0..=1.0).contains(&factor) {
+            panic!("shrink factor must be between 0.00 and 1.0")
```
airstrike, line 242:
```rust
pub fn shrinking_scrollbar(mut self, factor: f32) -> Self {
    debug_assert!((0.0..=1.0).contains(&factor), "shrink factor must be between 0.0 and 1.0");
    self.scrollbar_shrink_factor = factor.clamp(0.0, 1.0);
    self
}
```

hecrj:
Hmm... I don't think this is common enough to warrant built-in support.

Open to be convinced otherwise!

PR #2733, opened 2025-01-18
Add keyboard modifiers to mouse events

edwloef:
This PR makes the current state of keyboard modifiers available in many of the possible mouse events.

Before, implementing mouse+modifier shortcuts meant the need to save the state of the modifier keys in your app or widget state, and then track changes to the current modifier keys via `Event::Keyboard(keyboard::Event::ModifiersChanged(modifiers))`, which is cumbersome.

https://discourse.iced.rs/t/should-mouse-events-expose-the-state-of-the-modifier-keys/577
https://github.com/iced-rs/iced/discussions/1868
https://github.com/iced-rs/iced/discussions/1914


widget/src/combo_box.rs
```diff
@@ -744,9 +745,10 @@ where
                 // Unfocus the input
                 self.text_input.update(
                     &mut tree.children[0],
-                    Event::Mouse(mouse::Event::ButtonPressed(
-                        mouse::Button::Left,
-                    )),
+                    Event::Mouse(mouse::Event::ButtonPressed {
+                        button: mouse::Button::Left,
+                        modifiers: Modifiers::empty(),
```
edwloef, line 750: I'm not entirely sure hard-coding the modifiers to be empty here is correct, open to suggestions if it isn't :)

edwloef:
Another idea I've had recently is to pass the mouse position in the `ButtonPressed` message as it is in the `CursorMoved` message, since that seems like a logical thing to want when a mouse button is pressed, but it's considerably less inconvenient to get that so I just haven't done it yet. If anyone thinks that would be a good idea as well I can add it here or make a new PR.

airstrike:
> Another idea I've had recently is to pass the mouse position in the `ButtonPressed` message as it is in the `CursorMoved` message, since that seems like a logical thing to want when a mouse button is pressed, but it's considerably less inconvenient to get that so I just haven't done it yet. If anyone thinks that would be a good idea as well I can add it here or make a new PR.

Sounds to me like that's for a separate PR. I see the appeal and I think the library should default to provide as much information as possible with the messages that are produced, unless there is a meaningful cost associated with it. For example, `mouse_area` should also report the position in which a click happened, possibly also the `Point` in which the cursor entered the area, etc. Many won't care about this additional information, but it would be very valuable to applications with very fine grained interactivity without forcing them to create their own widget.

The counter argument is that creating your own widget isn't *that* hard and you could always make a button which includes this information in a third-party crate or directly in your app...

edwloef:
> The counter argument is that creating your own widget isn't that hard and you could always make a button which includes this information in a third-party crate or directly in your app...

~~This is exactly why I haven't done this yet, however it shouldn't be much work. I'll probably just implement it in the coming days, then Héctor can merge or close it depending on whether he finds it to be a good fit.~~ Never mind, winit doesn't even give us the current mouse position on a click event, so it wouldn't fit in neatly with the rest of the event conversions. Also the mouse position is already passed into various widget methods, so it wouldn't even be necessary.

core/src/mouse/event.rs
```diff
@@ -20,10 +20,17 @@ pub enum Event {
...
-    ButtonPressed(Button),
+    ButtonPressed {
+        /// the button that was pressed
+        button: Button,
+        /// the modifier keys at the time of the button press
+        modifiers: Modifiers,
+    },
 
     /// A mouse button was released.
     ButtonReleased(Button),
```
mfrench1989, line 36: I think `ButtonReleased` should also track the modifiers.
Sometimes you want to react on press and sometimes on release, this also keeps it in line with `keyboard::Event` being the same.

edwloef: Seems reasonable to me. When I implemented this I couldn't think of a case where the on-release modifiers mattered, but it doesn't really hurt to include them either way. Thanks for the feedback, I'll take care of it soon :)

edwloef:
Hmm, after going through all the built-in widgets and seeing where this could be used instead of the old workaround, I've noticed that for this to really be useful, the touch events would also need access to modifiers.

Which got me thinking: maybe this shouldn't be part of any events, and rather be its own type that's passed around similar to `Cursor`, so you can just know the modifier state at any time? It sure seems like widgets could make good use of this info for various events. Of course you'd have to decide how to best give subscriptions access to the modifiers as well, if they aren't contained in the events anymore.

Just throwing this out there to get opinions.

Alternatively I could just add the modifiers to the touch events as well.

mkrueger:
This is really required +1. Just adding the info to other events make it easier to use/discoverable IMO.

mahlquistj:
> maybe this shouldn't be part of any events, and rather be its own type that's passed around similar to Cursor, so you can just know the modifier state at any time?

@edwloef i do feel like having it as it's own type being passed around might be the "best" way to do it imo. But also, that would require a lot of changes to the existing API i'm guessing. Adding the modifiers to current events would be the easy fix, but might not be the best option long-term.

I guess we will have to wait for Hector to get his opinion on this.

edwloef:
Considering I've just found myself wishing for access to the modifiers in `fn draw`, adding them as a parameter is what I would ideally like happen.

hecrj:
I'm not sold on this. Widgets that need to track the keyboard modifiers are quite rare.

This couples the `mouse` module with `keyboard` and seems to actually introduce more code than it removes, which defeats the purpose of the PR.

Also, introducing dynamic parameters in `draw` is a bad idea; since `draw` must be requested in `update` first. In fact, we should remove the `cursor` argument eventually.

airstrike:
FWIW I do find myself <kbd>cmd</kbd>+<kbd>click</kbd>ing or <kbd>opt</kbd>+<kbd>click</kbd>ing daily. It's not used for simple cut-and-dry UI but for richer apps like, say, `vscode`, it's a game changer.

Even on Discord, we <kbd>shift</kbd>+<kbd>click</kbd> reply to someone without pinging them. And on macOS, almost every interaction is different if you <kbd>opt</kbd>+<kbd>click</kbd> it.

Having to track the modifier in widget state feels needlessly cumbersome for this basic of an interaction. IMHO processing something like a <kbd>shift</kbd>+<kbd>click</kbd> is "batteries included" behavior I would assume was available in the library, if I were a new user.

Moreover, even if you assume that only complex widgets use this <kbd>modifiers</kbd>+<kbd>click</kbd> combo, that could actually be an argument in favor of providing this out-of-the-box. Those are precisely the widgets in which a lot of other manual plumbing is already being done, so getting this out of the way is more valuable for those use cases than it is for simpler widgets.

Lastly, on a more prescriptive note, I'd argue we should promote such interactions because they make for better UX, and providing easy support is a way to promote them. Power users benefit, and basic users don't ever need to notice them.

edwloef:
Another bit to consider is that with the current solution you don't actually know the state of the modifiers when the widget is created, so you have to wait for them to change once to be able to do these modifier+click interactions properly. This is a case I would be able to hit super easily in my app, since one of my main interactions that creates widgets is to shift+click on something to duplicate, and it's fairly common to want to do this a few times in a row. With the current solution that is really only solvable by sharing the modifier state amongst a bunch of widgets and hoping that it's up-to-date.

mfreeborn:
Just to add another concrete usecase as a data point, it's typical to be able to change a number input by small increments if you scroll over it, and by larger increments if you ctrl + scroll. Ctrl + scroll comes up again sometimes for quickly adjusting the scale factor of the window.

hecrj:
@airstrike I am certainly not convinced at all.

> Having to track the modifier in widget state feels needlessly cumbersome for this basic of an interaction.

I think this is hyperbolic at best and misleading at worst. Most widgets that handle clicks are likely to have internal state already, and tracking the modifiers is a total of 4 trivial lines of code (1 field declaration and 3 for the new `match` branch).

This is a very low bar for cumbersome.

> Another bit to consider is that with the current solution you don't actually know the state of the modifiers when the widget is created

I think this is the only valid issue here.

hecrj:
Since all keys are affected by discontinuity, I think it may be interesting to provide a new `Keyboard` type to `update` that represents the current state of all keys.

mkrueger:
> I'm not sold on this. Widgets that need to track the keyboard modifiers are quite rare.
> 
> This couples the `mouse` module with `keyboard` and seems to actually introduce more code than it removes, which defeats the purpose of the PR.
> 
> Also, introducing dynamic parameters in `draw` is a bad idea; since `draw` must be requested in `update` first. In fact, we should remove the `cursor` argument eventually.

your opinion - but in case you need them it's really annoying to work with iced. In this area this library is rubbish compared to others. I use custom scroll bars  & scrolling  - that's one other thing where iced can improve -see:

https://github.com/mkrueger/icy_tools

And in that case I need the modifiers for mouse events badly - working with modifierschanged isn't a good replacement and has own bugs/issues.

When working on a UI library it's not the job of the UI library to impose restrictions on the APPS that can be written with it but to open possibilities & make things easy. If something obviously doesn't wrok (like your modifierschanged system) then it should change. Just closing issues users have and say: no, I don't think so is the worst thing you can do to your library. UI libraries should it make easy to write custom controls - esp. when the main set of controls is tiny it's a core feature to make a wide range of controls possible.

And mouse input is one of these things users want - for some reason.

btw. it's one reason I switched away from egui - they're bullshitting around like that too in some areas. The state of UI libs in rust isn't good and iced doesn't make it better which such development style.

The easiest thing would be indeed to add modifiers to the mouse events - that what all other UI libraries do - for a reason. The next thing would be to fix the modifier event system to work - there is a modelling issue there (see bug).

hecrj:
@mkrueger You are permanently banned. Please, do me a favor and switch away from `iced` too.

Thanks!

PR #2736, opened 2025-01-19
Allow Button widgets to be focusable

chrismanning:
Focusable buttons can be pressed with the space/enter key if they are focused.

A focused button renders the same as a hovered button in the default stylings.


hecrj:
This is something I will personally tackle in a unified way for `0.15`.

PR #2739, opened 2025-01-21
Add `max_width` to `Tooltip`

jhff:
## Summary

This change adds `max_width` to `Tooltip`

This is useful to avoid way too wide tooltips when text is long

### Without max width

without

### With max width

with


jhff:
Closing this as the actual tooltip content takes an `Element`, so wrapping the text with a `Container` and setting its `max_width` works as well

PR #2744, opened 2025-01-23
Do not capture released event inside `image::Viewer` when mouse is not dragged

rhysd:
This PR changes the event handler of `image::Viewer` not to capture a pressed event when a mouse was not dragged by checking the cursor position.


hecrj (approved):
Thanks!

`ButtonReleased` events should never be captured.

PR #2653, opened 2024-10-29
Add `.id` method and `text_editor::focus` task

sgued:
This functionality adds the ability to programmatically focus a `text_editor` widget.

- [x] Implemented following the focus handling of the `text_input` widget.
- [x] Updated an example to make use of the feature

PR #2745, opened 2025-01-23
Use widget::Id directly instead of duplicated subclasses

kiilerix:
There doesn't seem to be any benefit from encapsulating Id in widget specific types or to not allowing custom Ids for container and scrollable.

Keep the widget ::Id around as an alias for backwards compatibility and minimal diff.

### DIFF
--- a/core/src/widget/id.rs
+++ b/core/src/widget/id.rs
@@ -23,6 +23,18 @@ impl Id {
     }
 }
 
+impl From<&'static str> for Id {
+    fn from(id: &'static str) -> Self {
+        Self::new(id)
+    }
+}
+
+impl From<String> for Id {
+    fn from(id: String) -> Self {
+        Self::new(id)
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 enum Internal {
     Unique(usize),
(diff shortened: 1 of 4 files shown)


hecrj (commented):
There is a reason for the different `Id` types. They increase the type-safety of specific widget operations. For instance, it should be impossible for you to try to scroll a container.

kiilerix:
Ok, thanks for explaining the reason for keeping it as it is.

But since these IDs are very weak links that might get invalidated, and some of them can be constructed from any string, the different types doesn't seem to provide much guarantee of correctness. Especially since the underlying operations use plain IDs without any type safety. It thus seems like a non-trivial amount of code to duplicate for each widget type for little gain.

Anyway, https://github.com/iced-rs/iced/pull/2653 thus seems to bring exactly the right amount of code duplication to give text_editor the same functionality as other widgets have.

hecrj:
Hmm... I think you have a point.

I'll think about it. Let's reopen this.

sgued:
I think this can be closed. Now there is only one `widget::Id` type and it's used for focus through `iced::widget::operation::focus`.

hecrj:
Superseded by #3059.

PR #2749, opened 2025-01-25
Make `Color::from_rgb8` and `Color::from_rgba8` const

rhysd:
### Problem

I want to use `Color::from_rgb8` and `Color::from_rgba8` for constants to define my own color palette. However they are not `const fn`.

### Solution

They are not `const fn` because `From::from` is not const. This PR replaces it with `as` operator which is equivalent to it. As a bonus, the implementation of `LIGHT` and `DARK` palettes can be simplified.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        rust: [stable, beta, "1.81"]
+        rust: [stable, beta, "1.82"]
     steps:
     - uses: hecrj/setup-rust-action@v2
       with:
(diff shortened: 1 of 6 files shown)


Zarthus:
Neat, I ran into needing this a few days ago 🙌🏻

hecrj (changes requested):
I think we need to bump the MSRV to `1.82` or `1.83`.

rhysd:
I confirmed that 1.81 didn't compile and 1.82 compiled. Do you think it's okay to bump MSRV? If it is okay, I'll try to add a commit to bump MSRV to 1.82.

hecrj (approved):
Great!

I bumped the MSRV to `1.82` and made all the `Color` constructors const. This lets us simplify the `color!` macro considerably as well.

Thanks! :partying_face:

rhysd:
Thank you!

PR #2756, opened 2025-01-27
Fix `Color::new` was made private

rhysd:
ce07acf6fedd705d6380739fbc5da40f95739e04 accidentally made `Color::new` private. This PR fixes the bug.


hecrj:
It wasn't an accident! Use `from_rgba` instead.

rhysd:
Ah, I apologize for my misunderstanding. Thanks for your correction.

PR #2758, opened 2025-01-28
Implement `Mul<Transformation>` for `mouse::Cursor` and `mouse::Click`

airstrike:
This makes it easier to work with non-screen coordinates in custom widgets.

### DIFF
--- a/core/src/transformation.rs
+++ b/core/src/transformation.rs
@@ -1,3 +1,4 @@
+use crate::mouse::Cursor;
 use crate::{Point, Rectangle, Size, Vector};
 
 use glam::{Mat4, Vec3, Vec4};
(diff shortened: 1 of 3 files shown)


core/src/transformation.rs
```diff
@@ -106,6 +107,19 @@ impl Mul<Transformation> for Rectangle {
...
+
+    fn mul(self, transformation: Transformation) -> Self {
+        match self {
+            Cursor::Unavailable => Cursor::Unavailable,
+            Cursor::Available(point) => {
+                Cursor::Available(point * transformation)
+            }
+        }
+    }
+}
```
hecrj, line 121: Move this to the `mouse::cursor` module.

hecrj (changes requested):

core/src/transformation.rs
airstrike: Done!

hecrj (approved):
Thanks!

PR #2763, opened 2025-01-29
`keybind` widget

hecrj:
This PR introduces a brand new widget: `keybind`.

A `keybind` widget can be used to trigger a click interaction on its contents when a certain `Hotkey` is pressed.

A `Hotkey` is a keyboard combination: a key, some modifiers, and an optional location. By default, hotkeys use `keyboard::Modifiers::COMMAND` as the trigger modifiers.

For instance:

```rust
keybind('s', button("Save").on_press(Message::Save))
```

This will display a save button that can be clicked with the mouse and also by pressing `Ctrl+S` (or `⌘+S` on macOS). In both instances, a `Save` message will be produced.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2628,6 +2628,7 @@ dependencies = [
  "iced_highlighter",
  "iced_renderer",
  "iced_runtime",
+ "iced_test",
  "num-traits",
  "ouroboros",
  "pulldown-cmark",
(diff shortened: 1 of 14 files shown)


hecrj:
I am scratching this idea.

Since `keybind` listens to key presses unconditionally, it is possible to write UIs with conflicting keybinds very easily.

The current `Subscription` is superior because the most natural design encourages a single source of truth for key presses.

PR #2766, opened 2025-01-29
Implement `Transformation::stretch`, `Transformation::inverse` and the safer `Transformation::try_inverse`

airstrike:
`Transformation::stretch` seems like a non-issue since it's just scaling across a single axis. Useful for custom widgets that want to scale in that fashion, such as anything with a timeline: DAWs, video editors, charts...

While we're on the topic of `Transformation`, it can incredibly handy to be able to invert some previous complex transformation so we can go to-and-from different coordinate systems, so I think it would be helpful to add some inversion method. After iterating for a bit, I am still not sure about the API for `inverse` and `try_inverse`.

I have tests to add if you think it would be helpful.

### DIFF
--- a/core/src/transformation.rs
+++ b/core/src/transformation.rs
@@ -11,6 +11,9 @@ impl Transformation {
     /// A [`Transformation`] that preserves whatever is transformed.
     pub const IDENTITY: Self = Self(Mat4::IDENTITY);
 
+    /// An error margin for [`Transformation`] comparisons.
+    const EPSILON: f32 = 1e-6;
+
     /// Creates an orthographic projection.
     #[rustfmt::skip]
     pub fn orthographic(width: u32, height: u32) -> Transformation {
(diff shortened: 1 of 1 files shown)


airstrike:
@edwloef was kind enough to point out that `Transformation::scale_factor` needs to account for the non-uniform scaling. I've changed its return type to `(f32, f32)` and worked through the compiler errors.

It's looking good, but I want to do some more testing, likely by modifying the `custom_quad` example to help visualize it.

I just don't really have the time to wrap this up this week so I'll have to come back to it later.

core/src/transformation.rs
```diff
@@ -31,6 +34,37 @@ impl Transformation {
...
+    /// In release mode, it may return a valid inverse or produce NaNs.
+    pub fn inverse(&self) -> Self {
+        let det = self.0.determinant();
+        if cfg!(debug_assertions) {
+            debug_assert!(
+                det.abs() >= Transformation::EPSILON,
+                "attempted to invert a near-singular transform (det={})",
+                det
+            );
+        }
```
NyxAlexandra, line 54: this is the same as just using `debug_assert`

airstrike: thanks! Rust noob here, so appreciate the pointer. will fix in the next draft.

NyxAlexandra: very clean considering :D

mahlquistj:
@airstrike any progress on this? I would love to see `Transformation::stretch` - I wanted to make my own PR but discovered this soon after writing the code.

airstrike:
> @airstrike any progress on this? I would love to see `Transformation::stretch` - I wanted to make my own PR but discovered this soon after writing the code.

Unfortunately I haven't had time to wrap this up. I'm under the impression this will break layout in many weird ways, but have not yet tested. Also there is now an inverse operation for `Transformation` so this would need to be rebased.

It's unlikely I will have time to work on this for the next 4-6 weeks. No pride of authorship so feel free to fork, propose your own or anything in between if you have an urgent need.

hecrj:
This will not work, unfortunately.

The text pipeline in `iced_wgpu` does not support any kind of stretching.

PR #2768, opened 2025-01-30
fix rectangle snapping

### LINKED ISSUE #2753: Horizontal scrollable off-by-one error?

### What happened?

Here are some containers in a scrollable with their borders shown. Note that the scrollable area is fully expanded, but the right-hand border of the right-most container inside the scrollable area has been cut off.

Image

Equally, I can reduce the width of the window to activate the scrollable area, and note how it is scrolled all the way to the right, but again the last column of pixels is cut off.

Image

In the above screenshots, I would expect to be able to see the right-hand border of the right-most container. 

I observed this behaviour in 0.13.1 (installed via the crates.io release), and I have produced this example code by writing it as a new example in a locally cloned copy of the iced repository.

My current workaround for this issue is to wrap the contents of the scrollable with a container and add 1 pixel of padding to the right-hand side.

Code:

```rust
// main.rs
use iced::widget::scrollable::{Direction, Scrollbar};
use iced::widget::{center, column, container, row, Container, Scrollable};
use iced::{border, Element, Length, Padding, Theme};

    iced::run(
        "Scrollable Off By One",
        ScrollableOffByOne::update,
        ScrollableOffByOne::view,
    )
}

#[derive(Default)]
struct ScrollableOffByOne;

#[derive(Debug, Clone, Copy)]
enum Message {}

impl ScrollableOffByOne {
        match message {}
    }

        let scrollable_content = Scrollable::with_direction(
            row![
                group_box("Inside scrollable"),
                group_box("Inside scrollable"),
                group_box("Inside scrollable"),
                group_box("Inside scrollable"),
                group_box("Inside scrollable"),
                group_box("Inside scrollable"),
            ]
            .spacing(10),
            Direction::Horizontal(Scrollbar::default()),
        )
        .spacing(10);

        center(column![row![
            group_box("Outside scrollable area"),
            scrollable_content
        ]
        .spacing(20)
        .padding(Padding::ZERO.top(50).bottom(50))])
        .padding(12)
        .into()
    }
}

pub fn group_box_style(t: &Theme) -> container::Style {
    let p = t.palette();

    container::Style {
        border: border::rounded(5).width(1).color(p.text),
        ..Default::default()
    }
}

pub fn group_box<'a, Message>(
    content: impl Into<Element<'a, Message>>,
) -> Container<'a, Message>
where
    Theme: container::Catalog + 'a,
{
    container(content)
        .height(Length::Fill)
        .padding(8)
        .style(group_box_style)
}

```

### Version

crates.io release

### Operating System

Linux

```shell

```

edwloef:
fixes #2753 

Before, the whole rectangle was moved towards the top left (to align the top left point), then the bottom right point was aligned after being moved up. Now the top left point is aligned independently of the bottom right point.


core/src/rectangle.rs
```diff
@@ -203,7 +203,10 @@ impl Rectangle<f32> {
     }
 
     /// Snaps the [`Rectangle`] to __unsigned__ integer coordinates.
-    pub fn snap(self) -> Option<Rectangle<u32>> {
+    pub fn snap(mut self) -> Option<Rectangle<u32>> {
+        self.width += self.x.fract();
+        self.height += self.y.fract();
+
         let width = self.width as u32;
         let height = self.height as u32;
```
hecrj, line 211:
```suggestion
    pub fn snap(self) -> Option<Rectangle<u32>> {
        let width = (self.width + self.x.fract()) as u32;
        let height = (self.height + self.y.fract()) as u32;
```

hecrj (changes requested):
Maybe we should consider rounding instead of flooring?

edwloef:
> Maybe we should consider rounding instead of flooring?

I don't think that truncate is less correct than rounding in this case, either should be fine. Truncate seems more intuitive to me though since it is the default way of casting float to int.

edwloef:
After thinking a bit more, I decided to implement it with rounding. I thought it'd be easiest to do if I added a function to snap a point, so I did that as well. Hopefully this all looks good.

dtzxporter:
Tested this with my vertical list and can confirm it also fixes the issue for me as well.

edwloef:
oh I messed up a rebase real bad here, oops one sec

PR #2770, opened 2025-01-31
Animate the `radio` widget

bradysimon:
## Overview

This PR uses the new `animation` API to animate the `radio` widget. The current animation scales/fades the dot icon as the `is_selected` state changes.

This also gates the animation behind a new `animations` feature that [Hector specified](https://github.com/iced-rs/iced/pull/2483#issuecomment-2617667426) and adds an optional `animations` feature to the `scrollable` example to demonstrate this change. To test this change locally:
- with animations: `cargo run -p scrollable --features animations`
- without animations: `cargo run -p scrollable`

## Preview 

Here's what the animation looks like (in light mode since the radio in dark mode isn't as clear):

https://github.com/user-attachments/assets/8e29bb6a-d392-48ec-a953-f09cf5d85544

> Side note: Hector mentioned on Discord to avoid animating more complex types like colors/backgrounds, so this PR doesn't add any general `Style` animation. Only `f32`/`bool` animations for now.


Brady-Simon:
An alternative animation would be to make the dot scale from a size of zero to its final size. I think I prefer the current animation but I want to at least present both options:


https://github.com/user-attachments/assets/c3f6332a-a371-429c-8fe7-587019cc9cb8

PR #2757, opened 2025-01-27
`Animation` API for application code

hecrj:
This PR introduces a new `animation` module built on top of [`lilt`](https://github.com/cyypherus/lilt).

The `animation` module exposes an `Animation` type that wraps some state and can be used to trigger an animation when this state changes; as well as cancel it.

When combined together with a `window::frames` subscription, it can be effectively used to animate smoothly at maximum framerate!

As usual, I have created a new example to showcase this feature. The new `gallery` example lazily loads the most popular images of Civitai and displays them in a grid:

https://github.com/user-attachments/assets/69712948-454a-4f8f-ac5a-82fb9245a5ce

PR #2779, opened 2025-02-02
feat: 'Toggler' animation

lazytanuki:
Hi!

This is the continuation of #2483 for the 'Toggler' widget animation based on the recently merged #2757.

To respect the `animations` feature flag, we still use the animation code but the animation is just instantaneous, which IMO reduces the code complexity as opposed to having several features gates.

The toggler movement follows an "ease out" animation, and the background color smoothly transitions from one state to the other as well.

I've also included an example crate which is supposed to showcase all animated widgets as they land. ('cargo run -rp animations')

Thanks!

[Kooha-2025-02-02-20-08-42.webm](https://github.com/user-attachments/assets/06880b11-8083-4b97-86fd-5f3fb8d05ef0)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,6 +116,13 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
+name = "animations"
+version = "0.1.0"
+dependencies = [
+ "iced",
+]
+
+[[package]]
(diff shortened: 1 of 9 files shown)


TheESN:
Is there anything new about merging this pr?

hecrj:
This is waiting for me to explore an approach that would introduce an associated type to `Catalog` (and maybe a trait) to allow fine-tuning of animations in custom themes.

I'm aware the wait can be a bit frustrating, but hopefully it'll be worth it in the end!

I appreciate all the effort here.

PR #2787, opened 2025-02-04
implement Clone for text_editor::Content

JL710:


hecrj:
`Clone` cannot be implemented like this; since sharing internal buffer state should not be encouraged.

JL710:
alright @hecrj

PR #2790, opened 2025-02-05
Set correct text size for preedit window

rhysd:
## Problem

Currently the text size of preedit window is fixed to `renderer.default_size()`. However the text size of text input and text editor can be changed by user. Using same size for preedit window is more natural.

https://github.com/user-attachments/assets/8e3bb738-4f9a-4648-9005-29893241a102

## Solution

This PR propagates the text size from the widget which requests IME to the window manager which manages the event loop. The propagated text size is used when the preedit window is rendered if available.

https://github.com/user-attachments/assets/5629c1ea-0c5a-4f5f-b941-66ff6f9d57f9

### DIFF
--- a/core/src/input_method.rs
+++ b/core/src/input_method.rs
@@ -40,14 +40,11 @@ pub struct Preedit<T = String> {
 
 impl<T> Preedit<T> {
     /// Creates a new empty [`Preedit`].
-    pub fn new(text_size: Option<impl Into<Pixels>>) -> Self
+    pub fn new() -> Self
     where
         T: Default,
     {
-        Self {
-            text_size: text_size.map(Into::into),
-            ..Default::default()
-        }
+        Self::default()
     }
 
     /// Turns a [`Preedit`] into its owned version.
--- a/core/src/input_method.rs
+++ b/core/src/input_method.rs
@@ -1,5 +1,5 @@
 //! Listen to input method events.
-use crate::Point;
+use crate::{Pixels, Point};
 
 use std::ops::Range;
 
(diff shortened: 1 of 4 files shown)


core/src/input_method.rs
```diff
@@ -34,15 +34,20 @@ pub struct Preedit<T = String> {
...
-    pub fn new() -> Self
+    pub fn new(text_size: Option<impl Into<Pixels>>) -> Self
     where
         T: Default,
     {
-        Self::default()
+        Self {
+            text_size: text_size.map(Into::into),
+            ..Default::default()
+        }
```
hecrj, line 50: No need. Let's not ask for any argument and just use `Default`.

hecrj (changes requested):

kenz-gelsoft:
Thank you for working on follow-ups!

I had an idea on this, which text widgets pass the caret height as an preedit style hint.

Passing text style itself achieves high fidelity preedit, but we need to find balance to widgets requirement.

core/src/input_method.rs
rhysd: Sure. I'll make a change.

rhysd:
@kenz-gelsoft Yeah, it's even better idea but it would require more design decisions. In this PR, I'd like to focus on only the text size as a quick fix. And let me say thank you for the bunch of the first work for adding IME support. It's definitely needed for CJK users :+1:

core/src/input_method.rs
rhysd: @hecrj I fixed this at cf851e133ad6aaedaf07b58c68e7c41d41ee151a

hecrj (approved):
Thanks!

PR #2795, opened 2025-02-06
Request redraw in image viewer update

tarkah:
image viewer isn't requesting redraw after scaling / panning so it appears stuck. It doesn't produce messages to trigger redraws so they have to be manually requested.


hecrj (approved):
Thanks!

PR #2797, opened 2025-02-07
Fix system informations `cpu_brand` always being empty

### LINKED ISSUE #2794: System informations `cpu_brand` is always empty

### What happened?

Experimenting a bit I noticed that the `cpu_brand` is always an empty string. Running the `system_information` example with `cargo r --package system_information` for example. On my machine looks like this:

Image

Looking at the implementation shows that [here](https://github.com/iced-rs/iced/blob/master/winit/src/system.rs#L20) we fill out the `Information` struct like this:

```rust
use sysinfo::{Process, System};
    let mut system = System::new_all();
    system.refresh_all();

    let cpu = system.global_cpu_info();

    // ...

    Information {
        // ...
        cpu_brand: cpu.brand().into(),
        // ...
    }

```

Reading the documentation for `System::global_cpu_info()` from [here](https://docs.rs/sysinfo/0.30.13/sysinfo/struct.System.html#method.global_cpu_info) it states:

```md
...
⚠️ Important ⚠️

Information like [Cpu::brand](https://docs.rs/sysinfo/0.30.13/sysinfo/struct.Cpu.html#method.brand), 
[Cpu::vendor_id](https://docs.rs/sysinfo/0.30.13/sysinfo/struct.Cpu.html#method.vendor_id) or [Cpu::frequency]
(https://docs.rs/sysinfo/0.30.13/sysinfo/struct.Cpu.html#method.frequency) are not set on the “global” CPU.
```

which would explain why the `cpu_brand` always shows an empty String.

`cpu_brand` not being an empty string.

### Version

master

### Operating System

Linux

```shell
N/A
```

AMS21:
This was caused by a misuse of the `global_cpu_info` function, which does not contain a valid `cpu_brand` field.

To fix this, we now get the first cpu and return it's brand instead.

Fixes #2794

### DIFF
--- a/winit/src/system.rs
+++ b/winit/src/system.rs
@@ -17,7 +17,11 @@ pub(crate) fn information(
     let mut system = System::new_all();
     system.refresh_all();
 
-    let cpu = system.global_cpu_info();
+    let cpu_brand = system
+        .cpus()
+        .first()
+        .map(|cpu| cpu.brand().to_string())
+        .unwrap_or_default();
 
     let memory_used = sysinfo::get_current_pid()
         .and_then(|pid| system.process(pid).ok_or("Process not found"))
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks!

PR #2801, opened 2025-02-08
Implement macOS-specific key shortcuts for text input

rhysd:
macOS provides several standard key shortcuts with `Control` modifier key. However iced does not implement them for `TextInput` widget.

https://support.apple.com/en-us/102650#:~:text=and%20iCloud%20Drive.-,Text%2Dediting%20shortcuts,-The%20behavior%20of

This PR implements the following key shortcuts for macOS:

- `Control + B` (move cursor left)
- `Control + F` (move cursor right)
- `Control + A` (move cursor to start)
- `Control + E` (move cursor to end)
- `Control + H` (delete character before cursor)
- `Control + D` (delete character after cursor)


airstrike:
TIL... We should implement these for `text_editor` as well for consistency.

widget/src/text_input.rs
```diff
@@ -1853,3 +1856,32 @@ fn alignment_offset(
...
 }
+
+#[cfg(target_os = "macos")]
+fn convert_macos_shortcut(
+    key: &keyboard::Key,
+    modifiers: keyboard::Modifiers,
+) -> &keyboard::Key {
+    let control = modifiers.control();
+    match key.as_ref() {
+        keyboard::Key::Character("b") if control => {
```
Exidex, line 1867: this also catches ctrl + alt or any other combination of modifiers that uses ctrl, which is probably not intended

rhysd: Thanks. I fixed this and force-pushed the fix to this branch (with rebasing it onto the latest master branch).

PR #2804, opened 2025-02-11
add unfocus operation

edwloef:
Adds an operation that unfocuses all focusable widgets. Previously this could be done by focusing a guaranteed-to-be-missing widget Id, however having it be its own thing is more intuitive. A use case for this would be to unfocus a `text_input` on submit.

### DIFF
--- a/core/src/widget/operation/focusable.rs
+++ b/core/src/widget/operation/focusable.rs
@@ -61,7 +61,7 @@ pub fn focus<T>(target: Id) -> impl Operation<T> {
     Focus { target }
 }
 
-/// Produces an [`Operation`] that unfocuses the focused widget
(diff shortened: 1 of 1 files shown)


core/src/widget/operation/focusable.rs
```diff
@@ -61,6 +61,33 @@ pub fn focus<T>(target: Id) -> impl Operation<T> {
     Focus { target }
 }
 
+/// Produces an [`Operation`] that unfocuses the focused widget
```
hecrj, line 64:
```suggestion
/// Produces an [`Operation`] that unfocuses the focused widget.
```

hecrj (changes requested):

hecrj (approved):
Thanks.

PR #2812, opened 2025-02-22
`operation::focusable::is_focused` & `text_input::is_focused`

andymandias:
This PR consists of two additions:

1.  An `is_focused` function that produces an `Operation` that looks for a `focusable` widget by ID, and produces a `bool` that stores whether the widget is focused or not.  This function was modeled after the existing `find_focused`, and for my purposes was a means to implement the second addition.
2.  An `is_focused` function that produces a `Task` that looks for a `text_input` widget by ID, and produces a `bool` with whether the widget is focused or not.

Originally I tried to use the `find_focused` `Operation` for `text_input::is_focused`, but I couldn't figure out how to make it work as desired when there is no currently focused widget (and `find_focused` produces `Outcome::None`).

### DIFF
--- a/core/src/widget/operation/focusable.rs
+++ b/core/src/widget/operation/focusable.rs
@@ -257,3 +257,48 @@ pub fn find_focused() -> impl Operation<Id> {
 
     FindFocused { focused: None }
 }
+
+/// Produces an [`Operation`] that searches for the focusable widget
+/// and stores whether it is focused or not. This ignores widgets that
+/// do not have an ID.
+pub fn is_focused(target: Id) -> impl Operation<bool> {
+    struct IsFocused {
+        target: Id,
+        is_focused: Option<bool>,
+    }
+
+    impl Operation<bool> for IsFocused {
+        fn focusable(
+            &mut self,
+            id: Option<&Id>,
+            _bounds: Rectangle,
+            state: &mut dyn Focusable,
+        ) {
+            if id.is_some_and(|id| *id == self.target) {
+                self.is_focused = Some(state.is_focused());
+            }
+        }
+
+        fn container(
+            &mut self,
+            _id: Option<&Id>,
+            _bounds: Rectangle,
+            operate_on_children: &mut dyn FnMut(&mut dyn Operation<bool>),
+        ) {
+            operate_on_children(self);
+        }
+
+        fn finish(&self) -> Outcome<bool> {
+            if let Some(is_focused) = &self.is_focused {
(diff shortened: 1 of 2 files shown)


core/src/widget/operation/focusable.rs
```diff
@@ -257,3 +257,48 @@ pub fn find_focused() -> impl Operation<Id> {
...
+            }
+        }
+
+        fn container(
+            &mut self,
+            _id: Option<&Id>,
+            _bounds: Rectangle,
+            operate_on_children: &mut dyn FnMut(&mut dyn Operation<bool>),
+        ) {
+            operate_on_children(self);
```
hecrj, line 288: We could probably stop traversing the widget tree if `self.is_focused` is already `Some`:

```suggestion
            if self.is_focused.is_some() {
                return;
            }

            operate_on_children(self);
```

core/src/widget/operation/focusable.rs
```diff
@@ -257,3 +257,48 @@ pub fn find_focused() -> impl Operation<Id> {
...
+        }
+
+        fn finish(&self) -> Outcome<bool> {
+            if let Some(is_focused) = &self.is_focused {
+                Outcome::Some(*is_focused)
```
hecrj, line 293: Since `bool` is copy, we can avoid the dereference:

```suggestion
            if let Some(is_focused) = self.is_focused {
                Outcome::Some(is_focused)
```

hecrj (changes requested):
Thanks!

Looking good! Just a couple nits.

core/src/widget/operation/focusable.rs
hecrj: Maybe even something functional like:

```rust
self.is_focused.map(Outcome::Some).unwrap_or(Outcome::None)
```

core/src/widget/operation/focusable.rs
andymandias: Thank you for the suggestion!  Makes sense to me.  Outside the scope of this PR probably, but would it make sense to similarly stop traversing once a result is found for `find_focused` as well?

core/src/widget/operation/focusable.rs
andymandias: Much cleaner, thank you!

core/src/widget/operation/focusable.rs
hecrj: Yes, both nits should apply there as well.

Issue #2810, opened 2025-02-21
WGPU prefferencing a low power graphics adapter by default


laycookie:
### Is your issue REALLY a bug?

- [x] My issue is indeed a bug!
- [x] I am not crazy! I will not fill out this form just to ask a question or request a feature. Pinky promise.

### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Is this issue related to iced?

- [x] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

In cases where user has both a discrete, and integrated GPU, iced is currently preferring integrated GPU, which is a huge problem, as it isn't just the low powered version, but is also practically always not the GPU the user is actively using as there output device, which results in severe performance issues.

### What is the expected behavior?

Currently wgpu picks the GPU to utilize by first calling `wgpu::util::power_preference_from_env()` (which was removed in the latest version of wgpu) which checks for the users preference based on there environment performance, which in practice is never set returning `None`.
```rs
        let adapter_options = wgpu::RequestAdapterOptions {
            power_preference: wgpu::util::power_preference_from_env()
                .unwrap_or(if settings.antialiasing.is_none() {
                    wgpu::PowerPreference::LowPower
                } else {
                    wgpu::PowerPreference::HighPerformance
                }),
```
This results in the power preference being decided, by seeing if anti-aliasing is enabled. Not exactly sure what was the logic behind this decision back at the time, but I also don't know much about low powered devices so I digress (This gets relevant as bellow will be my speculations on when LowPower preference should be set).
Now intuitively one might think that it is checking if the user has anti-aliasing set as a system default, as if it isn't enabled this most likely means they are using a device which is either low powered, or is attempting to preserve battery as much as possible, where in both cases Low Power preference is probably preferred.
In reality after tracking down where the settings gets initialized you will be lead to the initialization of the Program, or Deamon, which means that in reality it is actually just checking if the app itself has anti-aliasing enabled, which it by default does not.
Sourced from iced::Deamon function:
```rs
    Daemon {
        raw: Instance {
            update,
            view,
            _state: PhantomData,
            _message: PhantomData,
            _theme: PhantomData,
            _renderer: PhantomData,
        },
        settings: Settings::default(),
    }

```
Sourced from wgpu Settings definition:
```rs
impl Default for Settings {
    fn default() -> Settings {
        Settings {
            present_mode: wgpu::PresentMode::AutoVsync,
            backends: wgpu::Backends::all(),
            default_font: Font::default(),
[...]

PR #2813, opened 2025-02-22
Power prefference patch

laycookie:
Fixing the issue mentioned in #2810. Currently I'm doing so by looking at the reported chastise type on Linux, and requesting pc system type enum for windows, to determine if the device is low powered, or high powered. In the cases of platform not being either of them it defaults to Low Power, which shouldn't matter in case of Macs, and is a logical default in my head for an unknown OS.


laycookie:
So far tested it on windows, and Linux desktop.

wgpu/src/window/compositor.rs
```diff
@@ -82,10 +84,60 @@ impl Compositor {
...
-                .unwrap_or(if settings.antialiasing.is_none() {
-                    wgpu::PowerPreference::LowPower
-                } else {
-                    wgpu::PowerPreference::HighPerformance
+                .unwrap_or_else(|| {
+                    #[cfg(target_os = "linux")]
+                    {
+                        let output = Command::new("cat")
+                            .arg("/sys/class/dmi/id/chassis_type")
+                            .output();
```
NyxAlexandra, line 92: is there a reason to call out to `cat` instead of normal fs io?

laycookie: No, not really. I just used shell on windows, and decided to use it on Linux as well just for consistency.

hecrj (approved):
Let's just use `HighPerformance` by default and see if anyone yells at us.

laycookie:
> Let's just use `HighPerformance` by default and see if anyone yells at us.

Respectable

hecrj:
@laycookie I'd still consider contributing the device detection code to `wgpu` if I were you. It may make sense there as a `util` and I'd feel better relying on it.

PR #2815, opened 2025-02-24
extend mouse interactions

edwloef:
This PR adds more mouse interactions, and orders them based on the [wayland cursor shape protocol](https://drafts.csswg.org/css-ui/#cursor) (https://drafts.csswg.org/css-ui/#cursor). I've decided to leave out all the `*-resize` cursors that weren't already present for now, but I can add those too if need be.

I implemented this because I found myself looking for the `NoDrop` cursor, and found that iced was missing a bunch of the cursors it could be exposing.


edwloef:
I've updated this to remove the `vertical_text` cursor, since iirc we can't even create vertical text in iced as of now, and added the `col_resize` and `row_resize` cursors.

Issue #2817, opened 2025-02-25
Rehighlight `TextEditor` when `Theme` changes


pml68:
### Is your issue REALLY a bug?

- [x] My issue is indeed a bug!
- [x] I am not crazy! I will not fill out this form just to ask a question or request a feature. Pinky promise.

### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Is this issue related to iced?

- [x] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

When the application `Theme` changes, `TextEditor` is redrawn, but is not rehighlighted.

### What is the expected behavior?

Since the `to_format` field of `TextEditor::highlight_with` accepts a closure with the Theme as an argument, and the Theme gets passed to to format fn in `TextEditor::draw` as well, I'd expect it to get rehighlighted when Theme changes. I'd open a PR, but I don't see a way to check if the Theme has actually changed, and I believe calling `state.highlighter.borrow_mut().change_line(0)` on each redraw isn't desired.

### Version

master

### Operating System

Linux

### Do you have any log output?

```shell

```

hecrj:
Fixed by #2818.

PR #2818, opened 2025-02-25
Simple „solution” to #2817

pml68:
It's far from perfect, but it's something. Maybe it's even acceptable.

### DIFF
--- a/widget/src/text_editor.rs
+++ b/widget/src/text_editor.rs
@@ -125,6 +125,7 @@ pub struct TextEditor<
         &Theme,
     ) -> highlighter::Format<Renderer::Font>,
     last_status: Option<Status>,
+    rehighlight_on_redraw: bool,
 }
 
 impl<'a, Message, Theme, Renderer>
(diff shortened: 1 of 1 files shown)


edwloef:
A better solution to this could be checking whether the theme has changed in the `draw` function, something like this: https://github.com/generic-daw/generic-daw/blob/d2d1c5f95733cd0169ffe03346491cfb959b9049/generic_daw_gui/src/widget/knob.rs#L205-L213

pml68:
I tried doing something like that, but adding PartialEq as a trait requirement to the Theme *generic* caused issues with the Widget trait, but I'll try again today. Maybe I just didn't read the error carefully.

pml68:
@edwloef I'm not sure what I did wrong last night, but it worked just fine this time around. Thanks for making me try again!

pml68:
This solves #2817

acidjazz:
Would be awesome if this got merged - really helpful 👍

PR #2819, opened 2025-02-26
Use theme's base colors for preedit iff program style's background is not opaque

kenz-gelsoft:
Some iced apps may have non-opaque background color as winit::Program::style().

For example, I'm trying to backport input method for COSMIC apps, those program styles have transparent background.

We're using that color for the background color of the preedit layer, but transparent background doesn't work well.

|before|
|-|
|Screenshot_2025-02-27_00-40-25|

|after|
|-|
|Screenshot_2025-02-27_00-43-27|

If that color is non-opaque color, should we use the theme's base color instead?


kenz-gelsoft:
This PR doesn't change iced examples behaviors as far as I tested.

hecrj (approved):
Thanks!

We can just hardcode the `a` channel of the background to `1.0`.

kenz-gelsoft:
> Thanks!
> 
> We can just hardcode the `a` channel of the background to `1.0`.

Thank you for cleaning up this. I confirmed your simplified solution works with my usecase (COSMIC ime backport).

PR #2820, opened 2025-02-28
Create a shared window handle struct that implements both window and display traits.

dtzxporter:
This is required to use a library like rfd now, where linux support requires the display handle as well, which is always apart of a window anyways. The handle struct prevents exposing the internal window.


Rimpampa:
A similar PR has already been submitted #2718, I don't know which is better but regardless I think that this should be implemented.

hecrj:
Thanks!

Superseded by #2718.

Issue #2708, opened 2024-12-29
Geometry example in the examples folder is broken


davidfufu:
### Is your issue REALLY a bug?

- [X] My issue is indeed a bug!
- [X] I am not crazy! I will not fill out this form just to ask a question or request a feature. Pinky promise.

### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

the geometry project in the examples doesn't work. the center vertex doesn't follow the mouse as the description says

### What is the expected behavior?

for the demo to work properly

### Version

master

### Operating System

macOS

13r0ck:
Does it not move at all, or are the movements just very laggy? I see the later on a macOS machine. Interestingly on the tiny-skia backend there is no rendered geometry at all

YoCraz:
In regards to this issue; the example for me does not render the geometry. I use Windows, and via the master version as well. All the other examples seem to work.

PR #2823, opened 2025-03-03
Geometry example in the examples folder is broken #2708

ArtHome12:
Since some update the window stopped receiving updates constantly, but only when the mouse entered and exited the window. 

Example `Geometry` broke - there was no reaction to mouse movement. Now the original correct behavior has been restored.

### DIFF
--- a/examples/geometry/src/main.rs
+++ b/examples/geometry/src/main.rs
@@ -8,7 +8,7 @@ mod rainbow {
     use iced::mouse;
     use iced::{
         Element, Length, Rectangle, Renderer, Size, Theme, Transformation,
-        Vector,
+        Vector, 
     };
 
     #[derive(Debug, Clone, Copy, Default)]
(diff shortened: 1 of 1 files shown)


ArtHome12:
# Benchmark 
## Before
- default size 120 FPS, 25TPS, 
- with size in whole world 3-4 FPS, 25 TPS.
- full screen (with filter, the world fits 10 times), refresh every 1.5 seconds. Full CPU utilization.

## With bitmap unoptimized
- default size 120 FPS, 25TPS, 
- with size in whole world 6 FPS, 25 TPS.
- full screen refresh every 1-2 FPS, 25-42 TPS.

ArtHome12:
Has this small PR not yet been considered or are there any comments on it?

hecrj:
Thanks!

But the proper fix is #3020.

PR #2824, opened 2025-03-03
add cursor show/hide actions

edwloef:


EmmanuelDodoo:
@edwloef did it get implemented elsewhere?

edwloef:
I noticed #2685 was implemented before this, it seems like a better solution to me.

PR #2826, opened 2025-03-04
make mouse area behave more like its documentation

edwloef:
Before, the mouse area wasn't emitting messages for events that its contents had captured. This made e.g. a mouse area wrapping a button never emit double-click events.

Additionally, the documentation of `MouseArea::interaction` make it seem like that interaction has priority over the interaction of the contents, I changed that as well.


edwloef:
Hmm, I now understand the use case for the way it was before. Should this behavior be configurable, or should this come down to "copy paste and modify the existing widget"?

edwloef:
Closed since I noticed this breaks the `modal` example.

PR #2828, opened 2025-03-06
Add an option for the wgpu feature to disable the default backends.

dtzxporter:
- This allows the user to specify which backends they want enabled, just like how the `image-without-codecs` feature works.

Allows significant reduction in binary size on platforms with multiple backends.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2510,6 +2510,7 @@ dependencies = [
  "iced_winit",
  "image",
  "thiserror 1.0.69",
+ "wgpu",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


dtzxporter:
Fixed for latest wgpu update.

StT191:
Thinking about this a lot more in depth while writing [#3064](https://github.com/iced-rs/iced/pull/3064) I quite like this approach as it doesn't increase the surface area of feature flags as much.

However rather than "wgpu-without-backends" the new feature flag should be named "wgpu-without-defaults" or "wgpu-no-defaults" or something similar that does convey what it's actually achieving: not just opting out of backends but also features like "std" and "parking_lot" etc.

edwloef:
It might be worth enabling wgpu's `std` feature regardless, since `iced` isn't `no-std` anyways.

hecrj (approved):
Thanks!

I ended up naming the feature simply `wgpu-bare` and introduced it in `iced_renderer` to avoid the redundant `wgpu` dependency in the root crate.

PR #2830, opened 2025-03-06
Fix WGPU backend when using chromium+linux (webgl only)

EleDiaz:
When using the WGPU backend on chromium with Linux, the instance tries to get the adapter, which is null (due to the lack of support by the browser).

So, this change forces the WebGL on browser when WebGPU is unsupported.

And the source of this wgpu function give more details about it.
[https://docs.rs/wgpu/latest/wgpu/util/fn.is_browser_webgpu_supported.html](https://docs.rs/wgpu/latest/wgpu/util/fn.is_browser_webgpu_supported.html)

### DIFF
--- a/wgpu/src/window/compositor.rs
+++ b/wgpu/src/window/compositor.rs
@@ -288,6 +288,10 @@ impl graphics::Compositor for Compositor {
                     settings.present_mode = present_mode;
                 }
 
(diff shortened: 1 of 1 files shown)


hecrj:
I believe this was fixed by #2686.

PR #2832, opened 2025-03-08
Updated wgpu to 24 and glyphon to 0.8

codewing:
- I've updated wgpu to 24 (very uneventful)
- I've switched glyphon to 0.8 which includes everything custom in hecjr's repository but the encoder. 
  If this is a dealbreaker then let me know and I see if we can have an additional prepare that also accepts an encoder and not just a queue


hecrj (approved):
Thanks!

I was just working on updating our own `glyphon` fork: [`cryoglyph`](https://github.com/iced-rs/cryoglyph).

It also seems we are getting a bit faster:

image

PR #2834, opened 2025-03-10
Update `cosmic-text` to `0.13`

hecrj:
Pretty straightforward!

Further testing and benchmarks pending.


hecrj:
Run the benchmarks a couple times and there does not seem to be any meaningful difference.

If anything, the `dynamic text (heavy)` is ~2% faster :tada:

PR #2837, opened 2025-03-11
Expose setting to disable vsync if requested.

dtzxporter:
- This improves rendering performance on windows 10x and doesn't result in noticeable tearing. The user can decide if they want it anyways.

### DIFF
--- a/core/src/settings.rs
+++ b/core/src/settings.rs
@@ -33,6 +33,13 @@ pub struct Settings {
     ///
     /// By default, it is disabled.
     pub antialiasing: bool,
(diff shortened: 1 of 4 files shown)


hecrj:
Ah, no. It doesn't. `ICED_PRESENT_MODE` will override.

dtzxporter:
> Ah, no. It doesn't. `ICED_PRESENT_MODE` will override.

You beat me to it, yeah, it will take over.

hecrj (approved):
Thanks!

PR #2839, opened 2025-03-13
get rid of palette dependency in iced-core

edwloef:
This brings iced-core down from 39 dependencies to 23.

Using `palette` here was overkill, and the crate constantly invalidated my incremental build cache which was annoying and the main reason I looked into this. Ran the tour example and everything looks correct to me, and tests pass locally as well, though I don't know whether anything tests screenshots etc.

I removed two of the tests, since they seemed to only exist to test conversions between iced `Color` and the `palette` types. Maybe the conversions could go behind a feature flag, but I left that out for now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2527,7 +2527,6 @@ dependencies = [
  "lilt",
  "log",
  "num-traits",
- "palette",
  "rustc-hash 2.1.1",
  "smol_str",
  "thiserror 1.0.69",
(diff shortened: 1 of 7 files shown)


hecrj (approved):
Sweet! I've added a snapshot test for each built-in `Theme` in `master`; and they all seem to pass.

Let's :ship:!

airstrike:
Any chance we make `Hsl` and all the helper functions in `core/src/theme/palette.rs` `pub`?

It would allow users creating their own `Theme` types which use these functions to also skip the `palette` dependency.

PR #2840, opened 2025-03-13
Clean up devShell templates

max-privatevoid:
TL;DR
- remove `flake-utils`
- use [the nix-systems pattern](https://github.com/nix-systems/nix-systems)
- remove unnecessary `buildInputs`
- replace usage of `LD_LIBRARY_PATH` with `-rpath` linker flag
- add `rustc` and `cargo`

This PR cleans up the Nix devShells provided in `DEPENDENCIES.md` to serve as a better getting started point for newbies. `flake-utils` was removed because it's an unnecessary abstraction. A simple call to `lib.genAttrs` replaces it.

During testing, I've discovered that all of the `buildInputs` were unnecessary to get some examples (specifically `arc` and `tour`) to run. It seems this project uses `dlopen` for everything, which is pretty cursed, but I'll roll with it.

`LD_LIBRARY_PATH` is a horrible hack and should never be used. The proper course of action is to have these libraries available via `RPATH`, which can be set at link time via the linker argument `-rpath`. With the `RUSTFLAGS` environment variable and some trickery, one can eventually pass down such arguments to `ld`. I've discovered the minimal set of dependencies to run some of the examples:
- `libxkbcommon`, always
- `vulkan-loader`, for Vulkan
- `libGL`, for OpenGL
- `wayland`, for Wayland
- `libX11`, `libXcursor` and `libXi` for X11

I've enabled Wayland and Vulkan by default because they're clearly superior, and I've left the dependencies for OpenGL and X11 next to them in a comment so people can easily pick them if needed.

Since you'll probably be using Rust when working with `iced-rs`, you most likely also want a Rust toolchain, so I've added `rustc` and `cargo` from Nixpkgs to the `nativeBuildInputs`. This project currently uses edition 2024, which requires rustc 1.85, which is currently only available in `nixos-unstable`.


max-privatevoid:
Repent for your sins and crimes against humanity @RGBCube

RGBCube:
LGTM, I was pretty new to Nix when I created that initial PR.

Gerg-L:
use `packages` over `nativeBuildInputs`

max-privatevoid:
I think it'd be better to explicitly use `nativeBuildInputs` and force people to learn the difference between that and `buildInputs`. I'm hoping for a future where we can enable `strictDeps` by default. I think adding an empty `buildInputs` list with a comment saying what to put in there would be a good idea. What do you think?

Gerg-L:
Seems good to me

PR #2841, opened 2025-03-13
don't pull in `futures-executor` when it's not the default executor

edwloef:
This avoids the need to pull in `futures-executor` when a different default executor is chosen. I've verified that app startup still works on all native executors. Hopefully I chose the correct approach.

`futures-executor` won't actually not be pulled in until https://github.com/hecrj/sipper/pull/1 is included as well


edwloef:
well, wasm-bindgen-futures was being annoying but it passes now. :shrug:

edwloef:
You may need to enable the `async-await` feature of `futures` in `iced-futures`.

hecrj:
I just did in 5d1de9dc95bd981be122f6442d7d77de80d5d176.

edwloef:
Yeah, I saw, I was just barely too late :sweat_smile:

karolisr:
Hmm, I do not understand why, but after this merge some examples compiled with "tiny-skia" disabled and "wgpu" backend enabled do not work (at least on macOS). "todos" example does work. "arc" and "bezier-tool" do not. They compile, but no window opens and program exits. I tried with "thread-pool", "tokio", and "smol". No go. Any ideas what's going on?

PR #2845, opened 2025-03-16
request redraw on mouse area hover change

edwloef:
Before, the mouse area's interaction wouldn't immediately change if it was set, since no redraw was requested when the cursor switched between being in the mouse area or outside of it.


hecrj:
A `mouse_area` does not draw anything but its contents; therefore requesting redraws makes no sense.

hecrj (approved):
Thanks!

PR #2850, opened 2025-03-19
no_link_checks

nico2sh:
Validating URLs on markdown rendering prevents adding custom logic for links.
I think link validation makes more sense on the application update, when consuming the message. This way the developer can decide how to react and use custom protocols or relative paths, which are not possible now.

This PR replaces the `Url` for `String` in markdown links.


hecrj:
Superseded by #2992.

PR #2852, opened 2025-03-20
Allow custom vertical spacing for wrapped rows

mtkennerly:
I have a use case where I'd like the spacing between wrapped lines to be different than the spacing within each line.

| before (inherited spacing = 20) | after (overridden spacing = 0) |
| - | - |
| Screenshot 2025-03-20 182819 | Screenshot 2025-03-20 182953 |

My specific use case where I'd like to reduce the spacing:

> Screenshot 2025-03-20 182953


hecrj (approved):
Seems reasonable! Thanks :partying_face:

PR #2854, opened 2025-03-22
fix: update serde dependency to include derive feature in Cargo.toml

davehorner:
simple fix, this example requires the derive macro.


davehorner:
that fixes it compiling; it doesn't work exactly.  maybe not an example?  code doesn't compile without derive.
2025-03-22T10:55:57.889587Z  WARN cosmic_text::font::system: failed to load font 'GB18030Bitmap'
2025-03-22T10:55:57.899802Z ERROR changelog: no GITHUB_TOKEN variable was set

PR #2858, opened 2025-03-25
make DEFAULT_PADDING public

JL710:
This PR makes the DEFAULT_PADDING for a button public, like the one for text_input already is. 

https://docs.rs/iced/latest/iced/widget/text_input/constant.DEFAULT_PADDING.html


hecrj (approved):
Seems reasonable.

PR #2859, opened 2025-03-25
Expose drawing cosmic_text::Buffer directly in text::Renderer

nate-trojian:
Adds `fill_buffer` method to text::Renderer trait that takes in a Weak pointer to a cosmic_text::Buffer.
Allows for more advanced use cases on top of the existing helper methods.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2522,6 +2522,7 @@ dependencies = [
  "approx",
  "bitflags 2.9.0",
  "bytes",
+ "cosmic-text",
  "dark-light",
  "glam",
  "lilt",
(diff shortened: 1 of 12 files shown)


pml68:
hecrj stands strongly against re-exporting other crates' API and types. Instead, you should probably create a fill_raw method that renders an instance of `iced_graphics::Text::Raw`

nate-trojian:
Hmm, okay.  I did this for my own project, but figured it might be useful for others.

I made this PR from my own master branch but I have more changes I want to push to that, so I'll close this for now.  I moved this to its own branch so I could update it with that change at some point.

PR #2860, opened 2025-03-25
Move `status` of button from widget to `State`

mmitteregger:
When the button is being recreated as part of an overlay like in an [iced_aw::ContextMenu](https://github.com/iced-rs/iced_aw/blob/def1db9aac1e58a47e0c3127d4d4e95d724ca8ad/src/widget/context_menu.rs#L228-L259), but the widget tree state is managed outside, then the button will be rendered with the wrong status (the default is always `None`, so it will fall back to `Status::Disabled` rendering).

By moving the `status` from the widget data to the persistent `State` of the widget tree, the rendering works as expected.

This `Status` was added to the widget data as part of the [Reactive Rendering](https://github.com/iced-rs/iced/pull/2662) rework.
I noticed this with a local (only partially updated) branch of `iced_aw`, which is based on the latest iced master branch.

I am not sure about the internals of iced. Is there a reason this was done outside of the persistent widget tree State?

### DIFF
--- a/widget/src/button.rs
+++ b/widget/src/button.rs
@@ -81,7 +81,6 @@ where
     padding: Padding,
     clip: bool,
     class: Theme::Class<'a>,
-    status: Option<Status>,
 }
 
 enum OnPress<'a, Message> {
@@ -118,7 +117,6 @@ where
             padding: DEFAULT_PADDING,
             clip: false,
             class: Theme::default(),
-            status: None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


hecrj:
It is a part of the widget contract that at least one `RedrawRequested` event _must always_ precede a `Widget::draw` call.

If a widget reconstructs part of the widget tree during a redraw, then it must uphold this guarantee by re-feeding the event.

In any case, reconstructing part of the widget tree on every `overlay` call is not good.

PR #2861, opened 2025-03-26
`aspect_ratio` helper for `Responsive`

tigerros:
A new `aspect_ratio` function to create a `Responsive` with a given ratio.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -206,6 +206,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "aspect_ratio"
+version = "0.1.0"
+dependencies = [
+ "iced",
+]
+
+[[package]]
 name = "async-broadcast"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


airstrike:
For the record—and I understand this is subjective—I'm not a huge fan of the double closures API as it "feels out of place" (again subjective) relative to how other widgets work.

My current half-baked view is that the ideal implementation would be one that modifies `Responsive` to include additional fields such that it allows for configuration of its sizing strategy between:

1. the current "`Fill` parent in both axes" approach
2. `Fill` only the width axis
3. `Fill` only the height axis
4. fill by using an aspect ratio

I haven't explored a draft implementation yet but that's my current thinking in terms of what I've seen people need, what I'd like to have myself, and what I think would feel idiomatic for `iced`.

tigerros:
> double closures API as it "feels out of place" (again subjective) relative to how other widgets work

Could this be because there's not a lot of helpers/widgets that work with `Responsive`? If you wanted to create other `Responsive` helpers, you might encounter this pattern again. One closure to generate the `Element`, then another closure to (optionally) wrap the modified `Element` from the first one, because wrapping the whole `Responsive` outside of the helper doesn't produce the same behavior.

edwloef:
Could this maybe also be a helper function in the `layout` module, similar to `layout::atomic` but with a fixed aspect ratio?

tigerros:
Changed it to just a simple `Size` function because making it a `Responsive` had issues. Now it's simpler, easier to use and more flexible.

PR #2867, opened 2025-03-29
Removed bezier_tool from list of examples in the Widget trait docs

berserkware:
The [`bezier_tool`](https://github.com/iced-rs/iced/blob/latest/examples/bezier_tool/src/main.rs) example doesn't use the Widget trait, it uses canvas.


hecrj (approved):
Thanks!

PR #2869, opened 2025-03-29
Re-expose winit features for window servers in Linux

edwloef:
This makes #1777 user-configurable from the top-level iced crate. In my case, disabling the Wayland features saves me 38 dependencies.

This depends on https://github.com/hecrj/window_clipboard/pull/31 to be as useful as possible, so it's just a draft for now.


TechnoPorg:
I also had the same idea, and didn't see this PR before implementing it in my fork: https://github.com/TechnoPorg/iced/tree/x11-wayland-feature-split.

Mine extends the X11/Wayland feature split to more dependencies (similarly to yours, it depends on unpublished revisions of the window-clipboard and rfd crates), and I end up with these numbers for dependencies:

- Current main: 526 dependencies
- My fork with X11 only: 484 dependencies
- My fork with Wayland only: 505 dependencies

Let me know if you think I should open a PR; alternatively, feel free to subsume all my changes into yours!

edwloef:
Your branch looks a bunch more up-to-date than what I have here, if you'd like I can give you write permissions to this branch (if I can figure out how) so this doesn't need a new PR :)

However, I'd advise you to check over everything again to see whether there might be feature flags that were on before but are now disabled and aren't exposed by `iced`, but should either be on or exposed (the `softbuffer` `kms` feature comes to mind, I can't find where it's documented but there must be a reason why it's on by default).

Also, you seem to have forgotten `iced_tester?/wayland` in the `wayland` feature.

edwloef:
Uh... not sure why that closed lol

Issue #2325, opened 2024-03-12
Unicode characters not showing up on the Text Input step in the tour


unlessgames:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

![iced_chars](https://github.com/iced-rs/iced/assets/38718816/9b979a14-7699-41ac-81a3-2f2fb8985a2c)

### What is the expected behavior?

Expected to see unicode characters in the output as well. Not familiar with Rust but I assume a different string type should be used in the example?

### Version

master

### Operating System

Linux

7sDream:
Needs a font supports the character.

See:  https://docs.rs/iced/latest/iced/settings/struct.Settings.html#structfield.default_font

hecrj:
Ah, we need to enable `text::Shaping::Advanced` for the text there!

hecrj:
This was fixed by #3048.

<img width="1234" height="992" alt="Image" src="https://github.com/user-attachments/assets/c766c3a6-cfb5-4195-87e4-212d9088b413" />

PR #2870, opened 2025-03-29
advanced text shaping for text_input content text in tour example

JL710:
fix https://github.com/iced-rs/iced/issues/2325


hecrj:
Superseded by #3048.

PR #2873, opened 2025-04-03
refactor(checkbox): make label optional for checkbox

Krahos:
As discussed [here](https://discourse.iced.rs/t/shouldnt-checkbox-be-without-a-label), in this PR I made the label for `CheckBox` optional. I also removed it as parameter from the `checkbox` helper function and added a way to add the label with a builder pattern instead.

The documentation and examples are updated accordingly.


hecrj (approved):
Thanks!

Issue #2814, opened 2025-02-23
Widgets' statuses are always `None` in the `integration` example


rctlmk:
### Is your issue REALLY a bug?

- [x] My issue is indeed a bug!
- [x] I am not crazy! I will not fill out this form just to ask a question or request a feature. Pinky promise.

### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Is this issue related to iced?

- [x] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

Widgets ignore mouse hover in the [integration](https://github.com/iced-rs/iced/tree/34314b3f576fdfffdd82968a685020d4dca47992/examples/integration) example:

https://github.com/user-attachments/assets/80580495-4fc4-4a31-b6bd-3287f47155ce

Supposedly the example has been broken ever since reactive rendering was implemented for each corresponding widget.

### What is the expected behavior?

Should work like this ([42a2cb6](https://github.com/iced-rs/iced/commit/42a2cb6d4f78343f43d6a68a28e5502d9426ed2c)):

https://github.com/user-attachments/assets/891ffd5a-14b2-4ae6-a50d-eb7a3f352388

### Version

master

### Operating System

Windows

### Do you have any log output?

```shell
Some dbg! output:
[widget\src\slider.rs:449:9] self.status = None
[widget\src\slider.rs:449:9] self.status = None
[widget\src\slider.rs:449:9] self.status = None
```

hecrj:
This is fixed in `master`.

PR #2875, opened 2025-04-06
Queue `RedrawRequested` event before redrawing in `integration` example

DerpDays:
This is required since some widgets depend on the event being fired to update their state (e.g. hover).

The `RedrawRequested` is queued and submitted independently from the events so that the state is correct when redrawing. (queuing it before submitting with received events causes flickering)

Fixes: #2814


hecrj:
This has already been fixed in `master`.

PR #2876, opened 2025-04-07
Introduce `shader::Pipeline` trait for easier `wgpu` resource management

AustinEvansWX:
As discussed [here](https://discord.com/channels/628993209984614400/1358256909634896003/1358477489290350885) on Discord, discarding of cached `wgpu` resources can be tricky without knowing when a frame has finished drawing. 

This PR introduces the `Pipeline` trait which will be required to be implemented for all types put into `Storage`. The `end_frame` method on this trait will be triggered following each frame akin to how the internal pipelines have [`trim()`](https://github.com/AtmosWX/iced/blob/fc9eb0247d0f3e4e6b93f6b54bffc43bcd72d673/wgpu/src/lib.rs#L151).


hecrj:
Thanks!

Merged manually since I didn't have permissions to push to your branch.

PR #2883, opened 2025-04-09
Fix WebGL WGPU crash

pml68:


wgpu/src/triangle/msaa.rs
```diff
@@ -254,6 +254,8 @@ impl Targets {
 struct Ratio {
     u: f32,
     v: f32,
+    p0: f32,
+    p1: f32,
```
hecrj, line 258:
```suggestion
    _padding: [f32; 2]
```

hecrj (changes requested):

pml68:
I committed the requested change, can't see it here though for some reason

pml68:
Ah it's still „processing updates”

hecrj (approved):
Sweet! Thanks.

PR #2884, opened 2025-04-09
implement `Wrapping` column

edwloef:
this implements a wrapping column that behaves analogous to a wrapping row, by wrapping into multiple columns when there's not enough space to show all elements in one column. the implementation is almost line-for-line identical to the wrapping row, I've verified it works in my own project.


edwloef:
Rebased.

hecrj (approved):
Thanks!

Issue #2318, opened 2024-03-09
TextInput paragraph state cache


tzemanovic:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

Hi, I’m using iced with bevy using https://github.com/tasgon/bevy_iced and I’m trying to update it from iced 0.10 to 0.12, but the integration  of `TextInput` seems to be affected by the state caching introduced in https://github.com/iced-rs/iced/pull/2058.

The issue is that the paragraph update applied in TextInput’s `on_event` in by the `update_cache` closure sometimes gets out of sync with the actual value, because in the integration the input message is passed via bevy’s event system which is sometimes delayed by a frame, which causes the input to flicker.

From what I can see the flicker is due to `fn offset` being calculated using the cache state before the actual value is updated via the propagated event. This is even worse with `.secure(true)` input where the raw input is shown for a frame. I’m not yet too familiar with the iced internals so I’m not sure what would be a right approach to fix this (and if there are any other widgets that may be affected by this?), but I’m happy to collaborate on a patch. When I removed the call to `update_cache` the problem in the integration goes away. 

### What is the expected behavior?

Visual integrity on text input. Cache should not affect the view before the event is processed by the app.

### Version

crates.io release

### Operating System

Linux

PR #2886, opened 2025-04-10
Use strong reference when sending `Paragraph` and `Editor` to the ren…

srussvoll:
I'm working on the `bevy_iced` plugin for Bevy that integrates Iced into Bevy. Since Bevy is using pipelined rendering and Iced assumes non-pipelined rendering, some things don't work. For instance, when typing into a text input widget, the text will often disappear for a frame when a letter is typed resulting in blinking, and is reported in #2318. I started a thread in [Discord](https://discourse.iced.rs/t/how-to-support-pipelined-rendering/931) to discuss this problem more generically.

This PR fixes the race condition that causes the flashing text.

The following PR is also required to get `bevy_iced` to work with the main branch of iced:
- https://github.com/iced-rs/iced/pull/2693

### DIFF
--- a/graphics/src/text.rs
+++ b/graphics/src/text.rs
@@ -11,7 +11,9 @@ pub use cosmic_text;
 
 use crate::core::alignment;
 use crate::core::font::{self, Font};
-use crate::core::text::{Alignment, Shaping, Wrapping};
+use crate::core::text::{
+    Alignment, Editor as _, Paragraph as _, Shaping, Wrapping,
+};
 use crate::core::{Color, Pixels, Point, Rectangle, Size, Transformation};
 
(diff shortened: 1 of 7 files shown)


srussvoll:
Fixes #2318.

hecrj:
This is planned, but it's unfortunately not as easy as cloning `Arc`.

Specifically, `Editor` assumes a unique strong owned reference:

https://github.com/iced-rs/iced/blob/41e69b0a0d04fc34d6a26feb5bb0283069983bc5/graphics/src/text/editor.rs#L265-L267

We need to address those first while making sure that for the `iced` use case no cloning occurs.

PR #2890, opened 2025-04-13
Make `text_input::State` `select_range` public

Remmirad:
I would like to build a wrapper around a text_input that only allows selecting part of the content. This would allow me to do so.


hecrj (approved):
Thanks!

PR #2892, opened 2025-04-17
fix doc comment of `Length::is_fill`

edwloef:
the missing slash made the docs cut off in the middle of the sentence:

image


alex-ds13:
@edwloef You could fix the `iff` typo as well...

edwloef:
> @edwloef You could fix the `iff` typo as well...

I didn't even notice, thanks 😅

PR #2893, opened 2025-04-21
use `fn` for `Subscription::map`

edwloef:
As per [rfc 1558](https://github.com/rust-lang/rfcs/blob/master/text/1558-closure-to-fn-coercion.md), any non-capturing closure can be coerced to a function pointer, so this replaces a runtime check for non-capturing-ness with a compile-time check.


hecrj:
It's a closure so that we can leverage the fn item itself as part of the identifier of the `Subscription`.

Using a function pointer can cause issues when multiple codegen units are used, as the same function may be present multiple times in the final binary.

See #2237.

PR #2896, opened 2025-04-23
fix number of rows in grid when evenly distributed

tafia:
Fix calculation of number of rows in grid, which should be rounded up.
Also use `Vec::with_capacity` for nodes to potentially save some re-allocations.


hecrj (approved):
Good catch! Thanks.

tafia:
Fixed the lint error (didn't know about the simpler `div_ceil` function!)

PR #2900, opened 2025-04-24
Add blurring on MacOS, Windows and Linux (setup dependent)

pml68:
Extends #2728 to support Windows in addition to MacOS and Linux.

### DIFF
--- a/core/src/window/settings.rs
+++ b/core/src/window/settings.rs
@@ -62,6 +62,21 @@ pub struct Settings {
     /// Whether the window should be transparent.
     pub transparent: bool,
 
+    /// Whether the window should have a blurred background.
+    ///
+    /// Note that the blurry effect is only applied to transparent windows. You need to enable
+    /// [`Settings::transparent`] and set a proper opacity value to the background color with
+    /// `Application::style`.
+    ///
+    /// On Windows, it sets the [backdrop][winit_backdrop] to `TransientWindow`, making the window [acrylic][acrylic].
+    ///
+    /// On MacOS and Linux, it uses `set_blur`. For more details, please refer to the [winit documentation][winit_set_blur].
+    ///
+    /// [winit_backdrop]: https://docs.rs/winit/latest/winit/platform/windows/enum.BackdropType.html
+    /// [winit_set_blur]: https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_blur
+    /// [acrylic]: https://learn.microsoft.com/en-us/windows/apps/design/style/acrylic
+    pub blur: bool,
+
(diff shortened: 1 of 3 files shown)


airstrike:
I was just playing around with this PR. For some reason, I can only really see the blur effect when I use a dark background theme. 

I could be wrong, but it is my understanding that the actual app background color comes from the `Theme`'s implementation of the `Base` trait, which for `iced::Theme` is:

```rust
/// The default [`Style`] of a built-in [`Theme`].
pub fn default(theme: &Theme) -> Style {
    let palette = theme.extended_palette();

    Style {
        background_color: palette.background.base.color,
        text_color: palette.background.base.text,
    }
}
```

Below is a screenshot of a sample app if you run it with an `iced::Theme::Custom` defined as: 

```rust
impl App {
    fn theme(&self) -> iced::Theme {
        let default = iced::Theme::default();
        let palette = iced::theme::Palette {
            background: default.palette().background.scale_alpha(0.05),
            ..default.palette()
        };
        iced::Theme::custom("Transparent".to_string(), palette)
    }
    // ...
}
```

image

The top one is what I get when the system theme is set to dark, bottom when set to light. Neither

pml68:
The checks will fail because this depends on iced-rs/winit#16

watsaig:
@airstrike I know this was a while ago but have you looked into it any more since your comment? This actually isn't tied to blurring, it also happens on master and 0.13 with a transparent window. It's easy to test with the `styling` example (tying slider value to background opacity, adding `.transparent(true)` to the app initialization and making basically the same changes as you to the theme)

But if you set the background color to `iced::Color::TRANSPARENT`, it does go completely clear. It seems to be related to the rgb values of the color, so that even setting alpha = 0 actually causes different results depending on the color's components. There's some weird mac shenanigans there because when I first ran into this, screenshotting the window made the issue disappear ([discord link with photo + screenshot](https://discord.com/channels/628993209984614400/1021828532189528094/1336123315164549121))

watsaig:
FWIW I get a similar issue on Windows, too, where the light background isn't transparent but the dark background is. I have a modified styling example [in this branch](https://github.com/watsaig/iced/tree/blur-troubleshooting) to quickly play around with this (slider value affects background alpha).

Also on Windows, a fully transparent background is still blurred whereas on mac, blur only kicks in with alpha > 0.0 (the latter is quite nice, would be great if that behavior could be achieved in windows too)

hecrj:
@watsaig We likely need to pre-multiply the alpha in the final framebuffer.

pml68:
I'll create a new branch and open a new PR that simply adds blurring on Windows.

PR #2901, opened 2025-04-24
Indent and Unindent Text Editor Actions

Ultrasquid9:
Cosmic-Text filters out tab characters when doing edits, advising the use of their indent/unindent actions instead. This provides a way to make use of those actions.

### DIFF
--- a/core/src/text/editor.rs
+++ b/core/src/text/editor.rs
@@ -106,6 +106,10 @@ pub enum Edit {
     Paste(Arc<String>),
     /// Break the current line.
     Enter,
+    /// Indent the current line.
+    Indent,
+    /// Unindent the current line.
+    Unindent,
     /// Delete the previous character.
     Backspace,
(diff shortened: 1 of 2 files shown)


hecrj (approved):
Thanks!

Atreyagaurav:
Hi @Ultrasquid9,

Can you provide an example on how to use this new feature? The `content.perform(action)` doesn't show any difference. I don't see any options to add what `indent` means for the editor. Are we supposed to handle it separate from other `Action` for editor contents?

Atreyagaurav:
Leaving it here if someone else needs it:

I made it work by having a key bind send an event while pressing tab to a custom binding because https://docs.rs/iced/0.14.0/iced/widget/text_editor/enum.Binding.html does not have corresponding Indent/Unindent events.

On that note, it might be nice to have user determine how much the tab width should be for indentation.

PR #2902, opened 2025-04-25
add iced logo to Iced-Icons.ttf

edwloef:
I did my best to emulate the background of the iced logo in the iced helper widget. The gradient doesn't match quite right, but I'm not sure how to make it better with the existing `Gradient` type.

image


edwloef:
A bunch of Firefox Eyedropper'ing later, I think it's now pretty much as close as it's gonna get.
image

PR #2905, opened 2025-04-26
fix combo-box panic after 6034a0c

edwloef:
Now that the default impl for `diff` clears all children, this would panic at line 481 when selecting the combo-box's text input.

### DIFF
--- a/widget/src/combo_box.rs
+++ b/widget/src/combo_box.rs
@@ -509,6 +509,10 @@ where
         vec![widget::Tree::new(&self.text_input as &dyn Widget<_, _, _>)]
     }
 
+    fn diff(&self, _tree: &mut widget::Tree) {
+        // do nothing so the children don't get cleared
+    }
+
     fn update(
         &mut self,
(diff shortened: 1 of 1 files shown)


widget/src/combo_box.rs
```diff
@@ -509,6 +509,10 @@ where
         vec![widget::Tree::new(&self.text_input as &dyn Widget<_, _, _>)]
     }
 
+    fn diff(&self, tree: &mut widget::Tree) {
+        tree.diff_children(&[&self.text_input as &dyn Widget<_, _, _>]);
```
hecrj, line 513: I think we can leave this as a no-op; since the children of a `ComboBox` never changes.

hecrj (changes requested):

widget/src/combo_box.rs
edwloef: Makes sense 👍

hecrj (approved):

PR #2906, opened 2025-04-27
Fix scrollable scrolling by touch

lmaxyz:
I ran into problem with scrollable by touch. The scrollable widget scrolled even when touch wasn't under the scroll area. This patch fixes the problem. And as I see it doesn't have side effects.


lmaxyz:
Here is a demo, how fix is looking on device.
Left device has fix, and right one hasn't and scrollable widget works incorrectly. It scrolls content even if touch starts not on the scrollable area.

https://github.com/user-attachments/assets/15350d6d-3283-45b2-a79d-da8c733e1265

PR #2908, opened 2025-04-27
Add the current toggle state to `toggler::Status::Disabled`

pml68:
This makes it similar to checkbox, which also has `is_checked` in its `Disabled` status.


hecrj (approved):
Thanks!

PR #2910, opened 2025-04-29
Time Travel Debugging

hecrj:
# Introducing Comet

I never properly introduced [`comet`], so please let me.

[`comet`] is a new companion debugging tool that can be summoned by pressing F12 when the `debug` feature is enabled—replacing the old (and hideous) debug overlay.

https://github.com/user-attachments/assets/5afb74f5-65e6-4fac-9124-6c64bf496304

[`comet`]: https://github.com/iced-rs/comet

When the `debug` feature is enabled, an `iced` application will send performance metrics through a socket to any server implementing the `iced_beacon` protocol.

[`comet`] is one of such servers. It is built with iced itself and you can find the code over here: https://github.com/iced-rs/comet

Of course, this all means you could build your own custom debugger if you wanted to! It should all be reusable enough.

# Custom Performance Metrics

As I introduced in #2891, you can decorate any piece of code with [`debug::time`](https://docs.iced.rs/iced/debug/fn.time.html) and [`debug::time_with`](https://docs.iced.rs/iced/debug/fn.time_with.html) to time and visualize different parts of your app.

https://github.com/user-attachments/assets/0d47b8b6-3ead-4561-a0f6-00faf5877c28

# Time Traveling

As you know, iced uses the Elm Architecture. One of the advantages of this architecture is that application state can only be mutated in a single place: `update`—and only in response of a `Message`. Furthermore, impure side effects are encouraged to happen inside `Task` and `Subscription` which are run indirectly by the runtime.

Effectively, this means that if we have an initial state and a list of messages, then we should be able to replicate any state the application has been at _any_ point in time. In other words, we can __time travel__.

https://github.com/user-attachments/assets/2e49695c-0261-4b43-ac2e-8d7da5454c4b

https://github.com/user-attachments/assets/278953fe-d190-4a15-8847-520509a90482

https://github.com/user-attachments/assets/5ccf67e5-68cc-43a9-950b-f317bf4d1a6b

Pretty cool, huh? Time traveling can be enabled with the new `time-travel` feature. Activating it will force a `Clone` bound on your `Message` type.

It goes without saying but... the feature is very experimental! It will only work well if your `update` logic is _pure_; meaning it does not rely on external state (e.g. calling `Instant::now`).

Use it carefully, it may eat your laundry!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2453,6 +2453,7 @@ version = "0.14.0-dev"
 dependencies = [
  "iced_beacon",
  "iced_core",
+ "iced_futures",
  "log",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2464,6 +2464,7 @@ dependencies = [
  "iced_debug",
  "iced_program",
  "iced_widget",
+ "log",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,8 @@ markdown = ["iced_widget/markdown"]
 lazy = ["iced_widget/lazy"]
 # Enables a debug view in native platforms (press F12)
 debug = ["iced_winit/debug", "iced_devtools"]
+# Enables time-travel debugging (very experimental!)
+time-travel = ["debug", "iced_devtools/time-travel"]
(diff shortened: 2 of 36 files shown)


Tahinli:
These update is more than awesome. Thanks a lot!

PR #2914, opened 2025-04-29
0.13  multilang keybindings

pashaish:
Replaced keyboard::Key::Character("...") with Physical::Code(Code::...) to ensure that keyboard shortcuts work independently of the current keyboard layout. This improves reliability for non-Latin layouts (e.g., Cyrillic).


idoric:
I wonder if this PR would resolve [this libcosmic bug](https://github.com/pop-os/libcosmic/issues/795) at the same time.

PR #2919, opened 2025-05-01
[Fix] GL backend interpreting atlas texture as cube map vs 2d tex array when atlas layer depth == 6

bungoboingo:
We have a hack implemented to "fix" this for when we create the atlas, but not when we grow it!

A user hit [this edge case](https://github.com/iced-rs/iced/pull/2259#issuecomment-1951428273) when dynamically loading a bunch of images

I can't really think of a better way to do this other than allocating an entire extra texture layer on GL backend when layer count is 6 since wgpu is hard-coding this, less retarded suggestions welcome! This should force wgpu to detect `is_cube_compatible` as `false` and interpret it as a `GL_TEXTURE_2D_ARRAY`. [Relevant WGPU gl backend code](https://github.com/gfx-rs/wgpu/blob/50eb207a77963a3818e8d05f65da34f1f6e6018c/wgpu-hal/src/gles/mod.rs#L442).

Underlying issue that will actually fix the issue: https://github.com/gfx-rs/wgpu/issues/7428#issuecomment-2809787735

GL user confirmed this fixes his issue, I don't have a GL supported device to test with if anyone else wants to give it a go

### DIFF
--- a/wgpu/src/image/atlas.rs
+++ b/wgpu/src/image/atlas.rs
@@ -19,6 +19,7 @@ use std::sync::Arc;
 
 #[derive(Debug)]
 pub struct Atlas {
+    backend: wgpu::Backend,
     texture: wgpu::Texture,
     texture_view: wgpu::TextureView,
     texture_bind_group: wgpu::BindGroup,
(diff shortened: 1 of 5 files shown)


hecrj (approved):
Thanks!

I stored the `backend` in the `Atlas` to avoid the additional plumbing, since it's immutable anyways!

PR #2920, opened 2025-05-01
refactor scroll_to

JL710:
This PR unifies the `scroll_to` and `scroll_by` implementations. Before, only the `scroll_by` implementation did a check weather the `offset` is out of bounds. Therefore another check existed in the offset getter. So in case of `scroll_by`, two checks where done.

I decided to remove the check in the offset getter and adding one in the `scroll_to` implementation. While doing that I added additional parameters to the `scroll_to` operation of the scrollable. Having bounds the the `scroll_to` of the operation further enables other widgets implementing that trait to access the bounds. (I have that use case for a custom widget of mine)

### DIFF
--- a/core/src/widget/operation/scrollable.rs
+++ b/core/src/widget/operation/scrollable.rs
@@ -8,7 +8,12 @@ pub trait Scrollable {
     fn snap_to(&mut self, offset: RelativeOffset);
 
     /// Scroll the widget to the given [`AbsoluteOffset`] along the horizontal & vertical axis.
-    fn scroll_to(&mut self, offset: AbsoluteOffset);
+    fn scroll_to(
+        &mut self,
+        offset: AbsoluteOffset,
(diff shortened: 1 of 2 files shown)


hecrj:
As I said in Discord, this does not work because the contents of a scrollable can change between views without an operation.

The internal state of widgets must be fault-tolerant.

hecrj:
Your `scroll_to` use case may be valid, though. What do you exactly need the bounds for?

JL710:
The check, in my case, is done when it is set or the content changes. But it might be a better approach correcting it whenever I access it (just one place to validate/correct it). 
But that raises the question: why is there a check in `state::scroll` if it works without? (it works only without it if you change `absolute.min((content - viewport).max(0.0))` to `absolute.min(content - viewport).max(0.0)` in the `absolute` method)

PR #2925, opened 2025-05-03
Fix `quad` shader blending

### LINKED ISSUE #2924: Unexpected black pixels if enable border radius

### What happened?

After enabling rounded borders, unexpected black pixels often appear, which usually occur when radius >= width

```rust
use iced::border::radius;
use iced::theme::Palette;
use iced::widget::{button, container};
use iced::{color, Border, Element, Padding, Theme};
use std::default::Default;

fn main() -> iced::Result {
    iced::application(MainState::default, MainState::update, MainState::view)
        .theme(|_| {
            Theme::custom("miao_theme", Palette {
                background: color!(219, 240, 240),
                text: color!(0xE4F9F5),

                primary: color!(0xA6E3E9),
                success: color!(0xEAFFD0),
                warning: color!(0xFCE38A),
                danger: color!(0xF38181),
            })
        })
        .window_size((100.0, 100.0))
        .run()
}

#[derive(Default, Debug)]
struct MainState {}

#[derive(Debug, Clone)]
enum MainMessage {}

impl MainState {
    fn update(&mut self, message: MainMessage) {
    }

    fn view(&self) -> Element<MainMessage> {
        container(button("cat"))
            .style(|theme: &Theme| container::Style {
                border: Border {
                    color: theme.palette().primary,
                    width: 1.0,
                    radius: radius(10.0),
                },
                ..Default::default()
            })
            .padding(Padding::new(10.0))
            .into()
    }
}
```

Image

There will be no unexpected black pixels

### Version

master

### Operating System

Windows

```shell

```

hecrj:
Fixes #2924.

### DIFF
--- a/wgpu/src/shader/color/oklab.wgsl
+++ b/wgpu/src/shader/color/oklab.wgsl
@@ -20,7 +20,7 @@ fn interpolate_color(from_: vec4<f32>, to_: vec4<f32>, factor: f32) -> vec4<f32>
     var color = to_rgb * (mixed * mixed * mixed);
 
     // Alpha interpolation
-    color.a = mix(from_.a, to_.a, factor);
+    color = to_ * factor + from_ * (1.0 - factor);
 
     return color;
 }
--- a/wgpu/src/shader/color/oklab.wgsl
+++ b/wgpu/src/shader/color/oklab.wgsl
@@ -20,7 +20,7 @@ fn interpolate_color(from_: vec4<f32>, to_: vec4<f32>, factor: f32) -> vec4<f32>
     var color = to_rgb * (mixed * mixed * mixed);
 
     // Alpha interpolation
-    color = to_ * factor + from_ * (1.0 - factor);
+    color.a = mix(from_.a, to_.a, factor);
 
     return color;
 }
--- a/wgpu/src/shader/quad.wgsl
+++ b/wgpu/src/shader/quad.wgsl
@@ -5,6 +5,10 @@ struct Globals {
 
 @group(0) @binding(0) var<uniform> globals: Globals;
 
+fn blend(over: vec4<f32>, under: vec4<f32>, alpha: f32) -> vec4<f32> {
+    return over * alpha + under * (1.0 - alpha);
+}
+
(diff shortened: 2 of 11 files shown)


wgpu/src/shader/color/oklab.wgsl
```diff
@@ -20,7 +20,7 @@ fn interpolate_color(from_: vec4<f32>, to_: vec4<f32>, factor: f32) -> vec4<f32>
     var color = to_rgb * (mixed * mixed * mixed);
 
     // Alpha interpolation
-    color.a = mix(from_.a, to_.a, factor);
+    color = to_ * factor + from_ * (1.0 - factor);
```
bungoboingo, line 23: Pretty sure you could just call `color = mix(from_, to_, factor)` here instead, but maybe this is more clear that we are lerping?

wgpu/src/shader/quad.wgsl
```diff
@@ -5,6 +5,14 @@ struct Globals {
 
 @group(0) @binding(0) var<uniform> globals: Globals;
 
+fn premultiply(color: vec4<f32>) -> vec4<f32> {
+    return vec4(color.xyz * color.a, color.a);
+}
+
+fn blend(over: vec4<f32>, under: vec4<f32>, alpha: f32) -> vec4<f32> {
+    return over * alpha + under * (1.0 - alpha);
```
bungoboingo, line 13: same here, could use `mix(under, over, alpha)`

bungoboingo: Or maybe just get rid of this `blend` fn entirely and use `mix` instead

hecrj: I see. Alpha blending is equivalent to mixing with premultiplied alphas.

hecrj:
Premultiplying should actually be faster, since the actual correct blending operations without premultiplication would be:

image

The premultiplication allows us to use `mix` directly for all the channels:

image

Saving a bunch of multiplications and divisions to the actual compositing stage; which is likely handled by hardware very efficiently.

PR #2928, opened 2025-05-04
feat: make devtools generated by macro

Decodetalkers:
I am making a platfrom plugin lib of iced, and I made another Program trait myself for that platform, but I cannot implement the debug attach because it needs iced_program::Program. but my trait is simillar with iced_program::Program, so I think about make the key code to macro, then I can use the macro to make my attach


Decodetalkers:
prepare to be used place is here https://github.com/waycrate/exwlshelleventloop/blob/iced14/iced_sessionlock/src/build_pattern.rs#L121-L141

Issue #2851, opened 2025-03-20
Container.clip doesn't work with tiny-skia


lufte:
### Is your issue REALLY a bug?

- [x] My issue is indeed a bug!
- [x] I am not crazy! I will not fill out this form just to ask a question or request a feature. Pinky promise.

### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Is this issue related to iced?

- [x] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

Usage of Container.clip has no effect when running iced apps with tiny-skia. The overflown content is always shown.

### What is the expected behavior?

It should work the same as with wgpu.

### Version

master

### Operating System

Linux

### Do you have any log output?

```shell

```

PR #2929, opened 2025-05-04
Implement tinyskia text bounds

### LINKED ISSUE #2740: [tiny-skia] does not truncate long text inside of a TextInput when the window size is changed

### What happened?

Text that is entered inside of the `TextInput` widget does not get clipped inside when adjusting the size of the window.

Compiling this snippet on `13.1` with `default-features = false` and `features = ["tiny-skia"]`

```rust
fn main() {
    let _ = iced::run("", update, view);
}

fn update(ctx: &mut String, msg: String) {
    *ctx = msg;
}

fn view(ctx: &String) -> iced::Element<String> {
    iced::widget::container(
        iced::widget::text_input(
            "This is a very long string that should be clipped inside of the text input",
            ctx,
        )
        .on_input(|msg| msg),
    )
    .width(iced::Length::Fill)
    .padding([0, 300])
    .into()
}

```

https://github.com/user-attachments/assets/c35a977f-9320-4411-89a9-a3aa715346ad

Text should be clipped inside the `TextInput`, as is the case when using the `wgpu` backend.

https://github.com/user-attachments/assets/7f65ea15-0fe7-4a4d-b2dc-ab59fc81ddb2

### Version

crates.io release

### Operating System

Windows

```shell

```

DKolter:
Fixes #2740 

Paragraph clip bounds are currently marked todo in the tiny skia backend. This PR makes an effort to change this by:
1. Finding the intersection between the clip_bounds argument (layer bounds) and the clip_bounds of the paragraph
2. Adjusting the clip mask to the intersection bounds if the physical bounds extend past the clip_bounds, otherwise don't update the clip mask to avoid expensive clip mask drawing.

The original issue only mentions `text_input`, however `text_editor` and cached text have the same todo comment and are also handled by this PR.

### Before
grafik

### After
grafik

### DIFF
--- a/tiny_skia/src/engine.rs
+++ b/tiny_skia/src/engine.rs
@@ -339,10 +339,15 @@ impl Engine {
                 paragraph,
                 position,
                 color,
-                clip_bounds: _, // TODO
+                clip_bounds: local_clip_bounds,
                 transformation: local_transformation,
             } => {
                 let transformation = transformation * *local_transformation;
+                let Some(clip_bounds) =
+                    local_clip_bounds.intersection(&clip_bounds)
+                else {
+                    return;
+                };
 
                 let physical_bounds =
                     Rectangle::new(*position, paragraph.min_bounds)
@@ -352,8 +357,13 @@ impl Engine {
                     return;
                 }
 
-                let clip_mask = (!physical_bounds.is_within(&clip_bounds))
-                    .then_some(clip_mask as &_);
+                let clip_mask = match physical_bounds.is_within(&clip_bounds) {
+                    true => None,
+                    false => {
+                        adjust_clip_mask(clip_mask, clip_bounds);
+                        Some(clip_mask as &_)
+                    }
+                };
 
                 self.text_pipeline.draw_paragraph(
                     paragraph,
@@ -368,10 +378,15 @@ impl Engine {
                 editor,
                 position,
                 color,
-                clip_bounds: _, // TODO
+                clip_bounds: local_clip_bounds,
                 transformation: local_transformation,
             } => {
                 let transformation = transformation * *local_transformation;
+                let Some(clip_bounds) =
+                    local_clip_bounds.intersection(&clip_bounds)
+                else {
+                    return;
+                };
 
                 let physical_bounds =
                     Rectangle::new(*position, editor.bounds) * transformation;
@@ -380,8 +395,13 @@ impl Engine {
                     return;
                 }
 
-                let clip_mask = (!physical_bounds.is_within(&clip_bounds))
-                    .then_some(clip_mask as &_);
+                let clip_mask = match physical_bounds.is_within(&clip_bounds) {
+                    true => None,
+                    false => {
+                        adjust_clip_mask(clip_mask, clip_bounds);
+                        Some(clip_mask as &_)
+                    }
+                };
 
                 self.text_pipeline.draw_editor(
                     editor,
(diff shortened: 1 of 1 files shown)


lufte:
This looks like it also fixes #2851 <3

lufte:
After more testing, though, I found a bug which is hard to explain. You can reproduce it by cloning my app from this repo https://git.sr.ht/~lufte/vimini/tree/tiny-skia-clipping (make sure you're on the tiny-skia-clipping branch), running it with tiny-skia, and scrolling down on the default page that opens up. You should notice an increasingly bigger rectangle, same color as the background, that is drawn on top of the text at the top. The tabs in the navigation bar at the top are using the new feature.

Let me know if it's hard to reproduce and I can provide assistance.

DKolter:
Hey @lufte, I had issues reproducing the visual bug. Do you think you could try to reproduce it based on this branch? 
If not, a short video would also be a great start. Thank you for checking the PR out!

lufte:
@DKolter I was able to write a short program that shows the behavior, code below. Just run it with tiny-skia and scroll down, you should see how less and less text is shown.

main.rs
```rust
#[derive(Default)]
pub struct App {
}

#[derive(Debug)]
pub enum Message {
}

impl App {
    fn update(&mut self, message: Message) -> iced::Task<Message> {
        let mut command = iced::Task::none();
        return command;
    }

    fn view(&self) -> iced::Element<Message> {
        let mut tab_bar = iced::widget::Row::new().width(iced::Length::Fill);
        tab_bar = tab_bar.push(
            iced::widget::Container::new(
                iced::widget::Text::new("Tab 1")
                .wrapping(iced::widget::text::Wrapping::None)
            )
            .width(iced::Length::Fill)
            .padding(2)
            .clip(true)
        );
        tab_bar = tab_bar.push(
            iced::widget::Container::new(
                iced::widget::Text::new("Tab 2")
                .wrapping(iced::widget::text::Wrapping::None)
            )
            .width(iced::Length::Fill)
            .padding(2)
            .clip(true)
        );
        let mut tabbed_browser = iced::widget::Column::new();
        tabbed_browser = tabbed_browser.push(tab_bar);
        let mut text_column = iced::widget::Column::new();

        for _ in [0; 100] {
            text_column = text_column.push(
                iced::widget::Text::new("A cross-platform GUI library for Rust focused on simplicity and type-safety. Inspired by Elm.")
                .size(17)
                .shaping(iced::advanced::text::Shaping::Advanced)
            );
        }
        let document_viewer = iced::widget::Container::new(
            iced::widget::scrollable::Scrollable::new(
                iced::widget::Container::new(
                    text_column
                    .max_width(800)
                    .width(iced::Length::Fill)
                    .padding(10)
                )
                .center_x(iced::Length::Fill),
            )
        )
        .height(iced::Length::Fill);
        tabbed_browser = tabbed_browser.push(document_viewer);
        return iced::widget::Container::new(tabbed_browser)
        .width(iced::Length::Fill)
        .height(iced::Length::Fill)
        .into();
    }
}

    return iced::run(App::update, App::view);
}
```

Cargo.toml
```toml
[package]
name = "tiny-skia-bug"

[dependencies]
iced = { git = "https://github.com/DKolter/iced/", branch = "issue_2740", features=["advanced"] }
```

lufte:
Glad to be of help @DKolter. There is another bug, related with scrolling and tiny-skia but not with this patch here, for which you might be of help. If you're interested I'll try to come up with a minimal example for that too and get in touch with you.

DBJim:
This works great for me. I tested with different scaling (dpi) options on windows and it always looks good. Thanks very much for working on this  @DKolter 🙂

PR #2930, opened 2025-05-05
fix missing request_redraw on `Component`

alex-ds13:
There was a missing `shell.request_redraw();` on the `Component` widget. If some internal state changed it would rebuild the element properly but since there wasn't this redraw request, the draw wouldn't be called and the view wouldn't actually update.

_The core team is busy and does not have time to mentor nor babysit new contributors. If a member of the core team thinks that reviewing and understanding your work will take more time and effort than writing it from scratch by themselves, your contribution will be dismissed. It is your responsibility to communicate and figure out how to reduce the likelihood of this!_

_Read the contributing guidelines for more details: https://github.com/iced-rs/iced/blob/master/CONTRIBUTING.md_


hecrj (approved):
Thanks!

PR #2931, opened 2025-05-05
feat: muda integration

doronnac:
Due to the integration example being insufficient for my requirements, I took some time to integrate `muda` into the project. it's not all there yet but there's a working example (custom_menu).

Figured I'll share it right now so you can be part of the design in case you're interested in merging.

Important note: This has only been tested so far on MacOS.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,22 +3,6 @@
 version = 4
 
 [[package]]
-name = "ab_glyph"
-version = "0.2.29"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec3672c180e71eeaaac3a541fbbc5f5ad4def8b747c595ad30d674e43049f7b0"
-dependencies = [
- "ab_glyph_rasterizer",
- "owned_ttf_parser",
-]
-
-[[package]]
-name = "ab_glyph_rasterizer"
-version = "0.1.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
-
-[[package]]
 name = "addr2line"
 version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -32,19 +16,6 @@ name = "adler2"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
-
-[[package]]
-name = "ahash"
-version = "0.8.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
-dependencies = [
- "cfg-if",
- "getrandom 0.2.15",
- "once_cell",
- "version_check",
- "zerocopy 0.7.35",
-]
 
 [[package]]
 name = "aho-corasick"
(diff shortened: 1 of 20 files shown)


karolisr:
Nice, I am also using `muda` with my `iced` app (in a less integrated way). There may be a problem on Linux as `muda` (when using `winit`) needs a raw GTK window handle to add itself to the window. Even in `muda` `winit` example they do not demonstrate Linux support (but they do in the `tao` example). Just something to consider. I really hope native menu support comes to `iced`!

karolisr:
> @karolisr Thanks for the heads up, iirc Windows requires a raw handle as well.

Not quite, Windows requires the `raw_id`: `iced::window::get_raw_id` (`u64`)  not the reference to the object.

airstrike:
> To whom it may concern - I've successfully managed to create a tao poc with relatively minimal feature drops and file changes, you can check the example in examples/tao.
> 
> I had a chance to dive deeper into the code, and I believe there's a significant opportunity to reduce many abstractions and enable a plugin ecosystem to benefit more from the rapid ecosystem improvements and evolving requirements.
> 
> Due to an already pretty significant fork and lack of interest from core contributors, I've decided to close this PR and continue in [my own repo](https://github.com/doronnac/iced).

Hey, I think it's awesome that you're doing this! I've also wanted a menu bar in the past and thought of `tao` and `muda` but never had the bandwidth to dig in.

I'm not a core team member, but I'm not sure I agree there's lack of interest from core contributors. There may or may not be, it's too early to tell. This PR is 4 days old! Makes me wonder what your expectations were.

It may be the case that switching to `tao` is a good move for `iced` but it may also not be. It's always about trade offs and which make sense for `iced`.

I think it might be simpler if we can seamlessly integrate with `muda` without having to drop `winit`. Or maintaining a minor `muda` fork might be simpler than dealing with the wholesale `winit` replacement.

It doesn't mean exploring the idea isn't worthwhile. At a minimum we will know more than what we did before attempting it.

Either way, I would leave this PR open. I'm sure there are many more like me watching from the sidelines, and this way you'll at least get an answer from the core team.

Thanks again for your work here. We all benefit from it.

hecrj:
The GTK dependencies on Linux are a big price to pay just for system menus.

I'd like to try drawing the menus ourselves first.

In any case, the contributing guide mentions that the core team wants contributors to build trust first, specially before dropping a PR with 5k new lines of code.

Things take time.

karolisr:
This sucks, in some ways, I was really hoping I was wrong regarding Linux + winit + muda. In general, my feeling is that the "native" menus really only matter on macOS as they are drawn by the OS. I was a able to minimize the amount of code required to integrate muda in my app by replacing `crossbeam-channel` with `futures` which was basically a drop-in as the APIs were almost identical. Although it did result in a mutable static and thus a use of `unsafe` block (https://github.com/tauri-apps/muda/commit/170d13df2207f476a43eb7eec34e184cb00121b9). I just listen to the `muda` events using `sipper`-based `iced` subscription. My plan for Linux was to use `iced_aw` menus.

karolisr:
> The other unmet requirement is window events introduced in winit 0.30.9, providing both path and position during window file drag events. Love to hear your insights if you have any before I start forking again.

@doronnac I am not familiar with that, sorry. My requirements were met by just getting the path of the file dropped anywhere.

P.S. You may want to associate a specific file format with your app on macOS. One side is packaging the `.app` bundle with the correct `Info.plist` entries, to handle the macOS events the solution I found for that was in the `neovide` project https://github.com/neovide/neovide/pull/2395.
I adapted it here: https://github.com/karolisr/treehouse/blob/reorg/app/src/app/platform/macos/nsappdel.rs

doronnac:
@karolisr I originally needed this functionality for media import but this might actually be useful. Thanks!

PR #2932, opened 2025-05-05
Make Syntect use `fancy-regex` instead of `onig`

Ultrasquid9:
The Oniguruma C library is archived, and fails to build correctly on many systems. Fancy-regex is Rust based and still actively updated, and is able to be built just fine.


hecrj:
Thanks!

Merged manually.

PR #2933, opened 2025-05-05
Adds the ability to disable `allows_automatic_window_tabbing` on macOS.

karolisr:
macOS intrusively adds certain global menu items if a certain menu item is present and if the underlying feature of the `AppKit` is enabled. For example, if a `View` global menu item is present and the [`allowsAutomaticWindowTabbing`](https://developer.apple.com/documentation/appkit/nswindow/allowsAutomaticWindowTabbing) value is set to `true`, the `Show/Hide Tab Bar` item is added to the `View` submenu (see images). By default, `allowsAutomaticWindowTabbing` is set to `true`. This makes any application with a `View` menu item the ability to show/hide potentially useless tab bar. See attached images that show the described menu item and the tab bar in the window.

This PR exposes this setting in `settings::Settings` (`core/src/settings.rs`) and applies it in `iced_winit::run` function.

tab_bar_visible
tab_bar_hidden

### DIFF
--- a/core/src/settings.rs
+++ b/core/src/settings.rs
@@ -33,6 +33,14 @@ pub struct Settings {
     ///
     /// By default, it is enabled.
     pub antialiasing: bool,
+
+    #[cfg(target_os = "macos")]
+    /// Set whether the system can automatically organize windows into tabs.
+    ///
+    /// <https://developer.apple.com/documentation/appkit/nswindow/1646657-allowsautomaticwindowtabbing>
+    ///
+    /// The default value is `true`
+    pub allows_automatic_window_tabbing: bool,
 }
 
 impl Default for Settings {
@@ -43,6 +51,9 @@ impl Default for Settings {
             default_font: Font::default(),
             default_text_size: Pixels(16.0),
             antialiasing: true,
+
+            #[cfg(target_os = "macos")]
+            allows_automatic_window_tabbing: true
         }
     }
 }
--- a/winit/src/lib.rs
+++ b/winit/src/lib.rs
@@ -26,6 +26,9 @@ pub use runtime::debug;
 pub use runtime::futures;
 pub use winit;
 
+#[cfg(target_os = "macos")]
+use winit::platform::macos::ActiveEventLoopExtMacOS;
+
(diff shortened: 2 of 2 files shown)


dtzxporter:
This belongs in platform specific settings:

`core/src/window/settings/macos.rs`

karolisr:
Hmm, I think/thought this was an application-wide setting for allowing tabbing in windows (although it is under `NSWindow` in Apple's API). If this is indeed a per window setting, it is a method of `winit::platform::macos::ActiveEventLoopExtMacOS` as opposed to `winit::platform::macos::WindowAttributesExtMacOS` where other settings in `core/src/window/settings/macos.rs` live. The only place in `iced` where we have a reference to `winit::event_loop::ActiveEventLoop` is in `winit/src/lib.rs`. I am not quite familiar with `iced` enough to immediately understand where the call to the `ActiveEventLoop::set_allows_automatic_window_tabbing(&self, enabled: bool)` method should be placed as it does not appear to operate on a specific window instance, but on application windows as a whole.

airstrike:
See https://docs.iced.rs/iced/window/struct.Settings.html#structfield.platform_specific

Which is what @dtzxporter was referring to in his previous comment

Also thanks for your work on this and menu bars in general. As a user, I truly appreciate it.

karolisr:
> See https://docs.iced.rs/iced/window/struct.Settings.html#structfield.platform_specific
> 
> Which is what @dtzxporter was referring to in his previous comment
> 
> Also thanks for your work on this and menu bars in general. As a user, I truly appreciate it.

Thanks, I think you may have missed my point, the link and the @dtzxporter comment references _window_ specific settings, which I am aware of. This is not a _window_ specific setting, it is an application-wide setting that applies to all windows.

pml68:
> This is not a _window_ specific setting, it is an application-wide setting.

In this case you'd ideally create a `PlatformSpecific` struct in a similar fashion to `core/src/window/settings/*` and add a corresponding `platform_specific` field to `core::Settings`, instead of adding a field that only exists on one platform to the struct directly.

PR #2936, opened 2025-05-07
add warning status to toast

mariinkys:
This PR adds a Warning status to the toast example, which is probably missing because Extended::warning got added later


hecrj (approved):
Thanks!

PR #2938, opened 2025-05-09
Remove unnecessary `Arc`

a1phyr:
Since `wgpu` 24, most types are ref-counted so this is not necessary anymore.


hecrj:
Thanks! This was fixed in #3092 already.

PR #2945, opened 2025-05-15
feat: `impl Default for keyboard::Key`

nik-rev:
This is useful when you need "just any `Key`", e.g. my usage in particular is to `#[derive(strum::EnumIter)` on a type which includes the `Key`


hecrj:
I don't think the `Default` implementation for `Key` being `Unidentified` is reasonable. Seems like a footgun.

PR #2950, opened 2025-05-18
Fix inconsistent naming of `tree` parameter in `Widget` trait

### LINKED ISSUE #2949: Inconsistent naming of `tree` parameter in `Widget` trait

### What happened?

https://docs.iced.rs/iced/advanced/trait.Widget.html

The provided methods of the `Widget` trait use the name `_state`, whereas the required methods use `tree`, and it is these names which the IDE uses when auto-completing new methods that I want to `impl` on the trait.

`tree` is nicer, because then I can write `let state = tree.state.downcast_ref::<...>()` without shadowing the original `Tree`.

For consistency, it would be nicer to rename occurrences of `_state` to `_tree`. Saves me having to rewrite the parameter's name in its entirety vs just deleting the leading underscore if I want to use it in my `impl`.

.

### Version

master

### Operating System

Linux

```shell

```

edwloef:
fixes #2949

### DIFF
--- a/core/src/widget.rs
+++ b/core/src/widget.rs
@@ -103,7 +103,7 @@ where
     /// Applies an [`Operation`] to the [`Widget`].
     fn operate(
         &self,
-        _state: &mut Tree,
+        _tree: &mut Tree,
         _layout: Layout<'_>,
         _renderer: &Renderer,
         _operation: &mut dyn Operation,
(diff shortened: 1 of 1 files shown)


mfreeborn:
Just doing a quick Ctrl + F, there's a bunch of places in widget code/some examples which still use the old `state` parameter name, which should be updated to match the changes in the `Widget` trait.

PR #2951, opened 2025-05-20
chore: adding support for nix devshell and direnv

Krahos:
With this PR I added a nix flake, which should make it easier for people using the nix package manager to build Iced and contribute. It includes all the dependencies needed to build Iced (as far as I know), the Rust toolchain and a few utilities (bacon, rust-analyzer, ...). It also supports integration with direnv, to automatically download these dependencies once entering the Iced directory.
Attaching a short video as demonstration of how it works, for those unfamiliar with it.

https://github.com/user-attachments/assets/afd9a983-1e36-48a4-a596-0f855f34c80d

It's common to keep this file at the root of the project and it doesn't affect users not using nix.

I tested it on a GNU + Linux machine with Wayland, maybe some extra tweaks are needed for X11.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,42 @@
+{
+  description = "Rust devshell for Iced";
+
+  inputs = {
+    nixpkgs.url      = "github:NixOS/nixpkgs/nixos-unstable";
+    rust-overlay.url = "github:oxalica/rust-overlay";
+    flake-utils.url  = "github:numtide/flake-utils";
+  };
+
+  outputs = { nixpkgs, rust-overlay, flake-utils, ... }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        overlays = [ (import rust-overlay) ];
+        pkgs = import nixpkgs {
+          inherit system overlays;
+        };
+      in
+      with pkgs;
+      {
+        devShells.default = mkShell rec {
+          nativeBuildInputs = [
+            bacon
+            cargo-udeps
+            cargo-edit
+            rust-analyzer
+            cargo-deny
+            rust-bin.stable.latest.default
+
+            pkg-config
+            dbus
+          ];
+          buildInputs = [
+            libxkbcommon
+          ];
+          runtimeDependencies = [
+            wayland
+          ];
+          LD_LIBRARY_PATH = lib.makeLibraryPath (nativeBuildInputs ++ buildInputs ++ runtimeDependencies);
+        };
+      }
+    );
+}
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ pkg/
 **/*.rs.bk
 dist/
 traces/
+/.direnv
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,96 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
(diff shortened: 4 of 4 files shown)


pml68:
You should probably check out the `DEPENDENCIES.md` file for flake dependencies and such.

flake.nix
```diff
@@ -0,0 +1,42 @@
...
+
+            pkg-config
+            dbus
+          ];
+          buildInputs = [
+            libxkbcommon
+          ];
+          runtimeDependencies = [
+            wayland
+          ];
```
olebedev, line 37: These three sections can simple be `packages` or `buildInputs`. 

Since it's not only for Linux users, I assume macOS (as myself) users will also tune in, we have to put linux only dependencies behind a condition.

```suggestion
          buildInputs = [
            bacon
            cargo-udeps
            cargo-edit
            rust-analyzer
            cargo-deny
            rust-bin.stable.latest.default
            pkg-config
            dbus
            libxkbcommon
          ] ++ lib.optionals (stdenv.isLinux) [
            wayland
          ];
```

flake.nix
```diff
@@ -0,0 +1,42 @@
...
+          LD_LIBRARY_PATH = lib.makeLibraryPath (nativeBuildInputs ++ buildInputs ++ runtimeDependencies);
```
olebedev, line 38:
```suggestion
          runtimeDependencies = [
            wayland
          ];
```

This is what is happening by default in `mkShell`, we don't have to be explicit. Also, the name of this environment variable vary, depending on the platform.

olebedev (changes requested):

flake.nix
Krahos: For some reason, if I remove this, I get hit with an error:
```
thread 'main' panicked at /home/seb/projects/iced/winit/src/lib.rs:84:10:
Create event loop: Os(OsError { line: 81, file: "/home/seb/.cargo/git/checkouts/winit-865c6d5afb3ec9c3/11414b6/src/platform_impl/linux/wayland/event_loop/mod.rs", error: WaylandError(Connection(NoWaylandLib)) })
```

olebedev: Yeah, this is because you use the `runtimeDependencies` which is an attribute to add libraries to rpath but not into `LD_LIBRARY_PATH`. Initially you've done it manually via `lib.makeLibraryPath` but this can be avoided altogether.

Krahos: So, I've been trying different things, but maybe I'm misunderstanding something.
I moved all the dependencies to `buildInputs` and also removed the whole `LD_LIBRARY_PATH = lib.makeLibraryPath (nativeBuildInputs);` line (see the updated file). I still get the same runtime error when running `cargo run --package tour`.

Do you have any suggestions on what I could be missing?

Krahos: >     2. The above flake does not account for the Rust toolchain (for my use-case I have Rust installed on my system for now)
I think it would be better to include it, because the idea of dev flakes is to have a working environment out of the box.

>     3. This assumes a Vulkan backend rather than something like OpenGL.
This doesn't take into account old systems that do not support Vulkan. Since Iced is using wgpu, which is essentially an abstraction layer on top of Vulkan, OpenGL, DirectX, Metal, it would be better to not include this info.

> I'm moreso curious if this flake structure is good practice and will generally work for environments like `iced`.

So for flakes, as far as I understand, we need to be more generic. It has to work for every system supported by Iced (and by Nix, of course). But I'm far from being an expert. I still don't grasp the best practices with Nix.

mahlquistj:
@Krahos to my knowledge, `cargo-udeps` can't run on the stable channel, so unless `rustup` or the nightly toolchain is added, there is no way for a nix user to use `cargo-udeps` without installing `rustup` and the nightly toolchain themselves? Are `cargo-udeps` a must for working on `iced`? Just my two-cents here :smile: 

Also want to say that i love the idea of adding support for nix users, like myself! :+1:

Krahos:
> @Krahos to my knowledge, `cargo-udeps` can't run on the stable channel, so unless `rustup` or the nightly toolchain is added, there is no way for a nix user to use `cargo-udeps` without installing `rustup` and the nightly toolchain themselves? Are `cargo-udeps` a must for working on `iced`? Just my two-cents here 😄
> 
> Also want to say that i love the idea of adding support for nix users, like myself! 👍

Ah you're right! I sometimes use nightly and I like to keep udeps around to clean up stuff. But it's mostly a CI tool. I'll remove it.

hecrj:
Superseded by #2840.

PR #2952, opened 2025-05-21
Expose underlying mobile support from winit via cfg changes

markkimsal:
I understand mobile support is not a priority, and possibly even a detriment if it attracts more spurious bug reports.  But, with just 6 lines of code changes `iced` can expose underling `winit` android support.  This goes a long way to closing #302 with no ongoing maintenance from iced core team.

4 lines are cfg changes that more closely align with the underlying winit library

```diff
--- a/winit/src/conversion.rs
+++ b/winit/src/conversion.rs
@@ -203,13 +203,25 @@ pub fn window_event(
         WindowEvent::KeyboardInput { is_synthetic, .. } if is_synthetic => None,
         WindowEvent::KeyboardInput { event, .. } => Some(Event::Keyboard({
             let key = {
-                #[cfg(not(target_arch = "wasm32"))]
+                #[cfg(any(
+                    target_os="windows",
+                    target_os="macos",
+                    all(feature = "x11", all(unix, not(any(target_os = "ios", target_os = "macos")), not(target_os = "android"), not(target_os = "emscripten")), not(target_os = "redox")),
+                    all(feature = "wayland", all(unix, not(any(target_os = "ios", target_os = "macos")), not(target_os = "android"), not(target_os = "emscripten")), not(target_os = "redox")),
+                    target_os="redox",
+                ))]
```
Checking only for wasm32 let's unsupported platforms slip through the feature gate of KeyEventExtModifierSupplement, because the underlying core winit library uses a different set of platform checks.
    
This will allow more platform support transparently by matching the cfg of the underlying winit library.
    
`winit` 0.31 plans to remove config aliases in favor of raw cfg macros.  https://github.com/rust-windowing/winit/issues/3539

Lastly, re-exporting 2 `winit` features into the iced_winit workspace Cargo allows the underlying `winit` + `android-activity` to build when targeting `aarch64-linux-android`

```diff
--- a/winit/Cargo.toml
+++ b/winit/Cargo.toml
@@ -22,6 +22,8 @@ x11 = ["winit/x11"]
 wayland = ["winit/wayland"]
 wayland-dlopen = ["winit/wayland-dlopen"]
 wayland-csd-adwaita = ["winit/wayland-csd-adwaita"]
+android-native-activity = ["winit/android-native-activity"]
+android-game-activity = ["winit/android-game-activity"]
 unconditional-rendering = []
 ```

There are all the changes to allow successful building on android - fonts don't work, but that is another small pr.

### DIFF
--- a/winit/Cargo.toml
+++ b/winit/Cargo.toml
@@ -22,6 +22,8 @@ x11 = ["winit/x11"]
 wayland = ["winit/wayland"]
 wayland-dlopen = ["winit/wayland-dlopen"]
 wayland-csd-adwaita = ["winit/wayland-csd-adwaita"]
+android-native-activity = ["winit/android-native-activity"]
+android-game-activity = ["winit/android-game-activity"]
 unconditional-rendering = []
 
 [dependencies]
--- a/winit/src/conversion.rs
+++ b/winit/src/conversion.rs
@@ -203,13 +203,25 @@ pub fn window_event(
         WindowEvent::KeyboardInput { is_synthetic, .. } if is_synthetic => None,
         WindowEvent::KeyboardInput { event, .. } => Some(Event::Keyboard({
             let key = {
-                #[cfg(not(target_arch = "wasm32"))]
+                #[cfg(any(
+                    target_os="windows",
+                    target_os="macos",
+                    all(feature = "x11", all(unix, not(any(target_os = "ios", target_os = "macos")), not(target_os = "android"), not(target_os = "emscripten")), not(target_os = "redox")),
+                    all(feature = "wayland", all(unix, not(any(target_os = "ios", target_os = "macos")), not(target_os = "android"), not(target_os = "emscripten")), not(target_os = "redox")),
+                    target_os="redox",
+                ))]
                 {
                     use winit::platform::modifier_supplement::KeyEventExtModifierSupplement;
                     event.key_without_modifiers()
                 }
 
-                #[cfg(target_arch = "wasm32")]
+                #[cfg(not(any(
+                    target_os="windows",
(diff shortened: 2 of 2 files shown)


srfsh:
With this, does inputting work on Android? If so, do you happen to have an example repo where we can learn how to utilize this pr?

markkimsal:
This does not add any functionality to iced.  It merely changes some of the `cfg` checks from basic "is WASM or not" to enumerating all the supported desktop platforms, making sure to exclude "linux-android" from the X11 or Wayland checks.

This makes it so that if you want to experiment with Android support, you are getting more features from the underlying winit dependencies made available to you.

The code changes required to make an android app is probably not appropriate for the core iced-rs project, since android requires you to compile a lib with an exposed, no-mangle function called "android_main".  Each example would require a lib.rs with some if[cfg()] checks...  This also requires injecting a special event loop from android into iced.

There's only one example - the integration example - that can run on android with minimal changes, because of the way the integration example injects an event loop into iced.

You can explore ibaryshnikov's example repo.  His changes to core `iced-rs` are very similar to this change, but I feel this change is more inline with winit's future direction of removing cfg aliases.

Core changes:
https://github.com/ibaryshnikov/iced/commit/009bf6cb62f7528c7149e4c5db815a85da129ba4

Example repo:
https://github.com/ibaryshnikov/android-iced-example

Quackdoc:
Does that mean iced will not be compatible with x11 or wayland when compiled with android targets? If so that is sad to hear. I know that x11 and wayland compilation for android is not great in the rust ecosystem, but it would be nice to not break it even further.

hecrj:
I can't maintain any kind of mobile support.

Issue #2825, opened 2025-03-04
SVG DPI scaling issue with tiny-skia


njust:
### Is your issue REALLY a bug?

- [x] My issue is indeed a bug!
- [x] I am not crazy! I will not fill out this form just to ask a question or request a feature. Pinky promise.

### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Is this issue related to iced?

- [x] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

When using an SVG on macOS with tiny-skia the layout is not correct:

```rust
pub const LOGO: &'static [u8] = include_bytes!("../res/circle.svg");

pub fn logo() -> Svg<'static> {
    svg(svg::Handle::from_memory(LOGO))
}

pub fn banner<'a, T: 'static>(title: &'static str) -> Element<'a, T> {
    let header_content = row![
        container(logo().height(64).width(64)).padding(Padding {
            right: 20.,
            ..Default::default()
        }),
        column![text("Header").size(24), text(title).size(20)]
    ];

    container(header_content)
        .padding(12)
        .width(Fill)
        .style(|_| container::Style {
            background: Some(Background::Color(Color::from_rgb8(246, 247, 250))),
            ..Default::default()
        })
        .into()
}

```

<img width="1041" alt="Image" src="https://github.com/user-attachments/assets/cfab2347-8d26-4522-935c-9e547d5788f7" />

It works fine using wgpu. I created a minimal repro for the issue here:
https://github.com/njust/iced-svg-issue.git

The issue exists on iced 0.13.1 and the current master (beddf49c76f3b43e7b33452831e1293cad28fd89).

### What is the expected behavior?

No layout issues. On Linux and Windows it's rendered as expected:

# Linux
![Image](https://github.com/user-attachments/assets/428a6578-14b7-497e-999b-e5c91f1bf0c6)

# Windows
![Image](https://github.com/user-attachments/assets/4b158dd1-4df6-4842-a711-5866565cdf35)

### Version

crates.io release

### Operating System

macOS

### Do you have any log output?

```shell

```

flakes:
It doesn't just affect SVG, but also other renderings that use transformations, even simple translations like are ignored by tiny-skia:

```rust
let bounds_translation = Vector::new(
    (bounds.x + (viewport.width - width_float) * 0.5).max(bounds.x), // center horizontally
    bounds.y + y_offset,
);

renderer.with_translation(bounds_translation, |renderer| {
     // ...
});
[...]

PR #2953, opened 2025-05-21
Fix scaling error of SVGs

semiversus:
This small pull-requests solves a problem when SVG are drawn with DPI scaling factor != 1.0 and tiny-skia. The problem is described in #2825.

In the `draw_image` function of `Engine` (in `tiny_skia/src/engine.rs`) the variable `pyhsical_bounds` is used, but it should be `*bounds` (same as in the `Image:Raster` arm).

```rust
                self.vector_pipeline.draw(
                    &svg.handle,
                    svg.color,
                    physical_bounds,  // <- *bounds is expected
                    svg.opacity,
                    _pixels,
                    transform,
                    clip_mask,
                );
```

### DIFF
--- a/tiny_skia/src/engine.rs
+++ b/tiny_skia/src/engine.rs
@@ -602,7 +602,7 @@ impl Engine {
                 self.vector_pipeline.draw(
                     &svg.handle,
                     svg.color,
-                    physical_bounds,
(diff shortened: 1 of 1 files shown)


semiversus:
Closed in favor of #2954

PR #2955, opened 2025-05-23
Implement `Debug` for `Task`

watsaig:
Just a small quality-of-life thing: implementing Debug for `Task` so that it's easy to to use in, say, an enum returned by `update` and to be able to derive Debug on that enum.

### DIFF
--- a/runtime/src/task.rs
+++ b/runtime/src/task.rs
@@ -17,7 +17,6 @@ pub use sipper::{Never, Sender, Sipper, Straw, sipper, stream};
 /// A set of concurrent actions to be performed by the iced runtime.
 ///
 /// A [`Task`] _may_ produce a bunch of values of type `T`.
-#[allow(missing_debug_implementations)]
 #[must_use = "`Task` must be returned to the runtime to take effect; normally in your `update` or `new` functions."]
 pub struct Task<T> {
     stream: Option<BoxStream<Action<T>>>,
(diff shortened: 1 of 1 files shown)


pml68:
It would be nice to remove the Debug trait bound on T, since you're not making use of it anyway, and T is supposed to be without bounds anyway - otherwise great!

runtime/src/task.rs
```diff
@@ -278,6 +277,17 @@ impl<T> Task<T> {
...
 }
 
+impl<T> std::fmt::Debug for Task<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Task<{}>, units={}",
+            std::any::type_name::<T>(),
+            self.units
+        )
```
hecrj, line 287: Let's use [`debug_struct`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct) here!

hecrj (changes requested):

runtime/src/task.rs
watsaig: Done. Including the type name in the struct name feels a bit unorthodox but including it as a field was worse imo. Let me know if you'd prefer to have it formatted differently!

hecrj (approved):
Looks good! Thanks.

PR #2957, opened 2025-05-25
feat: Pixel perfect quads

### LINKED ISSUE #2894: partially rounded buttons can sometimes produce odd-looking results

### What happened?

rounding only _some_ of the corners of a button produces odd-looking results:

```rs
#[derive(Default)]
struct State;

impl State {

    fn view(&self) -> iced::Element<'_, ()> {
        iced::widget::container(
            iced::widget::button("test")
                .style(|theme, status| {
                    let mut style = iced::widget::button::primary(theme, status);
                    style.border.radius = iced::border::left(5.0);
                    style
                })
                .on_press(()),
        )
        .width(iced::Fill)
        .height(iced::Fill)
        .align_y(iced::Alignment::Center)
        .into()
    }
}

fn main() -> iced::Result {
    iced::application(State::default, State::update, State::view)
        .theme(|_| iced::Theme::CatppuccinFrappe)
        .run()
}
```

When resizing the window, the button occasionally looks like this:
Image

this affects both the crates.io release and current master, and doesn't affect tiny-skia, only wgpu.

the button doesn't have a visible border between the two rounding types

### Version

master

### Operating System

Linux

```shell

```

Koranir:
Fixes #2894

This PR makes quads snap to the nearest pixel, and adapts the SDF method used to give consistent results in all situations.

This means that quads will no longer have blurry edges in almost all situations.

This has some drawbacks - quads don't have subpixel detail after this which means that while resizing or moving, they snap by one-pixel increments, which makes them a little less smooth. Other frameworks like Qt have this same issue however, so this is an acceptable tradeoff.

# TODO:
- [ ] Implement for gradients
- [ ] Clean up code
- [ ] Investigate whether using switch statements in shader code are necessary (could use a multiply-add approach)

### DIFF
--- a/examples/custom_quad/src/main.rs
+++ b/examples/custom_quad/src/main.rs
@@ -107,6 +107,7 @@ enum Message {
     ShadowXOffsetChanged(f32),
     ShadowYOffsetChanged(f32),
     ShadowBlurRadiusChanged(f32),
+    None,
 }
 
 impl Example {
@@ -148,6 +149,7 @@ impl Example {
             Message::ShadowBlurRadiusChanged(s) => {
                 self.shadow.blur_radius = s;
             }
+            Message::None => {}
         }
     }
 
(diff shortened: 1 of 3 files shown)


hecrj:
This should be superseded by #2962; which is cleaner and ensures alignment consistency with images.

PR #2959, opened 2025-05-25
Avoids artifacts when drawing small arcs by casting `f32` to `f64`.

karolisr:
I ran into an `f32` related bug while drawing large phylogenetic trees that required subdividing a circle into ~400k sections (see images). The solution is an `f64` cast on the `lyon_geom::Arc` object before getting quadratic bezier curves (`for_each_quadratic_bezier`) before drawing.

f32_artifacts
f64


hecrj (approved):
Thanks! That looks cool.

PR #2960, opened 2025-05-25
Make Tooltip show after a delay

kiedtl:
Require the user to hover over the tooltip source for a brief duration before the tooltip is made visible.

The time period is configurable, and can be disabled by setting it to zero.


botiapa:
If the conflicts were to be resolved, can this be merged? I'd be more than happy to resolve the merge conflicts.

PR #2966, opened 2025-05-29
Fix Rectangle::is_within inconsistency

### LINKED ISSUE #2965: Rectangle::is_within inconsistency causing clip mask issues with tiny skia

### What happened?

The current implementation of `Rectangle::is_within` relies on `Rectangle::contains`. 
After an [issue](https://github.com/iced-rs/iced/issues/2010) with being able to click two buttons at once, the `contains` function was [changed](https://github.com/iced-rs/iced/commit/f5b95629009ecde8c6f6388c8f33ec43d30d17d1#diff-55b572a71f92d84b0c7eb6a5054342e9a89fcd61a75a67fd7268ed664867f072L77) to return `false` for the bottom- and rightmost points, changing the behavior of `Rectangle::is_within` in the process without documentation. 

This leads to the following black rectangle to be considered "within" the outer rectangle:
Image

But this black rectangle is considered outside of the outer white rectangle:
Image

This causes issues with tiny skia's clip mask code, which unnecessarily updates clip masks for quads and text when any axis is at 100% of the outer clip bounds and decreases scrolling performance.

Return predictable values from `Rectangle::is_within` by not relying on `contains`.  

### Version

master

### Operating System

Linux

```shell

```

DKolter:
Fixes #2965 

Also documents the current behavior of `Rectangle::contains` .

### DIFF
--- a/core/src/rectangle.rs
+++ b/core/src/rectangle.rs
@@ -137,6 +137,7 @@ impl Rectangle<f32> {
     }
 
     /// Returns true if the given [`Point`] is contained in the [`Rectangle`].
+    /// Excludes the right and bottom edges.
     pub fn contains(&self, point: Point) -> bool {
         self.x <= point.x
             && point.x < self.x + self.width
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Makes sense! Thanks.

semiversus:
This change introduces a really stange bug. When compiled with`--release`, the window is just a copy of another screen area. Without `--release` it works as normal. I'll try to get a minimal example to show the effect. On wgpu it's working, only tiny-skia is affected.

PR #2967, opened 2025-05-29
Use a single SDF for all quad corners

### LINKED ISSUE #2894: partially rounded buttons can sometimes produce odd-looking results

Koranir:
Fixes #2894, extracted from #2957

This PR makes the quad shader just use a single SDF, instead of the weird stuff it was doing before. This also means we only need to run the SDF once when using a border instead of twice.

### DIFF
--- a/wgpu/src/shader/quad.wgsl
+++ b/wgpu/src/shader/quad.wgsl
@@ -5,31 +5,9 @@ struct Globals {
 
 @group(0) @binding(0) var<uniform> globals: Globals;
 
-fn distance_alg(
-    frag_coord: vec2<f32>,
-    position: vec2<f32>,
-    size: vec2<f32>,
-    radius: f32
-) -> f32 {
-    var inner_half_size: vec2<f32> = (size - vec2<f32>(radius, radius) * 2.0) / 2.0;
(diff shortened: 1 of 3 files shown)


hecrj (approved):
image

Looks great. Happy to see all that weird shader logic gone!

PR #2970, opened 2025-05-29
Expose a method to get the scroll_top/scroll_last_visible_line of the text_editor widget.

dtzxporter:
- Now we can implement virtual scrolling!


dtzxporter:
This needs to ideally be in a text_editor refactor. It's not reasonable to implement externally.

PR #2975, opened 2025-06-01
fix: Wgpu text bounds cutoff

### LINKED ISSUE #2963: Cut-off placeholder in combo box

### What happened?

The placeholder text in the input of a combo box is clipped off at the bottom, depending on the placeholder provided.

Image

No cut-off text

### Version

master

### Operating System

Linux

```shell

```

DKolter:
Fixes #2963 

The current wgpu text bounds implementation casts the right and bottom bound to an integer without rounding, leading to letters not being displayed fully. 

In this example there is a combo box on the left with a `0.2px` vertical space and an input box on the right without a vertical space. The default font size is `16px`, leading to a default line height of `1.3 * 16px = 20.8px`. With a `0.2px` vertical space the bottom bound equals to a whole integer coordinate, so the letters are displayed correctly. On the right the bound is cast from `20.8px` to `20px`, loosing almost a pixel of all letters extending to the bottom line:

Bildschirmfoto vom 2025-06-01 12-41-32

This PR matches the line bound calculation of [tinyskia](https://github.com/iced-rs/iced/blob/ca6d992d6751bb523eaf8f14a3d2678df77cea77/tiny_skia/src/text.rs#L226) and `round`s the text bounds:

grafik

### DIFF
--- a/wgpu/src/text.rs
+++ b/wgpu/src/text.rs
@@ -628,8 +628,8 @@ fn prepare(
                 bounds: cryoglyph::TextBounds {
                     left: clip_bounds.x as i32,
                     top: clip_bounds.y as i32,
-                    right: (clip_bounds.x + clip_bounds.width) as i32,
-                    bottom: (clip_bounds.y + clip_bounds.height) as i32,
+                    right: (clip_bounds.x + clip_bounds.width).round() as i32,
+                    bottom: (clip_bounds.y + clip_bounds.height).round() as i32,
                 },
                 default_color: to_color(color),
             })


hecrj:
Is this fixed by #2768?

edwloef:
No, however the result is a similar kind of snapping, the only difference being the top left point being truncated instead of rounded. Maybe that could be done here as well, though I can hardly say whether it'll be an issue in this case.

This needs a cryoglyph rectangle, which has signed coordinates (for some reason) so the existing snap fn wasn't used before or now.

wgpu/src/text.rs
```diff
@@ -628,8 +628,8 @@ fn prepare(
                 bounds: cryoglyph::TextBounds {
                     left: clip_bounds.x as i32,
                     top: clip_bounds.y as i32,
```
hecrj, line 630: Let's round these too as well while we are at it.

hecrj (changes requested):

hecrj (approved):
Thanks!

Issue #2332, opened 2024-03-16
tiny-skia crashes on long texts


Hyphrio:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

On long texts without a space, attempting to resize the window cause a crash.

Minimal reproduction:
```
use iced::{
    widget::{button, column, horizontal_space, row, text},
    Application, Settings, Size,
};

fn main() {
    App::run(Settings {
        window: iced::window::Settings {
            min_size: Some(Size {
                width: 500.0,
                height: 600.0,
            }),
            ..Default::default()
        },
        ..Default::default()
    })
    .unwrap();
}

struct App {}

impl iced::Sandbox for App {
    type Message = ();

    fn new() -> Self {
        Self {}
    }

    fn title(&self) -> String {
        "Iced crash".to_owned()
    }

    fn update(&mut self, _message: Self::Message) {}

    fn view(&self) -> iced::Element<'_, Self::Message> {
        column![
            row![
            text("Autametvoluptatibusaccusamusaspernaturessedolorelaboriosamfacilis.Dictadictaeaeaminimaamaioresetanimi.Estinventoreteneturomnis."),
            horizontal_space(),
            button("abcdef")
        ].padding(8)
        ].padding(20)
        .into()
    }
}
```

### What is the expected behavior?

Text either wraps or splits

### Version

crates.io release

### Operating System

Windows

### Do you have any log output?

```shell
Finished dev [unoptimized + debuginfo] target(s) in 0.36s
     Running `C:\Users\Hyphr\.cargo\target\debug\iced_ts_lt_crash.exe`
thread 'main' panicked at C:\Users\Hyphr\.cargo\registry\src\index.crates.io-6f17d22bba15001f\iced_tiny_skia-0.12.1\src\backend.rs:158:17:
Quad with non-normal width!
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\panicking.rs:645
   1: core::panicking::panic_fmt
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\core\src\panicking.rs:72
   2: iced_tiny_skia::backend::Backend::draw_primitive
   3: iced_tiny_skia::backend::Backend::draw<alloc::string::String>
[...]

PR #2977, opened 2025-06-03
Fix tiny skia primitive rendering

### LINKED ISSUE #2934: Missing redraw of canvas on window resize

### What happened?

I'm running into an issue where (former) regions of a canvas, which were moved due to a resize of the window, are not redrawn. The issue can be reproduced on Linux with the loading_spinners example:

https://github.com/user-attachments/assets/6f6bd034-0578-491f-9c98-cfd72d02a97e

Proper redraw after the canvas was moved

### Version

crates.io release

### Operating System

Linux

```shell

```

DKolter:
Fixes #2934 

The `loading_spinners` example is currently broken on master with tiny skia, not rendering the circular progress bars correctly and crashing after a second with a debug assert `Quad with non-normal width`. 

https://github.com/user-attachments/assets/14934b7e-fc89-4769-846e-c51d0c7125c3

The circular loading spinner is implemented with `Renderer::draw_geometry`, different from the linear spinner using `Renderer::fill_quad`, which is working fine.

Damage calculation does not currently respect the transformation of a cached primitive, leading to all circular spinners having their bounds set to `Rectangle { x: 0.0, y: 0.0, width: 40.0, height: 40.0 }`. 
The same issue applies to calculating the clip bounds, which also doesn't respect the primitive group transformation.

This PR ...
- [x] Applies transformation to clip_bounds and damage bounds for cached primitives
- [x] Renames bounds in tiny skia's draw function to avoid names like `new_clip_bounds`
- [x] Removes the debug assertion which causes issues with zero width quads, crashing the example. Replacing the debug check with `is_finite()` also is not a solution, since that crashes for layout explaining in the tour example.

Result:

https://github.com/user-attachments/assets/b97b546e-5dfa-44c0-af91-f09cd52b1f11

### DIFF
--- a/tiny_skia/src/engine.rs
+++ b/tiny_skia/src/engine.rs
@@ -36,15 +36,6 @@ impl Engine {
         clip_mask: &mut tiny_skia::Mask,
         clip_bounds: Rectangle,
     ) {
-        debug_assert!(
-            quad.bounds.width.is_normal(),
-            "Quad with non-normal width!"
-        );
-        debug_assert!(
-            quad.bounds.height.is_normal(),
-            "Quad with non-normal height!"
-        );
-
         let physical_bounds = quad.bounds * transformation;
 
         if !clip_bounds.intersects(&physical_bounds) {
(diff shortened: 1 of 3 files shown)


MCOfficer:
This should also fix #2332

PR #2979, opened 2025-06-03
Fix Path::circle artifact

### LINKED ISSUE #2976: Arc/Circle Stroke has break in path on wgpu backend

### What happened?

When `stroke`ing a circle or arc on wgpu backend (notably, this issue does not appear to be present when using tiny-skia), the path appears to have a small break in it:

Image

Minimal Example:
```rust
use iced::{mouse, Color, Element, Task, Theme};
use iced::widget::canvas;
use iced::{Rectangle, Renderer};

struct State {}

#[derive(Debug, Clone)]
enum Message {}

// First, we define the data we need for drawing
#[derive(Debug)]
struct Circle {
    radius: f32,
}

// Then, we implement the `Program` trait
impl<Message> canvas::Program<Message> for Circle {
    // No internal state
    type State = ();

    fn draw(
        &self,
        _state: &(),
        renderer: &Renderer,
        _theme: &Theme,
        bounds: Rectangle,
        _cursor: mouse::Cursor
    ) -> Vec<canvas::Geometry> {
        // We prepare a new `Frame`
        let mut frame = canvas::Frame::new(renderer, bounds.size());

        // We create a `Path` representing a simple circle
        let circle = canvas::Path::circle(frame.center(), self.radius);

        // And fill it with some color
        frame.stroke(&circle, canvas::Stroke::default().with_color(Color::WHITE).with_width(10.0));

        // Then, we produce the geometry
        vec![frame.into_geometry()]
    }
}

// Finally, we simply use our `Circle` to create the `Canvas`!
fn view(_state: &State) -> Element<Message> {
    canvas(Circle { radius: 20.0 }).into()
}

fn update(_state: &mut State, _message: Message) -> Task<Message> {
    Task::none()
}

fn main() {
    iced::application(|| State {}, update, view).run().unwrap();
}
```

The circle is unbroken.

### Version

master

### Operating System

macOS

```shell

```

DKolter:
Fixes #2976 

The `Path::circle` helper function does not close the path after constructing the circle, leading to a small artifact on the right between the start and end point.

Before | After
--- | ---
Bildschirmfoto vom 2025-06-03 15-00-59 | Bildschirmfoto vom 2025-06-03 15-02-26

### DIFF
--- a/graphics/src/geometry/path.rs
+++ b/graphics/src/geometry/path.rs
@@ -61,7 +61,10 @@ impl Path {
     /// Creates a new [`Path`] representing a circle given its center
     /// coordinate and its radius.
     pub fn circle(center: Point, radius: f32) -> Self {
-        Self::new(|p| p.circle(center, radius))
+        Self::new(|p| {
+            p.circle(center, radius);
+            p.close();
+        })
     }
 
     /// Returns the internal [`lyon_path::Path`].


hecrj (approved):
Thanks!

PR #2981, opened 2025-06-06
Add `filter_map` method to `Subscription`

watsaig:
Just a handy way to go from `Subscription<Option<T>>` to `Subscription<T>`, using [`StreamExt::filter_map`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.filter_map)


hecrj (approved):
I haven't been a fan of treating subscriptions like composable units in the past; but I can't seem to argue that position reasonably today.

So, what the heck. Let's go ahead with this and see how it goes.

Thanks!

PR #2982, opened 2025-06-06
Expose macOS ActivationPolicy and activate_ignoring_other_apps

Exidex:
Second attempt on https://github.com/iced-rs/iced/pull/2524, this time with global settings instead of window specific ones


hecrj:
Application-wide platform-specific APIs that are not purely visual (or very OS-specific) are a hard sell.

What is the use case this API is tackling and how can we support the other platforms as well? 

If it's an "always on top" setting for application launchers, then maybe we could try to tackle that directly instead of exposing the implementation details to achieve it in a specific OS.

I am closing this for now since it seems we need to have a design conversation first.

PR #2985, opened 2025-06-11
Add select_range operation/helpers for `TextInput`

wbk:
https://github.com/user-attachments/assets/29559378-0202-4887-a5a5-a4d61a977ef6

I have a use case for select_range. Currently, I'm wrapping a TextInput in a component, providing tab completion in the wrapper component, which is then selected after the value is replaced by the autocompleted text.

https://github.com/user-attachments/assets/1ff0a5b4-a9e8-4898-8de7-1ec8130bfe39

I'd like to select only the text that is autocompleted, so that the user is able to manipulate it without additional input to reposition their cursor in the widget.

This is clearly cargo culted, but straightforward and appears complete/working as intended.

I hope you consider merging it.

### DIFF
--- a/core/src/widget/operation/text_input.rs
+++ b/core/src/widget/operation/text_input.rs
@@ -13,6 +13,8 @@ pub trait TextInput {
     fn move_cursor_to(&mut self, position: usize);
     /// Selects all the content of the text input.
     fn select_all(&mut self);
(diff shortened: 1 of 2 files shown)


wbk:
I just realized this is essentially identical to #2890, sorry for the dupe

PR #2986, opened 2025-06-12
 Fix panic when drawing empty image in `iced_tiny_skia`

### LINKED ISSUE #2877: Crash on `tiny_skia` backend when rendering zero-width/height image

### What happened?

Rendering an image widget with zero pixels results in a panic in Iced's tiny_skia code.

Minimal example:

```
use iced::window;
use iced::{Element, Task};
use iced::widget::image;

fn main() -> iced::Result {
    iced::daemon(Example::title, Example::update, Example::view)
        .run_with(Example::new)
}

struct Example { }

#[derive(Debug, Clone)]
enum Message { Nuhuh }

impl Example {
    fn new() -> (Self, Task<Message>) {
        let (_id, open) = window::open(window::Settings::default());

        (Self {}, open.map(|_| Message::Nuhuh))
    }

    fn title(&self, _: window::Id) -> String {
        "Test".to_owned()
    }

    fn update(&mut self, message: Message) -> Task<Message> {
        match message {
            Message::Nuhuh => { }
        }
        Task::none()
    }

    fn view(&self, _: window::Id) -> Element<Message> {
        image(image::Handle::from_rgba(0, 0, iced::advanced::image::Bytes::new())).into()
    }
}
```

This results in an immediate panic of `Build pixmap from image bytes`. https://github.com/iced-rs/iced/blob/e462535f3ba0bc1cdaa4a101b48542816f777e53/tiny_skia/src/raster.rs#L115

Note that this doesn't appear(?) to panic when running with the `wgpu` backend.

Well, anything except panic.

Assuming this isn't expected behaviour (I don't think it is), ideally an empty element should be returned with no work done, i.e., the equivalent of `row![].into()`. Alternatively, `image` could return a `Result`.

If this *is* expected behaviour, it could be documented on the Image widget, and the error message could be more helpful. And, this behaviour should be standardized on the other backends.

### Version

crates.io release

### Operating System

Linux

```shell

```

13r0ck:
Always alocate the min size for the tiny_skia buffer

https://github.com/linebender/tiny-skia/blob/main/src/pixmap.rs#L295

Fixes #2877.

### DIFF
--- a/core/src/image.rs
+++ b/core/src/image.rs
@@ -348,6 +348,9 @@ pub enum Error {
     /// Loading images is unsupported.
     #[error("loading images is unsupported")]
     Unsupported,
+    /// The image is empty.
+    #[error("the image is empty")]
+    Empty,
     /// Not enough memory to allocate the image.
     #[error("not enough memory to allocate the image")]
     OutOfMemory,
--- a/tiny_skia/src/raster.rs
+++ b/tiny_skia/src/raster.rs
@@ -106,6 +106,10 @@ impl Cache {
                 }
             };
 
+            if image.width() == 0 || image.height() == 0 {
+                return Err(raster::Error::Empty);
+            }
+
             let mut buffer =
                 vec![0u32; image.width() as usize * image.height() as usize];
 


hecrj (approved):
Thanks!

We can just return early.

PR #2988, opened 2025-06-12
Add `font_maybe` Utility to `Text`

andymandias:
A small utility function to allow concise font setting when one has an `Option<Font>` instead of a `Font`, modeled after the `font_maybe` function for `Span` in `core/src/text.rs`.


hecrj (approved):
Thanks!

PR #2989, opened 2025-06-12
fix docs for Limits.lose

JL710:
Since `.loose` removes both width and height, saying it removes the width constraint is wrong. This PR removes just the word `width` from the docs.


mfreeborn:
Oi! #2980

JL710:
Sorry for the duplicate!

PR #2992, opened 2025-06-16
Add mailto as valid link scheme in markdown

semiversus:
Currently only link schemes "http" and "https" are valid. This pull requests adds "mailto" as valid link scheme.

Another question is, if the scheme should be checked here as it the verification could be left to the Message handler.

### DIFF
--- a/widget/src/markdown.rs
+++ b/widget/src/markdown.rs
@@ -562,7 +562,8 @@ fn parse_with<'a>(
                 match Url::parse(&dest_url) {
                     Ok(url)
                         if url.scheme() == "http"
-                            || url.scheme() == "https" =>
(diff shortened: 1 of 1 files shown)


nacl42:
I have the same issue with the markdown viewer. IMHO, the viewer should not select the valid schemes, but the click handler should decide whether to handle or ignore this. In my app, I am using a custom scheme (card://).

An easy solution would be to completely remove the match guard for the url.scheme().

Alternatively, one could add a 'valid_schemes' list, which can be set by the user and then check if the url scheme is contained in the list.

semiversus:
Yes, removing the check would be the best thing here. It should be left to the click handler.

PR #2993, opened 2025-06-19
[Scrollable] Added the ability of not to change a specific direction offset

rizzen-yazston:
Added the ability to not change a specific direction offset when updating the scrollable's offsets. Before this commit, both directions' offsets are updated. This commit resolves an issue with linked scrollables resetting the other scrollable offset to the start position, while zeroing out the one direction of a linked scrollable.


rizzen-yazston:
Forgot to run `cargo fmt` before doing commit and pull request.

PR #2994, opened 2025-06-19
Allow `scroll_to` and `snap_to` to operate on a single axis

rizzen-yazston:
[Scrollable] Added the ability to change a specific direction offset when updating the scrollable's offsets. Before this commit, both directions' offsets are updated. This commit resolves an issue with linked scrollables resetting the other scrollable offset to the start position, while zeroing out the one direction of a linked scrollable.

### DIFF
--- a/core/src/widget/operation/scrollable.rs
+++ b/core/src/widget/operation/scrollable.rs
@@ -5,10 +5,10 @@ use crate::{Rectangle, Vector};
 /// The internal state of a widget that can be scrolled.
 pub trait Scrollable {
     /// Snaps the scroll of the widget to the given `percentage` along the horizontal & vertical axis.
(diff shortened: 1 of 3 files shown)


rizzen-yazston:
If there is no desire to change the existing functions of the `Scrollable` trait, alternatively, additional new functions can be added to handle the optional. In the implementation in `widget/src/scrollable.rs`, the `f32` to `Option<f32>` change to the fields is kept, with the original function simply placing the `f32` into `Some(f32)`, thus still maintaining backwards compatibility with the current applications, requiring no changes as all the changes are within `iced`.

hecrj (approved):
Thanks!

I avoided all the `unwrap` stuff with some generic magic.

PR #2998, opened 2025-06-22
make `modified_key` the default for `key` of `keyboard::Event`

### LINKED ISSUE #2997: Named Keys not working in widgets

### What happened?

- Use Neo2 Keyboard Layout (or probably any other with named keys on higher layers)
- Run editor example
- In texteditor widget Arrow-Keys, Del, Backspace, Pos1, End ... from layer 4 of the Keyboard do not work.

The expected behavior for the cursor is to move according to the Arrow-Keys and delete characters with Del or Backspace.

The problem was introduced in https://github.com/iced-rs/iced/commit/9600954d8de57442dfe1269dcf80e3d23edbaf0a by checking only baselayer keys for shortcuts.
In https://github.com/iced-rs/iced/commit/94c8b9639ca2d287c504345390585b80dc0d78cf a modified_key got introduced that would work but it is not used in the widgets.
There are two easy solutions for that:
Either use the modified_key in the shortcuts
or revert behavior and make modified_key the default and the baselayer key the special case.
I prefer the second one. I think `key` should match the output of something like the wev app.
I tried to get opinions [here](https://discourse.iced.rs/t/support-multilayer-keyboards/950) so i can just make a PR with the preferred solution but it is pretty silent over there.

### Version

master

### Operating System

Linux

```shell

```

Gerd-Augsburg:
Proposal for #2997 
it fixes #2997

### DIFF
--- a/core/src/keyboard/event.rs
+++ b/core/src/keyboard/event.rs
@@ -12,11 +12,14 @@ use crate::keyboard::{Key, Location, Modifiers};
 pub enum Event {
     /// A keyboard key was pressed.
     KeyPressed {
-        /// The key pressed.
+        /// The key pressed. Tries to meet expectations, currently equal to `modified_key`.
         key: Key,
 
         /// The key pressed with all keyboard modifiers applied, except Ctrl.
         modified_key: Key,
+
+        /// The key on keyboard layer 0 pressed.
+        baselayer_key: Key,
 
         /// The physical key pressed.
         physical_key: key::Physical,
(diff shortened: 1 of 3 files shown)


Seppel3210:
I got another take of this issue over at https://github.com/iced-rs/iced/pull/3007

coroa:
Was closing it intended?

Gerd-Augsburg:
Not intentional.

Wanted to apply my fix to the iced version libcosmic is using, because cosmic-term is the app where it bothers me.
My git-fu wasn't strong enough so i deleted my fork and forked from them.
Apparently you can't have two forks of the same base Project.

I created a new PR at https://github.com/iced-rs/iced/pull/3083 from my new fork.

PR #3002, opened 2025-06-26
re-export runtime widget task

tzemanovic:
hi, just a tiny change to re-export runtime widget task for conveniece

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -540,7 +540,7 @@ pub mod debug {
 
 pub mod task {
     //! Create runtime tasks.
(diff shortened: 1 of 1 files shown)


pml68:
FYI it's already exported under `advanced`

https://github.com/iced-rs/iced/blob/master/src/advanced.rs#L13

tzemanovic:
> FYI it's already exported under `advanced`
> 
> https://github.com/iced-rs/iced/blob/master/src/advanced.rs#L13

ah, didn't notice that, thx! Closing

PR #3007, opened 2025-06-30
Fix navigation keys on higher layers not working

### LINKED ISSUE #2997: Named Keys not working in widgets

Seppel3210:
Some keyboard layouts have navigation keys in "unusual" positions on higher levels (e.g., Neo family layouts). These did not work at all, but this fixes them by not using `key_without_modifiers()` for `Named(…)` logical keys.

I also thought of implementing this by checking if `key_without_modifiers()` changes the `std::mem::discriminant` of the key (which would also be unexpected) and then you could return the logical key instead, but then you'd probably have to handle dead keys seperately, so I think doing this only for `Named` keys should be easier.

Fixes #2997.

### DIFF
--- a/widget/src/text_editor.rs
+++ b/widget/src/text_editor.rs
@@ -1173,9 +1173,6 @@ impl<Message> Binding<Message> {
             return None;
         }
 
-        #[cfg(target_os = "macos")]
-        let key = convert_macos_shortcut(&key, modifiers);
-
         let combination = match key.as_ref() {
             keyboard::Key::Character("c") if modifiers.command() => {
                 Some(Self::Copy)
@@ -1197,6 +1194,10 @@ impl<Message> Binding<Message> {
         if let Some(binding) = combination {
             return Some(binding);
         }
+
+        #[cfg(target_os = "macos")]
+        let modified_key =
+            convert_macos_shortcut(&key, modifiers).unwrap_or(modified_key);
 
         match modified_key.as_ref() {
             keyboard::Key::Named(key::Named::Enter) => Some(Self::Enter),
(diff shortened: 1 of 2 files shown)


hecrj (approved):
Thanks!

I think the proper fix is to use `key` without modifiers for key combinations (e.g. `Ctrl+C`) and the `modified_key` for single key bindings (e.g. motions).

Hope that works.

coroa:
Even though this was merged navigation keys at least on Neo2 and Colemak are not working for instance in cosmic term: https://github.com/pop-os/cosmic-term/issues/497 .

Can you describe what needs to be changed there?

Thank you

edwloef:
I believe pop-os hasn't rebased their iced fork onto a commit that includes this patch yet.

coroa:
> I believe pop-os hasn't rebased their iced fork onto a commit that includes this patch yet.

Sorry for coming back to this, but if i read https://github.com/pop-os/libcosmic/issues/1089 correctly, the rebase is in their fork since March, and the situation in all cosmic apps i checked is unchanged.

ie. navigational keys in colemak+extend are not moving the cursor (except for in cosmic-edit, which seems to be using modified_key directly, https://github.com/pop-os/cosmic-edit/commit/099e32bdc0636d6c9e86a451a63bf7d11e598ed5).

Does this need additional changes for instance in cosmic-term to work?

PR #3010, opened 2025-07-03
Add `push_under` method to `stack` widget

Remmirad:
This allows to set the element used by the `Stack` widget to determine its intrinsic size. At the moment it is always the first element which is problematic if you want a dynamic background at the base that is shown behind a (partial) transparent upper layer that should determine the size e.g. an image that preserves aspect ratio.


Remmirad:
That looks much cleaner than my version thanks!

PR #3018, opened 2025-07-18
`table` widget

hecrj:
This PR introduces a new built-in `table` widget.

https://github.com/user-attachments/assets/42ceb660-c836-48f4-aa42-3c504dffa9b8

The new `table` example shows how to use it:

```rust
let bold = |header| {
    text(header).font(Font {
        weight: font::Weight::Bold,
        ..Font::DEFAULT
    })
};

let columns = [
    table::column(bold("Name"), |event: &Event| text(&event.name)),
    table::column(bold("Time"), |event: &Event| {
        let minutes = event.duration.as_secs() / 60;

        text!("{minutes} min").style(if minutes > 90 {
            text::warning
        } else {
            text::default
        })
    })
    .align_x(Right)
    .align_y(Center),
    table::column(bold("Price"), |event: &Event| {
        if event.price > 0.0 {
            text!("${:.2}", event.price).style(
                if event.price > 100.0 {
                    text::warning
                } else {
                    text::default
                },
            )
        } else {
            text("Free").style(text::success).width(Fill).center()
        }
    })
    .align_x(Right)
    .align_y(Center),
    table::column(bold("Rating"), |event: &Event| {
        text!("{:.2}", event.rating).style(if event.rating > 4.7 {
            text::success
        } else if event.rating < 2.0 {
            text::danger
        } else {
            text::default
        })
    })
    .align_x(Right)
    .align_y(Center),
];

table(columns, &self.events)
```

Additionally, the `markdown` widget supports tables now!

This PR also includes a QoL change that implements `From<Option<T>>` for `Element`. Effectively, this allows users to directly use any `Option` containing a widget directly in `view` code—rendering `push_maybe` methods redundant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5634,6 +5634,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "table"
+version = "0.1.0"
+dependencies = [
+ "iced",
+]
+
+[[package]]
 name = "target-lexicon"
 version = "0.12.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/core/src/element.rs
+++ b/core/src/element.rs
@@ -532,3 +532,49 @@ where
         )
     }
 }
+
+impl<'a, T, Message, Theme, Renderer> From<Option<T>>
+    for Element<'a, Message, Theme, Renderer>
+where
+    T: Into<Self>,
+    Renderer: crate::Renderer,
+{
+    fn from(element: Option<T>) -> Self {
+        struct Void;
+
+        impl<Message, Theme, Renderer> Widget<Message, Theme, Renderer> for Void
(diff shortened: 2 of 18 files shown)


GyulyVGC:
That's awesome, thanks for all the work 🙏 

Question: would it be possible to truncate strings with ellipsis in case the content is too large for its column's width?
I don't know if iced / cosmic support this already...

hecrj:
@GyulyVGC That is better implemented as an ellipsized text widget at some point.

KodeBarista:
Can separators be used as anchor points to resize individual columns?

EmmanuelDodoo:
Not with the current implementation. It is more along the lines of a markdown table

PR #3021, opened 2025-07-23
Tray Icon integration

nate-trojian:
Adds a basic integration with [tray_icon](https://github.com/tauri-apps/tray-icon).  Usage has been put behind the `tray-icon` feature flag.

Added the ability to:
- Define a Tray Icon using either text or image
- Process events for:
  - Mouse entering the icon
  - Exiting the icon
  - Pressing the mouse over the icon
  - Releasing the mouse over the icon
  - Double clicking the icon 
  - Clicking a menu item
- Add a menu to the Icon
- Add a menu item with just text, a checkbox, an icon, or a predefined menu item

Current limitations:
- No ability to modify the tray icon during the running of the application
- Tray icon events must be processed from a subscription due to the fact they aren't tied to a window
- Menu has to be open in order for "accelerators" to be used (limitation of how menus work in tray icon)
- Clicking a predefined menu item does not generate a menu clicked event (this is a limitation inside of [muda](https://github.com/tauri-apps/muda))

Screenshot of running example:
Screenshot 2025-07-29 at 23 25 18


Cargo.toml
```diff
@@ -202,6 +204,7 @@ softbuffer = "0.4"
 syntect = "5.1"
 sysinfo = "0.33"
 thiserror = "1.0"
+tray-icon = "0.21"
```
n1ght-hunter, line 207: this should be optional right?
also need to test that without the tray-icon feature enabled you dont need gtk as i believe tray-icon requires gtk on linux

dtzxporter:
I'd like to see more of the code be disabled completely when the feature is disabled, IMO, nothing related to tray icon, even stubs, should be included because we have `#[cfgs]`.

n1ght-hunter:
> I'd like to see more of the code be disabled completely when the feature is disabled, IMO, nothing related to tray icon, even stubs, should be included because we have `#[cfgs]`.

agreed that was just the first thing that popped out to me. really i think there should be pretty much no code included without enabling the feature

Cargo.toml
nate-trojian: This is in the workspace dependencies section, which is for defining a common version that can then be used across the workspace.  It only gets built into the project if it is included in a crate within the workspace.

For reference: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table

nate-trojian:
> I'd like to see more of the code be disabled completely when the feature is disabled, IMO, nothing related to tray icon, even stubs, should be included because we have `#[cfgs]`.

Because of how tightly integrated this feature needs to be within the system, I didn't originally want to do this.  I think it hurts the readability of the code and will make it harder to add functionality to this feature in the future.

However, I've made the change to move the guards to all the public interfaces.

Cargo.toml
n1ght-hunter: mb lol github had hidden that info
image

Decodetalkers:
https://github.com/tauri-apps/tray-icon/blob/dev/examples/winit.rs#L117-L125

I think on linux it needs gtk to show the tray icon

defaultUser822:
I wonder if this implementation needs reconsideration since most popular Linux distros, including Ubuntu, default to Wayland nowadays. Even if it works on Wayland, the background portal should be put in mind.

> Note: I didn't see the implementation since I am still a beginner in Rust. I left this comment as a reminder.

PR #3025, opened 2025-07-30
Add SurrealDB example

Dhghomon:
Hi,

I recently gave Iced a try for the first time (have mostly used egui until now) in putting a demo UI together that uses a SurrealDB in-memory instance for things like vector embeddings and graph queries. It calls into the Wikipedia article summary API to get the text for the records to store.


examples/surrealdb/src/main.rs
```diff
@@ -0,0 +1,676 @@
...
+            Err(e) => e.to_string(),
+        }
+    }
+
+    fn update(&mut self, message: Message) {
+        use Message as M;
+
+        let rt = self.rt.take().unwrap();
+
+        rt.block_on(async {
```
cguentherTUChemnitz, line 419: I think the desired way to do this in iced is return iced::Task by update function, which passes those Tasks to the async iced runtime. Those can itself generate messages on-complete instead of blocking inside the update function. See  for the Task example:

https://github.com/iced-rs/iced/blob/d475ae5b454b82cb549d6dd961ae158eb5ac7f19/examples/pokedex/src/main.rs#L43

cguentherTUChemnitz (commented):
I think this example should use iced::Task<Message> as return of update instead of blocking inside the update function

mahlquistj:
> I think this example should use iced::Task as return of update instead of blocking inside the update function

This also applies to the App itself. Instead of saving a runtime within the app, you could use the internal runtime of Iced combined with tasks to do all of the DB operations you do.

`Task::perform` can take a `Future` and run it beside the main-thread.

Dhghomon:
Thanks @cguentherTUChemnitz and @mahlquistj, I'll give that a try! I was originally thinking of setting up a channel as I usually do but wanted to keep the code as simple as possible, and looks like a Task is the way to go.

hecrj:
Hey, it's great to see other cool projects play with iced!

I believe it would make more sense to host this example in the SurrealDB repository/organization rather than here.

We already have a lot of examples, and I don't think I would be able to easily maintain this one.

Dhghomon:
@hecrj Sounds good! We do indeed have the code on the site so it can stay there. I'll close this PR and leave this link to the blog post that discusses the demo tool which you could reference if the subject ever comes up.

https://surrealdb.com/blog/building-an-ai-native-multi-model-ui-with-surrealdb

(Will get to the Task business in due time, must write some other things today...)

Issue #2044, opened 2023-08-19
Unable to access any value from Id


sionzee:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

I encountered an issue while trying to get a specific value from an `Id` in the `iced` crate.
Here's the code snippet:

```rust
#[derive(Debug, Clone)]
pub enum Message {
  TextInputValueChanged(String, String),
}

// `attributes` is a hashmap <String, String>
let id = attributes.get("id").and_then(|value| Some(iced::widget::text_input::Id::new(value.clone()))).unwrap_or(iced::widget::text_input::Id::unique());

text_input.on_input(move |c: String| {
  // I need to access the custom or unique id here
  crate::Message::TextInputValueChanged(id.clone(), c)
});
```

The problem here is I can't access any value from the `Id` that I created. The `Id` is defined in the crate like this:

```rust
/// The identifier of a generic widget.
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub struct Id(Internal);

#[derive(Debug, Clone, PartialEq, Eq, Hash)]
enum Internal {
    Unique(usize),
    Custom(borrow::Cow<'static, str>),
}
```

### What is the expected behavior?

I was expecting to have a way to fetch a specific value from `Id` so that I can use it as an argument in `TextInputValueChanged`. If there's an `Id` variant that exposes a way to retrieve its value (or a method that allows accessing it), it would be great. If there's an alternative way to handle this, I would really appreciate your guidance.

### Version

crates.io release

### Operating System

Windows

hecrj:
Use `text_input::Id` instead of `String`.

```rust
#[derive(Debug, Clone)]
pub enum Message {
  TextInputValueChanged(text_input::Id, String),
}
```

PR #3027, opened 2025-08-03
Add getter for window::Id

Thetamesis:
This commit adds a getter method to window::Id.

The raw integer was already indirectly accessible via the type's Display implementation. This method is more conventient if one wants to work with the integer directly.


Thetamesis:
This resolves issue #2044

hecrj:
Thanks, but the issue mentioned seems unrelated, since it talks about `text_input::Id`.

I am not sure why exposing the internal integer that the runtime uses to identify windows would be useful.

PR #3042, opened 2025-08-25
Add option to reuse wgpu Renderer screenshot buffers to improve performance

oberien:
When taking several screenshots with the `iced_wgpu::Renderer`, the wgpu-buffers aren't reused currently. Buffer creation takes ~40% of the time of the whole screenshot. This PR adds the option to reuse the screenshot-buffers, including reusing buffers for color conversion.

Performance measurements on my machine:
* current `convert`: 1.3ms
* new `convert_buffered` when reusing buffers: 0.1ms
* current `screenshot`: 14ms
* new `screenshot_into` when reusing buffers including using `convert_buffered`: 8ms

### DIFF
--- a/wgpu/src/color.rs
+++ b/wgpu/src/color.rs
@@ -9,6 +9,7 @@ pub struct ConvertBuffers {
     constant_bind_group: wgpu::BindGroup,
     texture_layout: wgpu::BindGroupLayout,
     texture: wgpu::Texture,
+    needs_clear: bool,
 }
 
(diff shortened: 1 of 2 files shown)


hecrj:
This doesn't seem to actually reuse buffers in the current `screenshot` API itself, but instead extends `iced_wgpu` to allow an optimized path.

The changes also seem unnecessarily verbose and complex. I think we can do better.

Issue #2802, opened 2025-02-08
.window_size(iced::Size::INFINITY) crashes on MacOS


dcampbell24:
### Is your issue REALLY a bug?

- [x] My issue is indeed a bug!
- [x] I am not crazy! I will not fill out this form just to ask a question or request a feature. Pinky promise.

### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Is this issue related to iced?

- [x] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

I loaded my application on Linux and it works fine, but when I run it on MacOS it crashes. If I comment out the line `.window_size(iced::Size::INFINITY)` it works again.

### What is the expected behavior?

The application will run.

### Version

crates.io release

### Operating System

macOS

### Do you have any log output?

```shell

```

dcampbell24:
I tested it on windows 11 and it works.

JL710:
Why do you set the window size to infinite?

dcampbell24:
Because I want to maximize the window size, but I don't know what it is.

JL710:
Why not maximizing with [window::maximize](https://docs.rs/iced/latest/iced/window/fn.maximize.html)?

dcampbell24:
Then you have to take an action to maximize the window? (I wanted it to start out maximized, but anyway now I have it so the window starts out at the standard size, then you can maximize it if you want to.) I'm leaving this issue open as it seems like an error, even if I don't encounter it anymore.

pml68:
I'd think an app crashing when you set the window size to infinite is perfectly understandable. You can just return the task in your initializer method, so it happens before anything else.

dcampbell24:
Then why doesn't it crash on Windows or Linux?

pml68:
No idea. You check the source, should be in winit. Lots of windowing-related things are very OS-specific, so withoit checking I can't really tell you.

pml68:
@dcampbell24 you probably don't care about it at this point, but while going through some stuff in winit, I also found the source of this. Here's the Apple documentation: https://developer.apple.com/documentation/appkit/nswindow/init(contentrect:stylemask:backing:defer:)#parameters (specifically `contentRect`)

dcampbell24:
So set the window size to 10,000 on macOS when the user sets it to `.window_size(iced::Size::INFINITY)`?

pml68:
Clamp the value if it's provided by the user, yes.

dcampbell24:
[...]

PR #3044, opened 2025-08-27
Fix macos max window size.

dcampbell24:
See issue https://github.com/iced-rs/iced/issues/2802 for discussion.

This stops a crash from happening on macOS when the window size is set greater than 10,000.


hecrj (commented):
I don't think clamping is the right move here.

A huge `Size` is a programmer error and a panic is what I'd expect.

PR #3045, opened 2025-08-29
Prioritize `Shrink` over `Fill`

hecrj:
This PR makes a `Fill` inside a `Shrink` behave as if it was `Shrink`.

Effectively, this allows controlling and overriding `Fill` behavior from a higher level container widget.

As a consequence, `scrollable` can now handle `Fill` content by simply forcing it to behave as `Shrink`; so it will no longer panic in `debug` mode.

### DIFF
--- a/core/src/layout/flex.rs
+++ b/core/src/layout/flex.rs
@@ -155,7 +155,9 @@ where
                 axis.pack(size.width, size.height)
             };
 
-            if main_size.fill_factor() == 0 && cross_size.fill_factor() != 0 {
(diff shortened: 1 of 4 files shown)


abique:
With the following scenario:

A row with one widget fixed height 150, and another widget that has a fill height.
What do you set on the row to make the second widget reach the height 150? Before I'd set shrink height on the row and that'd work. Now it is broken it seems.

Is that a bug?

abique:
Fixed by https://github.com/iced-rs/iced/commit/9994f1f133cb22d5023a10619c14b4fb86524208

PR #3051, opened 2025-09-07
Replace `dark-light` with `mundy`

hecrj:
This PR replaces `dark-light` with `mundy`, which does not depend on the new defunct `async-std` (RIP).

Eventually, we'll want to leverage the preferences subscription to dynamically change the preferred color scheme at runtime.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,7 +77,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ef6978589202a00cd7e118380c448a08b6ed394c3a8df3a430d0898e3a42d046"
 dependencies = [
  "android-properties",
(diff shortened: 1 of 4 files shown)


hecrj:
> Eventually, we'll want to leverage the preferences subscription to dynamically change the preferred color scheme at runtime.

Ah... The test failures tell us that time has arrived.

hecrj:
I finally added a `linux-theme-detection` feature that leverages `mundy` to keep feature parity. It still brings down the dependencies from 439 to 401 for the `tour` on Linux.

Additionally, `mundy` let's us keep track of theme changes at runtime; so applications should automatically switch themes when using the system fallback (i.e. `theme` returns `None`).

PR #3065, opened 2025-09-27
Allow setting windows size based on monitor size

acul009:
This just adds the `FromScreensize` option to the window settings size.

I used the positioning option as a base.

I use this for my drowdown terminal, so the window size is automatically set.


pml68:
I think you could use `fn(Size) -> Size` in the added variant to avoid the Arc, since `Position::SpecificWith` also just uses a function pointer (which is `Send`, `Sync` and `Copy`)

acul009:
> I think you could use `fn(Size) -> Size` in the added variant to avoid the Arc, since `Position::SpecificWith` also just uses a function pointer (which is `Send`, `Sync` and `Copy`)

I did in the original PR, but for the tester I need to support closures, since I need to make the window wider than it would otherwise be:


https://github.com/iced-rs/iced/blob/0a34496c793fd87c235e764317351a0aacd4f6d1/tester/src/lib.rs#L67-L72

https://github.com/iced-rs/iced/blob/a9ac8ba1a997e8487c37d9be272e7a86f8bbc966/core/src/window/size.rs#L56-L67

The only other solution I could think of, would be to ignore the variable window size when the tester is used.
Maybe you have a better idea?

pml68:
Hmm

You could add an extra field to the variant which contains extra size, but that could make it a bit confusing to use, or just plain ugly.

acul009:
> Hmm
> 
> You could add an extra field to the variant which contains extra size, but that could make it a bit confusing to use, or just plain ugly.

The current API isn't really all user friendly either.
I added a helper function, but the enum variant is still there and requires a manual Arc.
It's not obvious you should use the helper.

@pml68 What do you think about creating a wrapper struct to hide the enum variants?
Then I'd just expose helper functions and nothing else.
That way I wouldn't expose any weird enum variants.

```rs
pub struct Size( pub(crate) SizeVariant)
```

Not sure these enum in struct types fir with the iced style.

pml68:
> @pml68 What do you think about creating a wrapper struct to hide the enum variants? Then I'd just expose helper functions and nothing else. That way I wouldn't expose any weird enum variants.
> 
> ```rust
> pub struct Size( pub(crate) SizeVariant)
> ```

Hmm not sure honestly. I can't really think of a better way, but I wonder what Héctor would think about the proposed solutions (user has to `Arc` / hidden enum / some other solution to the test recorder-size adding problem)

hecrj:
I think introducing further tasks to query available monitors and their sizes should be a more flexible and simpler approach than this.

Then, when leveraging `daemon`, it would be just a matter of chaining the appropriate tasks.

acul009:
> I think introducing further tasks to query available monitors and their sizes should be a more flexible and simpler approach than this.
> 
> Then, when leveraging `daemon`, it would be just a matter of chaining the appropriate tasks.

That would be a lot nicer, then I could easily implement changing the screen of a window too.

PR #3068, opened 2025-10-02
Make slider cursor consistent across os and align with web sliders.

dtzxporter:
Basically, this makes the slider use the pointer cursor, which aligns with how it's commonly done on web and other platforms, in addition this fixes the bad choices available on windows, and aligns all OSs to use the same icon.

### DIFF
--- a/widget/src/slider.rs
+++ b/widget/src/slider.rs
@@ -532,10 +532,8 @@ where
         let bounds = layout.bounds();
         let is_mouse_over = cursor.is_over(bounds);
 
-        if state.is_dragging {
-            mouse::Interaction::Grabbing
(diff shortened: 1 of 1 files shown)


dtzxporter:
Rebased for latest changes.

hecrj (approved):
This feels like a downgrade on Linux and macOS; and aligning with Web is not really a great argument.

I think this is a `winit` issue that we will inevitably hit if we want to implement any kind of drag and drop functionality. Turns out I reported it 6 years ago:

https://github.com/rust-windowing/winit/issues/1043

In any case, we can just fall back to `Pointer` on Windows for now.

PR #3070, opened 2025-10-04
make `Mesh`es cheaply cloneable by using `Arc<[T]>` in `iced_graphics::mesh::Indexed`

edwloef:
I have large meshes that I'd like to not have to recreate/clone every time I draw them. When using `wgpu` I can use a `iced_wgpu::geometry::Cache`, however this doesn't work when using `tiny-skia`. Therefore it makes sense to me to make reuse possible with this change.

### DIFF
--- a/examples/geometry/src/main.rs
+++ b/examples/geometry/src/main.rs
@@ -83,7 +83,7 @@ mod rainbow {
 
             let mesh = Mesh::Solid {
                 buffers: mesh::Indexed {
-                    vertices: vec![
+                    vertices: [
                         SolidVertex2D {
                             position: posn_center,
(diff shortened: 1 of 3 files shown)


edwloef:
Failing CI is from an upstream rustdoc change: https://github.com/rust-lang/rust/pull/138907

mahlquistj:
This is great. We have ~25% performance increase by reusing the meshes! Hoping this gets merged soon! :+1:

hecrj (approved):
Thanks!

I have moved the `Cache` in `wgpu` to `mesh` which lets renderers reuse buffers under the hood.

This also avoids redundant `Arc` when caching isn't desired.

PR #3072, opened 2025-10-09
Fixed hit span  for paragraph

EmmanuelDodoo:
The previous implementation of `hit_span` incorrectly handled visual breaks. Empty spaces at the end of a line were considered as part of the last span. This PR fixes it so empty spaces are ignored. The code is modified version of the `span_bounds` implementation.

**Before**
Screencast From 2025-10-08 11-47-09(1)

**After**
Screencast From 2025-10-08 20-30-45


hecrj (approved):
Thanks!

I've been able to fix it while keeping the original simpler implementation.

Well, technically there is still an edge case where half a glyph won't be clickable at the end of the line. But let me sleep on that.

PR #3081, opened 2025-10-14
Add `Mouse::Scroll` instruction to `tester`

thorn132:
This adds event recording and playback for `mouse::Event::WheelScrolled`, supporting any `ScrollDelta` and merging consecutive events.

### DIFF
--- a/test/src/instruction.rs
+++ b/test/src/instruction.rs
@@ -60,6 +60,9 @@ impl Interaction {
                     button: *button,
                     target: None,
                 },
+                mouse::Event::WheelScrolled { delta } => {
+                    Mouse::Scroll { delta: *delta }
+                }
                 _ => None?,
             }),
             Event::Keyboard(keyboard) => Self::Keyboard(match keyboard {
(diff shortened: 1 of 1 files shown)


hecrj:
I'm not convinced with the DSL. We want something easy to write for the common use case.

I think it may be worth having different instructions for vertical and horizontal scrolling (`scroll 10px` being the vertical one).

But let's see how the `tester` evolves first before expanding it further.

thorn132:
> scroll 10px being the vertical one

And maybe `swipe 10px` for horizontal? Should that still allow merging horizontal and vertical scrolling into one line though, e.g. `scroll 10px, swipe -20px`?

> But let's see how the tester evolves first before expanding it further.

I.e. 'please stop opening `tester` PRs'? 😅

hecrj:
> And maybe swipe 10px for horizontal? Should that still allow merging horizontal and vertical scrolling into one line though, e.g. scroll 10px, swipe -20px?

I don't know. I'm not convinced. I'd have to think about it.

> I.e. 'please stop opening tester PRs'? 😅

You can open them, but I won't be reviewing them until I know where I'm going with it. So, maybe in a year or so.

PR #3083, opened 2025-10-15
makes `modified_key` the default for `key` of `keyboard::Event`

Gerd-Augsburg:
Proposal for https://github.com/iced-rs/iced/issues/2997
it fixes https://github.com/iced-rs/iced/issues/2997
replacement of https://github.com/iced-rs/iced/pull/2998

Replacement needed because: me == bad at git.

Modified key as a better default. Added `baselaye_key` for those who want especially that.


hecrj:
Superseded by #3007.

PR #3087, opened 2025-10-17
Add `command` instruction to `tester`

thorn132:
This adds a `command` instruction to turn arbitrary text into a `Message` executed by the `Emulator`. The motivating usage is mocking user interactions outside of `iced`, e.g., system file dialogs.

Some open design questions:
- Should declaring commands use existing `Subscription` infrastructure, including `map`ping `Message`s?
- Should there be a strict/typed parsing format for arguments instead of passing a single `&str`?
- Should one `command` be able to send a sequence of `Message`s?

### DIFF
--- a/program/src/lib.rs
+++ b/program/src/lib.rs
@@ -97,6 +97,14 @@ pub trait Program: Sized {
         Subscription::none()
     }
 
+    fn command(
+        &self,
+        _state: &Self::State,
+        _command: &str,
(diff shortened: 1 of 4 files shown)


hecrj:
Nah, this isn't happening.

The DSL isn't meant to be application specific. We want to avoid defining testing logic at the application-level as much as possible. Presets should be enough!

PR #3090, opened 2025-10-23
Provide a `Display` handle to `graphics::Compositor`

### LINKED ISSUE #3089: [tiny-skia] Multi window demo: cannot close first window if a second window is open

### What happened?

On the current master branch (rev https://github.com/iced-rs/iced/commit/d5608162def4688052c75e43f5d1b04268af4d06)

If you force the multi_window example to use tiny skia and open a second window, then closing the first window fails, leaving it unresponsive. Closing the second window afterwards will also close the first window.

The first window should be closed

### Version

master

### Operating System

Linux

```shell

```

njust:
Fixes #3089.

### DIFF
--- a/graphics/src/compositor.rs
+++ b/graphics/src/compositor.rs
@@ -18,23 +18,25 @@ pub trait Compositor: Sized {
     type Surface;
 
     /// Creates a new [`Compositor`].
-    fn new<W: Window + Clone>(
+    fn new(
         settings: Settings,
-        compatible_window: W,
+        display: impl Display + Clone,
+        compatible_window: impl Window + Clone,
         shell: Shell,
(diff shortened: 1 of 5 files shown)


hecrj:
Would this be considered a `softbuffer` bug?

hecrj (approved):
Thanks!

However, I have been able to identify the root cause and keep a single shared `Context`.

The actual issue is that `softbuffer::Context` expects a `HasDisplayHandle` and not an actual window.

Therefore, a reference to the first window we use to create the `Compositor` is stored in the `Context` and is never dropped when `iced_winit` closes it.

The solution is to introduce an additional `display` argument to the `Compositor` constructors and leverage that instead to create the `Context`.

PR #3093, opened 2025-10-26
make `overlay::from_children` and `layout::flex::resolve` generic over the widget type

edwloef:
A bunch of other widget functions are already generic over `impl Borrow<dyn Widget>`, these were missing presumably because they need mutable access to the widget. This is useful when working with concrete widget types instead of type-erased `Element`s.

### DIFF
--- a/core/src/element.rs
+++ b/core/src/element.rs
@@ -9,7 +9,7 @@ use crate::{
     Vector, Widget,
 };
 
-use std::borrow::Borrow;
+use std::borrow::{Borrow, BorrowMut};
 
 /// A generic [`Widget`].
 ///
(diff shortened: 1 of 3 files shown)


hecrj:
Is there any specific reason you cannot type erase?

Wrapping existing widgets directly is generally discouraged.

edwloef:
I have them as concrete types in order to minimize renderer layers. Essentially, because I know my children never overlap but may themselves have overlapping children and therefore need multiple renderer layers, I can let each child draw as much as it can per layer in a loop, therefore "sharing" renderer layers between them.

https://github.com/generic-daw/generic-daw/blob/7571852b8c02e58831c50316fa6e3fb8b2684995/generic_daw_gui/src/widget/playlist.rs#L454-L482
https://github.com/generic-daw/generic-daw/blob/7571852b8c02e58831c50316fa6e3fb8b2684995/generic_daw_gui/src/widget/track.rs#L212-L252

hecrj:
I don't see the point in running `flex::resolve` for concrete types and hardcoded values. It's overkill.

It seems you could get away with a very simple hand-rolled `layout` implementation.

For your `overlay` implementation, I'd just copy the `from_children` code here directly. There is hardly any complex logic there. It's just a helper for the common case.

edwloef:
> hardcoded values

That's only temporary sadly :') but I guess that's something I can tackle when it becomes a problem

hecrj:
Let's close this for now, then!

PR #3095, opened 2025-10-28
add function to set mouse interactions for slider

royalmustard:
This PR allows to set a custom mouse interaction for a slider, allowing for further styling and customization of applications.

### DIFF
--- a/widget/src/slider.rs
+++ b/widget/src/slider.rs
@@ -95,6 +95,7 @@ where
     width: Length,
     height: f32,
     class: Theme::Class<'a>,
+    mouse_interaction: SliderMouseInteraction,
 }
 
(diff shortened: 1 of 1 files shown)


hecrj:
Thanks! I'm going to need use cases here.

royalmustard:
My use case was that in a music player application, using a slider as a seek bar, I wanted the pointing rather than the grabbing cursor when the user is scrubbing through the audio.

edwloef:
I would find this useful as well, for a similar use-case (fader in an audio mixer).

To note a few concrete examples of sliders not using the `Grab`/`Grabbing` cursors:
- Spotify uses the `Pointer` cursor for every slider I've tested
- Discord uses the `ResizingHorizontally` cursor for every slider I've tested

PR #3119, opened 2025-11-21
add efficient `Edit::ReplaceAll` variant for `TextEditor`

mfreeborn:
Closes [discord discussion](https://discord.com/channels/628993209984614400/1021827652321038529/threads/1436312651176345630).

There's some debate as to the most optimal type for the contents of the `Arc`. I think the result of that decision may well apply to `Edit::Paste<Arc<String>>`, too, but I don't want to get too bogged down in this specific PR myself.

### DIFF
--- a/core/src/text/editor.rs
+++ b/core/src/text/editor.rs
@@ -114,6 +114,8 @@ pub enum Edit {
     Backspace,
     /// Delete the next character.
     Delete,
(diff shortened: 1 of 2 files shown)


hecrj:
The intended way here would be to recreate the `text_editor::Content`.

PR #3120, opened 2025-11-21
make `LineHeight::to_absolute` take `impl Into<Pixels>`

mfreeborn:
Closes [discord discussion](https://discord.com/channels/628993209984614400/1021827652321038529/threads/1437072773993205870)

### DIFF
--- a/core/src/text.rs
+++ b/core/src/text.rs
@@ -204,9 +204,9 @@ pub enum LineHeight {
 
 impl LineHeight {
     /// Returns the [`LineHeight`] in absolute logical pixels.
-    pub fn to_absolute(self, text_size: Pixels) -> Pixels {
+    pub fn to_absolute(self, text_size: impl Into<Pixels>) -> Pixels {
         match self {
-            Self::Relative(factor) => Pixels(factor * text_size.0),
+            Self::Relative(factor) => Pixels(factor * text_size.into().0),
             Self::Absolute(pixels) => pixels,
         }
     }


hecrj:
What is the use case? This is a rather internal function.

mfreeborn:
The exact line of code is:

```rust
let text_content = column![current_version_text, next_version_text]
    .spacing(text::LineHeight::default().to_absolute(DEFAULT_FONT_SIZE.into()) / 2.0);
```

in which I'm setting the spacing of two pieces of text in a column relative to their font size.

i.e. calculating a visually appealing line spacing for these two lines:

image

hecrj:
Any reason you are not using `text!("...\n...")`?

mfreeborn:
Err, yea I could do that..!

janTatesa:
My use case of LineHeight is for example when I have to do manual layout calculations because of Length::Adaptive not being a thing yet. Or maybe it could be useful to smn when making a custom widget

mahlquistj:
We also use `LineHeight::to_absolute` in [iced_aksel](https://github.com/QuistHQ/iced_aksel) when rendering text via. the Mesh renderer - While this is a very odd use case to be fair (And we don't really *need* it to take `Into<Pixels>`), i could see others wanting to do the same in other custom widgets like @janTatesa is suggesting. And also, to me, taking `Into<Pixels>` does reflect the proper usage of the method more clearly.

https://github.com/QuistHQ/iced_aksel/blob/d4218de067bf898b103319387047f2ed22d870b0/iced_aksel/src/render/cache/mesh.rs#L360

PR #3121, opened 2025-11-21
Update `svg` example with best practices of storing `Handles` in app state

mfreeborn:
According to the `Comet`, this doesn't actually make much difference in terms of performance on my laptop (before and after are both laggy on a debug build, and both are basically equally performant on a release build).

That said, I believe it's best practice in the general case to avoid creating `Handle`s in your view function, and instead to create them once and put them in your app state. Subsequent `clone`ing  of the `Handle` in `view` is then just a cheap `clone` of a `usize` and an `Arc`. 

In the spirit of getting people started with the correct way of doing things, I think this should be reflected in the example. This particular footgun caught me out, as one of my very first interactions with `iced` was experimenting with some large SVGs as background images.

This might need to change again if the new `image::allocate` feature gets ported to `svg`, but I think this is still valid in the meantime.

EDIT: I've added a couple of lines to the `svg` docs to reflect this.

### DIFF
--- a/widget/src/helpers.rs
+++ b/widget/src/helpers.rs
@@ -1804,6 +1804,11 @@ pub fn image<Handle>(handle: impl Into<Handle>) -> crate::Image<Handle> {
 ///
 /// Svg widgets display vector graphics in your application.
 ///
+/// Note that creating a [`Handle`] can be an expensive operation, so it is
+/// best practice to create the [`Handle`] once and store it in your app state. Your
+/// view function can then simply take a cheap clone of the stored [`Handle`], rather
+/// than needing to recreate it each time.
+///
 /// [`Svg`]: crate::Svg
 /// [`Handle`]: crate::svg::Handle
 ///


pml68:
> Subsequent `clone`ing of the `Handle` in `view` is then just a cheap `clone` of a `usize` and an `Arc`.

On that topic, I wonder why `image::Handle` doesn't use `Arc` internally? At least for its `Bytes` and `Rgba` variants.

edwloef:
`Bytes` is already internally reference counted, as far as I understand.

mfreeborn:
For completion and general interest, here's the implementation for `Bytes`:

```rust
pub struct Bytes {
    ptr: *const u8,
    len: usize,
    // inlined "trait object"
    data: AtomicPtr<()>,
    vtable: &'static Vtable,
}

```

pml68:
Hmm https://docs.rs/bytes/latest/bytes/struct.Bytes.html#sharing

mfreeborn:
> Hmm https://docs.rs/bytes/latest/bytes/struct.Bytes.html#sharing

Hmm indeed, that's interesting

pml68:
For reference, here's the clone used for owned data: https://docs.rs/bytes/latest/src/bytes/bytes.rs.html#1123

pml68:
The iced methods take any `impl Into<Bytes>`, so the actual clone used will depend on what you create it from.

hecrj:
@pml68 @mfreeborn Why is this "Sharing" section relevant?

hecrj:
`from_path` is fine. This is a non-issue.

mfreeborn:
> @pml68 @mfreeborn Why is this "Sharing" section relevant?

It isn't relevant to this PR. This PR was about promoting best practice in the examples.

hecrj:
The best practice is to avoid having unnecessary GUI state in your application.

mfreeborn:
I understand that for this example. I still believe there's a gap in the documentation somewhere (whether that be the API docs, the book or the examples) regarding `view` wanting to be a cheap function versus the cost of uploading data to the GPU on every call (as creating a Handle does, if I'm not mistaken), but I will leave it at that in terms of pursuing this point further.

pml68:
> @pml68 @mfreeborn Why is this "Sharing" section relevant?

It describes how copying / cloning is handled depending on how you create the `Bytes`.

hecrj:
> I still believe there's a gap in the documentation somewhere (whether that be the API docs, the book or the examples) regarding view wanting to be a cheap function versus the cost of uploading data to the GPU on every call (as creating a Handle does, if I'm not mistaken)

It doesn't for paths.

I think this is an API problem, not a documentation problem.

> It describes how copying / cloning is handled depending on how you create the Bytes.

What is interesting about it? That it's cheaper than cloning an `Arc` in some cases?

pml68:
Interesting to my tired brain that couldn't fully comprehend it at the time of me writing the comment, mostly. My "issue" is non-existent anyway, with the Bytes impl being like this.

hecrj:
Ah, got it!

Since both of you seemed to acknowledge it, I thought I was missing something important.

mfreeborn:
> It doesn't for paths.

That's what I was missing, thanks for clarifying.

hecrj:
Maybe we could make the `from_path` constructor private since both `image` and `svg` take `Into<Handle>` and `From<Into<PathBuf>>` is implemented for them.

PR #3123, opened 2025-11-23
Prevent crashes from `svg` rendering

hammerlink:
When rendering certain SVGs, resvg::render can panic because resvg uses several assert! calls internally. This caused application crashes in Cosmic Desktop: https://github.com/pop-os/cosmic-epoch/issues/2452

This patch wraps the render call in std::panic::catch_unwind. If resvg panics, the panic is caught and None is returned instead of crashing the app.

Example problematic SVG:
```
<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">                                                    
  <defs>                                                                                                             
    <filter id="specular" width="50" height="50">                                              
      <feSpecularLighting result="specular" lighting-color="white"                                                   
                          x="10" y="10" width="5" height="5">                                                        
        <fePointLight x="50" y="50" z="200"/>                                                                        
      </feSpecularLighting>                                                                                          
    </filter>                                                                                                        
  </defs>                                                                                                            
  <rect width="100" height="100" fill="blue" filter="url(#specular)"/>                                               
</svg>
```

### DIFF
--- a/tiny_skia/src/vector.rs
+++ b/tiny_skia/src/vector.rs
@@ -7,7 +7,8 @@ use tiny_skia::Transform;
 
 use std::cell::RefCell;
 use std::collections::hash_map;
-use std::{fs, panic};
+use std::fs;
+use std::panic;
 use std::sync::Arc;
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


hecrj:
Isn't this a `resvg` bug?

hammerlink:
> Isn't this a `resvg` bug?

Yeah, it is a resvg bug. I mainly pushed this because it’s causing random crashes on Cosmic Desktop right now, and the fix for this issue in resvg has been sitting open since August. The fix should come from resvg, but this at least stops apps from crashing. Thanks for merging it.

PR #3128, opened 2025-11-29
Add `CornerPreference` window setting for Windows

B0ney:
I'm not a great fan of how windows 11 makes the corners comically round, so I've added the ability to customise this under the platform specific options for windows.

The `CornerPreference` enum was pretty much yoinked from winit.


mahlquistj:
@B0ney why not just re-export the `CornerPreference` enum? Any specific reason?

B0ney:
> @B0ney why not just re-export the `CornerPreference` enum? Any specific reason?

If we do that, `iced_core` will now depend on `winit`.

PR #3129, opened 2025-11-29
Enhance toggler styling

B0ney:
Makes it possible to use gradients for styling the background and foreground.

Also added the option to override its border radius, making squarish-looking togglers possible. 

I've chosen to use `Option<Radius>` instead of just `Radius` because while we could pass a large value to make it look round, doing it that way feels a bit hacky. I think passing `None` and letting the widget calculate the correct border radius is a bit more sensible.

I've also removed `SPACE_RATIO` because I found it to interfere with custom borders. And it does feel a bit of a hack to mitigate the background from bleeding through. 

While this does reintroduce bleeding, I've updated its default style to preserve how it looked originally.

The only difference is that the updated toggler has a slightly thicker outline.

**Before**:
before

**After**:
after


hecrj (approved):
Thanks!

The issue with leveraging foreground border to fake the padding is that rendering a quad is not an idempotent operation. Rendering two quads in the same spot is additive due to anti-aliasing and alpha blending. This can cause jagged rounded corners.

So, I have re-introduced the idea of a `padding_ratio`. It seems to work relatively well for reasonable scale factors (`master` produces misalignments as well).

I have exploratory work in the `hinting` branch that should let us query the final `scale_factor` during `draw`, which would allow us to perfectly align the toggle to the pixel grid. So, we can circle back when that lands.

Issue #3103, opened 2025-11-05
image::Viewer widget clipping the bottom and right sides when the zoomed image exceeds the container


stevekki:
### Is your issue REALLY a bug?

- [x] My issue is indeed a bug!
- [x] I am not crazy! I will not fill out this form just to ask a question or request a feature. Pinky promise.

### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Is this issue related to iced?

- [x] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

When using the Viewer widget to zoom an image, the bottom and right sides get clipped once the zoomed image exceeds the container size, and this only happens when using wgpu as the rendering backend.

https://github.com/user-attachments/assets/2611f619-bb9a-44d7-b167-987f703470ec

a minimal repo for the issue: [https://github.com/stevekki/iced_viewer_issue](https://github.com/stevekki/iced_viewer_issue)

Run the example with this command: ```ICED_BACKEND=wgpu cargo run```

main.rs

```rust
use iced::{
    Length,
    widget::image::{Handle, Viewer},
};

pub fn main() -> iced::Result {
    iced::application(App::new, App::update, App::view).run()
}

#[derive(Debug, Clone)]
enum Message {}

struct App {
    handle: Handle,
}

impl App {
    fn new() -> Self {
        Self {
            handle: Handle::from_path("images/ferris.png"),
        }
    }

    fn update(&mut self, _: Message) {}

    fn view(&self) -> iced::Element<'_, Message> {
        Viewer::new(self.handle.clone())
            .width(Length::Fill)
            .height(Length::Fill)
            .into()
    }
}
```

### What is the expected behavior?

When the zoomed image exceeds the container, it should not be cropped at the bottom or right edges.

https://github.com/user-attachments/assets/86bf4bc7-a651-419a-ada5-f74628884425

### Version

master

### Operating System

Linux

### Do you have any log output?

```shell

```

ghost:
This issue has only appeared since commit 7c11ccb0466594bbb4abaa83f16814f699e429c1
Tested on both Windows 10 and Linux, this issue only occurs when using wgpu as the rendering backend.

PR #3130, opened 2025-11-30
Fix image::Viewer widget always clipping the bottom and right sides when the zoomed image exceeds the container

ghost:
## Before the fix
https://github.com/user-attachments/assets/23c574ca-21f3-481c-9d51-453b4a377e18

## After the fix
https://github.com/user-attachments/assets/30bed46d-5eb2-4764-ae1f-2a0a322bf328

Fixes https://github.com/iced-rs/iced/issues/3103.

### DIFF
--- a/widget/src/image/viewer.rs
+++ b/widget/src/image/viewer.rs
@@ -311,7 +311,7 @@ where
         _style: &renderer::Style,
         layout: Layout<'_>,
         _cursor: mouse::Cursor,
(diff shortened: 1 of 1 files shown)


hecrj:
I can't reproduce this in `master`, and I'm fairly sure this solution breaks `Viewer` inside scrollables.

Could you open an issue with an SSCCE?

PR #3131, opened 2025-11-30
Implement `*Assign` traits for operations on `Point` and `Vector`

edwloef:
... and replace the manual `Default` impl with a `derive` (not sure why clippy didn't lint there, maybe I'm missing something)


hecrj (approved):
Thanks!

PR #3134, opened 2025-12-01
Fix copy/paste shortcuts for non latin keyboard layouts

sopvop:
Currently if I switch from US English layout to Russian, copy paste stops working. That's because keyboard input is matched on input character, which depends on current keyboard layout.

This patch will make it match keyboard input by physical key, so that CUA shortcuts still work even if you switch layout.

This is consistent with other gui frameworks, such as Qt or gtk, and browser behavior.

### DIFF
--- a/core/src/keyboard/key.rs
+++ b/core/src/keyboard/key.rs
@@ -30,6 +30,91 @@ impl Key {
             Self::Unidentified => Key::Unidentified,
         }
     }
+
+    /// Tries to convert this logical [`Key`] into its latin character, using the
+    /// [`Physical`] key provided for translation if it isn't already in latin.
+    ///
+    /// Returns `None` if no latin variant could be found.
+    ///
+    /// ```
+    /// use iced_core::keyboard::key::{Key, Named, Physical, Code};
+    ///
+    /// // Latin c
+    /// assert_eq!(
+    ///     Key::Character("c".into()).to_latin(Physical::Code(Code::KeyC)),
+    ///     Some('c'),
+    /// );
+    ///
+    /// // Cyrillic с
+    /// assert_eq!(
+    ///     Key::Character("с".into()).to_latin(Physical::Code(Code::KeyC)),
+    ///     Some('c'),
+    /// );
+    ///
+    /// // Arrow Left
+    /// assert_eq!(
+    ///     Key::Named(Named::ArrowLeft).to_latin(Physical::Code(Code::ArrowLeft)),
+    ///     None,
+    /// );
+    /// ```
+    pub fn to_latin(&self, physical_key: Physical) -> Option<char> {
+        let Self::Character(s) = self else {
+            return None;
+        };
+
+        let mut chars = s.chars();
+        let c = chars.next()?;
+
+        if chars.next().is_none() && c <= '\u{ff}' {
+            return Some(c);
+        }
+
+        let Physical::Code(code) = physical_key else {
+            return None;
+        };
+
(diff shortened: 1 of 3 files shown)


edwloef:
Do be aware that this could break other layouts, such as French AZERTY (Q and A are swapped). Maybe a good solution would be to first match on the character, and only match on the key code if none of the characters match? Though that might result in duplicate keybinds.

Alternatively, a crate like [sys-locale](https://crates.io/crates/sys-locale) could be used to implement full locale-specific keybinds (or at least decide between using `key` or `physical_key`) but I'm not sure whether that would still be considered in-scope.

sopvop:
I've just checked French AZERTY and indeed it breaks it and makes inconsistent with qt and browser behavior.

It does not really depend on locale. I guess to do it properly will require to get current layout from whatever os/compositor reports, which is more in a winit territory.
Qt has [this monstrosity](https://github.com/qt/qtbase/blob/dev/src/gui/kernel/qplatformtheme.cpp#L184) for standard keybinds matching.

Matching on on latin characters first, and physical keys second would probably be a good stopgap. Improvement over current behavior at least.

sopvop:
Made it match on logical key first, and physical key second.  Unfortunately this makes both Ctrl-A and Ctrl-Q select all text on AZERTY keyboards. But at least this does not force users of non latin layouts to switch just for copy-paste.

edwloef:
This would probably need to be two matches, not multiple cases in the arms of one match statement.

Imagine you have keybinds on both Ctrl+Q and Ctrl+A, and are using an AZERTY layout. Since matching happens from top to bottom, both Ctrl+Q and Ctrl+A would match the same keybind (once matching the `key` and once matching the `physical_key`) making the other one untriggerable.

This of course doesn't directly affect `text_input` in the case of AZERTY, but there are probably layouts where this applies.

Maybe it would be easier to have some conversion option to go from `keyboard::key::Code` to `keyboard::Key`, and just run the match first on the normal key and then on the "converted" `physical_key`?

hecrj:
Maybe we can just match the Cyrillic character?

If you add a `dbg!(event)`, what are the keyboard events you get when you press `Ctrl+C` in your Russian layout?

edwloef:
That seems unsustainable to me, since this should in theory affect every non-latin layout, not just the Russian one. Might be good enough for an interim solution though 🤷

hecrj:
It seems more correct and easier to implement than the hacky solutions proposed here.

edwloef:
[This random blog post I found](https://www.velopen.com/blog/keyboard-shortcuts-on-non-latin-alphabet-layouts/) recommends effectively to match on `key` for characters < 880 (end of unicode latin) and match on `physical_key` for everything else. Maybe that's worth a try?

sopvop:
Maybe something like that? It probably should also incorporate the movement and whatever convert_macos_shortcut does.
Or maybe map just map KeyX + DiginN keycodes to Key::Character for non latin codepoints and put it into modified_keys.

sopvop:
I decided to do it the simple way, as was discussed above.  If key above latin1, it will use physical key. 

I decided against uniting this with `convert_macos_shortcut`, because it uses `modified_key` for some reason, and I don't own a macos to test it.

Doing shortucts properly is seriously messed up. [Here](https://github.com/qt/qtbase/blob/dev/src/gui/platform/unix/qxkbcommon.cpp#L618) is how Qt does it on wayland. It tries to find latin layouts configured in xkb, and converts keycode with that to use in shortcuts.

PR #3135, opened 2025-12-02
Unify `keyboard` subscriptions into `listen`

edwloef:
This makes it possible to use `Key::to_latin` for top-level application keybinds as well.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -368,8 +368,8 @@
 //! visible widgets of your user interface, at every moment.
 //!
 //! As with tasks, some modules expose convenient functions that build a [`Subscription`] for you—like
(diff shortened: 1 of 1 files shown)


edwloef:
I also just noticed that the wrong cutoff value was used for `to_latin`, which led characters like Ł (which is on the polish keyboard) to fall back to the key code. `\u{370}` is the first character in the Unicode [Greek and Coptic](https://unicodeplus.com/block/0370) block.

hecrj:
We'll get rid of these eventually, so let's not break them further.

Use `Subscription::filter_map` for the time being.

hecrj:
Well, let me get rid of them now.

PR #3139, opened 2025-12-09
Add window blur support for Windows 11

pml68:
This sets the backdrop to [Acrylic](https://learn.microsoft.com/en-us/windows/apps/design/style/acrylic), similarly to my earlier PR, #2900.


pml68:
@hecrj you also commented about alpha pre-multiplication on my previous PR, mentioning watsaig. I'm not sure whether that is "fixed" by now, but I thought I'd mention it here anyway.

PR #3152, opened 2025-12-17
WIP: Keep mouse cursor in the middle

daniel-prause:
This is a WIP to keep the mouse cursor in the middle of the handle like it is the case for e.g. for HTML sliders.
There is one problem that we have here: We can't get the border radius for the handle in the update method.

Since we can't access the elements attributes in the update method (which seems to be by design to seperate update calls from draw calls), I wonder how we are going to continue with this whenever we need sizes of elements to do calculations 🤔 

@hecrj Maybe you can take this PR as a base and find a way to access style properties like handle_border_radius inside update to calculate the correct mouse position?

### DIFF
--- a/widget/src/slider.rs
+++ b/widget/src/slider.rs
@@ -259,28 +259,34 @@ where
             let locate = |cursor_position: Point| -> Option<T> {
                 let bounds = layout.bounds();
 
-                if cursor_position.x <= bounds.x {
-                    Some(*self.range.start())
-                } else if cursor_position.x >= bounds.x + bounds.width {
-                    Some(*self.range.end())
-                } else {
-                    let step = if state.keyboard_modifiers.shift() {
-                        self.shift_step.unwrap_or(self.step)
-                    } else {
-                        self.step
-                    }
-                    .into();
+                // FIX ME: Get real handle border radius from somewhere.
+                let handle_radius = 7.0;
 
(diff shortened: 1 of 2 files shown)


edwloef:
This was discussed in Discord once, and I'm also already working on a PR for this 😆 essentially the idea was something along the lines of providing a handle `Size` to the widget struct itself, and then replacing `handle.{shape, border_color, border_width}` with a `Border`, like it is in `Rail`.

daniel-prause:
> This was discussed in Discord once, and I'm also already working on a PR for this 😆 essentially the idea was something along the lines of providing a handle `Size` to the widget struct itself, and then replacing `handle.{shape, border_color, border_width}` with a `Border`, like it is in `Rail`.

This will be interesting regarding the calculation of the size of the Handle, because the border_width right now tampers with the actual handle size 😄 . Could you push the PR as a draft? I'd like to take a look at what you could do in the current state! The interesting thing will be to access the data in the update method regarding the mouse cursor position, because it seems that it might be too late to do this in draw 🤔 ... right now inside the update method, we can't access the attributes of the widget e.g. style, which I think should be accessible from every method inside the class 🤔

daniel-prause:
Closing in favour of https://github.com/iced-rs/iced/pull/3155

PR #3155, opened 2025-12-19
Center mouse cursor in sliders

edwloef:
This tries to solve the same issue as #3152, but differently. Additionally, it enables you to configure the length of the slider handle in the drag direction.

### DIFF
--- a/widget/src/slider.rs
+++ b/widget/src/slider.rs
@@ -97,7 +97,7 @@ where
     on_release: Option<Message>,
     width: Length,
     height: f32,
-    handle_size: Size,
+    handle: Size,
     class: Theme::Class<'a>,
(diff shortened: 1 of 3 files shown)


widget/src/slider.rs
```diff
@@ -93,6 +93,7 @@ where
     on_release: Option<Message>,
     width: Length,
     height: f32,
+    handle_size: Size,
```
hecrj: Let's go with just `handle`.

hecrj (changes requested):

edwloef:
One possible simplification left is that one could remove the `height` field/method on `Slider`, and just use the handle size for layout. I'm not sure whether that's actually desirable, though.

PR #3159, opened 2025-12-21
Table styling additions

EmmanuelDodoo:
This PR adds setters for the style and class for the `table` widget.

### DIFF
--- a/widget/src/table.rs
+++ b/widget/src/table.rs
@@ -495,15 +495,50 @@ where
         cursor: mouse::Cursor,
         viewport: &Rectangle,
     ) {
+        let bounds = layout.bounds();
+        let metrics = tree.state.downcast_ref::<Metrics>();
+        let table_style = theme.style(&self.class);
+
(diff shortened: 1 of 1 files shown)


EmmanuelDodoo:
Also added the option for alternating row backgrounds. 

image

bicarlsen:
I would love to have rows individually styled, e.g. for highlighting a particular row.

EmmanuelDodoo:
Interesting I hadn't considered that use case. 

Though seeing as columns are the main way tables are currently handled, I'm struggling to see how row highlighting could be implemented

bicarlsen:
Perhaps in `Table` we could hold some extra state for custom row styling, e.g. `HashMap<row index, styling>`. Then in [`draw`](https://github.com/EmmanuelDodoo/iced/blob/c5c9de3962eb5ea918987859769984418f59ef8b/widget/src/table.rs#L518) we check if a custom style exists for that row. If so apply it, if not, go to the default.

PR #3168, opened 2025-12-28
Add `modifiers` argument to `Widget::update`, `Overlay::update`, `canvas::Program::update`, `shader::Program::update`

edwloef:
An alternative to #2733. As mentioned in https://github.com/iced-rs/iced/pull/2733#issuecomment-3560100229, this adds the current keyboard modifiers to the arguments of various `update` methods. I've opted to not add a new `Keyboard` type, since I find it unlikely that anyone actually cares about the state of any other keyboard keys in a way that would be affected by the discontinuity mentioned in https://github.com/iced-rs/iced/pull/2733#issuecomment-3556055395 (I can't see a use-case for non-modifier keys modifying non-keyboard events in general though -- feel free to change my mind on this).


hecrj:
I have plans for this already.

Just sit tight and wait a bit.

edwloef:
Ah I wasn't aware, sorry.

hecrj:
No need to be sorry. I'm exploring a redesign of `Shell` that should tackle this.

PR #3193, opened 2026-01-09
Markdown text color

EmmanuelDodoo:
This PR adds an option for setting the text color of rendered markdown. `markdown::Style` has been updated to include two new fields, `text_color` to set the color for basic text and `strong_color` for strong and emphasized text color


EmmanuelDodoo:
Ideally, there would be some granularity in assigning text colors so a single rendered markdown could have paragraphs with different colors.

I attempted this approach by having `markdown::Text` pass its color to `rich_text` but that meant requiring the additional bound from `rich_text::color_maybe`. My proposed solution is to add an additional `text_color` method to `text::Catalog` to remove the need for extra bounds though I wanted to discuss the idea before moving forward.

PR #3194, opened 2026-01-12
Use unstable sort in `group`

mati865:
`distance` function uses `hypot` internally, which has non-deterministic precision [1]. This is a problem for stable sorting, which needs deterministic results. And indeed, with the right circumstances [2] this breaks stable sort guarantees.

[1]: https://doc.rust-lang.org/std/primitive.f32.html#method.hypot
[2]: https://github.com/pop-os/cosmic-launcher/issues/352

With this change, I no longer get `panic_on_ord_violation` panic mentioned in https://github.com/pop-os/cosmic-launcher/issues/352.


mati865:
This is not sufficient, as std's unstable search also requires deterministic comparison...

PR #3196, opened 2026-01-13
Sort using more deterministic algorithm

mati865:
Replaces https://github.com/iced-rs/iced/pull/3194 which turned out to be insufficient.

`distance` method uses `hypot` which while useful for most of the cases, may break sorting assumptions like deterministic comparison.
As seen in: https://github.com/pop-os/cosmic-launcher/issues/352

### DIFF
--- a/core/src/point.rs
+++ b/core/src/point.rs
@@ -34,6 +34,18 @@ impl<T: Num> Point<T> {
 
         a.hypot(b)
     }
+
+    /// Computes squared distance to another [`Point`], using a deterministic algorithm.
+    /// Suitable for sorting algorithms.
+    pub fn distance_squared(&self, to: Self) -> T
+    where
+        T: Float,
+    {
+        let a = self.x - to.x;
+        let b = self.y - to.y;
+
+        a * a + b * b
+    }
 }
 
 impl<T> From<[T; 2]> for Point<T>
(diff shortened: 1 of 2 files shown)


edwloef:
Since `distance_slow` is only used as a comparison value, you could save calling `sqrt` to make it *actually* deterministic, no? And call it `distance_squared` or something

mati865:
As far as I know, `sqrt` doesn't introduce additional nondeterminism. `hypot` may use FMA which I do believe is the source of the previous nondeterminism within the same execution.
I'm somewhat hesitant to call it fully deterministic because the results will wary between targets, and possibly when evaluating it at both build time and runtime (software vs HW implementation). Current version should already be deterministic between invocations on the same platform.

I like the suggestion to skip `sqrt`, but I'm slightly worried that somebody (or perhaps even code linters) would propose to turn `x*x + y*y` into `x.powi(2) + y.powi(2)` to make it faster, but nondeterministic between the invocations once again.

edwloef:
Ah yes I misremembered, sqrt is specified by IEEE to be the correctly rounded infinite precision result.

mati865:
Applied the suggestion to make it `distance_squared`. This doesn't make it any more deterministic, but it should be marginally faster.

mati865:
@hecrj sorry to bother you directly but could you review it, so [Cosmic launcher crash issue](https://github.com/pop-os/cosmic-launcher/issues/352) doesn't return over and over?

PR #3215, opened 2026-01-18
Fix panic when atlas upload size is zero

po1onius:
When `w` or `h` is 0, a `usize` overflow panic occurs here
https://github.com/iced-rs/iced/blob/master/wgpu/src/image/atlas.rs#L369
https://github.com/iced-rs/iced/blob/master/wgpu/src/image/atlas.rs#L387


po1onius:
@hecrj Could you review this crash fix? I verified the affected behavior with local regression tests. Is there anything else needed to make this ready to merge?

PR #3223, opened 2026-01-23
Change `image` widget to borrow `Handle`

### LINKED ISSUE #2941: Displayed image not refreshing when new handle is constructed using the same path but different image file

### What happened?

When a handle is created from a certain path (e.g. "test.png") and is displayed, if the file is changed (to another image) and a new handle is created from the same path (e.g. iced::widget::image::Hande::from_path("test.png")) and is set as the handle to be displayed image is not refreshed, the original image remains visible.

I attached a short example highlighting this behavior.

[iced_image_bug.zip](https://github.com/user-attachments/files/20148450/iced_image_bug.zip)

The expected behavior would be for the displayed image to refresh to the new image. Currently there is a workaround which involves loading the image as bytes using fs::read and then creating a handle from the bytes.

### Version

crates.io release

### Operating System

Windows

```shell

```

### LINKED ISSUE #3160: 0.14.0 – Images flicker / disappear when using Handle::from_bytes with active subscription

### What happened?

**This issue did not occur in iced 0.13.1 and started happening in iced 0.14.0.**

https://github.com/user-attachments/assets/934450e8-25c0-401c-beb9-fecb91556b49

When a subscription (such as event::listen()) is enabled and the application starts, the image initially renders correctly. However, as soon as window events begin to flow (for example, moving the mouse or triggering redraws), the image starts to flicker or completely disappear.

If the subscription is disabled, this behavior does not occur and the image remains stable.

The issue happens when loading images using Handle::from_bytes, especially when the handle is recreated during frequent view() calls triggered by subscriptions.

### Details:
- NVIDIA graphics card with up-to-date proprietary drivers
- System running under an X11 session
- Vulkan support

### Minimal Example

```toml
[package]
name = "iced-image-bytes-mre"
version = "0.1.0"
edition = "2024"

[dependencies]
iced = { version = "0.14.0", features = ["image"] }
```

```rust
use iced::widget::image::Handle;
use iced::widget::{Column, Image, column};
use iced::{ContentFit, Element, Event, Subscription, Task, event};

    // Start an iced application using the functional API.
    // State type: u64
    // update() is called for every incoming Message.
    // view() is called after each update.
    iced::application(u64::default, update, view)
        // Enabling this subscription is sufficient to trigger the issue.
        // When active, the image may flicker or temporarily disappear.
        .subscription(subscription)
        .run()
}

#[derive(Debug, Clone)]
enum Message {
    // All window/system events are forwarded here
    EventOccurred(Event),
}

fn update(_value: &mut u64, message: Message) -> Task<Message> {
    match message {
        // Even though no state is changed here,
        // receiving frequent events causes continuous update/view cycles.
        Message::EventOccurred(_event) => Task::none(),
    }
}

fn subscription(_state: &u64) -> Subscription<Message> {
    // Listen to ALL window events:
    // mouse movement, redraw requests, focus changes, resize, etc.
    // This results in view() being called many times per second.
    let subscriptions = vec![
        event::listen().map(Message::EventOccurred)
    ];

    Subscription::batch(subscriptions)
}

fn view(_value: &u64) -> Column<Message> {
    // The image file is read from disk every time view() is called.
    // With an active subscription, this happens very frequently.
    let path = std::path::PathBuf::from("assets/image.png");

    let icon_widget: Element<Message> = match std::fs::read(&path) {
        Ok(bytes) => {
            // IMPORTANT:
            // A NEW image Handle is created on every view() call.
            //
            // In iced 0.14.0, Handle::from_bytes() appears to not have
            // a stable identity across calls.
            // Each call seems to be treated by the renderer
            // as a completely new image resource.
            //
            // This suggests that the internal image/texture cache
            // may continuously drop and recreate the GPU texture
            // when view() is re-executed frequently.
            //
            // This behavior appears to be the cause of the
            // flickering / temporary disappearance of the image,
            // especially when subscriptions are active.
            //
            // NOTE:
            // This behavior did NOT occur in iced 0.13.1,
            // which suggests a possible regression in iced 0.14.0.
            let handle = Handle::from_bytes(bytes);

            Image::new(handle)
                .content_fit(ContentFit::ScaleDown)
                .width(512)
                .height(512)
                .into()
        }
        Err(_) => iced::widget::text("").into(),
    };

    // A new widget tree is returned on every call
    column![
        icon_widget,
    ]
        .into()
}
```

As in iced 0.13.1, images should not flicker or disappear when a subscription is enabled.

### Version

master

### Operating System

Linux

```shell

```

### LINKED ISSUE #3256: canvas::Frame::draw_image permanently blank for RGBA images ≥ 2 MiB (streaming/video use case broken)

### What happened?

 When using canvas::Frame::draw_image with Handle::from_rgba in a per-frame rendering loop (e.g. streaming video, remote desktop), images are permanently invisible once their raw byte count exceeds 2 MiB. Smaller images render correctly.

Reproduction

  - Create a canvas::Program that calls Handle::from_rgba(w, h, pixels) and frame.draw_image(rect, image) every frame
  - Images where w * h * 4 >= 2_097_152 bytes are permanently blank
  - Images below the threshold render correctly

  Threshold breakdown by resolution:

  ┌──────────┬────────┬───────────┬───────────┐
  │   Size   │ Pixels │ Raw bytes │  Result   │
  ├──────────┼────────┼───────────┼───────────┤
  │ 960×540  │ 518K   │ 2.07 MB   │ ✓ Visible │
  ├──────────┼────────┼───────────┼───────────┤
  │ 1200×675 │ 810K   │ 3.24 MB   │ ✗ Blank   │
  └──────────┴────────┴───────────┴───────────┘

  Root Cause

  Two independent design decisions interact fatally:

  1. iced_wgpu/src/image/cache.rs:235
  const MAX_SYNC_SIZE: usize = 2 * 1024 * 1024;

  if image.len() < MAX_SYNC_SIZE {
      // Sync GPU upload, visible this frame ✓
  } else {
      // Async worker upload, returns None this frame ✗
      self.worker.upload(handle, image);
      None
  }

Images ≥ 2 MiB are uploaded asynchronously. None is returned for the current frame, which is fine if the same handle ID is reused next frame (the GPU texture will be ready).

  2. iced_core/src/image.rs
  pub fn from_rgba(width: u32, height: u32, pixels: impl Into<Bytes>) -> Handle {
      Self::Rgba {
          id: Id::unique(),  // ← new atomic ID every call
          ...
      }
  }

  Handle::from_rgba generates a brand-new unique ID on every call. In a streaming loop, each frame gets a different cache key. The async upload result from the previous frame is stored under the old ID, never retrieved, and immediately purged by trim(). The image is
  permanently blank.

  The async path assumption, "the caller will present the same handle next frame", is broken by the unique-ID design.

  Why it wasn't obvious

  - Works perfectly for static images (handle created once, reused every frame)
  - Works for small streaming images (under the 2 MiB sync threshold)
  - Only breaks for large streaming RGBA, exactly the remote desktop / video use case
  - No error, no warning — just a blank canvas

  Workaround (caller side)

  Cap texture dimensions to stay under the sync threshold:
  const MAX_TEX_PIXELS: u32 = 520_000; // safely under 2MB/4 = 524_288
  let pixel_count = draw_w * draw_h;
  let (tex_w, tex_h) = if pixel_count > MAX_TEX_PIXELS {
      let sf = (MAX_TEX_PIXELS as f32 / pixel_count as f32).sqrt();
      (((draw_w as f32 * sf) as u32).max(1), ((draw_h as f32 * sf) as u32).max(1))
  } else {
      (draw_w, draw_h)
  };
  draw_image scales the smaller texture to fill the display rect, correct position/size at reduced quality.

  Suggested Fix (upstream)

  Either:

  1. Force sync upload for Handle::Rgba, the pixel data is already in memory, the async path provides no benefit. Skip the MAX_SYNC_SIZE branch for in-memory RGBA handles.
  2. Add a stable-ID constructor, Handle::from_rgba_stable(id: u64, ...) so callers doing streaming can provide a consistent cache key.

  Option 1 is the clean fix. The async path only makes sense for file-backed handles where decoding is expensive.

  Version

  - iced = 0.14.0
  - iced_wgpu = 0.14.0

 image renders, same as images under 2 MiB.

### Version

crates.io release

### Operating System

Windows

```shell
nothing. That's actually part of the bug report. No panic, no warning, no tracing output. Silent failure. Worth highlighting — it makes it nearly impossible to diagnose without reading iced_wgpu source.
```

EmmanuelDodoo:
This PR changes `widget::Image` so it holds a `&Handle`. This introduces a compile error when image handles are created within `view` functions, something which often trips new users

Fixes #3256.
Fixes #3160.
Fixes #2941.

### DIFF
--- a/widget/src/helpers.rs
+++ b/widget/src/helpers.rs
@@ -1759,9 +1759,7 @@ where
 /// # Example
 /// ```no_run
 /// # mod iced { pub mod widget { pub use iced_widget::*; } }
-/// # pub struct State {
-///     img: image::Handle
-/// }
+/// # pub type State = ();
 /// # pub type Element<'a, Message> = iced_widget::core::Element<'a, Message, iced_widget::Theme, iced_widget::Renderer>;
 /// use iced::widget::image;
 ///
@@ -1770,7 +1768,11 @@ where
 /// }
 ///
 /// fn view(state: &State) -> Element<'_, Message> {
-///     image(&state.img).into()
+///     use std::sync::LazyLock;
+///
+///     static IMAGE: LazyLock<image::Handle> = LazyLock::new(|| image::Handle::from_path("ferris.png"));
+///
+///     image(&IMAGE).into()
 /// }
 /// ```
 /// <img src="https://github.com/iced-rs/iced/blob/9712b319bb7a32848001b96bd84977430f14b623/examples/resources/ferris.png?raw=true" width="300">
--- a/widget/src/image.rs
+++ b/widget/src/image.rs
@@ -3,9 +3,7 @@
 //! # Example
 //! ```no_run
 //! # mod iced { pub mod widget { pub use iced_widget::*; } }
-//! # pub struct State{
-//!     img: image::Handle
-//! }
+//! # pub type State = ();
 //! # pub type Element<'a, Message> = iced_widget::core::Element<'a, Message, iced_widget::Theme, iced_widget::Renderer>;
 //! use iced::widget::image;
 //!
@@ -14,7 +12,11 @@
 //! }
 //!
 //! fn view(state: &State) -> Element<'_, Message> {
-//!     image(&state.img).into()
+//!     use std::sync::LazyLock;
+//!
+//!     static IMAGE: LazyLock<image::Handle> = LazyLock::new(|| image::Handle::from_path("ferris.png"));
+//!
+//!     image(&IMAGE).into()
 //! }
 //! ```
 //! <img src="https://github.com/iced-rs/iced/blob/9712b319bb7a32848001b96bd84977430f14b623/examples/resources/ferris.png?raw=true" width="300">
@@ -43,9 +45,7 @@ pub fn viewer<'a>(handle: &'a Handle) -> Viewer<'a, Handle> {
 /// # Example
 /// ```no_run
 /// # mod iced { pub mod widget { pub use iced_widget::*; } }
-/// # pub struct State{
-///     img: image::Handle
-/// }
+/// # pub type State = ();
 /// # pub type Element<'a, Message> = iced_widget::core::Element<'a, Message, iced_widget::Theme, iced_widget::Renderer>;
 /// use iced::widget::image;
 ///
@@ -54,7 +54,11 @@ pub fn viewer<'a>(handle: &'a Handle) -> Viewer<'a, Handle> {
 /// }
 ///
 /// fn view(state: &State) -> Element<'_, Message> {
-///     image(&state.img).into()
+///     use std::sync::LazyLock;
+///
+///     static IMAGE: LazyLock<image::Handle> = LazyLock::new(|| image::Handle::from_path("ferris.png"));
+///
+///     image(&IMAGE).into()
 /// }
 /// ```
 /// <img src="https://github.com/iced-rs/iced/blob/9712b319bb7a32848001b96bd84977430f14b623/examples/resources/ferris.png?raw=true" width="300">


hecrj:
I ended up reverting this, since borrowing a handle defeats the purpose of a handle to begin with. They are meant to be cheaply cloneable and easy to own.

PR #3242, opened 2026-02-10
[WIP] dawn backend

csmoe:
Use https://github.com/google/dawn as alternative of wgpu, which is battle-tested by chrome.

Current changes are based on wgpu directly, will be separated later.


hecrj:
Could you maintain this renderer yourself?

hecrj:
We could also maybe introduce this as a feature flag?

But ideally, it'd be a separate crate that is built nicely on top of `iced_wgpu`. Happy to explore any public API changes you may need for that.

PR #3099, opened 2025-11-01
Fix the build on Android.

dcampbell24:
A tiny fix for Android builds.

PR #3245, opened 2026-02-11
Basic iOS and Android support

Leinnan:
This PR adds basic support for building and running apps on the iOS and Android. Implements #302 

Changes made:
- put clipboard behind feature- arboard doesn't compile on Android
- recreating window surfaces- Android does destroy those when app is put into background
- fix window settings for iOS
- made possible to run program on Android with `cargo-apk` and on iOS as a library called from xcode project

What does not work:
- keyboard does not show up on Android when text input is pressed, on iOS it works

Potential improvements that could be done as next steps:
- mobile example
- update workflows to at least test compiling for the mobile targets


Leinnan:
In case the mantainers would not like to mantain any mobile specific code, are you open to PRs that would make it easier to write mobile specific code in other crate but not repeating code like I did initialy in this [repo](https://github.com/Leinnan/iced_mobile)?

hecrj:
See https://github.com/iced-rs/iced/pull/3099#issuecomment-3591876926.

> In case the mantainers would not like to mantain any mobile specific code, are you open to PRs that would make it easier to write mobile specific code in other crate but not repeating code like I did initialy in this [repo](https://github.com/Leinnan/iced_mobile)?

It depends, but probably not.

PR #3253, opened 2026-02-16
fix #3232

### LINKED ISSUE #3232: text_input throws error when enter 3 Chinese words

### What happened?

Hi,

I am running on Debian 12,

I get Iced code from github, switch to 0.14 release branch, running styling example,
when enter any 3 Chinese words, it throws error in winit→Window.rs file line 379.

Image

It can continue entering Chinese words without any exception.

### Version

crates.io release

### Operating System

Linux

```shell

```

jeffreyniu:
I made a fix myself, please review and merge.

### DIFF
--- a/winit/src/window.rs
+++ b/winit/src/window.rs
@@ -352,7 +352,7 @@ where
             Some(selection) => {
                 vec![
                     text::Span::new(&preedit.content[..selection.start]),
-                    text::Span::new(if selection.start == selection.end {
+                    text::Span::new(if selection.start >= selection.end {
                         "\u{200A}"
                     } else {
                         &preedit.content[selection.start..selection.end]


kenz-gelsoft:
I confirmed this change won't break the Japanese input by testing following examples with Pop!_OS 24.04, macOS **and Windows**:
- `todo`
- `styling` (originally reported in #3232)
- `combo_box`
- on macOS, following examples don't work with Kanji(Hanji) characters correctly with or without this change (separate problem)
  - `editor`
  - `markdown`

(**EDIT**: this change is OK on Windows too. I experienced link.exe error, that was because I forgotten installing required Windows SDK components.)

winit/src/window.rs
```diff
@@ -352,7 +352,7 @@ where
             Some(selection) => {
                 vec![
                     text::Span::new(&preedit.content[..selection.start]),
-                    text::Span::new(if selection.start == selection.end {
+                    text::Span::new(if selection.start >= selection.end {
```
kenz-gelsoft: Here, the `input_method::Preedit.selection` value comes from winit's `Preedit` event.

https://docs.rs/winit/latest/winit/event/enum.Ime.html#variant.Preedit

The winit's `Preedit` event doesn't specify negative Range specifically, and

https://doc.rust-lang.org/std/ops/struct.Range.html

> It is empty if `start >= end`.

`Range` document says negative Range means the range is empty, so this change looks correct for me.

kenz-gelsoft:
I commented this PR as a contributor in this area to help the library author understanding this change.

As I don't have merge right, please wait for the devs reviewing your change.

jeffreyniu:
thanks, it may also needs merge to 0.14 release and build out a package.

PR #3254, opened 2026-02-19
Ellipsis

hecrj:
This PR updates `cosmic-text` to `0.18`—featuring text ellipsis support!

image

Most text widgets have a new `ellipsis` method that can be used to configure the new `Ellipsis` strategy.


GyulyVGC:
Cool!
It'd be awesome to also have a configuration to show a tooltip with the full text on hover if the ellipsis are active.
Would this be possible?

PR #3257, opened 2026-02-21
fix: wgpu from_rgba generates a new unique Id on every call.

### LINKED ISSUE #3256: canvas::Frame::draw_image permanently blank for RGBA images ≥ 2 MiB (streaming/video use case broken)

sheridans:
When pixel data exceeds MAX_SYNC_SIZE the async upload path stores the GPU result under that Id, but the next frame presents a different Id the result is never retrieved and the image is permanently blank.

Rgba handles already have their pixel data in memory; the async worker buys nothing. Force synchronous upload for Handle::Rgba regardless of size, consistent with how load_image already treats Rgba handles.

Fixes per-frame streaming use cases (video, remote desktop) where a new Handle::from_rgba is created each frameix(wgpu): always upload Rgba handles synchronously

Fixes #3256

### DIFF
--- a/wgpu/src/image/cache.rs
+++ b/wgpu/src/image/cache.rs
@@ -223,7 +223,12 @@ impl Cache {
         const MAX_SYNC_SIZE: usize = 2 * 1024 * 1024;
 
         // TODO: Concurrent Wasm support
-        if image.len() < MAX_SYNC_SIZE || cfg!(target_arch = "wasm32") {
+        // RGBA handles have their pixel data already in memory, sending them
+        // to the async worker buys nothing and breaks per-frame streaming use
+        // cases where Handle::from_rgba is called every frame (each call
+        // produces a new unique ID, so the async result is never retrieved).
+        let is_rgba = matches!(handle, core::image::Handle::Rgba { .. });
+        if image.len() < MAX_SYNC_SIZE || cfg!(target_arch = "wasm32") || is_rgba {
             let entry =
                 self.atlas
                     .upload(device, encoder, belt, image.width(), image.height(), &image)?;


hecrj:
> the async worker buys nothing

lmao I hope you used a clanker.

In any case, use [`image::allocate`](https://docs.rs/iced/latest/iced/widget/image/fn.allocate.html).

Issue #3258, opened 2026-02-22
The IME's candidate window is misplaced on Linux fcitx5 after #2918


kenz-gelsoft:
### Is your issue REALLY a bug?

- [x] My issue is indeed a bug!
- [x] I am not crazy! I will not fill out this form just to ask a question or request a feature. Pinky promise.

### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Is this issue related to iced?

- [x] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

I noticed [recently](https://github.com/iced-rs/iced/pull/3253#issuecomment-3910632193) the input method's candidate window hides preedit (composing) text as it is misplaced only on Linux **and fcitx5 combination**.

|NG: COSMIC<br>(+ cosmic-comp<br>+ fcitx5)|NG: GNOME<br>(+ mutter<br>+ fcitx5)|
|-|-|
|<img width="517" height="642" alt="Image" src="https://github.com/user-attachments/assets/742fcda9-7d1c-4ec2-8ccb-e01a39f8078d" />|<img width="515" height="853" alt="Image" src="https://github.com/user-attachments/assets/fe9d3072-5bf7-4125-90f7-04fd813eac1a" />|

The same code works well on Windows and macOS. So I overlooked this when I tested this change in

https://github.com/iced-rs/iced/pull/2918#issuecomment-3577775117

From the winit's documentation, this probably is a bug of ~~the winit's Linux implementation or it's documentation~~. The document says:

https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_ime_cursor_area

> Set the IME cursor editing area, where the position is **the top left corner of that area** and size is the size of this area starting from the position. An example of such area could be a input field in the UI or line in the editor.

> X11: - **area is not supported, only position**.

but it works as if to specify **the bottom left position** of the `ime_cursor_area`.

---

**EDIT:** This seems **the bug of fcitx5** from my additional testing with GNOME + ibus configuration.

|OK: GNOME<br>(+ mutter<br>+ ibus)|
|-|
|<img width="512" height="846" alt="Image" src="https://github.com/user-attachments/assets/a11f27d9-9fec-4182-97bd-6ab4d5bb29b6" />|

I have some workarounds for this, I'll open PR.

### What is the expected behavior?

The ime's candidate window should open below of the preedit like this:

|COSMIC (+ cosmic-com + fcitx5)|GNOME (+ mutter + fcitx5)|
|-|-|
[...]

PR #3259, opened 2026-02-22
fix(ime): the candidate window position of fcitx5 which does't work well with winit's set_ime_cursor_area()

kenz-gelsoft:
This is a workaround fix of #3258.

~~As I noted in #3258, this probably be a bug of the winit's Linux implmentation.~~

As I noted in #3258, this seems to be **a bug of the fcitx5**.

The fcitx5 is just one of the input method frontends on Linux, but it is the de-facto standard on desktop environments other than GNOME which is tightly integrated with the ibus.

This workaround won't break if they will fix their ~~Linux~~ **fcitx5** implementation to handle the cursor area rectangle consistent with other platforms or **the ibus**. Because this change specifies zero height rect for Linux which is the left-top corner position (as they documented) == the bottom-left corner position (current actual behavior).

|Before|After|
|-|-|
|image|image|

### DIFF
--- a/winit/src/window.rs
+++ b/winit/src/window.rs
@@ -266,14 +266,16 @@ where
         }
 
         if self.ime_state != Some((cursor, purpose)) {
-            let (cursor_y, cursor_height) = if cfg!(any(target_os = "windows", target_os = "macos"))
-            {
-                (cursor.y, cursor.height)
-            } else {
-                // Specify the bottom-left position of the cursor because
-                // only the position is supported on Linux (or other platforms).
-                (cursor.y + cursor.height, 0f32)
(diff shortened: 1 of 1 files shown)


kenz-gelsoft:
Closes #3258

kenz-gelsoft:
rebased on master to fix Document check

kenz-gelsoft:
I changed this PR title and diff to match the PR that is merged in downstream(pop-os/iced).

kenz-gelsoft:
This was not a fcixt5 specific problem but a COSMIC specific problem. Should be fixed there.

Closing.

PR #3260, opened 2026-02-22
Sliders: replace Into<f64> with AsPrimitive<f64>

tritoke:
This change is intended to allow the Slider and VerticalSlider structs to work with larger integer types such as u64 and usize.

I had a use case for this on a personal project and @edwloef suggested the change on the discord so I've put up this PR for it.

### DIFF
--- a/core/src/angle.rs
+++ b/core/src/angle.rs
@@ -71,6 +71,18 @@ impl num_traits::FromPrimitive for Degrees {
     }
 }
 
+impl num_traits::AsPrimitive<f32> for Degrees {
+    fn as_(self) -> f32 {
+        self.0
+    }
+}
+
+impl num_traits::AsPrimitive<f64> for Degrees {
+    fn as_(self) -> f64 {
+        f64::from(self.0)
+    }
+}
+
 /// Radians
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd)]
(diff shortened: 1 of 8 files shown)


hecrj (approved):
Thanks!

hecrj:
Although on second thought, I'm not sure we want to pretend that `slider` can work with all ranges of `u64` and `usize`.

tritoke:
> Although on second thought, I'm not sure we want to pretend that `slider` can work with all ranges of `u64` and `usize`.

Yeah this was a thought I had as well, I think it might be worth mentioning in the docs that it uses conversions to/from f64 internally so it can represent up to 52(iirc) bits losslessly and the rest is imprecise? From what I can tell if you have multiple ways to control the same value the slider will not overwrite it's value if you have another way to change it. In my case I have a +/- 1 button and. a text entry box which is based on the true value.

I'm not around tonight but I'm happy to add this to the relevant docs if you think it's a good idea.

tritoke:
Took me a hot minute to get round to this but I've added a note to the docs for both mentioning the limitiation and the fact that if you are using any large u? to their limits f64 is not what is stopping you from achieving precision, its the fact you are using a slider lol. Happy to amend the wording ofc

hecrj (approved):

PR #3262, opened 2026-02-24
Add From for iced::widget::canvas::Path <-> lyon_path::Path

Pandapip1:
This is a trivial change that allows for direct conversion of `lyon_path::Path` objects into `iced::widget::canvas::Path` objects (which was not previously possible without iterating though the internals of the `lyon_path::Path` object) and vice versa, something that a use case I have (I am writing a CAM program for PCB manufacturing and am doing a lot of my processing using `lyon_path::Path` objects rather than iced's wrapped objects) requires.

~~Untested, hence the draft PR. Once tested, will undraft.~~ Actually, I genuinely can't see how this could fail. Undrafting anyway.

### DIFF
--- a/graphics/src/geometry/path.rs
+++ b/graphics/src/geometry/path.rs
@@ -77,3 +77,15 @@ impl Path {
         }
     }
 }
+
+impl From<lyon_path::Path> for Path {
+    fn from(raw: lyon_path::Path) -> Self {
+        Self { raw }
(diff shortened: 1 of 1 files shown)


Pandapip1:
Bump!

Pandapip1:
Looks like CI hasn't run, mind re-running it?

PR #3264, opened 2026-02-25
chore: improve counter example

janTatesa:
This pr:
- Changes formatting in update
- Centers the ui

### DIFF
--- a/examples/counter/src/main.rs
+++ b/examples/counter/src/main.rs
@@ -1,5 +1,5 @@
-use iced::Center;
-use iced::widget::{Column, button, column, text};
+use iced::widget::{Container, button, column, container, text};
+use iced::{Center, Length};
 
 pub fn main() -> iced::Result {
     iced::run(Counter::update, Counter::view)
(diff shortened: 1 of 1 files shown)


janTatesa:
Why aren't they? Container is arguably the second most used widget after text, so I think it's suitable for the first example one looks at. When it comes to formatting can you elaborate? I mean the change clearly simplifies it

hecrj:
Are you asking in good faith because you are genuinely curious? Or will you silently disagree when I provide an explanation?

If it's the former, then what do you think my concerns could be? What could be the reasons I don't like `container` used in the `counter` example? And why do I not like the one-liners in `update`?

Can you think of anything?

janTatesa:
In terms of container, I think you're trying to say that counter example should be as simple as possible. But I think the goal should be to introduce people to the library, which I think should involve one of the first elements they are going to use in their first app, which container will certainly. In terms of formatting I completely have no idea. Maybe that adding lines to non one liners is faster?

hecrj:
You are looking at it in a very short-sighted way.

> But I think the goal should be to introduce people to the library, which I think should involve one of the first elements they are going to use in their first app

The `counter` example is the example we use to introduce the basics of the library in [the overview](https://github.com/iced-rs/iced?tab=readme-ov-file#overview) and [the book](https://book.iced.rs/first-steps.html). The goal is to introduce the Elm Architecture. Layout is secondary here. `column` is already enough to introduce basic layout principles in a very natural way because the example organically needs it.

`container` just adds unnecessary noise. There are plenty of other examples that use it.

> Maybe that adding lines to non one liners is faster?

Shorter does not mean simpler. Relying on the fact that `+=` returns `()` is obscure, and may create the false notion that `update` handlers must be one-liners.

I've put quite some thought on this. I am even writing a book about it.

janTatesa:
Okay makes sense

PR #3271, opened 2026-03-06
feat: make default font settable at runtime

janTatesa:
Note that this is a breaking change as it introduces a new `set_default_font` method to `Renderer`


edwloef:
Maybe this should be a task instead of calling `settings` a bunch? `iced::font::load` is already a task, and the other settings can't be changed at runtime at all right now, so that'd make more sense to me.

PR #3274, opened 2026-03-11
Implement `animation::Interpolable` for `Color`

janTatesa:
Title

### DIFF
--- a/core/src/color.rs
+++ b/core/src/color.rs
@@ -148,7 +148,7 @@ impl Color {
         }
     }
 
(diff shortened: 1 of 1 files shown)


edwloef:
A bit simpler (and I think the same outcome) would be
```rs
crate::theme::palette::mix(self, other, ratio)
```

janTatesa:
> A bit simpler (and I think the same outcome) would be
> ```rs
> crate::theme::palette::mix(self, other, ratio)
> ```

Okay thanks

janTatesa:
I wonder if we should deprecate palette::mix now

PR #3279, opened 2026-03-12
feat: add scale factor to window opened event

edwloef:
... and remove one small helper on `Window` that wasn't consistently used anyways.

This makes it possible to immediately know the system scale factor of a window, in contrast to only after it changes once. With some (only slightly messy) math and only a little bit of flashing on scale factor change, this makes it possible to control the physical size of windows from Iced.

### DIFF
--- a/core/src/window/event.rs
+++ b/core/src/window/event.rs
@@ -17,6 +17,8 @@ pub enum Event {
         /// The size of the created window. This is its "inner" size, or the size of the
         /// client area, in logical pixels.
         size: Size,
+        /// The scale factor of the created window.
+        scale_factor: f32,
     },
 
     /// A window was closed.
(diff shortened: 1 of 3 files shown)


hecrj:
This is fine but, in my experience, the initial scale factor can be wrong often.

In Hyprland at least, the initial scale factor is always 1 and it's only after the opening animation finishes that the proper scale factor event is received.

edwloef:
Interesting. I've only tested this on Sway, where it works properly. Does Hyprland fire a `Rescaled` event when the "real" scale factor kicks in?

hecrj (approved):

hecrj:
> Does Hyprland fire a Rescaled event when the "real" scale factor kicks in?

Yes, but it takes a second; so if `view` relies on it there can be flashing. The `text` example is affected, for instance.

PR #3283, opened 2026-03-17
Feat: add gesture events

b3nten:
This PR adds gestures support for pan, pinch, double tap, and rotate, for trackpad and touch.

### DIFF
--- a/core/src/event.rs
+++ b/core/src/event.rs
@@ -1,5 +1,6 @@
 //! Handle events of a user interface.
 use crate::clipboard;
+use crate::gesture;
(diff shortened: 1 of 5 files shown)


mfreeborn:
I would just note that mobile isn't supported, so the comments re iOS seem extraneous.

b3nten:
> I would just note that mobile isn't supported, so the comments re iOS seem extraneous.

Yeah you're right. The pan gesture could also be removed since it's scoped to iOS at the moment but in the future it's possible it's implemented on other systems.

PR #3292, opened 2026-03-23
Fix early return in image `State::prepare` skipping layer finalization

### LINKED ISSUE #3272: wgpu image pipeline crash on window resize — possible `return` vs `continue` issue in image State::prepare

### What happened?

I'm getting a crash when resizing my app window. It happens when images get scaled below 1px during the resize.

Looking at `wgpu/src/image/mod.rs` in `State::prepare`, the issue appears to be the two `return` statements (one for raster on line 238, one for vector on line 288) that fire when an image's bounds are < 1px:

```rust
if bounds.width < 1.0 || bounds.height < 1.0 {
    return;
}
```

These were added in `0fe99b19` and `1463ec84` as part of the pixel snapping fixes. `return` here exits the entire `prepare` function instead of just skipping that one image — which means the final `layer.push()`, `layer.prepare()`, and the `.clear()` calls at the bottom don't run. That leaves stale instances in the vec and causes the buffer to be undersized on the next frame.

Changing both to `continue` fixes the crash:

```diff
 if bounds.width < 1.0 || bounds.height < 1.0 {
-    return;
+    continue;
 }
```

The window should resize without crashing. Sub-pixel images should be skipped gracefully.

### Version

master

### Operating System

Linux

```shell
wgpu error: Validation Error

Caused by:
  In a CommandEncoder, label = 'iced_wgpu encoder'
    In a draw command, kind: Draw
      Instance 49 extends beyond limit 48 imposed by the buffer in slot 0. Did you bind the correct `Instance` step-rate vertex buffer?
```

f-o-o-g-s:
The sub-pixel bounds guards added in 0fe99b19 and 1463ec84 use `return` instead of `continue` inside the image loop in `State::prepare`. When any image in the batch has bounds smaller than 1px, `return` exits the whole function instead of just skipping that image. This means `layer.push()`, `layer.prepare()`, and the instance vec `.clear()` calls at the bottom never run, which leaves stale instances around and causes a wgpu buffer validation error on the next frame.

This changes both to `continue` so we just skip the sub-pixel image and let the rest of the function finish normally.

Fixes #3272

### DIFF
--- a/wgpu/src/image/mod.rs
+++ b/wgpu/src/image/mod.rs
@@ -235,7 +235,7 @@ impl State {
                     let clip_bounds = (*clip_bounds * scale).round();
 
                     if bounds.width < 1.0 || bounds.height < 1.0 {
-                        return;
+                        continue;
                     }
 
                     if let Some((atlas_entry, bind_group)) =
@@ -284,7 +284,7 @@ impl State {
                     let clip_bounds = (*clip_bounds * scale).round();
 
                     if bounds.width < 1.0 || bounds.height < 1.0 {
-                        return;
+                        continue;
                     }
 
                     if let Some((atlas_entry, bind_group)) = cache.upload_vector(


f-o-o-g-s:
made a small standalone repro in case it helps: https://gist.github.com/f-o-o-g-s/b7d07dbc0bda748d9fda7f89bc2232a1

took me a while to get it to trigger reliably. turns out you need two images big enough that they land on separate bind_groups, so the pipeline does a `layer.push` partway through the batch, and then a third image going sub-pixel right after that. once those line up the early `return` skips the buffer resize and it falls over after a couple frames:

```
Instance N extends beyond limit M imposed by the buffer in slot 0
```

in the repro you just drag the window narrow (or click the button) and it goes. swapping the two `return`s for `continue` is the only change.

short clip of the crash:

https://github.com/user-attachments/assets/d789aa46-233b-4f7a-a5b7-fcbb11a732c2

i've also had this patch running on a fork for my own app (a navidrome client) for a few months now without any trouble, so at least anecdotally it seems safe.

hecrj (approved):
Thanks!

Issue #2102, opened 2023-09-27
RTL text requiring latin to be displayed in inputs


ElhamAryanpur:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

So I was checking around and saw that Cosmic-text have been added to iced, and because of that I thought RTL languages are supported now. So I started checking some demos if that's true and...

When writing in RTL language, they are not displayed:
![image](https://github.com/iced-rs/iced/assets/35463470/60cb9013-2bc4-4f23-8896-8fd2cf44574d)

**HOWEVER!** When you add a latin letter or text in addition, it starts displaying correctly and the language using correct glyphs too:
![image](https://github.com/iced-rs/iced/assets/35463470/6b233e59-de5d-4eda-a1e5-a8797dbbf1ef)

From my tests, only latin letters works, so special characters and such does not make it display.

### What is the expected behavior?

It should correctly display without need for latin letters.

### Version

master

### Operating System

Linux

### Do you have any log output?

```shell
2023-09-27T12:16:07.697761Z  INFO winit::platform_impl::platform::x11::window: Guessed window scale factor: 1.5    
2023-09-27T12:16:07.888346Z  INFO wgpu_hal::vulkan::instance: Instance version: 0x403108    
2023-09-27T12:16:07.888428Z  INFO wgpu_hal::vulkan::instance: Enabling debug utils    
2023-09-27T12:16:07.889117Z  INFO wgpu_hal::vulkan::instance: Enabling device properties2    
2023-09-27T12:16:07.900702Z  INFO wgpu_hal::gles::egl: Loading Wayland library to get the current display    
2023-09-27T12:16:07.901063Z  INFO wgpu_hal::gles::egl: Loading X11 library to get the current display    
2023-09-27T12:16:07.902241Z  INFO wgpu_hal::gles::egl: Using X11 platform    
2023-09-27T12:16:07.902906Z  INFO wgpu_hal::gles::egl: Enabling EGL debug output    
2023-09-27T12:16:07.954970Z  INFO wgpu_hal::gles::egl: Display vendor "Mesa Project", version (1, 5)    
2023-09-27T12:16:07.955797Z  INFO wgpu_hal::gles::egl:  EGL surface: +srgb    
2023-09-27T12:16:07.956421Z  INFO wgpu_hal::gles::egl:  Trying native-render    
[...]

Issue #1454, opened 2022-10-04
Support for non-Latin languages


koutheir:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Is this issue related to iced?

- [X] My hardware is compatible and my graphics drivers are up-to-date.

### What happened?

Running the `tour` package, and typing Arabic text on the `Text input` page, displays squared interrogation point characters.

```bash
$ cargo run --package tour
```

![Screenshot](https://user-images.githubusercontent.com/1145849/193712545-14823977-9ab9-4db1-9181-9fd0165b3753.png)

### What is the expected behavior?

Arabic text should be displayed correctly.

### Version

master

### Operative System

Linux

ids1024:
I believe with the `default_system_font` feature enabled, this should at least be able to find a font with the appropriate characters?

Though Arabic text still won't be right without RTL support (https://github.com/iced-rs/iced/issues/250) and text shaping (https://github.com/iced-rs/iced/issues/33).

koutheir:
I added `default_system_font` to the iced features in `iced/examples/tour/Cargo.toml`:
```toml
[dependencies]
iced = { path = "../..", features = ["image", "debug", "default_system_font"] }
env_logger = "0.8"
```

and then run:
```bash
$ cargo run --package tour
```

But no Arabic characters are displayed properly.

![Screenshot](https://user-images.githubusercontent.com/1145849/193715629-a0f83af9-cba9-431c-9a2b-033b3a3c3874.png)

Gabirel:
Same issue here on MacOS.

adals:
> I added default_system_font to the iced features in iced/examples/tour/Cargo.toml:

<img width="801" alt="Screen Shot 2022-10-08 at 9 59 36 PM" src="https://user-images.githubusercontent.com/169691/194723385-705c0a9e-41b7-4e6d-8862-5c2588ddffc5.png">

i think that your problem is lacking a font that contains the relevant glyphs.

but RTL is still not supported yet

koutheir:
>i think that your problem is lacking a font that contains the relevant glyphs.

I disagree, as I routinely write Arabic text in other applications (libreoffice, kate, etc.) and I don't have any font issues.

mmstick:
@koutheir Check this out! https://github.com/pop-os/libcosmic/pull/12 Please try out Arabic in the text example and help locate bugs in the rendering.

ids1024:
[...]

PR #3294, opened 2026-03-27
Fix RTL and mixed BiDi text in text_editor and text_input

### LINKED ISSUE #1877: Right-to-left script not showing in `TextInput` widget

### What happened?

Thanks to #1830, regular `Text` widgets can handle complex scripts, including left-to-right (LTR), right-to-left (RTL), and mixed LTR/RTL. But this is only partially true for the `TextInput` widget, such that complex scripts will be displayed properly only if the content of the text input starts with an LTR (alphabet) character. In other words, the content of text input is only displayed when the content should be aligned LTR.

The screen capture below demonstrates the issue using the Todos example of the library, but this issue is reproducible with any `TextInput` instance. The video covers LTR, RTL, LTR->RTL, and RTL->LTR.

https://github.com/iced-rs/iced/assets/39179450/899d857a-61bb-43d7-ab0c-63dc7941c82c

Ultimately, if the user input starts with an RTL (alphabet) character, the entire content should be displayed RTL and be right-aligned. However, since `TextInput` does not yet have horizontal text alignment support, a first step would be to just make the text show up, even if left-aligned.

I don't have enough knowledge of text rendering or the internals of this library to pinpoint where the issue stems from and how to fix it. But I did try a couple of basic modifications to [this](https://github.com/iced-rs/iced/blob/fcb1b454368638209862aeb5db41bc5f7d6d51a7/widget/src/text_input.rs#LL1122C9-L1122C26) function call:
* Changed the hard-coded `Alignment::Left` to `Alignment::Right`
* Changed `bounds.x` to a value expected by `Alignment::Right`

I also tried rendering with different values of `offset` with `renderer.with_translation(...)`. None of my attempts so far has made the missing text appear.

I would be happy to look into fixing the issue if someone can point me to the potential location of the issue in the codebase.

### Version

master

### Operative System

macOS

_No response_

hojjatabdollahi:
~~Draft until https://github.com/pop-os/cosmic-text/pull/491 is merged.~~
~~Draft until new cosmic-text release.~~
~~Draft until https://github.com/iced-rs/cryoglyph/pull/3 is merged.~~

This fixes issues with RTL text in `text_editor` and `text_input`.

Currently both hit detection and highlighting is wrong.

This PR switches from the custom highlight algorithm to the one in `cosmic-text` for `editor`.

before:

https://github.com/user-attachments/assets/a57d0a40-f013-46b2-b0d1-3d3ed086a5a8

after:

https://github.com/user-attachments/assets/d94eb5b1-695b-4ef4-97c0-d302e32bb837

This also updates text input widget and adds RTL and mixed bidi support. The changes to text input is bigger, because it's a specialized input (single line, password support, etc), so I had to copy bidi support for it from cosmic-text.

image

Fixes https://github.com/iced-rs/iced/issues/1877, https://github.com/iced-rs/iced/issues/2102, and https://github.com/iced-rs/iced/issues/1454

### DIFF
--- a/core/src/text.rs
+++ b/core/src/text.rs
@@ -278,16 +278,34 @@ impl Hash for LineHeight {
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum Hit {
     /// The point was within the bounds of the returned character index.
-    CharOffset(usize),
+    CharOffset(usize, Affinity),
 }
 
 impl Hit {
     /// Computes the cursor position of the [`Hit`] .
     pub fn cursor(self) -> usize {
         match self {
-            Self::CharOffset(i) => i,
+            Self::CharOffset(i, _) => i,
         }
     }
+
+    /// Returns the cursor [`Affinity`] of the [`Hit`].
+    pub fn affinity(&self) -> Affinity {
+        match self {
+            Self::CharOffset(_, a) => *a,
+        }
+    }
+}
+
+/// Cursor affinity for BiDi text. At the boundary between runs of different
+/// directions, the same byte index can map to different visual positions.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum Affinity {
+    /// Associate with the run before the cursor index.
+    #[default]
+    Before,
+    /// Associate with the run after the cursor index.
+    After,
 }
 
 /// The difference detected in some text.
--- a/core/src/text/paragraph.rs
+++ b/core/src/text/paragraph.rs
@@ -1,7 +1,7 @@
 //! Draw paragraphs.
 use crate::alignment;
 use crate::text::{
-    Alignment, Difference, Ellipsis, Hit, LineHeight, Shaping, Span, Text, Wrapping,
+    Affinity, Alignment, Difference, Ellipsis, Hit, LineHeight, Shaping, Span, Text, Wrapping,
 };
 use crate::{Pixels, Point, Rectangle, Size};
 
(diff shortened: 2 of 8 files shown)


abdu-benayad:
Tested this PR on a real use case — building an Arabic POS UI in iced. Cherry-picked the three commits cleanly onto current iced master (no conflicts) and rebuilt our workspace plus four downstream demo crates. All built clean.

**Test case** — `text_editor` pre-filled with `"إجمالي: 12.50 ر.س"` (mixed Arabic + Latin digits + Arabic — the case Slint #7267 also fails). Probed:

| Behavior | iced master | iced + this PR |
| --- | --- | --- |
| Click hit-test in mixed-script | caret rendered at wrong physical position vs the click | ✅ correct |
| Arrow keys move the visual caret as expected | caret pinned, decoupled from logical position | ✅ correct |
| Type at the visual cursor | inserted at a different logical position than the rendered caret | ✅ correct |
| Selection across an LTR↔RTL boundary | scattered / wrong region | ✅ multi-rect highlight covers the logical range correctly |
| Ctrl+C / Ctrl+V | n/a | ✅ working |

**Notes on the design**

- `Cursor::move_visual(forward, by_words, rtl, value)` with the `forward ^ rtl` XOR is the right model — matches what users expect (Right-arrow always moves the caret visually rightward, regardless of script direction).
- Replacing the custom glyph-iteration code in `graphics/src/text/editor.rs` with delegation to `LayoutRun::highlight()` / `LayoutRun::cursor_position()` is exactly the architectural call I'd hoped for. ~150 fewer lines of bidi-blind code.
- The `Affinity` addition to `Hit::CharOffset(usize, Affinity)` is a public API break, but it's the textbook fix for bidi cursor disambiguation and the scope is reasonable.

Happy to test additional scenarios if useful.

PR #3295, opened 2026-03-30
[Scrollable] Add `padding` field for `Scrollbar`

git-f0x:
This allows e.g. for the Scrollbar to be shorter than the scrollable.
Useful when wanting content to extend to the edge of the window, but not the scrollbar itself.
image
Example of how it's used in COSMIC.

Should the `width` field be renamed to `girth`, similar to `progress_bar`? It can be a bit confusing since the scrollbar can be horizontal.

### DIFF
--- a/widget/src/scrollable.rs
+++ b/widget/src/scrollable.rs
@@ -313,6 +313,7 @@ pub struct Scrollbar {
     scroller_width: f32,
     alignment: Anchor,
     spacing: Option<f32>,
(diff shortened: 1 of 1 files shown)


git-f0x:
A simpler implementation was merged in 81324eb and uses just an f32 to shrink the scrollbar on both ends (matching what I did in the cosmic iced fork), rather than individual control like here, which was probably unnecessary, so I'll close this.

PR #3298, opened 2026-03-31
fix: sync text_input modifier state from KeyPressed events

alecthomas:
Previously, text_input only updated keyboard_modifiers via
ModifiersChanged events. A widget created while a modifier
was already held (e.g. a search bar opened via Cmd+F) never
received the initial ModifiersChanged, leaving modifiers
stale. This caused shortcuts like Cmd+V to be missed and
the bare character inserted instead.

### DIFF
--- a/widget/src/text_input.rs
+++ b/widget/src/text_input.rs
@@ -836,15 +836,14 @@ where
                 text,
                 modified_key,
                 physical_key,
+                modifiers,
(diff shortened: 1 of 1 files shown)


alecthomas:
I'm not sure if this is the correct way to fix this, but it does solve the problem.

PR #3306, opened 2026-04-15
 Goodbye `tiny-skia`, Hello `vello_cpu`

hecrj:
This PR replaces our software renderer—the `tiny-skia` backend—with a brand new one on top of `vello_cpu`.

Why? [It's quite faster](https://laurenzv.github.io/vello_chart/). I haven't written any benches yet, but I will get to that soon.

I have also removed the damage tracking logic for now. It was always a bit glitchy, so we may re-introduce that at a later time.


hecrj:
Added a couple of `cpu` rendering benchmarks:

- A bunch of lore `ipsum` text.
- A basic `application` layout.

Results on my MacBook M4 Max _without_ `multithreading`:

### `ipsum`

image

### `application`

image

nicoburns:
You may also be interested in https://github.com/rust-windowing/softbuffer/pull/329, which makes a *huge* different to the "present" performance of `softbuffer` on macOS in my testing (taking it from 10s of ms down to a couple of hundred microseconds).

zxhd863943427:
But strangely, I tried to use this branch to compile the example tour, and it felt like it was a lot more stuttering than tina-skia.

PR #3313, opened 2026-04-23
implement From<u16> for Pixels

pythcoiner:


hecrj:
Unfortunately, adding this implementation breaks type inference when using integer literals.

PR #3318, opened 2026-04-27
Documentation grammar

mfreeborn:
Tiny fix

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -135,7 +135,7 @@
 //! There is no unified layout system in iced. Instead, each widget implements
 //! its own layout strategy.
 //!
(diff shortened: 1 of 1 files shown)


hecrj:
"consist in using" seems to be the correct wording here.

"consist of [something]" is normally followed by a list of parts; so it would have to be:

> consist of a combination of rows, columns, and containers.

"consist of using" does not seem correct.

mfreeborn:
Pre-edit, you mentioned a noun, which is supposedly what a gerund (i.e. "using") acts as. Either way, "of" is most definitely correct in British English; not that this is a hill worth dying on.

PR #3333, opened 2026-05-17
feat(line_height): accept into pixels

janTatesa:
Modifies `LineHeight::to_absolute` so it accepts `impl Into<Pixels>`


mfreeborn:
Like this https://github.com/iced-rs/iced/pull/3120?

PR #3336, opened 2026-05-22
Enable synchronous image loading to prevent flicker

dvhar:
Problem: Asynchronous image loading introduced here https://github.com/iced-rs/iced/pull/3092 causes images to not render immediately, which is a big problem for applications that need to seamlessly change images. The `image::allocate` task that was added to address that problem is cumbersome to use and clutters up the application logic. For an immediate mode renderer like Iced, immediate image rendering should work without jumping through any hoops.

Solution: Add a <s>feature flag "instant-images"</s> method to the image widget to disable asynchronous image loading. <s>The fix is based on this commit https://github.com/iced-rs/iced/commit/5216253998cadc29b0933ac4f30a8101b423941a which does the same thing for wasm32, but toggles it with a feature rather than an architecture.</s>

### DIFF
--- a/core/src/image.rs
+++ b/core/src/image.rs
@@ -30,6 +30,9 @@ pub struct Image<H = Handle> {
     ///
     /// 0 means transparent. 1 means opaque.
     pub opacity: f32,
+
+    /// Don't render the next frame until the image is ready to display.
+    pub load_blocking: bool,
 }
 
 impl Image<Handle> {
@@ -41,6 +44,7 @@ impl Image<Handle> {
             rotation: Radians(0.0),
             border_radius: border::Radius::default(),
             opacity: 1.0,
+            load_blocking: false,
         }
     }
 
@@ -59,6 +63,13 @@ impl Image<Handle> {
     /// Sets the opacity of the [`Image`].
     pub fn opacity(mut self, opacity: impl Into<f32>) -> Self {
         self.opacity = opacity.into();
+        self
+    }
+
+    /// When set to `true`, the renderer will block until the image is
+    /// ready to display to prevent flickering. Defaults to `false`.
+    pub fn load_blocking(mut self, load_blocking: bool) -> Self {
+        self.load_blocking = load_blocking;
         self
     }
 }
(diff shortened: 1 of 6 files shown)


edwloef:
Maybe this could, instead of being a feature flag, be a different, blocking, image allocation API for the renderer, and then an option you could set on the `image` widget to opt into the blocking behavior??

dvhar:
Good idea, working on it...

dvhar:
Done. Synchronous image loading is now activated with `.load_blocking(true)` on the image widget. I tested it with in-memory and path-based Handles and with a fork of iced_gif that I modified to use it and all work without flickering.

hecrj:
I will need use cases and proper elaboration on how `image::allocate` is "cumbersome" before I can justify a change that potentially allows users to block the runtime and freeze UX.

Videos and animated images need first-class support; not a blocking flag.

dvhar:
Adding complexity to the event loop always makes it easier to introduce bugs. In the case of my file picker, fixing flicker with `image::allocate` means I need to take more care to avoid races. For a security critical application that must never select a file other than the one the user sees, the simplicity of synchronous loading is clearly the best solution.

First-class animations would be great but aren't here yet. Synchronous loading provides a building block to make it work in the meantime. I'm currently blocked from updating my Iced version because it breaks gifs.

thorn132:
For a second example, I'm working on a scientific application that renders a pixel buffer via `Handle::Rgba` generated (at interactive or near-interactive speed) based on some physical properties provided by user inputs. It doesn't make sense to render frames where the physical properties indicated by the UI are fresh but the pixel buffer is stale.

hecrj:
@thorn132 Images with `Handle::Rgba` are actually uploaded synchronously if they are below 2MB.

If your pixel buffer is bigger and changes every frame, then `image` is not fit for it. Instead, I would implement a custom `shader` with a first-class shared buffer (i.e. `MAP_WRITE`) and skip atlas caching altogether.

hecrj:
@dvhar I don't see where this complexity to avoid races arises. If you are okay with blocking UX, you can make sure only a single `image::allocate` is being performed at a time with a single boolean flag.

Far from cumbersome?

dvhar:
The complexity is added to the application logic for keeping image previews and file selections perfectly in sync. It's doable, but the simplicity of version <14 is more foolproof. There's no solution for gifs though.

> Images with Handle::Rgba are actually uploaded synchronously if they are below 2MB.

Making that number configurable would actually be a perfect solution.

PR #3350, opened 2026-06-06
Set resolution limits as defined by adapter for non-wasm32

luca020400:
As of wgpu 29 update we've been hitting the following panic on Apple Silicon.

thread 'main' (15199) panicked at <snip>/wgpu-29.0.3/src/backend/wgpu_core.rs:3879:18:
wgpu error: Validation Error

Caused by:
In Surface::configure
Surface width and height must be within the maximum supported texture size. Requested was (2060, 1528), maximum extent for either dimension is 2048.

A similar issue can be found upstream https://github.com/gfx-rs/wgpu/issues/4576
and the comment recommends setting the resolution given the adapter limits.

We do that already for the wasm32 target, so all this does is extending it
to all targets.

### DIFF
--- a/wgpu/src/window/compositor.rs
+++ b/wgpu/src/window/compositor.rs
@@ -148,7 +148,10 @@ impl Compositor {
         let limits = [wgpu::Limits::downlevel_webgl2_defaults().using_resolution(adapter.limits())];
 
         #[cfg(not(target_arch = "wasm32"))]
-        let limits = [wgpu::Limits::default(), wgpu::Limits::downlevel_defaults()];
+        let limits = [
+            wgpu::Limits::default().using_resolution(adapter.limits()),
+            wgpu::Limits::downlevel_defaults().using_resolution(adapter.limits()),
+        ];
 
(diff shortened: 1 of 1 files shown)


luca020400:
For reference this is the halloy bug report
https://github.com/squidowl/halloy/issues/2068

hecrj:
[`Limits::default` has a `max_texture_dimension_2d` of `8192`](https://docs.rs/wgpu-types/29.0.3/src/wgpu_types/limits.rs.html#376).

For some reason, you must be hitting the `downlevel_defaults`. This should not be the case when using any of the modern `wgpu` backends; so I would certainly look into that if I were you.

luca020400:
> [`Limits::default` has a `max_texture_dimension_2d` of `8192`](https://docs.rs/wgpu-types/29.0.3/src/wgpu_types/limits.rs.html#376).
> 
> For some reason, you must be hitting the `downlevel_defaults`. This should not be the case when using any of the modern `wgpu` backends; so I would certainly look into that if I were you.

Thanks for the input, I don't have any of the affected device at hand, but if times permits I may try to find what limits we get :)

PR #3358, opened 2026-06-12
Draw `rich_text` decorations using font metrics

airstrike:
`rich_text` draws underlines and strikethroughs at a hardcoded fraction of the text size (`size * 0.08`), ignoring the font's real metrics. The quads are also translated by the layout position alone, so they drift away from the glyphs whenever the paragraph is aligned.

This PR adds `Decoration` to `core::text` and `Paragraph::decoration_bounds`, which returns the decoration rectangles cosmic-text resolves at layout from the font's own metrics. `rich_text` feeds span decorations into cosmic-text, including underline geometry for link spans, and fills the bounds it gets back: always for decorated spans, on hover for links, with no relayout on hover. Highlight and decoration quads are anchored the way `fill_paragraph` anchors glyphs, so they stay with the text under alignment.

It also adds a "Rich text" screen to the `tour`, cycling Articles 1 to 3 of the Universal Declaration of Human Rights through eleven languages with a fade transition, exercising wrapping, BiDi, and several scripts. The screen's heading is a link with no explicit underline, revealing one on hover.

Depends on pop-os/cosmic-text#511. `[patch.crates-io]` points at that branch until a release carries it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -995,8 +995,7 @@ dependencies = [
 [[package]]
 name = "cosmic-text"
 version = "0.19.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "be17b688510d934ce13f48a2beba700e11583e281e0fda99c22bb256a14eda73"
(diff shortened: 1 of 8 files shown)


hecrj (commented):
Looks promising!

Let's wait until the `cosmic-text` changes go in.

PR #3363, opened 2026-06-13
fix(svg): parse in-memory SVG with leading whitespace

### LINKED ISSUE #3357: SVG code with formatting fails to load from memory

### What happened?

`svg(svg::Handle::from_memory(bytes))` fails if the bytes are derived from a string that contains formatting, like carriage returns or spaces/tabs. It works if the bytes are included in the source using `include_bytes!` because apparently that sanitizes the string. 

The library might be expected to parse any valid SVG code, not only if it is compressed into a single line.
Here is a small project that shows the issue: https://codeberg.org/psobolik/iced-svg-fun. 

### Version

crates.io release

### Operating System

Linux

```shell

```

StefanoD:
`svg::Handle::from_memory` failed to parse a valid SVG string whenever any whitespace (newline, space, tab, CR) preceded the `<?xml ?>` declaration — e.g. a Rust raw string that starts on its own line. `usvg`'s XML parser rejects a declaration that is not at the very start of the document, so the same SVG worked only when compressed onto a single line or loaded via `include_bytes!` (whose file simply started with `<?xml`).

Both renderers now parse uncompressed in-memory bytes via `usvg::Tree::from_str` after `trim_start()`, mirroring the file-path branch (likewise trimmed for consistency). gzip-compressed data is still handed to `usvg::Tree::from_data` untouched.

Closes #3357

Created by Claude

### DIFF
--- a/tiny_skia/src/vector.rs
+++ b/tiny_skia/src/vector.rs
@@ -11,6 +11,23 @@ use std::fs;
 use std::panic;
 use std::sync::Arc;
 
+/// Parses raw in-memory SVG bytes into a [`usvg::Tree`].
+///
+/// `usvg`'s XML parser rejects an `<?xml ?>` declaration that is preceded by
+/// any whitespace, so a formatted, multi-line SVG string fails to load while
+/// the very same SVG compressed onto a single line succeeds. We trim the
+/// leading whitespace of uncompressed data so both parse. gzip-compressed data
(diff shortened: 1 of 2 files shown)


edwloef:
Shouldn't this be fixed in `usvg` instead of in `iced`?

StefanoD:
Good point.

PR #3365, opened 2026-06-17
Reject software wgpu adapters to allow renderer fallback

zao111222333:
This PR prevents `iced_wgpu` from selecting software-rendered wgpu adapters, such as `llvmpipe (LLVM 17.0.2, 256 bits)` on `Red Hat 8.10`, as the primary renderer.

Previously, when `Backend::Best` was used, the fallback compositor tried `wgpu` first. If wgpu successfully created an adapter backed by `llvmpipe`, iced treated it as a valid wgpu renderer and never fell back to `tiny-skia`.

The change now:

- Detects software adapters using `wgpu::DeviceType::Cpu`
- Also checks known software adapter names like `llvmpipe`, `lavapipe`, `softpipe`, and `swiftshader`
- Rejects these adapters before requesting a device in the window compositor
- Applies the same detection to the headless wgpu renderer
- Adds tests covering `llvmpipe` rejection and normal hardware adapter acceptance

### DIFF
--- a/renderer/Cargo.toml
+++ b/renderer/Cargo.toml
@@ -16,7 +16,7 @@ workspace = true
 [features]
 wgpu = ["iced_wgpu/default"]
 wgpu-bare = ["iced_wgpu"]
-tiny-skia = ["iced_tiny_skia"]
+tiny-skia = ["iced_tiny_skia", "iced_wgpu?/software-fallback"]
 image = ["iced_tiny_skia?/image", "iced_wgpu?/image"]
 svg = ["iced_tiny_skia?/svg", "iced_wgpu?/svg"]
 geometry = ["iced_graphics/geometry", "iced_tiny_skia?/geometry", "iced_wgpu?/geometry"]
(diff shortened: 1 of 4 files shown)


edwloef:
This feels like it should only happen if `tiny-skia` is actually enabled as a fallback, otherwise you just can't start up an iced app using software-rendered gpu backends.

zao111222333:
> This feels like it should only happen if `tiny-skia` is actually enabled as a fallback, otherwise you just can't start up an iced app using software-rendered gpu backends.

That's more reasonable, is there any "fallback or not" option in iced? So I can reuse that configuration

pml68:
> That's more reasonable, is there any "fallback or not" option in iced? So I can reuse that configuration

If `tiny-skia` and `wgpu` (or at least `wgpu-bare`) are both enabled then so is the fallback mechanism.

zao111222333:
Now it only rejects software wgpu adapters when the tiny-skia fallback is available.

PR #3195, opened 2026-01-13
Disable default features of `resvg`

d-itu:
Disable default features (relating to text, raster-images) of `resvg` to avoid code bloating.
The disabled features can be re-exported by some added feature gates in `iced_renderer` crate.

Closes #3369.

PR #3369, opened 2026-06-19
fix: also enable guillotiere for svgs

edwloef:
I had overlooked this in #3362. Weird that it didn't break CI, since it does break e.g. the `slider` example.


hecrj:
I fixed this in #3195.

hecrj:
Also, CI does not catch this because `--workspace` enables the superset of enabled features by all crates.

PR #3372, opened 2026-06-24
Implement `From<()>` for `Element`

airstrike:
`()` is sometimes more ergonomic than `None` for producing an empty `Element`. `None.into()` requires a type annotation because the compiler *cannot infer type of the type parameter `T` declared on the enum `Option`*, while `().into()` just works.

Delegates to `Option::<Self>::None.into()` to reuse the existing `Void` widget without duplication.

### DIFF
--- a/core/src/element.rs
+++ b/core/src/element.rs
@@ -518,6 +518,15 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


hecrj (changes requested):
We should just expose a `void` helper instead.

PR #3398, opened 2026-07-22
Emulator: load fonts and honor ICED_TEST_BACKEND, matching Simulator

egrig:
The headless Emulator is missing two behaviors that Simulator already has, which makes cross-machine screenshot testing of full applications impractical:

Fonts — Emulator ignores Settings::fonts and leaves runtime::Action::Font as a dbg! TODO, so applications with custom fonts render missing glyphs (tofu) in emulator screenshots. This PR loads Settings::fonts before creating the renderer and handles font::Action::Load (loading the bytes and acknowledging the channel), mirroring Simulator.

Backend selection — Simulator reads ICED_TEST_BACKEND to pick its headless renderer backend, but Emulator always passes None, so emulator screenshots cannot be pinned to the deterministic tiny-skia software renderer for golden comparisons (GPU output differs between machines). This PR makes Emulator honor the same variable.

Context: we use Emulator for full-app visual regression tests (boot on recorded data, screenshot, compare against committed goldens). With these two changes, the same goldens verify byte-identically across different machines.

### DIFF
--- a/test/src/emulator.rs
+++ b/test/src/emulator.rs
@@ -95,8 +95,13 @@ impl<P: Program + 'static> Emulator<P> {
         // TODO: Error handling
         let executor = P::Executor::new().expect("Create emulator executor");
 
+        let backend = std::env::var("ICED_TEST_BACKEND").ok();
+
         let renderer = executor
-            .block_on(P::Renderer::new(renderer::Settings::from(&settings), None))
+            .block_on(P::Renderer::new(
+                renderer::Settings::from(&settings),
+                backend.as_deref(),
+            ))
             .expect("Create emulator renderer");
 
         let runtime = Runtime::new(executor, sender);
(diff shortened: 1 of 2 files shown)


test/src/emulator.rs
```diff
@@ -87,11 +89,20 @@ impl<P: Program + 'static> Emulator<P> {
...
         let settings = program.settings();
 
+        for font in &settings.fonts {
+            load_font(font.clone()).expect("Font must be valid");
+        }
+
         // TODO: Error handling
         let executor = P::Executor::new().expect("Create emulator executor");
 
+        let backend = std::env::var("ICED_TEST_BACKEND").ok();
```
hecrj, line 99: We should eventually make this an extra argument and allow an `.ice` test to define the backend (e.g. when ice snapshots are possible).

hecrj (approved):
Thanks!

PR #3182, opened 2026-01-06
Add opacity implementation, basic widget, and example

Ericky14:
Features:
- New Opacity widget - Wraps any content and applies transparency
- Nested opacity support - Opacity values multiply when nested (e.g., 50% × 50% = 25%)
- GPU-accelerated - Implemented in both wgpu and tiny-skia renderers
- Affects everything - Backgrounds, borders, images, SVGs, and text all respect opacity

Example added demonstrating how simple it is to integrate and how to do fade-in/out animations.

Related issue: https://github.com/iced-rs/iced/issues/912

Issue #912, opened 2021-06-14
Is there are any way to programmatically change the opacity of a widgets?


AzimMuradov:
I need a way to programmatically change the opacity of an image.

13r0ck:
Depends on what exactly you are asking for. If you want to just take w hole widget and make it have an opacity then no.
But if you are just looking for a button with a transparent background they yes absolutely, see the [styling example](https://github.com/hecrj/iced/tree/master/examples/styling).

AzimMuradov:
Sadly, the first one.

AzimMuradov:
Is there are any chance to this be implemented in the future releases?

13r0ck:
Oh I'm sure that it will be implemented eventually, but I don't believe that it is even on the roadmap, so I wouldn't expect it anytime soon. 

Why do you want this? I am sure there is some way to get a similar effect with what is already a part of iced?

AzimMuradov:
I want to have a welcome screen in my application with my logo on it.  I want the logo to appear gradually (gradually increasing the opacity).  But I also need that when I click on the screen, the logo appears immediately, without waiting.

13r0ck:
Oh that is going to be tough. Iced doesn't support animations yet, so you would have to manually track and update the opacity too.

Iced has very basic support to stacking widgets, so you would have to place a `Container` ontop of the image widget (See the `Toolltip` or `Picklist` widgets to stack a containers). Then you would have to subscribe to `time::every` to update the `Container` `style`. Performance wouldn't be that great either. Then on click you would set the opacity to the last setting and remove the subscription.

Image dragging would require using the canvas widget. If it is possible at all. I haven't seen anyone build something with an image inside of a canvas though...

AzimMuradov:
Yeah, I already did all code related to subscription, the only thing left is opacity attribute, that I can't find(

13r0ck:
Ya there is no opacity attribute. You will have to overlay a `Container` above the image that has the same background color as the background behind the logo.
`Container` background colors support alpha values so you can set the background of that to have an opacity, thus giving you the desired effect.

AzimMuradov:
I see, that would probably work for me. Thanks for help!

Although, I still hope that this issue will be added to the milestones.

13r0ck:
Yep. I'm sure it will be once Hecrj sees this issue. He is good at that. He is probably just busy now :smiley:

wgxh-cc:
[...]

PR #3400, opened 2026-07-23
Add opacity implementation, basic widget, and example

Ericky14:
Re-opening this PR with correct source branch: https://github.com/iced-rs/iced/pull/3182
And fully rebased on latest master.

Features:

New Opacity widget - Wraps any content and applies transparency
Nested opacity support - Opacity values multiply when nested (e.g., 50% × 50% = 25%)
GPU-accelerated - Implemented in both wgpu and tiny-skia renderers
Affects everything - Backgrounds, borders, images, SVGs, and text all respect opacity
Example added demonstrating how simple it is to integrate and how to do fade-in/out animations.

Related issue: https://github.com/iced-rs/iced/issues/912

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3883,6 +3883,13 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d6790f58c7ff633d8771f42965289203411a5e5c68388703c06e14f24770b41e"
 
 [[package]]
+name = "opacity"
+version = "0.1.0"
+dependencies = [
+ "iced",
+]
+
+[[package]]
 name = "open"
 version = "5.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/core/src/renderer.rs
+++ b/core/src/renderer.rs
@@ -50,6 +50,30 @@ pub trait Renderer {
         self.with_transformation(Transformation::translate(translation.x, translation.y), f);
     }
 
+    /// Starts recording a new opacity group.
+    ///
+    /// All primitives drawn until [`end_opacity`](Self::end_opacity) is called
+    /// will be rendered to an offscreen texture and then composited with the
+    /// given opacity value.
+    ///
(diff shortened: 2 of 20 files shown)


EmmanuelDodoo:
Great work!

One thing to note, the image path in the example won't work on Windows. "/" is not the path separator so ` env!("CARGO_MANIFEST_DIR").replace("/examples/opacity", "")` returns the wrong path.

EmmanuelDodoo:
Also, I shouldn't the opacity be handled by the layer stack like how transformations are?

Ericky14:
> Also, I shouldn't the opacity be handled by the layer stack like how transformations are?

Hmm I did an implementation with layer stack, I liked the other impl because of how simple it was and didn't need to render to an offscreen texture, easy win for perf.
But this shouldn't be so bad either, tried to keep each opacity group bounded to a smaller rect instead of whole window, seems to be working well.

EmmanuelDodoo:
Yep. Overlapping opacity looks to be better. Were you able to benchmark this?

Also it might be worth creating a post on this in the discussions channel of the iced discord so others can give more feedback

Ericky14:
> Yep. Overlapping opacity looks to be better. Were you able to benchmark this?
> 
> Also it might be worth creating a post on this in the discussions channel of the iced discord so others can give more feedback

I did this morning, wasn't sure if I should commit the benchmark code or not since `benches` folder is empty.
I added another cool bonus to this PR, with the layer stack implementation I was seeing bad perf at first, but then I implemented a way to batch and group layers and apply effects to them such as opacity, I am getting great performance now.

This is on the GPU drawing a row with cards.
At opacity 1, its no additional cost since we skip the opacity pipeline: 1000 groups at 0.51ms
At opacity 0.5:
100      0.59ms
1000    5.97ms

This is on the CPU (tiny_skia).
At opacity 1, its no additional cost since we skip the opacity pipeline: 1000 groups at 3.72ms
At opacity 0.5:
100      6.92ms

Also made that discord post.

PR #3422, opened 2026-08-13
Scroll `combo_box` options on keyboard navigation

EmmanuelDodoo:
The current `combo_box` widget supports navigating its options using the arrow keys but doesn't scroll up/down when the options are out of the viewport. This PR fixes that.

Before
https://github.com/user-attachments/assets/12e24c69-7966-4fb2-b6a0-975e24034f0b

After
https://github.com/user-attachments/assets/48ee6059-8cee-45be-8142-fa33cabfaf0b

### DIFF
--- a/widget/src/combo_box.rs
+++ b/widget/src/combo_box.rs
@@ -381,7 +381,6 @@ struct Internal<T, R: text::Renderer> {
     option_matchers: Vec<String>,
     filtered_options: Vec<T>,
     version: u64,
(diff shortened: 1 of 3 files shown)


EmmanuelDodoo:
I was able to fix the scroll so it only happens on the top/bottom edges of the menu

https://github.com/user-attachments/assets/d8231ee1-8a2e-432e-b959-221ac76c5c22

PR #3436, opened 2026-08-27
Fix macOS deletion shortcuts in `Binding::from_key_press`

JokerXiL:
This restores the semantics added to text_input in #2862, which were lost when #3404 unified text editing under core::text::editor. The mapping now uses jump() for word deletion and macos_command() for line deletion, as from_key_press already does for ⌘+arrows.


hecrj (approved):
Thanks!

PR #3438, opened 2026-08-28
Bless lazy widget and throw out ouroboros

B0ney:
With `Component` recently blessed, I figured we might as well try the same thing with `Lazy`, and send ouroboros back to the shadow realm.

With this approach, the computed `Element` is exclusively stored in the widget tree state, which completely removes the complexity of dealing with self-references required for overlays. :grin:

The only disadvantage is the slight overhead of downcasting, but I think it's worth it imo.

While implementing this, I needed to make `tree.diff_children` a free function, i.e `tree::diff_children`, as the borrow checker will complain that `tree.state` is being borrowed. Luckily, `diff_children` does not need it!

I have explored storing the computed `Element` in the impl struct (similar to `Component`, `Transition`, etc) but found that the contents will not initially show unless a rebuild is triggered via a dependency change. It might be possible to achieve this, but I haven't fully verified this.


widget/src/lazy.rs
```diff
@@ -101,48 +71,34 @@ where
...
-        });
+        tree::diff_children(
+            &mut tree.children,
+            std::slice::from_mut(&mut current.element.as_widget_mut()),
+        );
     }
 
     fn size(&self) -> Size<Length> {
-        self.with_element(|element| element.as_widget().size())
+        self.size
```
B0ney, line 88: I'm not sure if this would work as intended, should `fn size()` take `widget::Tree`?

hecrj: Since #3348, `diff` is guaranteed to be called before anything else after every `view` call.

hecrj (approved):
This is fantastic. No more hacky self-references. Thanks!

PR #3455, opened 2026-09-06
Keep extending selection when dragging outside text_editor bounds

A-Disruption:
`position_in` returns `None` once the cursor leaves the widget bounds, so the `?` short-circuits the `CursorMoved` arm and no `Action::Drag` is produced. Selection therefore stops growing the moment the pointer crosses the edge, even though the left button is still held.

`position_from` keeps reporting a position relative to the same origin regardless of whether the cursor is inside the bounds, so the drag keeps extending the selection, matching the behavior of other text editors.


hecrj (approved):
Thanks!

PR #3456, opened 2026-09-06
enable image clipboard and image::allocate on image-without-codecs

B0ney:
Allows you to use the image clipboard and `image::allocate` without needing the extra codecs that come with the image crate.


hecrj (approved):
Ah, I see. `arboard/image` only enables the `png` feature for the `image` crate.

Good catch! Thanks.

PR #3457, opened 2026-09-06
remove max bind groups limitation

B0ney:
I've encountered a validation error in my custom shader that was caused by having more than 2 bind groups (mine has 3). This pr removes the hard-coded limit.

We could do the same thing to `max_non_sampler_bindings` but I'll leave it as is.


hecrj (approved):
It seems like 4 is the default; let's roll with it.

PR #3460, opened 2026-09-07
Fix dropped events when an overlay disappears during update

cuishuang:
## Summary

Fixes an event loss bug in `UserInterface::update` when an overlay disappears
after invalidating the layout during event processing.

## Problem

When an overlay invalidates the layout and the rebuilt overlay is `None`, `UserInterface::update` breaks out of the overlay event loop.

This leaves `overlay_statuses` shorter than the original event batch. The later `events.iter().zip(overlay_statuses)` then stops at the shorter iterator, so events after the event that closed the overlay are never dispatched to the base widget.

This can happen when multiple events are passed to `UserInterface::update`, including the event sequences generated by `iced_test::Simulator`.

The cached `self.overlay` could also remain populated after the actual overlay had disappeared.

### DIFF
--- a/runtime/src/user_interface.rs
+++ b/runtime/src/user_interface.rs
@@ -263,6 +263,7 @@ where
                         .map(overlay::Nested::new);
 
                     if maybe_overlay.is_none() {
+                        event_statuses.resize(events.len(), event::Status::Ignored);
                         break;
                     }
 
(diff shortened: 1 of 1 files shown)


hecrj (approved):
Thanks!

I simplified the test a bit.

